PR #83, opened 2018-12-17
[nrf noup] dts: Add nRF52810 to secureboot.overlay

oyvindronningstad:
This provides support for BOARD=nrf52810_pca10040.

### DIFF
--- a/dts/common/secureboot.overlay
+++ b/dts/common/secureboot.overlay
@@ -1,32 +1,4 @@
-#ifdef CONFIG_SOC_NRF52840
-&flash0 {
-	partitions {
-		compatible = "fixed-partitions";
-		#address-cells = <1>;
-		#size-cells = <1>;
-		b0_partition: partition@0 {
-			label = "secure-boot";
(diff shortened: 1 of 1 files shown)


oyvindronningstad:
cc @hakonfam @ioannisg

hakonfam:
While you are in here, could you re-order the file so that the order is a bit more intuitive?

oyvindronningstad:
>While you are in here, could you re-order the file so that the order is a bit more intuitive?

Reordered and added nrf51822_qfaa which has the same flash size as qfac and is used in some boards (e.g. microbit)

oyvindronningstad:
The Bamboo is marked as failed, but the linked run was green, I don't know how to fix it, can we merge anyways?

Anything preventing us merging this, @ioannisg and @hakonfam?

hakonfam:
LGTM

oyvindronningstad:
The build failure seems to be related to samples/application_development/code_relocation/test. This test fails on my branch, and on the nrf91 branch, but not on zephyr upsteam.

PR #88, opened 2019-01-24
[nrf noup] ci: Add Jenkins-based CI for zephyr

myrdyr:
Build on Jenkins

Signed-off-by: Ulrich Myhre <ulmy@nordicsemi.no>


myrdyr:
With the current state of the master, the sanitycheck does not try to build any nrf91 samples, because it thinks the board doesn't exist. Whenever the changes for nrf91 are upmerged, we need to make sure that the nrf91 samples do build.

The CI will also fail the KConfig check, because the NRF91* symbols are not existing. This we should just accept IMO.

To speed up things, I've also divided the sanitycheck into three parts, utilizing the subset mechanism already present in the script. This should triple the speed.

PR #90, opened 2019-01-30
[nrf noup] kconfig: move nRF Connect entry at the top

lemrey:
Move ~~nRF Connect SDK~~ Zephyr modules Kconfig entries at the top of the menu.
They are currently in a random spot.

### DIFF
--- a/Kconfig.zephyr
+++ b/Kconfig.zephyr
@@ -7,6 +7,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+source "$(PROJECT_BINARY_DIR)/Kconfig.modules"
+
 # Include these first so that any properties (e.g. defaults) below can be
 # overriden in *.defconfig files (by defining symbols in multiple locations).
 # After merging all the symbol definitions, Kconfig picks the first property
@@ -37,8 +39,6 @@ source "lib/Kconfig"
 source "subsys/Kconfig"
 
 source "ext/Kconfig"
-
-source "$(PROJECT_BINARY_DIR)/Kconfig.modules"
 
 menu "Build and Link Features"
 


Qbicz:
I don't know original intent of @lemrey, but I think it looks better when a customer downloads Nordic nRF Connect SDK and see nRF-specific options on top, rather than 3rd from the bottom...

lemrey:
@ioannisg no, it does not solve any practical issue. I think it's worth doing though, because the current situation is that that entry is in the middle of the top level menu, in what really is a random spot.

ioannisg:
Could you take care of CI failure here, @lemrey ?

lemrey:
@ioannisg I think the CI failures are unrelated to this PR and coming from upstream, not sure I can do much about them.

ioannisg:
> 
> 
> @ioannisg I think the CI failures are unrelated to this PR and coming from upstream, not sure I can do much about them.

Would it make sense to re-check?

lemrey:
Rebased, hopefully will be greener.

ioannisg:
Looks like there were 2 tests failing in mps2_an385. We can try once more, or just merge this one, IMHO.

lemrey:
Rebased. @SebastianBoe @carlescufi This patch moves the Kconfig entries of Zephyr modules at the top of the menu. The rationale is that they now appear in a random spot in the menu and have poor visibility.

SebastianBoe:
Since there is no rush to get this in, wouldn't it be less paperwork to push this upstream, and then let it trickle down automatically?

Then we don't need to go through review here and upstream.

carlescufi:
@lemrey agreed with @SebastianBoe. How about we try to push this upstream first?

lemrey:
Will do.

lemrey:
Pushed upstream.

PR #91, opened 2019-01-31
[nrf noup] add field required for http DFU client

hakonfam:
Add fields required for HTTP GET DFU protocol.


hakonfam:
@lemrey

hakonfam:
Close since this should be added with together with other feature.

PR #92, opened 2019-01-31
[nrf noup] scripts: west: nrfjprog: Support nRF91

carlescufi:
This file will disappear when we upmerge next, but we need this fix in
the meantime to be able to flash the nRF91.


carlescufi:
>Why don't we just upmerge again, then?

Because upstream has introduced west and we need a bit of time to prepare for it.

PR #93, opened 2019-01-31
[nrf noup]: logging: Logger fixes

Qbicz:
This PR includes fixes for following issues:
* asserts not printed
* device hangs on going to deep sleep mode when JLink is not connected

Fixes are needed to enable proper automated testing on nRF52 Desktop devices.

With the next mergeup, these commits shouldn't be included anymore (they come from upstream).

### DIFF
--- a/include/misc/__assert.h
+++ b/include/misc/__assert.h
@@ -81,16 +81,7 @@
 
 #if __ASSERT_ON
 #include <misc/printk.h>
-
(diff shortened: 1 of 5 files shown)


Qbicz:
Holding off because of upmerge https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/94

PR #95, opened 2019-02-01
[RFC] Adding pca20035 board support

joakimtoe:
Depending on upstream PR https://github.com/zephyrproject-rtos/zephyr/pull/9798


lemrey:
@joakimtoe if we're not planning to upstream this, we should rather commit it to fw-nrfconnect-nrf.

joakimtoe:
@lemrey This should go upstream.

lemrey:
Ok, then I think you'll need to add the `[nrf toup]` prefix to the commit message :)

joakimtoe:
CC @carlescufi @ioannisg

PR #96, opened 2019-02-01
[DNM] drivers: flash: soc_flash_nrf: add support for nrf9160

lemrey:
This PR is intended to give visibility to the workaround to necessary to use Zephyr's flash driver on nRF9160.

FYI: @gbakke @osaether

### DIFF
--- a/drivers/flash/soc_flash_nrf.c
+++ b/drivers/flash/soc_flash_nrf.c
@@ -14,6 +14,7 @@
 #include <soc.h>
 #include <flash.h>
 #include <string.h>
+#include <nrfx.h>
 
 #if defined(CONFIG_SOC_FLASH_NRF_RADIO_SYNC)
 #include <misc/__assert.h>
@@ -77,6 +78,17 @@ static struct k_sem sem_lock;
 #define SYNC_UNLOCK()
 #endif
 
+#if defined(CONFIG_SOC_SERIES_NRF51X)
+#define CODEPAGESIZE 1024
+#define CODESIZE 256
+#elif defined(CONFIG_SOC_NRF52832)
+#define CODEPAGESIZE 4096
+#define CODESIZE 128
+#elif defined(CONFIG_SOC_NRF9160) || defined(CONFIG_SOC_NRF52840)
+#define CODEPAGESIZE 4096
+#define CODESIZE 256
+#endif
+
 static int write(off_t addr, const void *data, size_t len);
 static int erase(u32_t addr, u32_t size);
 
@@ -87,10 +99,10 @@ static inline bool is_aligned_32(u32_t data)
 
 static inline bool is_regular_addr_valid(off_t addr, size_t len)
 {
-	if (addr >= NRF_FICR->CODEPAGESIZE * NRF_FICR->CODESIZE ||
+	if (addr >= CODEPAGESIZE * CODESIZE ||
 	    addr < 0 ||
-	    len > NRF_FICR->CODEPAGESIZE * NRF_FICR->CODESIZE ||
-	    addr + len > NRF_FICR->CODEPAGESIZE * NRF_FICR->CODESIZE) {
+	    len > CODEPAGESIZE * CODESIZE ||
+	    addr + len > CODEPAGESIZE * CODESIZE) {
 		return false;
 	}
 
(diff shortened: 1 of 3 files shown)


drivers/flash/soc_flash_nrf.c
```diff
@@ -77,6 +78,17 @@ static struct k_sem sem_lock;
 #define SYNC_UNLOCK()
 #endif
 
+#if defined(CONFIG_SOC_SERIES_NRF51X)
```
nvlsianpu: you resign form ficr usage, maybe it is better.

drivers/flash/soc_flash_nrf.c
```diff
@@ -127,8 +139,33 @@ static void nvmc_wait_ready(void)
 	}
 }
 
-static int flash_nrf_read(struct device *dev, off_t addr,
-			    void *data, size_t len)
+enum nvmc_config {
```
nvlsianpu: by default enums are int size, but it possible to switch to other size.
Fro the other hand nice to use this in func. declaration.
Maybe no worth to be so carefully.

drivers/flash/soc_flash_nrf.c
```diff
@@ -217,12 +254,7 @@ static int flash_nrf_write_protection(struct device *dev, bool enable)
 {
 	SYNC_LOCK();
 
-	if (enable) {
-		NRF_NVMC->CONFIG = NVMC_CONFIG_WEN_Ren << NVMC_CONFIG_WEN_Pos;
-	} else {
-		NRF_NVMC->CONFIG = NVMC_CONFIG_WEN_Wen << NVMC_CONFIG_WEN_Pos;
-	}
-	nvmc_wait_ready();
+	nvmc_config_set(enable ? REN : WEN);
```
nvlsianpu: `(void)` before will suppress warning

drivers/flash/soc_flash_nrf.c
```diff
@@ -432,8 +464,7 @@ static int write_in_timeslice(off_t addr, const void *data, size_t len)
 
 static int erase_op(void *context)
 {
-	u32_t prev_nvmc_cfg = NRF_NVMC->CONFIG;
```
nvlsianpu: variable should be kept

drivers/flash/soc_flash_nrf.c
```diff
@@ -483,8 +517,7 @@ static int erase_op(void *context)
 
 	} while (e_ctx->len > 0);
 
-	NRF_NVMC->CONFIG = prev_nvmc_cfg;
-	nvmc_wait_ready();
+	nvmc_config_set(REN);
```
nvlsianpu: `nvmc_config_set(prev_nvmc_cfg);`

drivers/flash/soc_flash_nrf.c
```diff
@@ -446,9 +477,8 @@ static int erase_op(void *context)
 	}
 #endif /* CONFIG_SOC_FLASH_NRF_RADIO_SYNC */
 
-	/* Erase uses a specific configuration register */
-	NRF_NVMC->CONFIG = NVMC_CONFIG_WEN_Een << NVMC_CONFIG_WEN_Pos;
-	nvmc_wait_ready();
+	/* Enable erase operation */
+	nvmc_config_set(EEN);
```
nvlsianpu: prev_nvms_cfg = nvmc_config_set(EEN);

drivers/flash/soc_flash_nrf.c
```diff
@@ -513,6 +546,9 @@ static int write_op(void *context)
 	}
 #endif /* CONFIG_SOC_FLASH_NRF_RADIO_SYNC */
 
+	/* Enable write operation */
```
nvlsianpu: f. flash_nrf_write_protection() is for enable write explicite. Pleas remove this form here - its breaks API compatibility.

IDEMIA-Pessac:
If I may jump in, there are some issues with the proposed patch.
The flash write and erase features won't be working with 
`CONFIG_ARM_SECURE_FIRMWARE=y`

I had to slightly modify the usage of `NRF_NVMC->CONFIG` like this : 

```
enum nvmc_config {
#if defined(CONFIG_SOC_NRF9160) && !defined(CONFIG_ARM_SECURE_FIRMWARE)
	REN = NVMC_CONFIGNS_WEN_Ren << NVMC_CONFIGNS_WEN_Pos,
	WEN = NVMC_CONFIGNS_WEN_Wen << NVMC_CONFIGNS_WEN_Pos,
	EEN = NVMC_CONFIGNS_WEN_Een << NVMC_CONFIGNS_WEN_Pos,
#else
	REN = NVMC_CONFIG_WEN_Ren << NVMC_CONFIG_WEN_Pos,
	WEN = NVMC_CONFIG_WEN_Wen << NVMC_CONFIG_WEN_Pos,
	EEN = NVMC_CONFIG_WEN_Een << NVMC_CONFIG_WEN_Pos,
#endif
};

static u32_t nvmc_config_set(enum nvmc_config mode)
{
#if defined(CONFIG_SOC_NRF9160) && !defined(CONFIG_ARM_SECURE_FIRMWARE)
	const u32_t cfg = NRF_NVMC->CONFIGNS;
	NRF_NVMC->CONFIGNS = mode;
#else
	const u32_t cfg = NRF_NVMC->CONFIG;
	NRF_NVMC->CONFIG = mode;
#endif
	nvmc_wait_ready();
	return cfg;
}
```

nvlsianpu:
replaced by #118

PR #101, opened 2019-02-28
[nrf temphack] include: arm: linker.ld

sigvartmh:
Fix up for linker script to place firmware metadata in the built hexfile
for secure bootloader.

### DIFF
--- a/include/arch/arm/cortex_m/scripts/linker.ld
+++ b/include/arch/arm/cortex_m/scripts/linker.ld
@@ -172,6 +172,11 @@ SECTIONS
 	KEEP(*(".kinetis_flash_config.*"))
 
 	_vector_end = .;
+	/* This is used for Secure Boot */
+	#if defined(CONFIG_SECURE_BOOT)
+	. = CONFIG_SB_FIRMWARE_INFO_OFFSET;
+	KEEP(*(.firmware_info))
+	#endif
 	} GROUP_LINK_IN(ROMABLE_REGION)
 
 #ifdef CONFIG_CODE_DATA_RELOCATION


include/arch/arm/cortex_m/scripts/linker.ld
```diff
@@ -172,6 +172,11 @@ SECTIONS
 	KEEP(*(".kinetis_flash_config.*"))
 
 	_vector_end = .;
+	/* This is used for Secure Boot */
+	#if defined(CONFIG_SB_FIRMWARE_INFO_OFFSET)
```
nvlsianpu: Is it posible to not define CONFIG_SB_FIRMWARE_INFO_OFFSET while building using fw-nrfconnect-nrf?

nvlsianpu: Where is the switch to turn this off?

sigvartmh: It's turned on of by `CONFIG_SECURE_BOOT` I guess setting `CONFIG_SECURE_BOOT` would be more sensible I'll change it. It was a copy pasta from old linker script.

sigvartmh: As for where it is set it's set here https://github.com/NordicPlayground/fw-nrfconnect-nrf/blob/9816a684f31d62aad02fb6f7ead41dbb9320282a/subsys/bootloader/fw_metadata/Kconfig#L17

which is included here https://github.com/NordicPlayground/fw-nrfconnect-nrf/blob/3a15a14d17dbe3129514a228b34729d00e23ee09/subsys/bootloader/Kconfig#L104

sigvartmh: Changed it so it should make more sense now @nvlsianpu

nvlsianpu:
See no obstacles to merge this - CI failure is related to CI framework itself.

PR #104, opened 2019-03-07
[RFC] cmake: Add Cmake function for adding named sections to the linker file

oyvindronningstad:
Call add_named_section(".foo") to add a FLASH section with the name .foo.
Will also create the symbols _foo_start, _foo_end, and _foo_size.

### DIFF
--- a/cmake/extensions.cmake
+++ b/cmake/extensions.cmake
@@ -689,6 +689,248 @@ function(zephyr_check_compiler_flag lang option check)
   endif()
 endfunction()
 
+
+# add_named_section(section_name [parameters])
+#
+# This function creates a linker section with the given name placed in the given
+# region, and creates symbols for the start, end, and size of the section.
+#
+# See also add_named_ro_section() and add_named_rw_section() to avoid specifying
+# the region.
+#
+# This function will give an error if the same section name is added multiple
+# times, since code will often expect only its own data to be present in the
+# section.
+#
+# Parameters:
+# No-argument parameters:
+# NOLOAD: Make the section NOLOAD, i.e. not initialized.
+# OPTIONAL: Make the section OPTIONAL.
+# MPU_ALIGN: The special MPU_ALIGN macro will be used for aligning the "start"
+#            and end of the section. This alignment happens within the section.
+#
+# Single-argument parameters:
+# REGION: The region to place the section in. The region must be specified in
+#         the linker script. Default: RAMABLE_REGION
+# LOAD_REGION: The region to load from. Using this parameter also adds a symbol
+#              e.g. _foo_rom_start that has the start of the load source.
+# GROUP_START: The name of the group that starts with this section. This
+#              parameter must be given exactly once per group.
+# GROUP_END: The name of the group that ends with this section. This parameter
+#            must be given exactly once per group. When GROUP_START and
+#            GROUP_END has been given for a group, the sections in between as
+#            well as the once in which GROUP_START and GROUP_END were specified
+#            are considered inside this group. There are symbols for the start,
+#            end, and size of the group.
+# ALIGN: The alignment of the start of the section. This cannot be used together
+#        with LOAD_REGION because with LOAD_REGION, ALIGN_WITH_INPUT is used
+#        instead. Default: 4 (without LOAD_REGION)
+# SUBALIGN: The alignment of each input section within this section.
+#
+# Multi-argument parameters:
+# INPUT_SECTIONS: A list of input sections to be put into this section.
+#                 Default: KEEP(*(<section_name>))
+#                          KEEP(*("<section_name>.*"))
+#
+# Example 1:
+# add_named_section(".foo")
+# creates a section named .foo in RAM, and the symbols __foo_start, __foo_end, and
+# __foo_size. Notice that the '.' is converted to a '_'
+#
+# To access the symbols from code:
+#   extern const mytype_t __foo_start;
+#   extern const u32_t __foo_size;
+#   ...
+#   do_something_with_foo(&__foo_start, (u32_t)&__foo_size);
+#
+# To add c variables to the section:
+#   const mytype_t var_placed_in_foo _GENERIC_SECTION(.foo) __attribute__((used)) = something;
+#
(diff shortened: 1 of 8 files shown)


oyvindronningstad:
cc @hakonfam @sigvartmh @osaether

cmake/extensions.cmake
```diff
@@ -689,6 +689,80 @@ function(zephyr_check_compiler_flag lang option check)
...
+# To add c variables to the section:
+#   const mytype_t var_placed_in_foo _GENERIC_SECTION(.foo) __attribute__((used)) = something;
+#
+# Example 2:
+# add_named_section("bar", ROMABLE_REGION)
+# creates a section named bar in flash, and the symbols _bar_start, _bar_end,
+# and _bar_size.
+#
+set(generated_sections_filename "${PROJECT_BINARY_DIR}/include/generated/generated-sections.ld")
+file(WRITE ${generated_sections_filename} "")
```
SebastianBoe: So far, unless I've missed a patch, we have been able to limit this file to functions and macros. I'd like this to continue to hold if possible.

include/arch/arm/cortex_m/scripts/linker.ld
```diff
@@ -515,6 +515,10 @@ SECTIONS
 #include <custom-sections.ld>
 #endif
 
+/* Located in generated directory. This file is populated by the
+ * add_named_section() Cmake function. */
+#include <generated-sections.ld>
+
```
SebastianBoe: We need to add this to all linker.ld unfortunately.

cmake/extensions.cmake
oyvindronningstad: Done

include/arch/arm/cortex_m/scripts/linker.ld
oyvindronningstad: Will do

oyvindronningstad:
@SebastianBoe Thanks, I started addressing your comments. I ported CCM_*, TI_CCFG, and ramfunc so far. Take a look to see whether you think it's going in the right direction.

SebastianBoe:
Can you elaborate on the motivation?

Is it because we think it is easier for users to interface with the add_named_section API than the LD API, or is it to support out-of-tree use-cases?

oyvindronningstad:
> Can you elaborate on the motivation?

Yep, these are my main points:
 - Keep the linker definitions of a module together with everything else to do with that module.
 - Make it easier to add a section, syntax-wise.
 - Make it easier to enforce "best-practices", f.ex. the restriction on usage of `SECTION_DATA_PROLOGUE`
 - Make it less invasive to add a section, since you're less likely to interfere with what is already there. Right now it's hard/impossible to know if you are messing with the surrounding sections. It needs to be easier to see what sections/symbols must be placed in a specific position globally.
 - Make the linker script cleaner and more readable.

The out-of-tree point is a bonus, hadn't thought of that :).

sigvartmh:
I really like this change as it cleans up the vendor specific stuff out of the linkerscript where it doesn't really belong.

SebastianBoe:
I'll get back to this after multi-image is merged.

There are definitely some good ideas here that we want to take advantage of :)

soc/arm/st_stm32/common/CMakeLists.txt
```diff
@@ -1,2 +1,8 @@
 zephyr_sources_ifdef(CONFIG_ARM_MPU arm_mpu_regions.c)
 zephyr_sources(stm32cube_hal.c)
+
+if (DEFINED DT_CCM_BASE_ADDRESS)
+  add_named_section(".ccm_bss" REGION CCM GROUP_START CCM)
```
SebastianBoe: Are you missing NOLOAD?

include/arch/arm/cortex_m/scripts/linker.ld
```diff
@@ -472,45 +472,6 @@ SECTIONS
...
-	{
-		__ccm_data_start = .;
-		*(.ccm_data)
-		*(".ccm_data.*")
-	} GROUP_LINK_IN(CCM AT> ROMABLE_REGION)
-
-	__ccm_data_end = .;
-	__ccm_end = .;
-
-	__ccm_data_rom_start = LOADADDR(_CCM_DATA_SECTION_NAME);
```
SebastianBoe: Are we missing this symbol?

SebastianBoe:
When I read this I am finding it easier to predict the behaviour of the original LD scripts than the new API.

My suggestion would therefore be to keep these improvements:

> Keep the linker definitions of a module together with everything else to do with that module.
> 
> Make it less invasive to add a section, since you're less likely to interfere with what is already there.
> 
> Right now it's hard/impossible to know if you are messing with the surrounding sections. It needs to be easier to see what sections/symbols must be placed in a specific position globally.
> 
> Make the linker script cleaner and more readable.
> 
> Support injecting multiple out-of-tree LD script snippets.

And leave these be as-is:

> Make it easier to add a section, syntax-wise.
> 
> Make it easier to enforce "best-practices", f.ex. the restriction on usage of SECTION_DATA_PROLOGUE
> 

by having the API append/concatenate local LD script snippets instead of generating LD script itself.

oyvindronningstad:
> When I read this I am finding it easier to predict the behaviour of the original LD scripts than the new API.
> 
> My suggestion would therefore be to keep these improvements:
> 
> > Keep the linker definitions of a module together with everything else to do with that module.
> > Make it less invasive to add a section, since you're less likely to interfere with what is already there.
> > Right now it's hard/impossible to know if you are messing with the surrounding sections. It needs to be easier to see what sections/symbols must be placed in a specific position globally.
> > Make the linker script cleaner and more readable.
> > Support injecting multiple out-of-tree LD script snippets.
> 
> And leave these be as-is:
> 
> > Make it easier to add a section, syntax-wise.
> > Make it easier to enforce "best-practices", f.ex. the restriction on usage of SECTION_DATA_PROLOGUE
> 
> by having the API append/concatenate local LD script snippets instead of generating LD script itself.

I think this is a good call. I opened a separate pull request for the updated work (#109)

PR #106, opened 2019-03-12
Revert "[nrf temphack] cmake: Add nRF specific hook"

carlescufi:
This reverts commit 86599ab896febfcdf9a58f6026f49e4a45c3c103.


carlescufi:
See https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/544

Jenkinsfile
```diff
@@ -84,14 +83,6 @@ pipeline {
             }
           }
         }
-
-        stage('Build nrf samples') {
```
lemrey: This was removed because it's now done by CI in `fw-nrfconnect-nrf`, correct?

carlescufi: Correct

carlescufi:
@lemrey @SebastianBoe the error in sanitycheck is the usual suspect. An unstable test. I need to find a way to exclude that board.

PR #107, opened 2019-03-14
[nrf mergeup] Merge upstream up to commit 19de7ec

rlubos:
Regular upmerge.

Signed-off-by: Robert Lubos <robert.lubos@nordicsemi.no>


hakonfam:
FYI this upmerge is needed for CI not to fail on PRs against fw-nrfconnect-nrf

lemrey:
Arg, this is gonna be tough.

SebastianBoe:
> FYI this upmerge is needed for CI not to fail on PRs against fw-nrfconnect-nrf

A separate PR is now posted so we can restore CI while we debug the upmerge.

lemrey:
The issue is like introduce in:  af10d16.

lemrey:
@hakonfam could you please verify the issue has been solved?

hakonfam:
> @hakonfam could you please verify the issue has been solved?

The issue has been resolved.

SebastianBoe:
I believe there is an unresolved stack overflow somewhere in this mergeup branch (possibly from net_pkt.c).

Will look into it on monday.

Enabling CONFIG_STACK_SENTINEL causes the mqtt_simple sample to go from passing to 

```
Secure Boot: prepare to jump to Non-Secure image
***** HARD FAULT *****
  Fault escalation (see below)
***** Hardware exception *****
Current thread ID = 0x200204e4
Faulting instruction address = 0x45ffe
Fatal fault in ISR! Spinning...

```

Which implies to me that a stack overflow is being detected, but I can't explain that the error message is so poor.

Also, adding some padding after one of the net_pkt.c noinit.c section's causes the build to go from crashing to passing. Which implies a buffer in net_pkt.c is overflowing by a few bytes.

Also, the STACK_SENTINEL infrastructure reports a Stack Check Fail for certain orderings of the .noinit sections.

rlubos:
@SebastianBoe It seems that enabling stack sentinel causes sample to crash even on current master, so the upstream changes might not be a case here.

Still this should be investigated - the crash seems to happen at `bsd_init` call so it can be tough to debug :|

lemrey:
Check that the projects that hard-fault implement the bsdlib error handlers, if they don't, the weak implementation inside the library will hard-fault on errors. `mqtt_simple` is one sample that lacks the handlers, and thus hard-faults. I don't know why the library fails to initialize though, but it seems related to the patch that I reverted.

SebastianBoe:
> it seems related to the patch that I reverted

My working theory is that there has existed a stack overflow in something the mqtt_simple sample uses for quite some time, and that changing where put stuff in memory has changed it from a benign stack overflow to a malignant stack overflow.

The real question is where is the stack overflow, struggling to narrow it down. Tried to bisect, but I couldn't figure out which SHA's of nrf are compatible with which SHA's of Zephyr.

SebastianBoe:
> Check that the projects that hard-fault implement the bsdlib error handlers

What needs to be done to have the mqtt_sample to implement the bsdlib error handler?

rlubos:
I've just added the following in the main for the sake of debugging:
```
/**@brief Recoverable BSD library error. */
void bsd_recoverable_error_handler(uint32_t err)
{
	__ASSERT_NO_MSG(false);
}

/**@brief Irrecoverable BSD library error. */
void bsd_irrecoverable_error_handler(uint32_t err)
{
	__ASSERT_NO_MSG(false);
}
```

As a short term solution, we should have these in each sample. Long term - bsdlib should get rid of the default error handlers, and we should have a proper weak handles somewhere in `bsd_os`.

rlubos:
It's a main stack size that causes the problem. This can be clearly seen when manually stepping over the `bsd_init`:
```
***** Stack Check Fail! *****
Current thread ID = 0x200205c4
Faulting instruction address = 0x4737a
```
```
(gdb) info symbol 0x200205c4
_main_thread_s in section bss
```
Increasing main stack size to 4k resolved the crash (2k is still not enough). The reason that it was not seen rightaway is that for me, the crash ouput is somewhat rubbish:

```
Secure Boot: MSP_NS 20024030
Secure Boot: pre***** HARD FAULT *****
  Fault escalation (see below)
***** Hardware exception *****
Current thread ID = 0x200205c4
Faulting instruction add
```

Any ideas what buffers need to be increased to get a decent output? Logger is disabled in this sample, hence I'm lost here.

rlubos:
Given above, is https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/107/commits/15d31b5bbf3681a8d2ef10df399f5109fcde8a8e actually needed?

The sample seems to pass the initialization w/o errors with stack increased and no reverted commits. Still, this should be verifed by someone with a proper LTE connection (I'm stuck at `LTE Link Connecting ...`), but if there's no need to mess up with the git history, we should avoid that.

SebastianBoe:
15d31b5bbf3681a8d2ef10df399f5109fcde8a8e

is no longer needed if we have a reasonable fix.

SebastianBoe:
> Still, this should be verifed by someone with a proper LTE connection 

I can confirm that I get a connection when STACK_SENTINEL and MAIN_STACK_SIZE=4096

SebastianBoe:
So, for this upmerge, we get rid of the revert to align with Zephyr. Increase the main stack size to fix the stack overflow that was brought to light by the upmerge, and then long term;

Ensure that bsd lib does not corrupt the kernel's error reporting, e.g. stack check.
Maybe reduce the size of main stack so new application's don't need to remember to set it. Is there anything in the main stack that we can put elsewhere perhaps, in another thread, or in a buffer?

lemrey:
I agree with @SebastianBoe, to sum it up:
To sum it up:
- drop the `temphack` commit
- implement weak handlers in `mqtt_simple` project (temporary fix)
- increase stack size in `mqtt_simple` project (temporary fix?)

Then we can change the bsdlib own error handlers' behavior to prevent confusing messages.

rlubos:
MQTT sample fixes are submitted in a separate PR (https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/559), let's finish the upmerge

ioannisg:
> 
> 
> Something is off here.
> NACK, while investigating:
> 
> samples/nrf9160/mqtt_simple:
> 
> ```
> Secure Boot: prepare to jump to Non-Secure image
> Exception occurred in Secure State
> ***** HARD FAULT *****
>   Fault escalation (see below)
> ***** BUS FAULT *****
>   Precise data bus error
>   BFAR Address: 0x50008120
> ***** Hardware exception *****
> Current thread ID = 0x20020224
> Faulting instruction address = 0x0
> Fatal fault in ISR! Spinning...
> ```

@lemrey @SebastianBoe and all. this is a SecureFault exception

ioannisg:
@SebastianBoe

>So, for this upmerge, we get rid of the revert to align with Zephyr. Increase the main stack size to fix the stack overflow that was brought to light by the upmerge, and then long term;

>Ensure that bsd lib does not corrupt the kernel's error reporting, e.g. stack check.
Maybe reduce the size of main stack so new application's don't need to remember to set it. Is there anything in the main stack that we can put elsewhere perhaps, in another thread, or in a buffer? 

To investigate stack overflow: have you enabled `CONFIG_HW_STACK_PROTECTION`?

SebastianBoe:
We are done debugging. The sample was using more main stack than it had enabled.

PR #109, opened 2019-03-20
[RFC] cmake: Add function for adding snippets to the linker file

oyvindronningstad:
add_linker_snippet() allows snippets to be placed in different locations:
 - The noinit, rwdata and rodata output sections
 - Two different locations for placing custom output sections, one location for RAM and another for all other sections.

This PR removes the custom-\*.ld and soc-\*.ld mechanisms which are similar, but more cumbersome to use.


cmake/extensions.cmake
```diff
@@ -689,6 +689,48 @@ function(zephyr_check_compiler_flag lang option check)
   endif()
 endfunction()
 
+function(add_linker_snippet location snippet)
+  #set(snippet_base "${PROJECT_BINARY_DIR}/include/generated/snippets-")
```
hakonfam: remove

include/arch/arc/v2/linker.ld
```diff
@@ -114,6 +114,9 @@ SECTIONS {
 		*(".rodata.*")
 		*(.gnu.linkonce.r.*)
 
+/* Located in generated directory. This file is populated by the
+ * add_linker_snippet() Cmake function. */
```
hakonfam: I think you have to have the ending `*/` on a new line as discussed earlier.

cmake/extensions.cmake
oyvindronningstad: Done

include/arch/arc/v2/linker.ld
oyvindronningstad: Done

SebastianBoe:
wow, nice check

carlescufi:
@oyvindronningstad any reason this cannot go upstream?

oyvindronningstad:
> @oyvindronningstad any reason this cannot go upstream?

No, I wanted to RFC it here first, but I'm hoping it can go upstream. I'll open a PR.

arch/arm/core/cortex_m/CMakeLists.txt
```diff
@@ -9,3 +9,12 @@ zephyr_library_sources(
...
   exc_manage.c
   )
+
+add_linker_snippet_ifdef(CONFIG_SW_VECTOR_RELAY RAM_SECTIONS
+  "/* Reserved 4 bytes to save vector table base address */
+  SECTION_PROLOGUE(.vt_pointer,(NOLOAD),)
+  {
+          *(.vt_pointer_section)
+          *(\".vt_pointer_section.*\")
+  }"
```
SebastianBoe: Thoughts on having the LD script code in .ld files instead?

Benefits as I see it;

- Easy to find LD script code, (look for .ld files).
- Easier for an editor to help with formatting and syntax highlighting.
- User does not need to know which symbols need to be escaped.
- Consistent with how we add other files, e.g. c files.

oyvindronningstad: Sure, I see no big reasons not to. So the Cmake would look something like this maybe?

    zephyr_linker_sources(<location> <files>)
e.g:

    zephyr_linker_sources(RODATA
      ${CMAKE_CURRENT_LIST_DIR}/mylinkerscript.ld
    )

oyvindronningstad:
PR open upstream

hakonfam:
https://github.com/zephyrproject-rtos/zephyr/pull/15112

Kconfig.zephyr
```diff
@@ -143,24 +143,6 @@ config CUSTOM_LINKER_SCRIPT
...
-	help
-	  Include a customized linker script fragment for inserting additional
-	  data and linker directives into the data section.
-
-config CUSTOM_SECTIONS_LD
-	bool "Include custom-sections.ld"
-	help
-	  Include a customized linker script fragment for inserting additional
-	  arbitrary sections.
-
```
SebastianBoe: Before dropping support for CUSTOM_RODATA_LD we might want to first mark it as deprecated to give users a reasonable window of time to migrate.

arch/arm/core/cortex_m/CMakeLists.txt
SebastianBoe: yes

SebastianBoe: check out how zephyr_sources is implemented to see how to get rid of the CMAKE_CURRENT_LIST_DIR.

SebastianBoe:
Closing as this is now an upstream RFC.

arch/arm/core/cortex_m/CMakeLists.txt
oyvindronningstad: Fixed in upstream PR

PR #111, opened 2019-03-21
[nrf temphack] cmake: kconfig: Support building without kernel

hakonfam:
Certain components (e.g. bootloaders) will get conflicts
with symbols defined in 'drivers' and 'arch', create kconfig
which allows them to opt out of these.


hakonfam:
I assume that the failed CI is a false negative.
This because of the nature of the changes, and the fact that no other PR to this repo has passed CI.

PR #115, opened 2019-03-28
[nrf temphack] cortex_m: linker: Add ext_abi section

oyvindronningstad:
This will be changed when linker snippets (PR #109) is merged.

This change is required for https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/556, because it seems unlikely that #109 will be merged in the near future. If and when #109 is merged, this change is no longer needed.

### DIFF
--- a/include/arch/arm/cortex_m/scripts/linker.ld
+++ b/include/arch/arm/cortex_m/scripts/linker.ld
@@ -252,6 +252,11 @@ SECTIONS
 #include <linker/priv_stacks-rom.ld>
 #include <linker/kobject-rom.ld>
 
+	. = ALIGN(4);
+	_ext_abis_start = .;
+	KEEP(*(.ext_abis))
+	_ext_abis_size = ABSOLUTE((. - _ext_abis_start) / 4);
+
 	/*
 	 * For XIP images, in order to avoid the situation when __data_rom_start
 	 * is 32-bit aligned, but the actual data is placed right after rodata


oyvindronningstad:
cc @hakonfam @carlescufi

hakonfam:
Could you add some rationale for this change in this PR, so its more visible?

oyvindronningstad:
> Could you add some rationale for this change in this PR, so its more visible?

Yep, edited top comment now.

oyvindronningstad:
cc @sigvartmh

sigvartmh:
> cc @sigvartmh

Any specific reason we can't just keep this in a custom section (and give a linker path in CMake) until #109 is merged? Then use `select CUSTOM_SECTIONS_LD`  on secure boot in Kconfig?

There are pros and cons with both ways but I guess this way has the least overhead when changing back(git revert should work fine for both cases). So I'm not really opposed to it I just wondered if it could be solved another way so we don't have to diverge from the zephyr linker.ld downstream too much.

oyvindronningstad:
> Any specific reason we can't just keep this in a custom section (and give a linker path in CMake) until #109 is merged? Then use `select CUSTOM_SECTIONS_LD` on secure boot in Kconfig?

CUSTOM_SECTIONS_LD works poorly for this purpose since the custom_sections.ld file has to be placed in the correct directory depending on the sample, and it will collide with samples that use custom_sections.ld for something else.

sigvartmh:
> > Any specific reason we can't just keep this in a custom section (and give a linker path in CMake) until #109 is merged? Then use `select CUSTOM_SECTIONS_LD` on secure boot in Kconfig?
> 
> CUSTOM_SECTIONS_LD works poorly for this purpose since the custom_sections.ld file has to be placed in the correct directory depending on the sample, and it will collide with samples that use custom_sections.ld for something else.
> 
> Note that the sections need to be in both the app and the bootloader.

Yes, the biggest problem is conflicts. The 2nd part could be fixed by including a shared directory with `custom-sections.ld` but I won't push further on this and I'm fine with this solution 👍

carlescufi:
@oyvindronningstad any reason this cannot go upstream?

sigvartmh:
> @oyvindronningstad any reason this cannot go upstream?

@carlescufi : As far as I can see there is no reason for it not going upstream except that it doesn't make sens to put it upstream. However there is already a lot of board/soc specific linker stuff upstream. However we need this in by D4.

oyvindronningstad:
> @oyvindronningstad any reason this cannot go upstream?

In short: This isn't needed if we have #109.

sigvartmh:
@oyvindronningstad fix CI check(basically rebase onto master)

oyvindronningstad:
> @oyvindronningstad fix CI check(basically rebase onto master)

Ok, thanks, rebased now.

oyvindronningstad:
The failure seems unrelated.

oyvindronningstad:
@carlescufi ready to merge if CI failure is unrelated

carlescufi:
@SebastianBoe can you please take a look and merge?

oyvindronningstad:
> When this conflicts it will be useful for the person resolving the conflict if there was a comment explaining how long this patch is expected to live (including which patch is expected to allow it to be removed).
> 
> And a oneliner explaining what _ext_abis is.

Updated the commit message now.

PR #119, opened 2019-04-02
cmake: automatically board change from 'xns' to 'x'

hakonfam:
This is done so that sub-images of a non-secure image
is built as secure.

This is needed for https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/613

### DIFF
--- a/cmake/app/boilerplate.cmake
+++ b/cmake/app/boilerplate.cmake
@@ -247,7 +247,18 @@ else() # NOT FIRST_BOILERPLATE_EXECUTION
 
   # Have the child image select the same BOARD that was selected by
   # the parent.
-  set(BOARD ${CACHED_BOARD})
+  # Unless parent was "ns" in which case we assume that the child images are
+  # all secure, and should be build using the secure version of the board.
+  # It is assumed that only the root app will be built as non-secure.
+  # This is not a valid assumption as there might be multiple non-secure
+  # images defined. With this technique, it is not possible to have multiple
+  # images defined as non-secure.
+  # TODO: Allow multiple non-secure images by using Kconfig to set the
+  # secure/non-secure property rather than using a separate board definition.
+  if(${BOARD} STREQUAL nrf9160_pca10090ns)
+    set(BOARD nrf9160_pca10090)
+    message("Changed board to secure nrf9160_pca10090 (NOT NS)")
+  endif()
 
   unset(${IMAGE}DTC_OVERLAY_FILE)
   if(EXISTS              ${APPLICATION_SOURCE_DIR}/${BOARD}.overlay)


hakonfam:
CI Failure is expected, check the other PRs to this repo.

cmake/app/boilerplate.cmake
```diff
@@ -247,7 +247,15 @@ else() # NOT FIRST_BOILERPLATE_EXECUTION
 
   # Have the child image select the same BOARD that was selected by
   # the parent.
+  # Unless parent was "ns" in which case we assume that the child images are
+  # all are secure, and should be build using the secure version of the board.
   set(BOARD ${CACHED_BOARD})
+  if (${BOARD} MATCHES  ".*ns")
+    string(LENGTH ${BOARD} len)
+    MATH(EXPR len "${len}-2")
+    string(SUBSTRING ${BOARD} 0 ${len} BOARD)
```
hakonfam: If there is a cleaner way to do this I will be very happy :D

carlescufi:
Missing sauce tag `[nrf ]`

hakonfam:
> Missing sauce tag `[nrf ]`

fixed

cmake/app/boilerplate.cmake
SebastianBoe: I believe the cleanest solution would be to not have two boards.

But this is out-of-scope for this release.

Not sure what the best temporary solution would be ...

cmake/app/boilerplate.cmake
```diff
@@ -247,7 +247,15 @@ else() # NOT FIRST_BOILERPLATE_EXECUTION
 
   # Have the child image select the same BOARD that was selected by
   # the parent.
+  # Unless parent was "ns" in which case we assume that the child images are
+  # all are secure, and should be build using the secure version of the board.
```
SebastianBoe: We need a TODO here explaining what we are assuming, why this is an invalid assumption, and how we intend to re-write this in the future.

cmake/app/boilerplate.cmake
hakonfam: Do you have a proposal for how to do the substring operation "remove two from end" without taking the length of the string first?

cmake/app/boilerplate.cmake
hakonfam: will add

cmake/app/boilerplate.cmake
SebastianBoe:
```
if(${BOARD} STREQUAL nrf9160_pca10090ns)
set(BOARD nrf9160_pca10090)
endif()

```

If this code lives long enough to become a portability issue we deserve the pain of having to fix it.

cmake/app/boilerplate.cmake
hakonfam: check it now

cmake/app/boilerplate.cmake
```diff
@@ -247,7 +247,21 @@ else() # NOT FIRST_BOILERPLATE_EXECUTION
...
+  # It is assumed that only the root app will be built as non-secure.
+  # This is not a valid assumption as there might be multiple non-secure
+  # images defined. With this technique, it is not possible to have multiple
+  # images defined as non-secure.
+  # TODO: Allow multiple non-secure images by using Kconfig to set the
+  # secure/non-secure property rather than using a separate board definition.
   set(BOARD ${CACHED_BOARD})
+  if (${BOARD} MATCHES  ".*ns")
```
SebastianBoe: Won't this break CI if we integration test Zephyr with qemu_xtensa ?

hakonfam: does it end with "ns" ? I think it ends with "nsa"

cmake/app/boilerplate.cmake
hakonfam: I agree, changed it to this now.

SebastianBoe:
I'm unable to test this. Master revision of zephyr/samples/mcuboot appears to be broken.

Will try to get it fixed and then re-test.

SebastianBoe:
I suspect that we want to move this code into zephyr_add_executable. It'l give us nearly the same semantics, but will cause fewer conflicts with upstream (zephyr_add_executable isn't upstream yet, but boilerplate.cmake is).

hakonfam:
> I suspect that we want to move this code into zephyr_add_executable. It'l give us nearly the same semantics, but will cause fewer conflicts with upstream (zephyr_add_executable isn't upstream yet, but boilerplate.cmake is).

AFAICS this is not trivial, since the location of this change needs to be correct. Note that `zephyr_add_executable` is a function which is called from the image app CmakeLists.txt, which again calls `add_subdirectory(zephyr boilerplate cmakelists.txt)`. In other words, simply setting the `${BOARD}` - cmake variable will not suffice. 

One option would be to use some CACHE variable, but IMO that will cause a bigger diff compared to upstream than what we currently have.

hakonfam:
> I'm unable to test this. Master revision of zephyr/samples/mcuboot appears to be broken.
> 
> Will try to get it fixed and then re-test.

Follow these steps to have it build:
Check out nrf repo  https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/613  
west update
cd zephyr
cherry-pick https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/124

SebastianBoe:
> Follow these steps to have it build:
> Check out nrf repo NordicPlayground/fw-nrfconnect-nrf#613
> west update
> cd zephyr
> cherry-pick #124 

I'm getting this error after following those steps:

```
sebo@mach:~/ncs/zephyr/samples/mcuboot$ cb
-- Using application from '/home/sebo/ncs/zephyr/samples/mcuboot'
Zephyr version: 1.14.0
-- Found PythonInterp: /home/sebo/bin/python3.4 (found suitable version "3.5.2", minimum required is "3.4") 
-- Selected BOARD nrf52840_pca10056
-- Found west: /home/sebo/.local/bin/west (found suitable version "0.5.7", minimum required is "0.5.6")
-- Cache files will be written to: /home/sebo/.cache/zephyr
-- Loading /home/sebo/ncs/zephyr/boards/arm/nrf52840_pca10056/nrf52840_pca10056.dts as base
-- Overlaying /home/sebo/ncs/zephyr/dts/common/common.dts
nrf52840_pca10056.dts.pre.tmp:312.23-315.5: Warning (simple_bus_reg): /soc/virtualcom: missing or empty reg/ranges property
Parsing Kconfig tree in /home/sebo/ncs/zephyr/Kconfig
Loading /home/sebo/ncs/zephyr/boards/arm/nrf52840_pca10056/nrf52840_pca10056_defconfig as base
Merging /home/sebo/ncs/zephyr/samples/mcuboot/prj.conf
Configuration written to '/home/sebo/ncs/zephyr/samples/mcuboot/b/zephyr/.config'
-- The C compiler identification is GNU 8.2.1
-- The CXX compiler identification is GNU 8.2.1
-- The ASM compiler identification is GNU
-- Found assembler: /home/sebo/Downloads/gcc-arm-none-eabi-8-2018-q4-major/bin/arm-none-eabi-gcc
-- Performing Test toolchain_is_ok
-- Performing Test toolchain_is_ok - Success
Including module: nrf in path: /home/sebo/ncs/nrf
Including module: mcuboot in path: /home/sebo/ncs/mcuboot/zephyr
-- Using application from '/home/sebo/ncs/mcuboot/boot/zephyr'
Zephyr version: 1.14.0
CMake Error at /home/sebo/ncs/zephyr/cmake/kconfig.cmake:132 (message):
  File not found:
  /home/sebo/ncs/zephyr/boards/arm/nrf52840_pca10056/qemu_x86_defconfig
  /home/sebo/ncs/zephyr/cmake/app/boilerplate.cmake:487 (include)
  /home/sebo/ncs/mcuboot/boot/zephyr/CMakeLists.txt:51 (include)


-- Configuring incomplete, errors occurred!
See also "/home/sebo/ncs/zephyr/samples/mcuboot/b/CMakeFiles/CMakeOutput.log".
See also "/home/sebo/ncs/zephyr/samples/mcuboot/b/CMakeFiles/CMakeError.log".

```

hakonfam:
This issue was introduced when mcuboot started setting its BOARD by default. 

Luckily, there is a comment stating that it should only do this unless the board is already set, which they currently don't check. So I will create a PR fixing that.

cmake/app/boilerplate.cmake
```diff
@@ -247,7 +247,18 @@ else() # NOT FIRST_BOILERPLATE_EXECUTION
 
   # Have the child image select the same BOARD that was selected by
   # the parent.
-  set(BOARD ${CACHED_BOARD})
+  # Unless parent was "ns" in which case we assume that the child images are
+  # all are secure, and should be build using the secure version of the board.
```
ioannisg: remove double "are"

cmake/app/boilerplate.cmake
```diff
@@ -247,7 +247,18 @@ else() # NOT FIRST_BOILERPLATE_EXECUTION
...
+  # It is assumed that only the root app will be built as non-secure.
+  # This is not a valid assumption as there might be multiple non-secure
+  # images defined. With this technique, it is not possible to have multiple
+  # images defined as non-secure.
+  # TODO: Allow multiple non-secure images by using Kconfig to set the
```
ioannisg: BTW, isnt' the BSDLIB built as non-secure, too?

SebastianBoe: bsdlib is an app library, isn't it?

ioannisg: Yes, as far as I know.

SebastianBoe: Then it will be built as non-secure in the app image.

ioannisg: thanx

PR #120, opened 2019-04-02
scrips: cmake: subsys: introduce partition manager

hakonfam:
Partition Manager is a component which uses yaml files to
resolve flash placement with a wholistic view of the
device


hakonfam:
recheck

hakonfam:
TODO Is there a way to avoid using `PARTITION_MANAGER_TARGET_PROPER` ?

carlescufi:
Missing a sauce tag `[nrf ]`

hakonfam:
> Missing a sauce tag `[nrf ]`

fixed, from now on I will try really hard to not forget this, as it seems that I am not capable of not forgetting :)

cmake/app/boilerplate.cmake
```diff
@@ -532,6 +534,84 @@ set_property(TARGET ${IMAGE}app PROPERTY ARCHIVE_OUTPUT_DIRECTORY ${IMAGE}app)
...
+    PARTITION_MANAGER_CONFIG_FILES
+    )
+  print(partition_manager_config_files)
+  if(partition_manager_config_files)
+    get_property(
+      partition_manager_config_targets
+      GLOBAL PROPERTY
+      PARTITION_MANAGER_CONFIG_TARGETS
+      )
+    file(GLOB_RECURSE autoconf_files "${PROJECT_BINARY_DIR}/**/autoconf.h")
```
SebastianBoe: CMAKE_BINARY_DIR

SebastianBoe: globbing like this should be avoided.

Better to iterate over image build dirs.

scripts/partition_manager.py
```diff
@@ -0,0 +1,328 @@
...
+''' % (hex(conf['address']), hex(conf['size'])))
+
+
+def get_flash_size(config):
+    config.seek(0)  # Ensure that we search the entire file
+    for line in config.readlines():
+        match = re.match(r'#define CONFIG_FLASH_SIZE (\d*)', line)
+        if match:
+            return int(match.group(1)) * 1024
+    raise RuntimeError("Unable to find 'CONFIG_FLASH_SIZE' in any of: %s" % config.name)
```
SebastianBoe: @mbolivar : I'd like to use west's BuildConfiguration instead of what we are doing here.

How can we import BuildConfiguration from here?

scripts/partition_manager.py
```diff
@@ -0,0 +1,328 @@
...
+        open(path.join(adr_map[area_name]['out_dir'], pm_config_file), 'w').write('\n'.join(lines))
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Parse given input configurations and generate override header files.",
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+
+    parser.add_argument("-i", "--input", type=argparse.FileType('r', encoding='UTF-8'), nargs="+",
+                        help="List of JSON formatted config files. See tests in this file for examples.")
```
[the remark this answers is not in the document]
hakonfam: It is an abbreviation. The official documentation uses JSON, IMO it should be this way when mentioning it in a natural language.

scripts/partition_manager.py
```diff
@@ -0,0 +1,328 @@
...
+    for area_name, area_props in adr_map.items():
+        if 'out_path' not in area_props.keys() or 'out_dir' not in area_props.keys():
+            continue  # The 'image' being inspected is not an executable, just a place-holder.
+        # TODO replace 'out_dir' with 'out_path' and change the semantics of the latter
+        open(path.join(adr_map[area_name]['out_dir'], pm_config_file), 'w').write('\n'.join(lines))
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Parse given input configurations and generate override header files.",
```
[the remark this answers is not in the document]
hakonfam: Updated docs

cmake/app/boilerplate.cmake
hakonfam: why should it be avoided? How should I populate the list of all autoconf.h files if not by globbing?

scripts/partition_manager.py
hakonfam: Instead of parsing the kconfig file?

IMO it should be kept like this because 1 - we are already parsing this file, 2 - there exists a set of people which prefer not to keep west up to date, or think about west at all, and I don't see any benefit of mixing west in to the partition manager here.

mbolivar: > How can we import BuildConfiguration from here?

The `runners` package just needs to be on `sys.path`.

> IMO it should be kept like this because 1 - we are already parsing this file, 2 - there exists a set of people which prefer not to keep west up to date, or think about west at all, and I don't see any benefit of mixing west in to the partition manager here.

The `runners` package is used by the west extension commands provided with Zephyr, but it isn't part of west at all. It is automatically kept up to date because it's in the zephyr repository in `zephyr/scripts/west_commands/runners`. We could promote it to `scripts` to use it elsewhere at only a slight extra cost of having to manipulate `sys.path` to get at it in the west extensions.

So this doesn't require keeping west up to date, or for people to install west, or mixing west into the partition manager.

mbolivar: Wait, no, I am totally lying: the runners package does import west packages, even though it's not part of west. Sorry; I keep forgetting that.

However, we could factor out the common bits from `runners.core` that other scripts in zephyr want to use, and I think that would be a good idea. Thoughts?

scripts/partition_manager.py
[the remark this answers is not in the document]
SebastianBoe: I thought that we were using yaml?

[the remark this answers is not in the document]
hakonfam: hhahahahahahahahahah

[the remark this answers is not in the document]
hakonfam: Fixed

scripts/partition_manager.py
SebastianBoe: > However, we could factor out the common bits from runners.core that other scripts in zephyr want to use, and I think that would be a good idea. Thoughts?

I agree. Sounds like this won't be ready for this PR though :(

SebastianBoe:
AFAICT what we are doing here is going into users of DT partition information and changing (under an ifdef of USE_PARTITION_MANAGER) to use Partition Manager information instead.

I have a pretty good idea, but for the written record, I would like to see something that describes why we can't use DT information directly.

Also, I would like to see some thoughts on possible strategies for co-existence between the partitioning information in DT and PM in both the short term and the long term.

e.g. some text that touches on these questions:

Should we replace (instead of adding an optional alternative implementation) the information in DT with partition manager information?

Should/can PM feed information into DT instead of having users choose between the two?

Should we introduce an abstract "partition information" concept that is either implemented by DT or PM? (Then users would only need to be ported to use this concept, instead of having to do the multiplexing themselves).

How do we explain to users that the partition information in DT is dead code?

etc.

hakonfam:
Not a complete answer to your query @SebastianBoe , but some thoughts:

In the multi image context the partitioning of flash depends on kconfig heavily (include SPM, include MCUBOOT, include B0 etc). My goal with partition manager is to replace the partitioning information currently encoded in DT. 

> I would like to see something that describes why we can't use DT information directly.

Using DT directly for partitioning does not scale. AFAICS we have to replace it with something which can depend on Kconfig.

>  I would like to see some thoughts on possible strategies for co-existence between the partitioning information in DT and PM in both the short term and the long term.

Short term and long term: Provide partitioning information in partition manager, not in DT.

> Should we replace (instead of adding an optional alternative implementation) the information in DT with partition manager information?

IMO PM should replace the partitioning information currently found in DT.

> Should/can PM feed information into DT instead of having users choose between the two?

I don't see how we can do this as PM depends heavily on kconfig (as it must IMO). The user should not have a choice when it comes to partitioning info, and we should remove all partitioning info in the DT source files.


> Should we introduce an abstract "partition information" concept that is either implemented by DT or PM? (Then users would only need to be ported to use this concept, instead of having to do the multiplexing themselves)

I don't see how partitioning information can be contained in DT as it depends on the application configuration.

> How do we explain to users that the partition information in DT is dead code?

We should remove the partitioning information in DT.

carlescufi:
Please rebase on top of the current master

carlescufi:
Fix typo in commit message: "scrips"

SebastianBoe:
> I don't see how we can do this as PM depends heavily on kconfig (as it must IMO).

Noted. We can discard this option then.

Intuitively, one would think PM can not exist in a fork long term. I understand that we
are going to use the same strategy for PM as for multi-image? A temporary fork.

hakonfam:
> Intuitively, one would think PM can not exist in a fork long term. I understand that we
are going to use the same strategy for PM as for multi-image? A temporary fork

My understanding after discussing this in a zephyr dev-meeting was that they acknowledged that "something" like PM is needed, but that this "something" would need a lot of planning, and be able to solve a lot of complex use cases. 

Anyway, AFAICS we need this downstream for now, as we do with multi-image as you say :)

SebastianBoe:
Running samples/mcuboot in a loop triggers a failure every so often.

```
FAILED: zephyr/mcuboot/mcuboot/zephyr/linker.cmd 
cd /home/sebo/ncs/zephyr/samples/mcuboot/b/zephyr/mcuboot/mcuboot/zephyr && /home/sebo/zephyr-sdk/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc -x assembler-with-cpp -undef -MD -MF linker.cmd.dep -MT zephyr/mcuboot/mcuboot/zephyr/linker.cmd -I/home/sebo/ncs/zephyr/kernel/include -I/home/sebo/ncs/zephyr/arch/arm/include -I/home/sebo/ncs/zephyr/include -I/home/sebo/ncs/zephyr/include/drivers -I/home/sebo/ncs/zephyr/samples/mcuboot/b/zephyr/mcuboot/mcuboot/zephyr/include/generated -I/home/sebo/ncs/zephyr/soc/arm/nordic_nrf/nrf52 -I/home/sebo/ncs/zephyr/soc/arm/nordic_nrf/include -I/home/sebo/zephyr-sdk-0.10.0/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/8.3.0/include -I/home/sebo/zephyr-sdk-0.10.0/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/8.3.0/include-fixed -I/home/sebo/ncs/zephyr/lib/libc/minimal/include -I/home/sebo/ncs/zephyr/ext/hal/cmsis/Include -I/home/sebo/ncs/zephyr/ext/hal/nordic/nrfx -I/home/sebo/ncs/zephyr/ext/hal/nordic/nrfx/drivers/include -I/home/sebo/ncs/zephyr/ext/hal/nordic/nrfx/hal -I/home/sebo/ncs/zephyr/ext/hal/nordic/nrfx/mdk -I/home/sebo/ncs/zephyr/ext/hal/nordic/. -I/home/sebo/ncs/nrf/include -D__GCC_LINKER_CMD__ -DUSE_PARTITION_MANAGER -E /home/sebo/ncs/zephyr/soc/arm/nordic_nrf/nrf52/linker.ld -P -o linker.cmd
In file included from /home/sebo/ncs/zephyr/soc/arm/nordic_nrf/nrf52/linker.ld:9:
/home/sebo/ncs/zephyr/include/arch/arm/cortex_m/scripts/linker.ld:21:10: fatal error: override.h: No such file or directory
 #include <override.h>

```

```
FAILED: zephyr/mcuboot/mcuboot/zephyr/mcuboot_zephyr_prebuilt.elf 
: && ccache /home/sebo/zephyr-sdk/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc    zephyr/mcuboot/mcuboot/zephyr/CMakeFiles/mcuboot_zephyr_prebuilt.dir/misc/empty_file.c.obj  -o zephyr/mcuboot/mcuboot/zephyr/mcuboot_zephyr_prebuilt.elf  -T zephyr/mcuboot/mcuboot/zephyr/linker.cmd -Wl,-Map=/home/sebo/ncs/zephyr/samples/mcuboot/b/zephyr/mcuboot/mcuboot/zephyr/zephyr.map -u_OffsetAbsSyms -u_ConfigAbsSyms -Wl,--whole-archive zephyr/mcuboot/mcuboot/mcuboot_app/libmcuboot_app.a zephyr/mcuboot/mcuboot/zephyr/libmcuboot_zephyr.a zephyr/mcuboot/mcuboot/zephyr/arch/arm/core/libmcuboot_arch__arm__core.a zephyr/mcuboot/mcuboot/zephyr/arch/arm/core/cortex_m/libmcuboot_arch__arm__core__cortex_m.a zephyr/mcuboot/mcuboot/zephyr/arch/arm/core/cortex_m/mpu/libmcuboot_arch__arm__core__cortex_m__mpu.a zephyr/mcuboot/mcuboot/zephyr/lib/libc/minimal/libmcuboot_lib__libc__minimal.a zephyr/mcuboot/mcuboot/zephyr/ext/lib/crypto/mbedtls/libmcuboot_ext__lib__crypto__mbedtls.a zephyr/mcuboot/mcuboot/zephyr/drivers/gpio/libmcuboot_drivers__gpio.a zephyr/mcuboot/mcuboot/zephyr/drivers/flash/libmcuboot_drivers__flash.a zephyr/mcuboot/mcuboot/zephyr/drivers/serial/libmcuboot_drivers__serial.a -Wl,--no-whole-archive zephyr/mcuboot/mcuboot/zephyr/kernel/libmcuboot_kernel.a zephyr/mcuboot/mcuboot/zephyr/CMakeFiles/mcuboot_offsets.dir/arch/arm/core/offsets/offsets.c.obj -L"/home/sebo/zephyr-sdk-0.10.0/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/8.3.0/thumb/v7e-m/nofp" -L/home/sebo/ncs/zephyr/samples/mcuboot/b/zephyr/mcuboot/mcuboot/zephyr -lgcc -Wl,--print-memory-usage -mthumb -mcpu=cortex-m4 -nostdlib -static -no-pie -Wl,-X -Wl,-N -Wl,--gc-sections -Wl,--build-id=none -Wl,--sort-common=descending -Wl,--sort-section=alignment -Wl,--orphan-handling=warn -mabi=aapcs -march=armv7e-m && :
Memory region         Used Size  Region Size  %age Used
           FLASH:       34148 B        16 KB    208.42%
            SRAM:       23448 B       256 KB      8.94%
        IDT_LIST:          72 B         2 KB      3.52/home/sebo/zephyr-sdk-0.10.0/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/8.3.0/../../../../arm-zephyr-eabi/bin/ld: zephyr/mcuboot/mcuboot/zephyr/mcuboot_zephyr_prebuilt.elf section `_TEXT_SECTION_NAME_2' will not fit in region `FLASH'
/home/sebo/zephyr-sdk-0.10.0/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/8.3.0/../../../../arm-zephyr-eabi/bin/ld: region `FLASH' overflowed by 17764 bytes
collect2: error: ld returned 1 exit status
%

```

scripts/partition_manager.py
```diff
@@ -0,0 +1,356 @@
...
+
+    for area_name, area_props in adr_map.items():
+        lines.append("#define PM_%s_ID %d" % (area_name.upper(), area_props['flash_area_id']))
+    lines.append("#define PM_NUM %d" % flash_area_id)
+
+    lines.append("\n/* Direct look up list of flash areas */")
+    for area_name, area_props in sorted(adr_map.items(), key=lambda key_value: key_value[1]['address']):
+        lines.append("#define PM_%s_ADDRESS 0x%x" % (area_name.upper(), area_props['address']))
+        lines.append("#define PM_%s_SIZE 0x%x" % (area_name.upper(), area_props['size']))
+        lines.append("#define PM_%s_DEV_NAME \"NRF_FLASH_DRV_NAME\"" % area_name.upper())
```
SebastianBoe: NRF_FLASH_DRV_NAME

Any plans for how to make this portable in the future?

scripts/partition_manager.py
```diff
@@ -0,0 +1,353 @@
...
+    lines.append(get_header_guard_start(pm_config_file))
+    flash_area_id = 0
+
+    lines.append("\n/* Indirect, iterable list of flash areas */")
+    for area_name, area_props in sorted(adr_map.items(), key=lambda key_value: key_value[1]['address']):
+        lines.append("#define PM_%d_LABEL %s" % (flash_area_id, area_name.upper()))
+        adr_map[area_name]['flash_area_id'] = flash_area_id
+        flash_area_id += 1
+
+    for area_name, area_props in adr_map.items():
```
SebastianBoe: How come the other two are sorted, but not this one?

hakonfam: IMO these should be sorted by ID.

scripts/partition_manager.py
hakonfam: No, I will investigate why this is needed.

CMakeLists.txt
```diff
@@ -461,6 +461,23 @@ function(construct_add_custom_command_for_linker_pass linker_output_name output_
...
+      set_property(GLOBAL PROPERTY PROPERTY_PARTITION_MANAGER_TARGET
+        PARTITION_MANAGER_TARGET)
+      # Duplicates are removed before this list is used.
+      set_property(GLOBAL APPEND PROPERTY
+        ${IMAGE}PROPERTY_LINKER_SCRIPT_DEFINES
+        -DUSE_PARTITION_MANAGER)
+      # Below is needed to ensure that the origin app is placed correctly.
+      set_property(GLOBAL APPEND PROPERTY
+        PROPERTY_LINKER_SCRIPT_DEFINES
+        -DUSE_PARTITION_MANAGER)
```
SebastianBoe: What's going on here?

won't IMAGE evaluate to '' when the app is processed and make the second invocation redundant?

CMakeLists.txt
```diff
@@ -461,6 +461,23 @@ function(construct_add_custom_command_for_linker_pass linker_output_name output_
     set(linker_pass_define -DLINKER_PASS2)
   elseif (${linker_output_name} MATCHES "^linker_app_smem_unaligned$")
     set(linker_pass_define -DLINKER_APP_SMEM_UNALIGNED)
+  elseif (${linker_output_name} MATCHES "^linker$")
+    if(EXISTS ${APPLICATION_SOURCE_DIR}/pm.yml)
+      zephyr_compile_definitions(USE_PARTITION_MANAGER)
```
SebastianBoe: Why is this code placed here of all places?

CMakeLists.txt
```diff
@@ -461,6 +461,23 @@ function(construct_add_custom_command_for_linker_pass linker_output_name output_
...
+        PARTITION_MANAGER_TARGET)
+      # Duplicates are removed before this list is used.
+      set_property(GLOBAL APPEND PROPERTY
+        ${IMAGE}PROPERTY_LINKER_SCRIPT_DEFINES
+        -DUSE_PARTITION_MANAGER)
+      # Below is needed to ensure that the origin app is placed correctly.
+      set_property(GLOBAL APPEND PROPERTY
+        PROPERTY_LINKER_SCRIPT_DEFINES
+        -DUSE_PARTITION_MANAGER)
+      add_dependencies(${OFFSETS_LIB} PARTITION_MANAGER_TARGET)
```
SebastianBoe: move this to the common "OFFSETS_LIB" dependencies.

CMakeLists.txt
SebastianBoe: Move this out of this function.

hakonfam:
In progress: Remove override.h and only generate pm_config.h file and use info from this to set ROM_ADDR and ROM_SIZE correctly.

cmake/app/boilerplate.cmake
hakonfam: Refactored the script to not need these files, so this is now removed as requested.

hakonfam:
Note: I added a change to `Kconfig.zephyr`. Please review this as well.

scripts/partition_manager.py
```diff
@@ -0,0 +1,406 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+
+import argparse
+import yaml
+import re
+from os import path
+
+
+def remove_item_not_in_list(list_to_remove_from, list_to_check):
```
SebastianBoe: items, not item.

And it does more than this, since it is 'app' aware, so this abstraction isn't really working. This function should be inlined I suspect.

cmake/app/boilerplate.cmake
```diff
@@ -535,6 +540,34 @@ if(CONFIG_QEMU_TARGET)
...
+  string(LENGTH ${IMAGE} len)
+  MATH(EXPR len "${len}-1")
+  string(SUBSTRING ${IMAGE} 0 ${len} image_name)
+  set_property(
+    GLOBAL APPEND PROPERTY
+    PARTITION_MANAGER_CONFIG_FILES
+    "${image_name}:${PROJECT_BINARY_DIR}/include/generated/pm.yml:${PROJECT_BINARY_DIR}:${PROJECT_BINARY_DIR}/include/generated"
+    )
+endif()
+
```
SebastianBoe: This hunk should be moved to nrf/subsys/partition_manager

cmake/app/boilerplate.cmake
```diff
@@ -543,6 +549,45 @@ set_property(TARGET ${IMAGE}app PROPERTY ARCHIVE_OUTPUT_DIRECTORY ${IMAGE}app)
...
+    )
+  if(partition_manager_config_files)
+    set(pm_cmd
+      COMMAND
+      ${PYTHON_EXECUTABLE}
+      ${ZEPHYR_BASE}/scripts/partition_manager.py
+      --input ${partition_manager_config_files}
+      --app-pm-config-dir ${PROJECT_BINARY_DIR}/include/generated
+      )
+    execute_process(${pm_cmd})
```
SebastianBoe: This is bad-practice, the variable 'pm_cmd' should be inlined.

cmake/app/boilerplate.cmake
hakonfam: moved

PR #121, opened 2019-04-03
[nrf noup] jenkinsfile: Multiple fixes

carlescufi:
See commit messages

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,6 @@
 def IMAGE_TAG = "ncs-toolchain:1.08"
 def REPO_CI_TOOLS = "https://github.com/zephyrproject-rtos/ci-tools.git"
+def REPO_CI_TOOLS_SHA = "9f4dc0be401c2b1e9b1c647513fb996bd8abd057"
 
 pipeline {
   agent {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -25,19 +26,28 @@ pipeline {
...
       // ENVs for sanitycheck
       ARCH = "-a arm"
-      SANITYCHECK_OPTIONS_ALL = "--inline-logs --enable-coverage -N"
+      SANITYCHECK_OPTIONS = "--inline-logs --enable-coverage -N"
+      SANITYCHECK_RETRY = "--only-failed --outdir=out-2nd-pass"
+      SANITYCHECK_RETRY_2 = "--only-failed --outdir=out-3rd-pass"
 
       // ENVs for building (triggered by sanitycheck)
-      ZEPHYR_TOOLCHAIN_VARIANT = 'gnuarmemb'
+      ZEPHYR_TOOLCHAIN_VARIANT = 'zephyr'
```
SebastianBoe: Are we certain that the zephyr SDK can be used with all platforms?

e.g. nrf91 ?

carlescufi: Yes, since 0.10.0

metc: Nice. We should also move to `sdk-ng` on our local machines then.
https://github.com/zephyrproject-rtos/sdk-ng/releases/tag/v0.10.0

carlescufi: See the list of tests built by sanitycheck in the logs, there's quite a few for nrf91

carlescufi: @metc @SebastianBoe I don't think we should move to the SDK ourselves, we should stick to GNU Arm Embedded. There are a few reasons:

- There's no SDK in Windows
- I have no plans to switch the fw-nrfconnect-nrf CI to the SDK, since we want to standardize on a single compiler and Windows only has GNU Arm Embedded
- The only reason I have switched here is that we already test this tree in CI in fw-nrfconnect-nrf, and this particular CI here is to verify we are still compatible with upstream.

metc: Ack. Understood. Thanks.

PR #122, opened 2019-04-03
[nrf temphack] bootloader: cmake: Fix kernel_elf target for multi-image

sigvartmh:
This will make sure that the `${IMAGE}kernel_elf` target is created for
secure bootloader(b0).


SebastianBoe:
This doesn't look right.

Why does b0 require kernel_elf?

I wonder if b0 should be using logical_target_for_zephyr_elf instead.

hakonfam:
I agree @SebastianBoe , we will try to come up with something better

carlescufi:
Please rebase on top of the current master

sigvartmh:
Also agree :) hence [temphack] but I'll look with Håkon for another solution

PR #124, opened 2019-04-05
cmake: re-apply patch needed for multi image

hakonfam:
Without this patch multi image builds will fail
as the modules' build directories are the same
for each image, resulting in an error.

With this patch, each module build dir is placed
within each images build dir, making them unique.


Qbicz:
Will this be taken into account in manifest?
I reproduce the issue on NCS master, when trying to build nrf_desktop with MCUBoot support.

SebastianBoe:
I'm doing a manifest PR now.

SebastianBoe:
https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/634

hakonfam:
Sorry, forgot about the manifest pr :]

PR #126, opened 2019-04-10
drivers: nrf: Fix UARTE interrupt driven operation.

Mierunski:
Bad order of interrupt calls made it unable to transmitt data
using interrupt driven API.

Signed-off-by: Mieszko Mierunski <mieszko.mierunski@nordicsemi.no>


carlescufi:
This can be closed now I think since it's merged upstream?

PR #129, opened 2019-04-12
[nrf noup] scripts: cmake: partition_manager: Bug fix for windows

sigvartmh:
This is coupled with
https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/657
where a split on ':' would result in a split on 'c:\' on windows.
Changing the split deliminator to '|' will fix this problem as you
are not allowed to named folders on windows or linux to '|'.

### DIFF
--- a/scripts/partition_manager.py
+++ b/scripts/partition_manager.py
@@ -261,7 +261,7 @@ def write_pm_config(adr_map, app_output_dir):
         area_props['pm_config'].append("#define PM_SIZE 0x%x" % area_props['size'])
         area_props['pm_config'].insert(0, get_header_guard_start(pm_config_file))
         area_props['pm_config'].append(get_header_guard_end(pm_config_file))
-        
+
     # Store complete size/address configuration to all input paths
     for area_name, area_props in adr_map.items():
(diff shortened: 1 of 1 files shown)


SebastianBoe:
> Changing the split deliminator to '|' will fix this problem as you
> are not allowed to named folders on windows or linux to '|'.

Fun-fact. Was not aware, thanks!

stig-bjorlykke:
Are you sure about the ```|```? I'm able to create directories containing ```|``` on both Linux and macOS.

SebastianBoe:
:( Was it too good to be true?

sigvartmh:
wait what..

sigvartmh:
Will it actually be named `|` or will it be named `'|'` the directory ?

> Are you sure about the `|`? I'm able to create directories containing `|` on both Linux and macOS.

@stig-bjorlykke  Yup you are correct only `/` is the only illegal character in Linux and it wouldn't be very smart to use `/` as a splitting delimiter. Seems my assumption was incorrect as i tried `mkdir |`where what I should have tried was `mkdir "|"` and `touch "newfile|"`as both are valid use-cases.

PR #133, opened 2019-04-19
[nrf temphack] pm: Add partition manager support to 'settings'

SebastianBoe:
Add partition manager support to the settings subsystem.

### DIFF
--- a/subsys/settings/src/settings_fcb.c
+++ b/subsys/settings/src/settings_fcb.c
@@ -13,6 +13,18 @@
 #include "settings/settings_fcb.h"
 #include "settings_priv.h"
 
+#if USE_PARTITION_MANAGER
+
+#include <pm_config.h>
+#define FLASH_AREA_STORAGE_ID       PM_MCUBOOT_STORAGE_ID
+
+#else
+
+#include <generated_dts_board.h>
+#define FLASH_AREA_STORAGE_ID       DT_FLASH_AREA_STORAGE_ID
+
+#endif /* USE_PARTITION_MANAGER */
+
 #include <logging/log.h>
 LOG_MODULE_DECLARE(settings, CONFIG_SETTINGS_LOG_LEVEL);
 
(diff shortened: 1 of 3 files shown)


pdunaj:
With the change above settings still do not initialize. I checked that problem is error in one of the function calls initializing settings that causes panic (explicit panic on error).
The same problem is in other areas in zephyr - in our case `zephyr/subsys/settings/src/settings_fcb.c` also has invalid id.

subsys/settings/src/settings_init.c
```diff
@@ -59,7 +71,7 @@ static void settings_init_fcb(void)
 	int rc;
 	const struct flash_area *fap;
 
-	rc = flash_area_get_sectors(DT_FLASH_AREA_STORAGE_ID, &cnt,
+	rc = flash_area_get_sectors(FLASH_AREA_STORAGE_ID, &cnt,
```
pdunaj: Also needed in `zephyr/subsys/settings/src/settings_fcb.c` at line 45.

SebastianBoe:
> The same problem is in other areas in zephyr - in our case zephyr/subsys/settings/src/settings_fcb.c also has invalid id.

Fixed.

subsys/settings/src/settings_init.c
SebastianBoe: Fixed.

subsys/settings/src/settings_fcb.c
```diff
@@ -13,6 +13,18 @@
 #include "settings/settings_fcb.h"
 #include "settings_priv.h"
 
+#if USE_PARTITION_MANAGER
```
nvlsianpu: just why this is not incorporated into a header file?

SebastianBoe: It should be.

I'll add that to the partition manager TODO list.

nvlsianpu: :D

PR #135, opened 2019-04-25
[nrf toup] cmake: zephyr_modules: Add west executable argument

sigvartmh:
On some systems where you don't have access to `PATH` and you can't set the `ENV{PATH}` variable to what you need it to be. This means that the path where west is won't be passed down to the python script this should have been set explicit.

This is a bugfix to get NCS to work with SES on OSX

### DIFF
--- a/cmake/zephyr_module.cmake
+++ b/cmake/zephyr_module.cmake
@@ -20,12 +20,17 @@ endif()
 
 set(KCONFIG_MODULES_FILE ${CMAKE_BINARY_DIR}/Kconfig.modules)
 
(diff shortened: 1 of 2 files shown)


sigvartmh:
@SebastianBoe does [toup] make sense here? Also review if you have time :)

SebastianBoe:
> @SebastianBoe does [toup] make sense here? Also review if you have time :)

Yes

sigvartmh:
Upstream PR here https://github.com/zephyrproject-rtos/zephyr/pull/15666

PR #136, opened 2019-04-25
cmake: only unset CMAKE_C_COMPILER in first boilerplate execution

hakonfam:
The C compiler does not change between the different images.
Add a check to ensure that the C compiler variable is only
unset for the first boilerplate execution.

### DIFF
--- a/cmake/target_toolchain.cmake
+++ b/cmake/target_toolchain.cmake
@@ -38,9 +38,12 @@ endif()
 
 # The 'generic' compiler and the 'target' compiler might be different,
 # so we unset the 'generic' one and thereby force the 'target' to
-# re-set it.
-unset(CMAKE_C_COMPILER)
-unset(CMAKE_C_COMPILER CACHE)
+# re-set it. This is only needed for the first boilerplate execution as
+# the remaining boilerplate executions will use the same C compiler.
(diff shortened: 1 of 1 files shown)


cmake/target_toolchain.cmake
```diff
@@ -39,8 +39,10 @@ endif()
 # The 'generic' compiler and the 'target' compiler might be different,
 # so we unset the 'generic' one and thereby force the 'target' to
 # re-set it.
-unset(CMAKE_C_COMPILER)
-unset(CMAKE_C_COMPILER CACHE)
+if (NOT FIRST_BOILERPLATE_EXECUTION)
```
SebastianBoe: The above comment needs to be expanded to explain the if-statement.

hakonfam: Something line

```
# The 'generic' compiler and the 'target' compiler might be different,
# so we unset the 'generic' one and thereby force the 'target' to
# re-set it. This is only needed for the first boilerplate execution as
# the remaining boilerplate executions will use the same C compiler.
```
?

SebastianBoe: (Y)

SebastianBoe: LGTM

PR #138, opened 2019-04-29
Partition Manager - Sub-partition improvements

oyvindronningstad:
Add two new partition manager keywords:
 - 'inside': (inversion of 'span') Place this partition inside the span of the referenced partition.
 - 'share_size': (more flexible version of 'sub_partitions') Make this partition the same size as the referenced partition.

Both keywords take a list. The list works the same as for e.g 'after' where the first partition in the list that can be found in the project is selected.

Also added:
 - Support for nested sub-partitions.
 - Asserts on the tests, so they become "automatic".
 - "Pretty" printing of partition address and size.

Related PRs:
 - https://github.com/NordicPlayground/fw-nrfconnect-mcuboot/pull/32
 - https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/705


oyvindronningstad:
Changes in the last push:
 - Split into more commits.
 - Move printing into its own file and run it when running `ninja rom_report`.
 - partition_manager now writes an output YAML file which contains the merged contents of all pm.ymls as well as the addresses and sizes they were given. This file is parsed by the report printing script.
 - Some small refactoring in the file generation code.

oyvindronningstad:
> Missing sauce tags

Done. Also squashed, as discussed.

scripts/partition_manager.py
```diff
@@ -24,60 +25,112 @@ def item_is_placed(d, item, after_or_before):
...
+    # Get candidates
+    unsorted = [x for x in reqs.keys() if 'placement' in reqs[x] and type(reqs[x]['placement']) == dict
+                 and ('before' in reqs[x]['placement'].keys() or 'after' in reqs[x]['placement'].keys())]
+
+    # Sort sub_partitions by whether they are inside other sub_partitions. Innermost first.
+    sorted_subs = sorted(sub_partitions.values(), key=lambda x: outermost(sub_partitions, x), reverse=True)
+
+    # Sort candidates by whether they are part of a sub_partitions.
+    # sub_partition parts come last in the result list so they are more likely
+    # to end up being placed next to each other, since they are inserted last.
```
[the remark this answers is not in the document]
oyvindronningstad: If you have two partitions placed after the same third partition, you can have two orders:

    foo:
        placement: {after: app}

    bar
        placement: {after: app}

Order 1:

    app foo bar

Order 2:

    app bar foo

Which order ends up as the result is dependent on which of `foo` and `bar` is processed before the other. The one that is processed last, ends up immediately after the app.

Since PM fails if partitions in a span are not consecutive, this sorting minimizes the risk that PM fails in these ambiguous situations.

CODEOWNERS
```diff
@@ -259,6 +259,7 @@
 /scripts/sanity_chk/expr_parser.py        @andrewboie @nashif
 /scripts/gen_app_partitions.py            @andrewboie
 /scripts/partition_manager.py             @SebastianBoe
+/scripts/partition_manager_report.py      @SebastianBoe
```
hakonfam: You can change these to be you or me I guess. @SebastianBoe  was involved in the build code, but not the scripts themself. I am not sure if downstream code owners can be listed here though?

scripts/partition_manager_report.py
```diff
@@ -0,0 +1,71 @@
...
+    print("Partition Manager report")
+    args = parse_args()
+    with open(args.input, 'r') as f:
+        pm_config = yaml.safe_load(f)
+    min_address = min((part['address'] for part in pm_config.values() if 'address' in part))
+    max_address = max((part['address'] + part['size'] for part in pm_config.values() if 'address' in part))
+    print_region('FLASH', max_address - min_address, pm_config)
+
+if __name__ == "__main__":
+    main()
```
hakonfam: Missing newline

scripts/partition_manager.py
```diff
@@ -244,61 +269,49 @@ def fn(a, b):
...
             write_pm_config_to_file(path.join(area_props['out_dir'], pm_kconfig_file), config_lines)
 
-    # Store to root app
-    if not only_sub_image_is_being_built(adr_map, app_output_dir):
-        write_pm_config_to_file(path.join(app_output_dir, pm_kconfig_file), config_lines)
 
+def write_yaml_out_file(pm_config, path):
+    yamldump = yaml.dump(pm_config)
+    with open(path, 'w') as f:
+        f.write(yamldump)
```
hakonfam: pep8 add 2 newlines before parse_args() declaration

scripts/partition_manager.py
```diff
@@ -244,61 +269,49 @@ def fn(a, b):
...
-    for area_name, area_props in adr_map.items():
+    for _, area_props in adr_map.items():
         if 'out_dir' in area_props.keys():
             write_pm_config_to_file(path.join(area_props['out_dir'], pm_kconfig_file), config_lines)
 
-    # Store to root app
-    if not only_sub_image_is_being_built(adr_map, app_output_dir):
-        write_pm_config_to_file(path.join(app_output_dir, pm_kconfig_file), config_lines)
 
+def write_yaml_out_file(pm_config, path):
```
hakonfam: `path` is defined by python, use other keyword.

scripts/partition_manager_report.py
```diff
@@ -0,0 +1,71 @@
...
+        bcolors = bcolors_ansi
+
+    # Print header
+    print(bcolors["OKBLUE"] + "%s (0x%x):" % (region, size) + bcolors["ENDC"])
+
+    # Sort partitions three times:
+    #  1. On whether they are a container (has a 'span'), containers first.
+    #  2. On size, descending.
+    #  3. On address, ascending.
+    sorted_reqs = sorted(sorted(sorted(reqs.keys(), key=lambda x: int('span' in reqs[x]), reverse=True), key=lambda x: reqs[x]['size'], reverse=True), key=lambda x: reqs[x]['address'])
```
hakonfam: One liners are awesome. But this is surely illegal in some countries :D

IMO it would improve readability if this was split and inlined.

scripts/partition_manager_report.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Create text lines
+    sizes = ["%s%s (0x%x)" % ((("| " + bcolors["WARNING"] if 'span' not in reqs[part] else "+---" + bcolors["OKBLUE"]) + "0x%x: " % reqs[part]['address']), part, reqs[part]['size']) + bcolors["ENDC"]  for part in sorted_reqs]
```
hakonfam: This line is way too long.120 is max per pep8

scripts/partition_manager_report.py
```diff
@@ -0,0 +1,71 @@
...
+
+def main():
+    print("Partition Manager report")
+    args = parse_args()
+    with open(args.input, 'r') as f:
+        pm_config = yaml.safe_load(f)
+    min_address = min((part['address'] for part in pm_config.values() if 'address' in part))
+    max_address = max((part['address'] + part['size'] for part in pm_config.values() if 'address' in part))
+    print_region('FLASH', max_address - min_address, pm_config)
+
```
hakonfam: 2 lines before if check, same for main declaration

scripts/partition_manager.py
```diff
@@ -193,16 +199,9 @@ def set_sub_partition_address_and_size(reqs, sub_partitions):
         if size == 0:
             raise RuntimeError("No compatible parent partition found for %s" % sp_name)
         address = min([reqs[part]['address'] for part in sp_value['span']])
```
hakonfam: Should we remove the "span" functionality?

cmake/reports/CMakeLists.txt
```diff
@@ -16,3 +16,12 @@ foreach(report ram_report rom_report)
     DEPENDS ${logical_target_for_zephyr_elf}
     )
 endforeach()
+
+add_custom_command(
+  TARGET ${IMAGE}rom_report
```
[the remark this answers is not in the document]
oyvindronningstad: Yes, the intention is that that writing `ninja rom_report` will print both.

scripts/partition_manager.py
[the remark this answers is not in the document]
oyvindronningstad: The semantics of after are such that it only cares about the first existing entry, right?

scripts/partition_manager.py
oyvindronningstad: As discussed f2f: 'sub_partitions' are removed in a later commit. 'span' will stay since it still has uses.

scripts/partition_manager.py
```diff
@@ -244,61 +269,49 @@ def fn(a, b):
...
-    # Store output dir to app
-    config_lines.append('PM_APP_OUT_DIR "%s"' % app_output_dir)
-
     # Store complete size/address configuration to all input paths
-    for area_name, area_props in adr_map.items():
+    for _, area_props in adr_map.items():
         if 'out_dir' in area_props.keys():
             write_pm_config_to_file(path.join(area_props['out_dir'], pm_kconfig_file), config_lines)
 
-    # Store to root app
```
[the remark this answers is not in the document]
oyvindronningstad: Since the app's dirs have been added to adr_map.

scripts/partition_manager_report.py
oyvindronningstad: Haha :D
No problem, will do.

scripts/partition_manager_report.py
oyvindronningstad: will fix

scripts/partition_manager_report.py
oyvindronningstad: will fix

scripts/partition_manager_report.py
oyvindronningstad: will fix

scripts/partition_manager.py
oyvindronningstad: will change

scripts/partition_manager.py
oyvindronningstad: will fix

scripts/partition_manager.py
```diff
@@ -335,68 +351,116 @@ def main():
...
+        pm_config['app']['out_dir'] = args.app_pm_config_dir
+        write_pm_config(pm_config)
+        write_kconfig_file(pm_config)
+        if args.output:
+            write_yaml_out_file(pm_config, path.join(args.app_pm_config_dir, args.output))
     else:
         print("No input, running tests.")
         test()
 
+from pprint import pformat
```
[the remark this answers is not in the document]
oyvindronningstad: Will do

scripts/partition_manager.py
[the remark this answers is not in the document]
hakonfam: Yes

cmake/reports/CMakeLists.txt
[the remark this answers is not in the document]
hakonfam: But how will rom_report work if partition manager is not enabled then?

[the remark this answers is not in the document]
hakonfam: It will fail. You need to add a check for PARTITION_MANAGER presence before extending rom_report with this.

scripts/partition_manager.py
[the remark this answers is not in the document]
hakonfam: Got it.

cmake/reports/CMakeLists.txt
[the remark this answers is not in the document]
oyvindronningstad: Will fix

scripts/partition_manager.py
```diff
@@ -24,60 +25,112 @@ def item_is_placed(d, item, after_or_before):
...
+    comparison = 0
+    while 'inside' in elem.keys() and elem['inside'][0] in reqs.keys():
+        comparison += 1
+        elem = reqs[elem['inside'][0]]
+    return comparison
+
+
+def get_images_which_need_resolving(reqs, sub_partitions):
+    # Get candidates
+    unsorted = [x for x in reqs.keys() if 'placement' in reqs[x] and type(reqs[x]['placement']) == dict
```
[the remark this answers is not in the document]
oyvindronningstad: Fixed

CODEOWNERS
oyvindronningstad: Fixed

cmake/reports/CMakeLists.txt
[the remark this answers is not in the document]
oyvindronningstad: Fixed

scripts/partition_manager.py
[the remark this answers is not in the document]
hakonfam: Any update on this? AFAICT the order should not matter given my explanation above

scripts/partition_manager.py
```diff
@@ -318,85 +360,161 @@ def parse_args():
                              "image-name:pm.yml-path:build_dir:out_dir")
     parser.add_argument("--app-pm-config-dir", required=True,
                         help="Where to store the 'pm_config.h' of the root app.")
+    parser.add_argument("--app-build-dir", required=True,
+                        help="Path to of the root app's build directory.")
+    parser.add_argument("-o", "--output", required=False, type=str,
```
hakonfam: After testing this, I don't see the point of having this parameter. This path is relative to the output dir of the active image, and must be unique. As with the other two generated files, this should be hard coded in the script.

scripts/partition_manager.py
```diff
@@ -220,84 +283,63 @@ def get_header_guard_end(filename):
...
+        write_config_lines_to_file(path.join(dirs['out_dir'], pm_kconfig_file), config_lines)
+
 
-    # Store to root app
-    if not only_sub_image_is_being_built(adr_map, app_output_dir):
-        write_pm_config_to_file(path.join(app_output_dir, pm_kconfig_file), config_lines)
+def write_yaml_out_file(pm_config, input_config, filename):
+    yamldump = yaml.dump(pm_config)
+    for dirs in input_config.values():
+        write_config_lines_to_file(path.join(dirs['out_dir'], filename), [yamldump])
```
hakonfam: It is cumbersome to read this file when everything is in decimal, would it be possible to convert to hex representation?

hakonfam: Looking at the output, I'm not able to deduce what partitions are phony and not, any comments to this? For this to be useful for other scripts it would be necessary to identify what partitions are phony and not.

hakonfam: It looks like everything which does not have `span` `share_size` or `sharers` is a proper partition. Will this always be true?

oyvindronningstad: Everything without a `span` is a proper partition.

scripts/partition_manager.py
oyvindronningstad: Removed

scripts/partition_manager.py
oyvindronningstad: All numbers are now printed as hex

cmake/reports/CMakeLists.txt
```diff
@@ -16,3 +16,13 @@ foreach(report ram_report rom_report)
...
+  POST_BUILD
+  COMMAND
+  ${PYTHON_EXECUTABLE}
+  ${ZEPHYR_BASE}/scripts/partition_manager_report.py
+  --input ${PROJECT_BINARY_DIR}/include/generated/partitions.yml
+  "$<$<NOT:$<TARGET_EXISTS:partition_manager>>:--quiet>"
```
SebastianBoe: I believe you mentioned one could have empty COMMAND's.

If so, let's put the command in a variable and do

"$<$<TARGET_EXISTS:partition_manager>:${the_command}>"

Then we can avoid invoking partition_manager_report.py for non-PM builds.

oyvindronningstad: It seems this is not really possible to do in a good way.

scripts/partition_manager.py
hakonfam: FYI `span` is not sufficient.
'mcuboot_secondary` for instance does not have `span` but it has `share_size`, and that is not proper.

oyvindronningstad: mcuboot_secondary is not a phony partition.

hakonfam: Correct, I forgot that that was added now, my bad.

PR #142, opened 2019-05-03
[nrf noup] mgmt: mcumgr: Bugfix switch to use PM instead of DT if PM …

sigvartmh:
…is used

If partition manager(PM) is enabled one should switch to use the defines
set by PM instead of using the device three(DT) defines. As the DT
defines would be wrong.


sigvartmh:
@hakonfam

ChristopherBittner:
@nvlsianpu

ext/lib/mgmt/mcumgr/cmd/img_mgmt/port/zephyr/src/zephyr_img_mgmt.c
```diff
@@ -111,14 +133,14 @@ img_mgmt_impl_erase_slot(void)
     bool empty;
     int rc;
 
-    rc = zephyr_img_mgmt_flash_check_empty(DT_FLASH_AREA_IMAGE_1_ID,
+    rc = zephyr_img_mgmt_flash_check_empty(FLASH_AREA_IMAGE_PRIMARY,
```
hakonfam: Shouldn't this be `SECONDARY` ?

ext/lib/mgmt/mcumgr/cmd/img_mgmt/port/zephyr/src/zephyr_img_mgmt.c
```diff
@@ -89,16 +111,16 @@ zephyr_img_mgmt_flash_area_id(int slot)
...
 
     case 1:
-        fa_id = DT_FLASH_AREA_IMAGE_1_ID;
+        fa_id = FLASH_AREA_IMAGE_SECONDARY;
         break;
 
     default:
         assert(0);
-        fa_id = DT_FLASH_AREA_IMAGE_1_ID;
+        fa_id = FLASH_AREA_IMAGE_PRIMARY;
```
hakonfam: Shouldn't this be `SECONDARY` ?

sigvartmh: that's correct will fix it

sigvartmh: I don't know why you would default to the secondary after an assert but I assume it checks the primary first

subsys/fs/nffs_fs.c
```diff
@@ -19,6 +19,18 @@
 
 #define NFFS_MAX_FILE_NAME 256
 
+#if USE_PARTITION_MANAGER
+#include <pm_config.h>
+#define NFFS_FLASH_AREA_STORAGE_OFFSET PM_MCUBOOT_STORAGE_ADDRESS
```
sigvartmh: I'm a bit unsure if this is correct, since you could have the use case where you would use NFFS in your application. This should probably be changed to something more image specific?

hakonfam: The semantics mirror that of DT right? If so there will be no regression and IMO this should be OK.

nvlsianpu: @hakonfam  Is that possible that storage partition is not described via mcuboot pm.yaml, while it is well described via application dts? (so lack of PM_MCUBOOT_STORAGE_ADDRESS macro). Does it mak sense along witch PM usage.

hakonfam: the storage partition is specified by pm.yml, so pm_config.h should have a define for that the same way that the app dts has it.

sigvartmh: I'm just wondering if maybe we could reuse some of the `#define FLASH_MAP_OFFSET(i) UTIL_CAT(PM_, UTIL_CAT(PM_##i##_LABEL, _ADDRESS))` macro to give `STORAGE_ADDRESS` etc if the **application** and **mcuboot** both have their *own* file system for storage? I don't know if this is a valid use-case.

nvlsianpu: Currently I don't see need for storage partition for mcuboot.

nvlsianpu:
@hakonfam did you recheck this?

PR #144, opened 2019-05-06
[nrf toup] cmake: Improve access to other images in multi-image builds.

oyvindronningstad:
Global properties:
 - ${IMAGE}PROJECT_BINARY_DIR.

Helper function:
 - get_image_name() to strip the '_'.

### DIFF
--- a/cmake/app/boilerplate.cmake
+++ b/cmake/app/boilerplate.cmake
@@ -87,6 +87,12 @@ set(__build_dir ${CMAKE_CURRENT_BINARY_DIR}/zephyr)
 
 set(PROJECT_BINARY_DIR ${__build_dir})
 
+define_property(GLOBAL PROPERTY ${IMAGE}PROJECT_BINARY_DIR
+  BRIEF_DOCS "Build directory (PROJECT_BINARY_DIR) for the ${IMAGE} image."
+  FULL_DOCS "To be used to access e.g. this image's hex file."
+  )
+set_property(GLOBAL PROPERTY ${IMAGE}PROJECT_BINARY_DIR ${PROJECT_BINARY_DIR})
+
 add_custom_target(${IMAGE}code_data_relocation_target)
 
 # CMake's 'project' concept has proven to not be very useful for Zephyr
--- a/cmake/extensions.cmake
+++ b/cmake/extensions.cmake
@@ -889,6 +889,12 @@ function(import_kconfig prefix kconfig_fragment)
   endforeach()
 endfunction()
 
+function(get_image_name image out_var)
+  string(LENGTH ${image} len)
+  MATH(EXPR len "${len}-1")
+  string(SUBSTRING ${image} 0 ${len} ${out_var})
+endfunction()
+
 ########################################################
 # 3. CMake-generic extensions
 ########################################################


oyvindronningstad:
cc @SebastianBoe @hakonfam

cmake/extensions.cmake
```diff
@@ -889,6 +889,12 @@ function(import_kconfig prefix kconfig_fragment)
   endforeach()
 endfunction()
 
+macro(get_image_name image out_var)
```
SebastianBoe: Function's should be used whenever possible, AFAICT it is possible here.

(less chance of aliasing issues due to the creation of a new scope)

oyvindronningstad: Will fix

SebastianBoe:
I agree that these properties should be image'ified, I don't know how we missed them (we grepped for all properties at some point).

After migrating all users I can approve this PR.

cmake/app/boilerplate.cmake
```diff
@@ -87,6 +87,12 @@ set(__build_dir ${CMAKE_CURRENT_BINARY_DIR}/zephyr)
 
 set(PROJECT_BINARY_DIR ${__build_dir})
 
+define_property(GLOBAL PROPERTY ${IMAGE}PROJECT_BINARY_DIR
+  BRIEF_DOCS "Build directory (PROJECT_BINARY_DIR) for the ${IMAGE} image."
+  FULL_DOCS "To be used to access e.g. this image's hex file."
+  )
+set_property(GLOBAL PROPERTY ${IMAGE}PROJECT_BINARY_DIR ${PROJECT_BINARY_DIR})
```
SebastianBoe: Introducing ${IMAGE}PROJECT_BINARY_DIR is sane.

From the CMakeCache.txt I see that ${project}_BINARY_DIR is already available. But that isn't quite what we want unfortunately.

cmake/app/boilerplate.cmake
```diff
@@ -517,6 +523,12 @@ set(KERNEL_EXE_NAME   ${KERNEL_NAME}.exe)
 set(KERNEL_STAT_NAME  ${KERNEL_NAME}.stat)
 set(KERNEL_STRIP_NAME ${KERNEL_NAME}.strip)
 
+define_property(GLOBAL PROPERTY ${IMAGE}KERNEL_NAME
+  BRIEF_DOCS "Name (KERNEL_NAME) for the ${IMAGE} image."
+  FULL_DOCS "To be used to access e.g. this image's hex file."
+  )
+set_property(GLOBAL PROPERTY ${IMAGE}KERNEL_NAME ${KERNEL_NAME})
```
SebastianBoe: This is sane as well.

I would await introducing it until there was a user though.

In case this isn't quite what you are looking for. E.g. it might turn out that

${${IMAGE}logical_name_for_zephyr_elf} is what you actually need.

cmake/extensions.cmake
```diff
@@ -889,6 +889,12 @@ function(import_kconfig prefix kconfig_fragment)
   endforeach()
 endfunction()
 
+function(get_image_name image out_var)
+  string(LENGTH ${image} len)
+  MATH(EXPR len "${len}-1")
+  string(SUBSTRING ${image} 0 ${len} ${out_var})
+endfunction()
```
SebastianBoe: This is good to have.

cmake/app/boilerplate.cmake
oyvindronningstad: name -> target?
`${${IMAGE}logical_target_for_zephyr_elf}`

oyvindronningstad: I see what you mean.

oyvindronningstad: Removed

SebastianBoe:
Is this still an RFC?

oyvindronningstad:
> Is this still an RFC?

No, I'm happy if you're happy :).

PR #145, opened 2019-05-08
dfu/image_util: take trustzone and PM into account

nvlsianpu:
bugfix PR - image_utill need to take trustzone and PM into account when assignes proper iamge area for the image collection module.


nvlsianpu:
PR for `fromlist` commit: https://github.com/zephyrproject-rtos/zephyr/pull/16014

subsys/dfu/img_util/flash_img.c
```diff
@@ -18,6 +18,19 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
...
+#include <generated_dts_board.h>
+/* DT_FLASH_AREA_IMAGE_XX_YY values used below are auto-generated by DT */
+#ifdef CONFIG_TRUSTED_EXECUTION_NONSECURE
+#define FLASH_AREA_IMAGE_SECONDARY DT_FLASH_AREA_IMAGE_1_NONSECURE_ID
+#else
+#define FLASH_AREA_IMAGE_SECONDARY DT_FLASH_AREA_IMAGE_1_ID
+#endif /* CONFIG_TRUSTED_EXECUTION_NONSECURE */
+#endif /* USE_PARTITION_MANAGER */
+
 BUILD_ASSERT_MSG((CONFIG_IMG_BLOCK_BUF_SIZE % DT_FLASH_WRITE_BLOCK_SIZE == 0),
```
hakonfam: Is it correct to have `DT_FLASH_WRITE_BLOCK_SIZE` here?

subsys/dfu/img_util/flash_img.c
```diff
@@ -18,6 +18,19 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 #include <dfu/flash_img.h>
 #include <inttypes.h>
 
+#if USE_PARTITION_MANAGER
+#include <pm_config.h>
+#define FLASH_AREA_IMAGE_SECONDARY  PM_MCUBOOT_PARTITIONS_SECONDARY_ID
+#else
+#include <generated_dts_board.h>
+/* DT_FLASH_AREA_IMAGE_XX_YY values used below are auto-generated by DT */
+#ifdef CONFIG_TRUSTED_EXECUTION_NONSECURE
```
hakonfam: IMO we should not support non-secure building of the bootloader, so this check is superfluous.
Only the secure version should be used.

ioannisg:
It looks to me this is already merged upstream. Shall we close this and get it from upstream?

hakonfam:
I don't think the upstream version contains the full patch (IE the partition manager part.)

ioannisg:
> 
> 
> I don't think the upstream version contains the full patch (IE the partition manager part.)

Should we then upmerge, first, and rebase this PR right after?

sigvartmh:
Yes

nvlsianpu:
> Should we then upmerge, first, and rebase this PR right after?

That I prefer.

ioannisg:
OK, temporarily adding DNM to the PR

PR #148, opened 2019-05-14
Partition manager: predefined configuration

hakonfam:
…manager

This commit adds support for predefined configuration.
This is a feature intended for deployed products consisting of multiple images.
It is crucial that any new versions of these images are linked to the correct
location. This is solved by this commit.

Also, this fixes the issue of building sub-images with mixed build strategies.

### DIFF
--- a/cmake/partition_manager.cmake
+++ b/cmake/partition_manager.cmake
@@ -13,8 +13,14 @@ Also, the each image's hex file will be automatically associated with its partit
 if(FIRST_BOILERPLATE_EXECUTION)
   get_property(PM_IMAGES GLOBAL PROPERTY PM_IMAGES)
 
-  if(PM_IMAGES)
-    # Partition manager is enabled because we have populated PM_IMAGES.
+  set(static_configuration_file ${APPLICATION_SOURCE_DIR}/pm_static.yml)
+
+  if(PM_IMAGES OR (EXISTS ${static_configuration_file}))
+    # Partition manager is enabled because we have populated PM_IMAGES,
+    # or because the application has specified a static configuration.
+    if (EXISTS ${static_configuration_file})
+      set(static_configuration --static-config ${static_configuration_file})
+    endif()
 
     set(generated_path include/generated)
 
@@ -42,6 +48,7 @@ if(FIRST_BOILERPLATE_EXECUTION)
       --input-files ${input_files}
       --flash-size ${flash_size}
       --output ${CMAKE_BINARY_DIR}/partitions.yml
+      ${static_configuration}
       )
 
     set(pm_output_cmd
--- a/scripts/partition_manager.py
+++ b/scripts/partition_manager.py
@@ -196,13 +196,13 @@ def app_size(reqs, total_size):
     return size
 
 
-def set_addresses(reqs, sub_partitions, solution, flash_size):
-    set_shared_size(reqs, sub_partitions, flash_size)
+def set_addresses(reqs, sub_partitions, solution, size, start=0):
+    set_shared_size(reqs, sub_partitions, size)
 
-    reqs['app']['size'] = app_size(reqs, flash_size)
+    reqs['app']['size'] = app_size(reqs, size)
 
-    # First image starts at 0
-    reqs[solution[0]]['address'] = 0
+    # First image starts at 0, or at start of dynamic area if static configuration is given.
+    reqs[solution[0]]['address'] = start
     for i in range(1, len(solution)):
         current = solution[i]
         previous = solution[i - 1]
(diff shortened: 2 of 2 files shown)


scripts/partition_manager.py
```diff
@@ -262,14 +262,67 @@ def get_flash_size(reqs):
...
+    return pre_defined_start, pre_defined_size
+
+
+def get_pm_config(input_config, pre_defined_config):
+    to_resolve = dict()
+    pre_defined_size = None
+    pre_defined_start = None
+
+    load_reqs(to_resolve, input_config)
+    flash_size = get_flash_size(input_config) if not pre_defined_size else pre_defined_size
```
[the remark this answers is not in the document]
hakonfam: `flash_size` is used inside the if block.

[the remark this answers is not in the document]
oyvindronningstad: Right, but `pre_defined_size` is used before being set. Probably needs a bit of mangling.

hakonfam:
@oyvindronningstad  your comments should be addressed.

scripts/partition_manager.py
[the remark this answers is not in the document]
oyvindronningstad: What is the verdict on this? We probably don't need to check `pre_defined_size` since it can't be anything other than `None` (AFAICT).

scripts/partition_manager.py
```diff
@@ -195,13 +195,13 @@ def app_size(reqs, total_size):
     return size
 
 
-def set_addresses(reqs, sub_partitions, solution, flash_size):
-    set_shared_size(reqs, sub_partitions, flash_size)
+def set_addresses(reqs, sub_partitions, solution, size, start=None):
```
oyvindronningstad: `start=0` => you don't need the ternary if at line 204.

scripts/partition_manager.py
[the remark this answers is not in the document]
hakonfam: I get your point.

what about this

```
def get_pm_config(input_config, pre_defined_config):
    ...
    flash_size = get_flash_size(input_config)

    if pre_defined_config:
        pre_defined_start, flash_size = get_pre_defined_start_size(pre_defined_config, flash_size)
```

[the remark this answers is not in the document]
oyvindronningstad: LGTM.

Can we rename it to `size` or `free_size` or something?

scripts/partition_manager.py
```diff
@@ -262,14 +262,45 @@ def get_flash_size(reqs):
...
+
+    assert len(gaps) == 1, "Incorrect amount of gaps found"
+
+    start, end = gaps[0]
+    return start, end-start
+
+
+def get_pm_config(input_config, pre_defined_config):
+    to_resolve = dict()
+    pre_defined_size = None
```
oyvindronningstad: I think this is unused now.

scripts/partition_manager.py
```diff
@@ -262,14 +262,44 @@ def get_flash_size(reqs):
...
+
+    assert len(gaps) == 1, "Incorrect amount of gaps found"
+
+    start, end = gaps[0]
+    return start, end-start
+
+
+def get_pm_config(input_config, pre_defined_config):
+    to_resolve = dict()
+    pre_defined_start = None
```
oyvindronningstad: `start = 0`?

scripts/partition_manager.py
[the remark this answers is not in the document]
hakonfam: Yes, this is not very future proof :) `free_size` is best IMO

cmake/partition_manager.cmake
```diff
@@ -11,22 +11,35 @@ if(FIRST_BOILERPLATE_EXECUTION)
     PARTITION_MANAGER_CONFIG_FILES
     )
 
-  if(partition_manager_config_files)
+  set(partitions_yml ${APPLICATION_SOURCE_DIR}/prj.partitions.yml)
```
SebastianBoe: partitions_yml should be renamed to prj_partitions_yml or predefined_configuration_file

cmake/partition_manager.cmake
```diff
@@ -11,22 +11,35 @@ if(FIRST_BOILERPLATE_EXECUTION)
...
+
+  if(partition_manager_config_files OR (EXISTS ${partitions_yml}))
     # Partition manager is enabled because we have populated config
-    # files.
+    # files or because the application has pointet to pre defined partition
```
SebastianBoe: has specified a predefined configuration.

cmake/partition_manager.cmake
```diff
@@ -11,22 +11,35 @@ if(FIRST_BOILERPLATE_EXECUTION)
...
+    # configurations.
 
-    execute_process(
-      COMMAND
-      ${PYTHON_EXECUTABLE}
+    if (EXISTS ${partitions_yml})
+      set(pre_defined_conf --pre-defined-config ${partitions_yml})
```
SebastianBoe: we should consistently use pre-defined or predefined. The docs use predefined.

scripts/partition_manager.py
```diff
@@ -258,18 +258,48 @@ def get_config(reqs, config):
...
+                         if 'span' not in config.keys() and name != 'app']
+
+    starts = {flash_size} | {config['address'] for config in proper_partitions}
+    ends = {0} | {config['address'] + config['size'] for config in proper_partitions}
+    gaps = list(zip(sorted(ends - starts), sorted(starts - ends)))
+
+    assert len(gaps) == 1, "Incorrect amount of gaps found"
+
+    start, end = gaps[0]
+    return start, end-start
```
SebastianBoe: surround - with whitespace

scripts/partition_manager.py
```diff
@@ -258,18 +258,48 @@ def get_config(reqs, config):
...
 
-def get_pm_config(input_config):
-    reqs = dict()
-    load_reqs(reqs, input_config)
-    flash_size = get_flash_size(input_config)
-    solution, sub_partitions = resolve(reqs)
-    set_addresses(reqs, sub_partitions, solution, flash_size)
-    set_sub_partition_address_and_size(reqs, sub_partitions)
-    return reqs
+def get_pre_defined_start_size(pre_defined_config, flash_size):
```
SebastianBoe: either rename this to start_and_size, or if possible, find a way to have the function(s) return one value.

hakonfam:
> But how does this interact with *_BUILD_STRATEGY_SKIP_BUILD ?
> 
> Does the user need to make sure he enables this for static partitions?
> 
> If no, do we still need the option? If yes, should we document this in the static configuration docs somewhere?

If `*_BUILD_STRATEGY_SKIP_BUILD` is selected, the user must define a static partition in order to ensure correct placement of the dynamic partitions. I will state this more clearly in the docs.

cmake/partition_manager.cmake
hakonfam: will be `static`

cmake/partition_manager.cmake
hakonfam: static_configuration_file

hakonfam:
@SebastianBoe  your comments should be resolved

PR #149, opened 2019-05-15
[nrf toup] cmake: fix regex for west version check in host-tools.cmake

thst-nordic:
This change enables the nrfconnect CI to test tags.  

'west --version' output has the path to the west manifest directory along with the version.
If there is a semantic version string in this path name (like when CI tests a tag), it raises
a false version check failure.


mbolivar:
Is this breaking anything? If not, can you submit upstream first? I'm happy to +1 there

thst-nordic:
ok, will move upstream.

PR #158, opened 2019-05-29
[nrf toup] ci: enable windows 'sanitycheck --build-only'

thst-nordic:
Small changes that enable windows sanitycheck --build-only.
Enables CI to build apps/samples on windows agents.


scripts/sanitycheck
```diff
@@ -3173,6 +3176,13 @@ def main():
     global run_individual_tests
     options = parse_arguments()
 
+    if (os.name == 'nt'):
+        print("Running sanitycheck on Windows is not supported yet.")
+        print("https://github.com/zephyrproject-rtos/zephyr/issues/2664")
+        if (not options.build_only):
+            print("Running sanitycheck on Windows with '--build-only' might work.")
```
SebastianBoe: Why only "might work" ?

If this is useful enough to be upstreamed then we would want to do so to avoid conflicts.

thst-nordic: I tested windows sanitycheck on zephyr samples and it requires one more flag: '--disable-size-report'.   

I've added a check for this in the script.

tstilz:
The Kconfig error from CI is on master already, and was introduced from upmerge on friday.  I fixed half of the problems on PR-160,, but we just have to ignore the others until we can upgrade west.   

It passed all other tests on CI.

thst-nordic:
The Kconfig error from CI is on master already, and was introduced from upmerge on friday. I fixed half of the problems on PR-160,, but we just have to ignore the others until we can upgrade west.

It passed all other tests on CI.

scripts/sanitycheck
```diff
@@ -3208,6 +3211,15 @@ def main():
     global run_individual_tests
     options = parse_arguments()
 
+    if (os.name == 'nt'):
+        print("Running sanitycheck on Windows is not supported yet.")
+        print("https://github.com/zephyrproject-rtos/zephyr/issues/2664")
```
SebastianBoe: If I'm using a supported operation, like --build-only, then I don't want to be warned that what
I'm doing is not supported.

This should either be removed or merged into the below message.

SebastianBoe:
Please add a link from here to the upstream PR when available.

The upstream PR being created at some point is necessary for this to be merged, but it is not necessary to create it before this is merged.

PR #167, opened 2019-06-19
[nrf toup] cmake: imageify OVERLAY_CONFIG

hakonfam:
Currently what is provided in -DOVERLAY_CONFIG will be applied
to all images, this is incorrect. This patch ensures that the
overlay config is only applied to a single image.

Note that it is also possible to specify which image it should
apply to by providing the image name as a prefix in the command.

E.g. '-Dspm_OVERLAY_CONFIG'


metc:
cc @rlubos

metc:
@hakonfam Could you please rebase so we can merge soon?

hakonfam:
@metc done

metc:
Could we merge this PR?

PR #168, opened 2019-06-28
[nrf toup] arch: Allow removing drivers while keeping arch and kernel

oyvindronningstad:
Add dummy interrupt handlers and driver functions.
Rename CONFIG_KERNEL to CONFIG_DRIVERS and have it affect drivers only.

This helps with porting B0 to use more native Zephyr code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -457,7 +457,7 @@ endif()
 add_subdirectory(boards)
 add_subdirectory(ext)
 add_subdirectory(subsys)
-add_subdirectory_ifdef(CONFIG_KERNEL drivers)
+add_subdirectory(drivers)
 
 # Include zephyr modules generated CMake file.
 if(EXISTS ${CMAKE_BINARY_DIR}/zephyr_modules.txt)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -457,7 +457,7 @@ endif()
 add_subdirectory(boards)
 add_subdirectory(ext)
 add_subdirectory(subsys)
-add_subdirectory(drivers)
+add_subdirectory_ifdef(CONFIG_DRIVERS drivers)
 
 # Include zephyr modules generated CMake file.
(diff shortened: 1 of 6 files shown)


oyvindronningstad:
cc @sigvartmh @hakonfam @osaether

sigvartmh:
I get the idea, but do we want to have this dummy stuff in the zephyr repo?

drivers/Kconfig
```diff
@@ -7,6 +7,7 @@
 #
 
 
+if KERNEL
```
hakonfam: Isn't this symbol now replaced with `DRIVERS` ?

arch/arm/core/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ zephyr_library_sources(
   thread_abort.c
 )
 
+if (NOT DEFINED CONFIG_DRIVERS)
```
hakonfam: Would `if (NOT CONFIG_DRIVERS)` work?

Kconfig.zephyr
```diff
@@ -50,8 +50,8 @@ menu "Build and Link Features"
 
 menu "Linker Options"
 
-config KERNEL
-	bool "Include Zephyr kernel"
+config DRIVERS
```
hakonfam: Actually, we don't need to have any kconfig like this. Including the `driver` folder works without compilation error now!

CMakeLists.txt
```diff
@@ -457,7 +457,7 @@ endif()
 add_subdirectory(boards)
 add_subdirectory(ext)
 add_subdirectory(subsys)
-add_subdirectory_ifdef(CONFIG_KERNEL drivers)
+add_subdirectory_ifdef(CONFIG_DRIVERS drivers)
```
hakonfam: Try to make this non-conditional, AFAICS it will work without the kconfig check.

oyvindronningstad:
Not needed now.

hakonfam:
Is closing this related to my PR upstream? I am not sure that this will be accepted.

PR #171, opened 2019-07-03
Fix "undefined reference to `net_addr_pton'" error.

gordonklaus:
socket.h includes net_ip.h which declares net_addr_pton, however it is not defined when CONFIG_NET_RAW_MODE=y.

See this thread for reference https://devzone.nordicsemi.com/f/nordic-q-a/48792/function-net_pton---compiling-issue-on-macos/


lemrey:
Hi @gordonklaus could you address the gitlint issues?

rlubos:
@gordonklaus I think that this change should target upstream Zephyr repository. We aim to minimize differences between upstream and fw-nrfconnect-zephyr.

gordonklaus:
@rlubos I tried to reproduce this upstream but ran into too many unrelated errors along the way and gave up.

In any case, I think I identified the issue as being related to [this larger one](https://github.com/zephyrproject-rtos/zephyr/issues/15626).

PR #173, opened 2019-07-12
Post 1.0 rebase

mbolivar:
Now that NCS v1.0.0 is out, this pull request rewrites our downstream history (and is thus a destructive operation that must be force-pushed) as a linear set of patches on top of the latest upstream base commit.

The resulting tree contains exactly the same code as the current west manifest revision for this repository (i.e. `git diff this-pull-request a8933a2ead` is empty).

Since this is the first time we've done this, please refer to our internal development documentation for details on how and why we would do such a thing; feel free to ping me if you need a pointer. Also since this is the first time we do it, the entire process is up for debate :).

In my opinion, we will want to revert some of these patches immediately after merge:

- merged upstream: `[nrf toup] drivers: sensor: Update DT symbols in qdec_nrfx driver`
- merged upstream: `[nrf toup] ext: hal: nordic: Update DT symbols in nrfx config file`
- no longer needed per its commit message: `[nrf temphack] cortex_m: linker: Add ext_abi section`
- should not have been made downstream: `[nrf temphack] REVERTME: carry whitespace changes from a mergeup`

However, they're preserved here to get the same code in the tree as we have now.

There are some other questionable things:

- split out from a mergeup and didn't make sense to squash into something else: `[nrf temphack] cmake: carry forward dts.cmake patches from a mergeup`, probably should get a better git log
- some patches are multi-image related and might want to get squashed into the patch adding support for it (e.g. `[nrf toup] cmake: IMAGEify extra_post_build_byproducts`)

### DIFF
--- a/include/arch/arm/cortex_m/scripts/linker.ld
+++ b/include/arch/arm/cortex_m/scripts/linker.ld
@@ -188,8 +188,8 @@ SECTIONS
 
 	_vector_end = .;
 	/* This is used for Secure Boot */
-	#if defined(CONFIG_SECURE_BOOT)
-	. = CONFIG_SB_FIRMWARE_INFO_OFFSET;
+	#if defined(CONFIG_FW_METADATA)
+	. = CONFIG_FW_FIRMWARE_INFO_OFFSET;
 	KEEP(*(.firmware_info))
 	#endif
 	} GROUP_LINK_IN(ROMABLE_REGION)


mbolivar:
Separating this into its own comment since it's a side note:

@SebastianBoe @ioannisg @rlubos:

This time, the process was a bit painful, because several mergeup commits were what is known as "evil merges":

https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefevilmergeaevilmerge

That is, the merge commits contained extra code, such as fixes for multi-image, which were neither downstream nor upstream, and were added as part of the merge commit itself.

In cases where code like that is needed, it's better to revert the downstream patches which conflict, then merge upstream, then cherry-pick the downstream patches and apply conflict resolution in those separate commits. That makes it obvious what code comes from where, and also makes it easier to use internal tooling to make rewriting history easier in these types of "cleanup" PRs.

I knew this was happening, but I didn't want to complain because we needed to get a 1.0 release out the door. Since I knew I could clean things up later, I figured we could discuss when the time was right. I think that is now :), so can we talk offline about how to manage this in the future?

For example:

- a288e8b8ab54c667c6ed9776042200411710a9cf: contains multi-image changes to arch/x86/CMakeLists.txt
- 1378a1c7ac89d13fa08bd45b1272a70060d90663: various changes to cmake/dts.cmake, which should be in their own patch. I've added a temphack for that here.
- 36ca7b1a0c2e8d31dc5352e2567fde882b996686: adds IMAGE prefixes in cmake/linker/ld/target_relocation.cmake
- 816f8795890bc4947691e943dc49be9d88e4bc70:
  - changes CMAKE_BINARY_DIR to CMAKE_CURRENT_BINARY_DIR in zephyr_modules.txt
    processing, i.e. re-applies d70aa052bc4b333e97a36b4e5832a8509b4aa417
  - changes REPO_CI_TOOLS_SHA in the Jenkinsfile
- f69d8d61a934533de2dfc5c3ca915b82910acd8d: changes kernel to ${KERNEL_LIBRARY} in cmake/linker/ld/target.cmake, adds IMAGE prefixes

mbolivar:
Second site note:

@carlescufi:

In order to keep nrf/west.yml working, we will need to tag our current master before force-pushing this so we don't make that commit eligible for garbage collection in the git history.

PR #176, opened 2019-07-23
Jenkinsfile: Build only nRF Platforms for PR's

thst-nordic:
When testing a PR, we only need to build nRF platforms.
Branches and Tags test all platforms (nRF & non-nRF) 

Based on changes from Kamil Gawor's  [PR-154](https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/154)

Signed-off-by: Kamil Gawor <Kamil.Gawor@nordicsemi.no>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,8 +42,10 @@ pipeline {
       // ENVs for sanitycheck
       ARCH = "-a arm"
       SANITYCHECK_OPTIONS = "--inline-logs --enable-coverage -N"
-      SANITYCHECK_RETRY = "--only-failed --outdir=out-2nd-pass"
-      SANITYCHECK_RETRY_2 = "--only-failed --outdir=out-3rd-pass"
+      SANITYCHECK_RETRY_CMDS = """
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -13,6 +13,7 @@ pipeline {
    booleanParam(name: 'RUN_DOWNSTREAM', description: 'if false skip downstream jobs', defaultValue: true)
    booleanParam(name: 'RUN_TESTS', description: 'if false skip testing', defaultValue: true)
    booleanParam(name: 'RUN_BUILD', description: 'if false skip building', defaultValue: true)
+   string(name: 'PLATFORMS', description: 'Default Platforms to test', defaultValue: 'nrf9160_pca10090 nrf52_pca10040 nrf52840_pca10056')
```
carlescufi: Should we add a few more here? nrf51, nrf52810, etc?

thst-nordic: With no platforms defined, Sanitycheck runs each testcase once with a random platform.  
When we supply the platform if runs all tests on all platform args. 

If we add every nRF platform the time to run increases dramatically.  
The plan is to add parallel build/test for each platform, and then add nrf51, nrf52810, etc..

carlescufi: ok, sounds good

PR #169, opened 2019-06-28
Multi-image: Fixes in kconfig.cmake

oyvindronningstad:
Allow CONFIG parameters to be specified in command line to cmake

This will e.g. make cmake [...] -DCONFIG_BOOTLOADER_MCUBOOT work again
It also allows e.g. cmake [...] -Dmcuboot_CONFIG_xxx to enable configs in mcuboot.

The second commit I'm a bit unsure if of. Double check if it is ok.

PR #177, opened 2019-07-26
[nrf-toup] set config of sub image

hakonfam:
Allow setting config of sub image

Note that this is an alternative to #169 


hakonfam:
#169 is a better solution as it avoids congesting the sub-images' namespace.

PR #184, opened 2019-08-13
[nrf-noup] cmake: remove check for including 'drivers'

hakonfam:
Previously the build code in the 'drivers' folder included
some files which caused a compilation error for B0.

This problem has been fixed, and the folder can now
safely be included in the build.

This aligns with upstream code.


SebastianBoe:
tag is wrong, shouldn't be a hyphen in there

hakonfam:
@SebastianBoe tag issue fixed

PR #185, opened 2019-08-14
[nrf toup] drivers/flash: leverage nsc flash read

hakonfam:
When current image is non-secure and address being read is configured
as secure, use the Non Secure Callable function spm_request_read to
try and read the secure flash area.

### DIFF
--- a/drivers/flash/soc_flash_nrf.c
+++ b/drivers/flash/soc_flash_nrf.c
@@ -16,6 +16,11 @@
 #include <string.h>
 #include <nrfx_nvmc.h>
 
+#if CONFIG_ARM_NONSECURE_FIRMWARE
+#include <secure_services.h>
+#include <pm_config.h>
+#endif
+
 #if defined(CONFIG_SOC_FLASH_NRF_RADIO_SYNC)
 #include <sys/__assert.h>
 #include <bluetooth/hci.h>
@@ -141,6 +146,12 @@ static int flash_nrf_read(struct device *dev, off_t addr,
 	if (!len) {
 		return 0;
 	}
+
+#if CONFIG_ARM_NONSECURE_FIRMWARE
+	if (addr < PM_APP_ADDRESS) {
+		return spm_request_read(data, addr, len);
+	}
+#endif
 
 	memcpy(data, (void *)addr, len);
 


oxalistriangularis:
Should the patch include similar changes for other flash operations like write and erase?

Also, I am not sure we want to allow flash regions from Non-secure application to have access rights to operations. Do we have a requirement and analysis on this?

hakonfam:
> Should the patch include similar changes for other flash operations like write and erase?

Currently there is no need for that. Those operations sounds scary to allow from non secure side as well :)
 
> Also, I am not sure we want to allow flash regions from Non-secure application to have access rights to operations. Do we have a requirement and analysis on this?

There is a requirement for reading metadata of mcuboot slots. This is currently the only data which is allowed to read.

oxalistriangularis:
The current code will fault when write and erase are attempted on a secure side, is this by design? Would you introduce an error code instead?

Also, any operation including read sounds scary the minute it is in secure region. So, if we permit read, we can permit other operations too. Remember the secure side, based on access rights can return a failure to the non-secure side. So, secure side must always guard all operations on its flash and any other resource and not expect the non-secure side to the right thing. 

I am fine with a need driven development, as long as it is well thought through. In this case, I find it absurd that we check if the address is in secure region only in case of read.

hakonfam:
> The current code will fault when write and erase are attempted on a secure side, is this by design? Would you introduce an error code instead?

This PR does not try to address issues around write or erase, so it is not by design, it is more by "not implemented yet".
Currently you would get an exception which explains that a security exception occured, which is sufficient IMO.

> Also, any operation including read sounds scary the minute it is in secure region.

The only area which we allow reads from (see https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1039) is metadata created my MCUboot.

>So, if we permit read, we can permit other operations too. Remember the secure side, based on access rights can return a failure to the non-secure side.

Sure, this is already implemented in https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1039 as mentioned above.
I agree that we should permit other operations as well when they are needed.

> I am fine with a need driven development, as long as it is well thought through. In this case, I find it absurd that we check if the address is in secure region only in case of read.

That is because only in case of a read can it be allowed, as write and erase from non-secure to secure is not supported.

drivers/flash/soc_flash_nrf.c
```diff
@@ -142,6 +147,12 @@ static int flash_nrf_read(struct device *dev, off_t addr,
 		return 0;
 	}
 
+#if CONFIG_ARM_NONSECURE_FIRMWARE
+	if (addr < PM_APP_ADDRESS) {
```
nvlsianpu: I see silent assumption that secure flash partition to read always begin at 0 and ends up at PM_APP_ADDRESS

hakonfam: That is a sane assumption, also made in the SPM and secure services.

In theory it is not always like this, in practice (for these solutions) it will always be contiguous.

nvlsianpu: Ok, can live with this.

hakonfam:
@SebastianBoe  could we get this merged?

SebastianBoe:
I don't have the domain knowledge to merge this, requesting review from @lemrey /  @anangl  / @ioannisg .

drivers/flash/soc_flash_nrf.c
sigvartmh: Maybe add a comment which documents this so that customers who looks to modify the code can understand some of the reasoning in case they have another use-case.

lemrey:
@oxalistriangularis comments are addressed.
- We don't blindly allow the non-secure application to read at any address, the secure side implements the logic to determine whether the access is allowed or not, see this [PR](https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1039).
- The TrustZone hardware generates the fault when write and erase operations are made from Non-Secure domain on flash areas which are configured as Secure. To return an error code we'd have to actually introduce a couple of APIs in the Secure domain _and_ disallow these operations explicitly. We don't need this right now and I don't think it's a good idea, in general since a bug in that logic could be very nasty.

hakonfam:
Ping @SebastianBoe @nvlsianpu - should be ready to merge

PR #186, opened 2019-08-21
Build s1 variant of mcuboot

hakonfam:
To support upgrade of mcuboot we need to build a candidate of mcuboot which is linked against the partition "s1".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1252,6 +1252,69 @@ set(logical_target_for_zephyr_elf ${logical_target_for_zephyr_elf} PARENT_SCOPE)
 # 2. it can be defined in Kconfig
 set_target_properties(${logical_target_for_zephyr_elf} PROPERTIES OUTPUT_NAME ${KERNEL_NAME})
 
+if (IMAGE STREQUAL mcuboot_ AND CONFIG_MCUBOOT_BUILD_S1_VARIANT)
+# Create second MCUBoot executable for the second slot of the second stage
+# bootloader. This is done inside this file since it is non-trivial to add
+# executable targets outside the root CMakeLists.txt. The problem is that
+# most of the variables requried for compiling/linking is only available
+# in the scope of this file.
+
+  # Create linker script which has an offset from S0 to S1.
+  configure_linker_script(
+    ${PROJECT_BINARY_DIR}/linker_mcuboot_s1.cmd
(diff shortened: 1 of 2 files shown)


oyvindronningstad:
Could we solve this by instead having two separate projects for mcuboot with two different `pm.yml` files?

hakonfam:
> Could we solve this by instead having two separate projects for mcuboot with two different `pm.yml` files?

Wouldn't it be re-compiled then?

oyvindronningstad:
> > Could we solve this by instead having two separate projects for mcuboot with two different `pm.yml` files?
> 
> Wouldn't it be re-compiled then?

Yes, but not with ccache

hakonfam:
> Yes, but not with ccache

Good point. Having another pm.yml is, however, not sufficient to get another executable built. To do that we need to include `zephyr/cmake/app/boilerplate.cmake` from somewhere. I don't see how this should be done, but hopefully there is a way.

hakonfam:
I thought about your suggestion @oyvindronningstad , and AFAICS it is better do what is proposed in this PR. The main reason is that you want to have the two mcuboot variants (s0 and s1) as similar as possible. In order to use a separate pm.yml one would (as I wrote above) have to create a new sample. This will create a new kconfig instance, and then it is not trivial to keep these two kconfig instances in sync (mcuboot-s0 and mcuboot-s1). 

Another question is: Should pm and the build system allow you to perform these types of re-linking for partitions? I don't think so as I don't see any other use case than mcuboot for s0 and s1.

CMakeLists.txt
```diff
@@ -1252,6 +1252,67 @@ set(logical_target_for_zephyr_elf ${logical_target_for_zephyr_elf} PARENT_SCOPE)
 # 2. it can be defined in Kconfig
 set_target_properties(${logical_target_for_zephyr_elf} PROPERTIES OUTPUT_NAME ${KERNEL_NAME})
 
+if (IMAGE STREQUAL mcuboot_ AND CONFIG_MCUBOOT_BUILD_S1_VARIANT)
+# Create second MCUBoot executable for the second slot of the second stage
+# bootloader. This is done inside this file since it is non-trivial to add
+# executable targets outside the root CMakeLists.txt.
```
SebastianBoe: We need to elaborate on, in the comment, why this must be in the root so that we can move it if it happens to become easier in the future.

hakonfam: Agree. Also Perhaps try to come up with an upstream solution for this (create executable from anywhere)

hakonfam: Extended with 

```
#  ... The problem is that
# most of the variables requried for compiling/linking is only available
# in the scope of this file.

hakonfam:
Ping @ioannisg

PR #187, opened 2019-08-26
lwm2m: support socket_offload in hostname resolution

junqingzou:
Required in the context of nRF91 modem used
Resolution is implemented in nRF BSD lib

DNM: this PR is raised for internal discussion


carlescufi:
Thanks for the PR @junqingzou . I think this probably belongs directly upstream, but I will let @rlubos and @mike-scott decide.

junqingzou:
@mike-scott I've pushed to Zephyr master [PR#18690](https://github.com/zephyrproject-rtos/zephyr/pull/18690). Thanks!

PR #189, opened 2019-08-30
cmake: fix incorrect path in wrapper target

hakonfam:
The dependency in the wrapper target 's1_image_hex' did not
point to the correct path. Fix this to make the target
work correctly as a dependency.


hakonfam:
Ping @SebastianBoe - should be ready to merge

hakonfam:
cc @oyvindronningstad  @sigvartmh

PR #192, opened 2019-09-03
[nrf toup] drivers/flash: limit secure read inclusion

hakonfam:
Only include secure read call if SPM is enabled and partition manager is
in use.

### DIFF
--- a/drivers/flash/soc_flash_nrf.c
+++ b/drivers/flash/soc_flash_nrf.c
@@ -16,10 +16,12 @@
 #include <string.h>
 #include <nrfx_nvmc.h>
 
-#if CONFIG_ARM_NONSECURE_FIRMWARE
+#if CONFIG_ARM_NONSECURE_FIRMWARE && CONFIG_SPM
 #include <secure_services.h>
+#if USE_PARTITION_MANAGER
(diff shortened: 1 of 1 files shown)


nvlsianpu:
Looks good, will merge after upmerge.

lemrey:
This works but I suggest to `&&` the macros and `#ifdef` once, for readability.

hakonfam:
> This works but I suggest to `&&` the macros and `#ifdef` once, for readability.

fxed

nvlsianpu:
> This works but I suggest to && the macros and #ifdef once, for readability.

@hakonfam I missed that, better to use #ifdef as a define might doesn't exist while is =n in Kconfig.

hakonfam:
I did `#if` for now as that makes it easier to read (no split line). The only difference from using `#ifdef` is that if, for some reason, the `autoconf.h` file writes `CONFIG_BOOTLOADER_MCUBOOT=n` instead of simply `/* CONFIG_BOOTLOADER_MCUBOOT is not set*/` then the `#if` would still work.

nvlsianpu:
^^ clarified off-channel

PR #194, opened 2019-09-05
[nrf toup] subsys/testsuite: make tc_util overridable

torsteingrindvik:
**What**:
A Kconfig variable (under menuconfig->Testing) is added.
Checking this prompts subsys/testsuite/include/tc_util.h to look for tc_util_user_override.h.
This user override include can define various things in tc_util.h to control how output is produced during execution of test suites.

If not checked, no changes are made to current projects.

**Why**:
Some test suites with many tests (hundreds) are overly verbose---this feature allows projects to tune the output as necessary.

Signed-off-by: Torstein Grindvik <torstein.grindvik@nordicsemi.no>

### DIFF
--- a/subsys/testsuite/include/tc_util.h
+++ b/subsys/testsuite/include/tc_util.h
@@ -16,7 +16,7 @@
 #include <sys/printk.h>
 
 #if defined CONFIG_ZTEST_TC_UTIL_USER_OVERRIDE
-#include "tc_util_user_override.h"
+#include <tc_util_user_override.h>
 #endif
 
 #ifndef PRINT_DATA
@@ -66,27 +66,25 @@
 #define TC_FAIL 1
 #define TC_SKIP 2
 
+#ifndef  TC_PASS_STR
+#define TC_PASS_STR "PASS"
+#endif
+#ifndef  TC_FAIL_STR
+#define TC_FAIL_STR "FAIL"
+#endif
+#ifndef  TC_SKIP_STR
+#define TC_SKIP_STR "SKIP"
+#endif
+
 static inline const char *TC_RESULT_TO_STR(int result)
 {
 	switch (result) {
 	case TC_PASS:
-		#ifndef TC_PASS_STR
-		return "PASS";
-		#else
 		return TC_PASS_STR;
-		#endif
 	case TC_FAIL:
-		#ifndef TC_FAIL_STR
-		return "FAIL";
-		#else
 		return TC_FAIL_STR;
-		#endif
 	case TC_SKIP:
-		#ifndef TC_SKIP_STR
-		return "SKIP";
-		#else
 		return TC_SKIP_STR;
-		#endif
 	default:
 		return "?";
 	}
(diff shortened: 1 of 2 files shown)


nvlsianpu:
looks reasonable. Can you fix CI and open PR upstream as well?
cc @rakons

subsys/testsuite/ztest/Kconfig
```diff
@@ -41,6 +41,13 @@ config ZTEST_MOCKING
 	  Enable mocking support for Ztest. This allows the test to set
 	  return values and expected parameters to functions.
 
+config ZTEST_TC_UTIL_USER_OVERRIDE
+       bool "Override tc_util.h"
+       depends on ZTEST
+       help
+               Enable overriding some defines in tc_util.h.
```
nvlsianpu: Would be great to point any documentation which should helps to get know what should be provided as the tc_util_user_override.h. file content. Also on which exactly directory the file has to be located.

nvlsianpu: ^^ doc might be even a header template file, or a header in an example which shows this feature

torsteingrindvik: > 
> 
> Would be great to point any documentation which should helps to get know what should be provided as the tc_util_user_override.h. file content. Also on which exactly directory the file has to be located.

I updated the help section with more details. Should I change it further?

torsteingrindvik: > 
> 
> ^^ doc might be even a header template file, or a header in an example which shows this feature

I am working on a PR for fw-nrfconnect-nrf which will be using this feature, so at a later stage I could point to it.

Or I could just provide a template header. In that case do you have a suggestion on where to put it?

nvlsianpu: I would add and extra simple sample to upstream which shows usage and has template you can point. Otherwise put it to ztest includes somewhere.

subsys/testsuite/include/tc_util.h
```diff
@@ -62,41 +70,69 @@ static inline const char *TC_RESULT_TO_STR(int result)
 {
 	switch (result) {
 	case TC_PASS:
+		#ifndef TC_PASS_STR
 		return "PASS";
+		#else
+		return TC_PASS_STR;
+		#endif
```
SebastianBoe: Refactor so that it is just

return TC_PASS_STR;

inside of the case statement.

subsys/testsuite/include/tc_util.h
```diff
@@ -15,7 +15,13 @@
 #include <shell/shell.h>
 #include <sys/printk.h>
 
+#if defined CONFIG_ZTEST_TC_UTIL_USER_OVERRIDE
+#include "tc_util_user_override.h"
```
SebastianBoe: Change to from " to <.

" includes should only be used when the include path is relative from the file being processed.

torsteingrindvik:
Closing and creating PR upstream.

PR #198, opened 2019-09-13
cmake: partition_manager: define pm.yml for settings subsystem

hakonfam:
This pm.yml defines a partition 'settings_storage'.


nvlsianpu:
LGTM - need to clarify/fix CI failure (the file is in the tree...)

subsys/settings/pm.yml
```diff
@@ -0,0 +1,5 @@
+#include <autoconf.h>
```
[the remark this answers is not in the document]
hakonfam: As far as I can tell, it should be shared. Then it is up to the settings module to support multiple users, which I guess it does. But as you say, we should check that with someone more experienced with the settings module.

subsys/settings/pm.yml
```diff
@@ -0,0 +1,5 @@
...
+
+storage:
```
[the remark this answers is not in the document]
hakonfam: I guess so, there was an issue with other components using this partition, but then I guess those should define their own.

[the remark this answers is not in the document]
hakonfam: Fixed

subsys/fs/pm.yml
```diff
@@ -0,0 +1,7 @@
+#include <autoconf.h>
+
+#ifdef CONFIG_FILE_SYSTEM_NFFS
```
oyvindronningstad: Is this needed? I would guess it's covered by the corresponding `if` in the CmakeLists.txt.

hakonfam: true, but it shows how to handle multiple subsystems defined in `fs`. That way each of them can configure in their own partition.

oyvindronningstad: In that case, should the other be ` if (CONFIG_FILE_SYSTEM_NFFS)` or something?

hakonfam: Not sure I get what you mean.  

One example could be
```
#ifdef CONFIG_FILE_SYSTEM_NFFS
nffs_storage:
  blabalbla
#endif

#ifdef CONFIG_FILE_SYSTEM_LITTLEFS
littlefs_storage:
  something: blabal
#endif
```

Does that answer your question?

oyvindronningstad: Sorry, I meant `if (CONFIG_FILE_SYSTEM_FS)`

hakonfam: Correct

subsys/fs/CMakeLists.txt
```diff
@@ -16,6 +16,16 @@ if(CONFIG_FILE_SYSTEM)
   target_link_libraries_ifdef(CONFIG_FAT_FILESYSTEM_ELM   FS INTERFACE ELMFAT)
   target_link_libraries_ifdef(CONFIG_FILE_SYSTEM_LITTLEFS FS INTERFACE LITTLEFS)
   target_link_libraries_ifdef(CONFIG_FILE_SYSTEM_NFFS     FS INTERFACE NFFS)
+
+  if (CONFIG_FILE_SYSTEM_NFFS)
```
oyvindronningstad: If multiple filesystems should be supported as stated in the comment below, I would say this `if` should be removed.

hakonfam: agree

PR #201, opened 2019-09-17
[nrf noup] linker: add symbol for image 1 primary slot

hakonfam:
This symbol is used to ensure that s0 mcuboot points to s1 as
image 1 primary slot, and s1 mcuboot points to s0 as image 1
primary slot.

### DIFF
--- a/include/arch/arm/cortex_m/scripts/linker.ld
+++ b/include/arch/arm/cortex_m/scripts/linker.ld
@@ -43,10 +43,18 @@
 #if USE_PARTITION_MANAGER
 #include <pm_config.h>
 #ifdef PM_ADDRESS_LOAD_OFFSET
+/* We are linking mcuboot against S1, S0 is Image 1 primary */
+_image_1_primary_slot_id = PM_S0_ID;
+
 #define ROM_ADDR PM_ADDRESS + PM_ADDRESS_LOAD_OFFSET
 #else
+#ifdef PM_S1_ID
+/* We are linking mcuboot against S0, S1 is Image 1 primary */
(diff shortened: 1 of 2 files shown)


include/arch/arm/cortex_m/scripts/linker.ld
```diff
@@ -43,10 +43,18 @@
 #if USE_PARTITION_MANAGER
 #include <pm_config.h>
 #ifdef PM_ADDRESS_LOAD_OFFSET
+/* We are linking mcuboot against S1, Image 0 is S0 */
```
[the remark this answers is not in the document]
hakonfam: You are correct that something is off here. 

S0/S1 is Image 1, because Image 0 is what is booted by mcuboot. 

So it should be `Image 1 is Sx`

include/arch/arm/cortex_m/scripts/linker.ld
```diff
@@ -43,10 +43,18 @@
 #if USE_PARTITION_MANAGER
 #include <pm_config.h>
 #ifdef PM_ADDRESS_LOAD_OFFSET
+/* We are linking mcuboot against S1, S0 is Image 1 primary */
+_image_1_primary_slot_id = PM_S0_ID;
```
sigvartmh: Where is the code which uses this symbol located?

hakonfam: It will be in mcuboot

sigvartmh: Okei.

PR #204, opened 2019-09-27
[nrf noup] net: socket: add SO_DFU_ERROR socket option

lemrey:
Add SO_DFU_ERROR socket option for DFU socket.
The option can be used when a operation on a DFU socket
returns -ENOEXEC, indicating that the modem has rejected
the operation, to retrieve the reason for the error.

Signed-off-by: Emanuele Di Santo <emdi@nordicsemi.no>


lemrey:
This PR source branch is on the remote, which was not intended.
I am closing this.

PR #206, opened 2019-09-27
DNM: Revert subsys partition

hakonfam:


carlescufi:
Do not merge yet

carlescufi:
These commits are at the top and no one is referencing them yet. So please do not merge yet, let's talk about it first on Monday.

SebastianBoe:
We have decided to re-write the history to not have these commits instead of reverting them.

PR #209, opened 2019-10-01
[nrf noup] cmake: add function add_overlay_conf to extensions.cmake

hakonfam:
This function adds the specified config path to the 'OVERLAY_CONFIG'
variable of the given image name. The function will ensure that
the config path will not be added twice.


hakonfam:
Will fix by adding FW_METADATA=y to prj.conf directly instead.

PR #210, opened 2019-10-02
NCSDK-NONE Run all sanitychecks on every PR using zephyr compiler.

LeHack:
Signed-off-by: Łukasz Hejnak <lukasz.hejnak@nordicsemi.no>


mbolivar:
This should have had a sauce tag.

PR #219, opened 2019-10-17
[nrf fromlist] cmake: zephyr_library_amend feature

tejlmand:
This PR allows amending to a Zephyr library created inside Zephyr repo in a generic way.

Currently, if no files are added to a library, as example the drivers/entropy zephyr library because no drivers (c-files) are selected using Kconfig, the following error when running cmake / menuconfig:

```
CMake Error at /projects/github/ncs/zephyr/cmake/extensions.cmake:411 (add_library):                       
  No SOURCES given to target: drivers__entropy
  /projects/github/ncs/zephyr/cmake/extensions.cmake:388 (zephyr_library_named)
  /projects/github/ncs/zephyr/drivers/entropy/CMakeLists.txt:3 (zephyr_library)
```

This PR allows to amend drivers files out-of-tree to an existing zephyr library.

As example:
`drivers/entropy/CMakeLists.txt` creates a zephyr library as:
```
 zephyr_library()
```
but due to kconfig settings, no files are added to the lib in Zephyr.
   
The amend function allows to amend to such a lib, by creating a
CMakeLists.txt file following identical folder structure in a Zephyr Module:
```
    <zephyr_module_oot>/drivers/entropy/CMakeLists.txt
    zephyr_library_amend()
    zephyr_library_sources() # Sources are amended to the original library
```
which means that the library created in Zephyr is no longer empty.

### DIFF
--- a/cmake/extensions.cmake
+++ b/cmake/extensions.cmake
@@ -441,7 +441,7 @@ macro(zephyr_library_amend)
 
   zephyr_library_get_current_dir_lib_name(${ZEPHYR_CURRENT_MODULE_DIR} lib_name)
 
-  set(ZEPHYR_CURRENT_LIBRARY ${lib_name})
+  set(ZEPHYR_CURRENT_LIBRARY ${IMAGE}${lib_name})
 endmacro()
 
 function(zephyr_link_interface interface)
--- a/cmake/extensions.cmake
+++ b/cmake/extensions.cmake
@@ -384,15 +384,15 @@ endmacro()
 
 # Constructor with a directory-inferred name
 macro(zephyr_library)
-  zephyr_library_get_current_dir_lib_name(lib_name)
+  zephyr_library_get_current_dir_lib_name(${ZEPHYR_BASE} lib_name)
   zephyr_library_named(${lib_name})
 endmacro()
 
-# Determines what the current directory's lib name would be and writes
-# it to the argument "lib_name"
-macro(zephyr_library_get_current_dir_lib_name lib_name)
+# Determines what the current directory's lib name would be according to the
+# provided base and writes it to the argument "lib_name"
+macro(zephyr_library_get_current_dir_lib_name base lib_name)
   # Remove the prefix (/home/sebo/zephyr/driver/serial/CMakeLists.txt => driver/serial/CMakeLists.txt)
-  file(RELATIVE_PATH name ${ZEPHYR_BASE} ${CMAKE_CURRENT_LIST_FILE})
+  file(RELATIVE_PATH name ${base} ${CMAKE_CURRENT_LIST_FILE})
 
   # Remove the filename (driver/serial/CMakeLists.txt => driver/serial)
   get_filename_component(name ${name} DIRECTORY)
@@ -415,6 +415,34 @@ macro(zephyr_library_named name)
   target_link_libraries(${ZEPHYR_CURRENT_LIBRARY} PUBLIC ${IMAGE}zephyr_interface)
 endmacro()
 
+# Provides amend functionality to a Zephyr library for out-of-tree usage.
+#
+# When called from a Zephyr module, the corresponding zephyr library defined
+# within Zephyr will be looked up.
+#
+# Note, in order to ensure correct library when amending, the folder structure in the
+# Zephyr module must resemble the structure used in Zephyr, as example:
+#
+# Example: to amend the zephyr library created in
+# ZEPHYR_BASE/drivers/entropy/CMakeLists.txt
+# add the following file:
+# ZEPHYR_MODULE/drivers/entropy/CMakeLists.txt
+# with content:
+# zephyr_library_amend()
+# zephyr_libray_add_sources(...)
+#
+macro(zephyr_library_amend)
+  # This is a macro because we need to ensure the ZEPHYR_CURRENT_LIBRARY and
+  # following zephyr_library_* calls are executed within the scope of the
+  # caller.
+  if(NOT ZEPHYR_CURRENT_MODULE_DIR)
+    message(FATAL_ERROR "Function only available for Zephyr modules.")
+  endif()
+
+  zephyr_library_get_current_dir_lib_name(${ZEPHYR_CURRENT_MODULE_DIR} lib_name)
+
+  set(ZEPHYR_CURRENT_LIBRARY ${lib_name})
+endmacro()
 
 function(zephyr_link_interface interface)
   target_link_libraries(${interface} INTERFACE ${IMAGE}zephyr_interface)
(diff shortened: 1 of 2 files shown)


tejlmand:
@SebastianBoe Please review this.
The PR is placed here to allow this to be used for the upcoming release.

Please comment if you believe code should be changed to get it approved upstream later.

tejlmand:
> changelog needs to begin with `[nrf toup]`, please see https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/ug_dev_model.html#oss-repositories-downstream-project-history

Thanks for notifying.
[nrf toup] added.

thomasstenersen:
Some background. As I originally did this, the `BT_LL_NRFXLIB` selects `CONFIG_ENTROPY_NRF_FORCE_ALT` _and_ [adds its source to the Zephyr entropy library](https://github.com/NordicPlayground/fw-nrfconnect-nrf/blob/master/drivers/bt_ll_nrfxlib/entropy/CMakeLists.txt#L7). 
As discussed in the original PRs (https://github.com/zephyrproject-rtos/zephyr/pull/14509, https://github.com/zephyrproject-rtos/zephyr/pull/13631) we chose to go with a Nordic-only workaround.

@tejlmand :
- I'm a bit out of the loop here, what is the scenario where the library will be empty?
- Is there something else selecting this config without providing a source?

tejlmand:
> 
> 
> PR must be created upstream before this is merged, and @thomasstenersen should review.

yes, I know, but wanted to make this PR first to allow others and especially @thomasstenersen to provide feedback before going upstream.

tejlmand:
I our case with the cc310, an entropy driver will be provided internally in the fw-nrfconnect-nrf repo, similar to entropy_ll_nrfxlib.c.
I did consider to directly use the "magic" name `drivers__entropy` to place the file, but was under impression from @SebastianBoe at earlier talks that users are not supposed to directly use those names outside the zephyr CMake wrapped calls.
(I had not discovered you where already doing it that way)

If this approach is generally accepted (or at least accepted in nrf repo), I will add the cc310 entropy driver in same way.

SebastianBoe:
You are correct that one should not use the library name when it can be avoided, but for this use-case we were unable to avoid it.

tejlmand:
> 
> 
> You are correct that one should not use the library name when it can be avoided, but for this use-case we were unable to avoid it.

I'm happy to use the library name directly, but it is possible to avoid using the library name with this PR.
If we take this PR upstream, then the `zephyr_library` will not be created when `CONFIG_ENTROPY_FORCE_ALT=y` due to the safeguard in CMakeLists.txt, and instead it is possible `fw-nrfconnect-nrf` to do:
```
zephyr_library()
zephyr_library_sources(entropy_alt.c)
```
inside the `fw-nrfconnect-nrf` repo.

The principle in this PR will also allows a more generic approach for Zephyr community to provide alt implementation of entropy if they so wish.

@SebastianBoe and @thomasstenersen Please let me know if you think it has value to go ahead with this PR upstream for a more generic approach and to avoid directly using the lib name `drivers__entropy`, or if I should just follow same use-case as @thomasstenersen

tejlmand:
> 
> 
> Why this patch cannot go upstream as is? Is there any controversial content?

Due to the fact that we already did this in slightly different way internally, I just wanted to give people opportunity to give feedback first.

SebastianBoe:
I'm sceptical, what if we want entropy_handlers.c and also a different implementation of the nrf driver?

I think we should follow @thomasstenersen 's work.

thomasstenersen:
@tejlmand thanks for the context. It's not pretty and I myself would prefer your solution, really. But I don't think this will be accepted upstream until there are others than Nordic that needs this functionality (that was the argument last time).

mbolivar:
> But I don't think this will be accepted upstream until there are others than Nordic that needs this functionality (that was the argument last time).

If that is the case, please use `[nrf noup]` (noup != toup) so we can track it as such

tejlmand:
> > But I don't think this will be accepted upstream until there are others than Nordic that needs this functionality (that was the argument last time).
> 
> If that is the case, please use `[nrf noup]` (noup != toup) so we can track it as such

Completely agree, if this is not going upstream.
But I still expect to push this upstream.

tejlmand:
Thanks for all the comments and based on this, and especially this from @SebastianBoe 

> I'm sceptical, what if we want entropy_handlers.c and also a different implementation of the nrf driver?

So I have now made a more generic way of doing this.

I have reworked the PR, so read the updated description at the top.
@thomasstenersen @SebastianBoe please take a fresh look.

With the changes, it is now possible to create a `<NCS>/nrf/drivers/entropy/CMakeLists.txt` file with content:
```
zephyr_library_amend()
zephyr_library_sources_if_kconfig(entropy_cc310.c)
zephyr_library_sources_ifdef(CONFIG_ENTROPY_NRF_LL_NRFXLIB entropy_ll_nrfxlib.c)
```
@thomasstenersen for this to also work for `entropy_ll_nrfxlib` this file must be relocated, and secondly, if the Kconfig flag or the file is renamed, then you can also use the `if_kconfig` syntax instead.

cmake/extensions.cmake
```diff
@@ -415,6 +415,47 @@ macro(zephyr_library_named name)
   target_link_libraries(${ZEPHYR_CURRENT_LIBRARY} PUBLIC ${IMAGE}zephyr_interface)
 endmacro()
 
+# Determines what the current directory's lib name would be and writes
+# it to the argument "lib_name" as if ZEPHYR_BASE had been equal to
+# ZEPHYR_CURRENT_MODULE_DIR.
+macro(zephyr_library_amend_module_lib_name lib_name)
+  message(FATAL_ERROR "Function only available for Zephyr modules.")
```
SebastianBoe: Is this error missing an if-statement ?

SebastianBoe:
The approach LGTM, we just need a PR posted upstream and then we can merge this.

cmake/extensions.cmake
tejlmand: good spotted, thansk.
This was added at last minut, cause I wanted to safe guard accidential calls from Zephyr repo.
Guess it was done a little to fast.

tejlmand: Fixed

tejlmand:
> The approach LGTM, we just need a PR posted upstream and then we can merge this.

https://github.com/zephyrproject-rtos/zephyr/pull/19980

SebastianBoe:
The tag should be fromlist according to Marti.

mbolivar:
Thanks for bearing with me on the patch bookkeeping! These tags make it a lot easier to track patches by category, and help bring attention to patches which may have been merged upstream during mergeups.

SebastianBoe:
DNM until https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/220 is merged. It is policy to block PR's to fw-nrfconnect-zephyr when an upmerge PR is ongoing.

mbolivar:
> NACKing to block the merge

the mergeup is in, please revisit @SebastianBoe

thst-nordic:
The CI is failing because of QEMU bug that is on master also.    This is ok to merge.

PR #221, opened 2019-10-24
Subsys partitions

hakonfam:
see https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1337


mbolivar:
Zephyr history was rewritten in preparation for the next release. Please rebase this PR.

hakonfam:
@pdunaj  I verified that a change compared to last time fixed the issue you reported. Removing DNM

subsys/fs/pm.yml
```diff
@@ -0,0 +1,7 @@
+#include <autoconf.h>
+
+#ifdef CONFIG_FILE_SYSTEM_NFFS
+nffs_storage:
+  placement: {after: [mcuboot_storage, app]}
```
nvlsianpu: --mcuboot_storage, elsewhere

subsys/settings/Kconfig
```diff
@@ -16,6 +16,14 @@ if SETTINGS
 module = SETTINGS
 module-str = settings
 source "subsys/logging/Kconfig.template.log_config"
+
+# This config must be compatible with
+# nrf/subsys/partition_manager/Kconfig.template.partition_size
+config PM_PARTITION_SIZE_SETTINGS_STORAGE
```
nvlsianpu: What will happen if i introduce similar change for usage pure NVS/FCB into code?

subsys/fs/pm.yml
hakonfam: This will actually not break if mcuboot_storage is removed (and at this point it is not, as that PR is blocked by this). 

I will make these fixes in https://github.com/NordicPlayground/fw-nrfconnect-mcuboot/pull/65

subsys/settings/Kconfig
hakonfam: Then those would have to declare their own storage partitions. Which makes sense IMO, as the `settings_storage` partition is occupied by the settings subsystem.

nvlsianpu: Ok - I have tested local patch according to your gide - It have been working well.
I think need to add similar changes for each storage. I can do that in another patch.

PR #233, opened 2019-11-07
[nrf noup] revert: "943a8651224470b6594e"

SebastianBoe:
This reverts commit 943a8651224470b6594e992fd2c58a454540796a.

This commit should have been applied simultaneously with a commit in
nrf but it was not and is now causing problems.


SebastianBoe:
We can't apply the commit that 943 depends on because it is blocked by an obscure error in CI that I don't know how to reproduce or debug.

PR #237, opened 2019-11-15
Multi image

tejlmand:
This PR  drops the following commits, which are no longer needed with the new strategy for multiimage:
- 5aa1e0d5fd
- 1865dcf370
- 29b04065fc
- 47e991060b
- bbcbd8f016
- 9e7a9516a24

Thus in order to get this PR in, we must tag and rewrite history.


mbolivar:
Will we rewrite history for this patch series, or not? @tejlmand what would be easier for you?

tejlmand:
> 
> 
> Will we rewrite history for this patch series, or not? @tejlmand what would be easier for you?

@mbolivar it doesn't matter to me.
Both of them are just as easy to do for me, so I think it matters what suits the normal upmerge strategy best.

tejlmand:
Closing, as superseded by PR#240

PR #239, opened 2019-11-18
Ncs nrf53

ioannisg:
Contains the following patches
- From tree:
   - Enable DCDC
   - Fix CACHE and NVMC macros
   - [BLE] buffer init after allocation
- From list:
   - Fix partitions for nRF5340 CPU NET
   - **HCI over openAMP For nRF5340** (squashed commits take from https://github.com/zephyrproject-rtos/zephyr/pull/20467)
- Revert Local temp-hacks:
   - Remove setting the default controller to LEGACY


ioannisg:
@kapi-no can you test that the setup still works now that we moved this to NCS?

kapi-no:
@ioannisg, yes, sure.

Could you open matching PR to the nrf repository (with west manifest pointing to this PR)?

PR #240, opened 2019-11-19
Multi image commit revert

tejlmand:
This PR supersedes PR#237.
It has been rebased against latest master after upmerge, and now uses `git revert` instead of dropping commits.


mbolivar:
> [ @tejlmand ](/tejlmand)
> ` [[nrf noup] cmake: Adopted to new multi image approach](/NordicPlayground/fw-nrfconnect-zephyr/pull/240/commits/c9c368ca1f54fca1fcee4b2ad6d612de98a5fd11) `

Just recording some discussion from a chat here:

It looks like this commit was only necessary because some other commits mixed in fixes to manipulate IMAGE prefixing with other changes which we still want to preserve. So we could get rid of it by reverting those and reapplying them without the IMAGE changes.

mbolivar:
This is fantastic! Thanks @tejlmand !

arch/arm/core/cortex_m/tz/CMakeLists.txt
```diff
@@ -27,11 +27,6 @@ if(CONFIG_ARM_FIRMWARE_HAS_SECURE_ENTRY_FUNCS)
     )
 endif()
 
-# Link the entry veneers library file with the Non-Secure Firmware that needs it.
-zephyr_link_libraries_ifdef(CONFIG_ARM_FIRMWARE_USES_SECURE_ENTRY_FUNCS
-  ${CMAKE_BINARY_DIR}/${CONFIG_ARM_ENTRY_VENEERS_LIB_NAME}
-  )
-
```
tejlmand: Not sure what your intentions are with this temphack.

If we disregard the `CONFIG_ARM_FIRMWARE_USES_SECURE_ENTRY_FUNCS` and do not set it, this code will not be activated.

Instead we could do similar to what was earlier prototyped here:
https://github.com/tejlmand/fw-nrfconnect-nrf/blob/da2af5e2ca77d12f4f20295fbfeac9ff47c36c04/subsys/spm/Kconfig#L13
https://github.com/tejlmand/fw-nrfconnect-nrf/blob/39007f4caf2e610ca8e745a27e3aa01288745e41/cmake/multi_image.cmake#L145-L151

of course with the addition of allowing the user to not link with `veneers` if desired.
That way we don't need a temphack locally.

CMakeLists.txt
```diff
@@ -1295,6 +1295,7 @@ if (CONFIG_MCUBOOT AND CONFIG_MCUBOOT_BUILD_S1_VARIANT)
 
   add_custom_target(
     s1_image_hex
+    ALL
```
tejlmand: Should we include this as a revert + reapply of this commit: bfd1ae364c
and then include this change as part of the reapply commit ?
I believe this would minimize the numbers of [nrf noup] commit to handle during an upmerge.

@mbolivar guess as the master of upmerges you have an opinion on this.

arch/arm/core/cortex_m/tz/CMakeLists.txt
SebastianBoe: I can't recall my plans for how to get rid of the temp hack, I will look into this.

I don't want to unset CONFIG_ARM_FIRMWARE_USES_SECURE_ENTRY_FUNCS because that would be incorrect, the image does use secure entry functions.

CMakeLists.txt
SebastianBoe: I'm unable to tell if that makes sense or not.

arch/arm/core/cortex_m/tz/CMakeLists.txt
SebastianBoe: I got rid of the temp hack now.

The default string is now changed based on whether the image is using SPM or is SPM:

https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1508/files#diff-4098c32be5d8477048a5e432e1d5066cR225

CMakeLists.txt
mbolivar: @SebastianBoe -- I think we agreed yesterday that we'd work together to get this in and then clean it up at rebase time. Is that right? If so, I think you two should just do something that feels reasonable to you and we can decide what happens later.

tejlmand: @mbolivar @SebastianBoe Agreed. Let's keep as is for now.

arch/arm/core/cortex_m/tz/CMakeLists.txt
tejlmand: > I got rid of the temp hack now.

Great.

tejlmand:
Great work.
Approved when CI is green.

@SebastianBoe can you retrigger CI, it seems to have failed with disk space:
```
OSError: [Errno 28] No space left on device
```

tejlmand:
Approved.

PR #244, opened 2019-11-20
[nrf noup] west: multi-image: Change variable used to find app dir

SebastianBoe:
APPLICATION_SOURCE_DIR is not in the cache due to multi-image
requiring cache variables to be image-prefixed and due to a desire to
not image-prefix this variable.

But west expects it to be so. To resolve this, have west use a
different variable that points to the same thing.

### DIFF
--- a/scripts/west_commands/build.py
+++ b/scripts/west_commands/build.py
@@ -293,8 +293,8 @@ class Build(Forceable):
         if not self.cmake_cache:
             return          # That's all we can check without a cache.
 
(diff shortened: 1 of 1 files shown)


SebastianBoe:
see https://github.com/NordicPlayground/fw-nrfconnect-nrf/pulls

mbolivar:
I'm OK with this as an nrf temphack

SebastianBoe:
Do you want me to change the tag? The patch will be removed when multi-image is migrated.

PR #245, opened 2019-11-22
Fix s1 variant

SebastianBoe:
This PR makes the attached code compatible with both ncs master and with the upcoming re-write of multi-image.

This is done to reduce the size of the re-write PR.

Details of the changes are in the commit messages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1278,7 +1278,7 @@ if (CONFIG_MCUBOOT AND CONFIG_MCUBOOT_BUILD_S1_VARIANT)
   # Create linker script which has an offset from S0 to S1.
   configure_linker_script(
     linker_mcuboot_s1.cmd
-    "-DPM_ADDRESS_LOAD_OFFSET=($<TARGET_PROPERTY:partition_manager,PM_S1_ADDRESS>-$<TARGET_PROPERTY:partition_manager,PM_S0_ADDRESS>);-DLINKER_PASS2"
+    "-DLINK_MCUBOOT_INTO_S1;-DLINKER_PASS2"
     ${PRIV_STACK_DEP}
     ${CODE_RELOCATION_DEP}
     ${ZEPHYR_PREBUILT_EXECUTABLE}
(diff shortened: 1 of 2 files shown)


tejlmand:
Just noticed.
Although this is a small patch, would it be preferable to do a revert of: 
https://github.com/NordicPlayground/fw-nrfconnect-zephyr/commit/bfd1ae364c01fbe576d36ee21af00cdeb87feb6c

and then a re-apply, including the proposed changes as a single [nrf noup] ?
In order to keep the number of [nrf noup] patches as small as possible.

Afterall, this PR introduces 3 new [nrf noup] + plus the existing.

@mbolivar thoughts ?

SebastianBoe:
I don't have a good enough understanding of the upmerge procedure to say if this makes sense or not.

tejlmand:
> 
> 
> I don't have a good enough understanding of the upmerge procedure to say if this makes sense or not.

me neither, that's why I asked for @mbolivar comments ;)

mbolivar:
> > I don't have a good enough understanding of the upmerge procedure to say if this makes sense or not.
> 
> me neither, that's why I asked for @mbolivar comments ;)

If you want to keep the west ncs-loot output clean, revert and reapply is nice. But for something like this, where we're basically replacing one patch series with another, I think I might just do a bunch of squashing at the next rebase anyway, so don't worry too much about reverts if they are really slowing you down. Just please keep commits logically separate so they can be rebased around later and I'll be grateful :).

mbolivar:
> Although this is a small patch, would it be preferable to do a revert of:
> [bfd1ae3](https://github.com/NordicPlayground/fw-nrfconnect-zephyr/commit/bfd1ae364c01fbe576d36ee21af00cdeb87feb6c)


FTR, looking at that patch again, I see this comment:

```
# Create second MCUBoot executable for the second slot of the second stage
# bootloader. This is done inside this file since it is non-trivial to add
# executable targets outside the root CMakeLists.txt. The problem is that
# most of the variables requried for compiling/linking is only available
# in the scope of this file.
```

Above, "this file" is zephyr/CMakeLists.txt. I think this is evidence that defining the zephyr executable in there instead of boilerplate forces users to edit this file when they might prefer not to. It's certainly a layering violation.

SebastianBoe:
>  I think this is evidence that defining the zephyr executable in there instead of boilerplate forces users to edit this file when they might prefer not to. It's certainly a layering violation.

You might be right about that, I should look into this later.

PR #250, opened 2019-12-13
net: mqtt: Maintain count of PINGREQ awaiting response

jbrzozoski:
Maintain a simple count of how many PINGREQ have been sent for the
current connection that have not had a corresponding PINGRESP.  Nothing
is done with this information internal to the MQTT driver, but it is
exposed to the application layer to monitor as desired.

Signed-off-by: Justin Brzozoski <justin.brzozoski@signal-fire.com>

### DIFF
--- a/include/net/mqtt.h
+++ b/include/net/mqtt.h
@@ -499,6 +499,9 @@ struct mqtt_client {
 	/** MQTT protocol version. */
 	u8_t protocol_version;
 
+	/** Unanswered PINGREQ count on this connection. */
+	s8_t unacked_ping;
+
 	/** Will retain flag, 1 if will message shall be retained persistently.
 	 */
 	u8_t will_retain : 1;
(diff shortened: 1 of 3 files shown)


jbrzozoski:
I wasn't sure if it made more sense to do the pull request to you guys or directly to zephyrproject.  I went this way since it was closer to something I could actually build/test.  If you want me to open a pull request up there instead, let me know and I will.

jtguggedal:
>  wasn't sure if it made more sense to do the pull request to you guys or directly to zephyrproject. I went this way since it was closer to something I could actually build/test. If you want me to open a pull request up there instead, let me know and I will.

Thanks for the PR, @jbrzozoski! I think you could direct it upstream right away and save a couple of days of delay in review :) Rationale for going directly upstream is that it touches MQTT lib in a way that's not specific to Nordic devices and could be useful to other Zephyr users as well.

jbrzozoski:
Heading upstream...

PR #255, opened 2020-01-08
Post v2.1 history rewrite

mbolivar:
This Zephyr tree is the same code as the current manifest-rev  (9a0a85cf67602e4fa5b0c6d7cef2c5e58ca4ac34), but with history cleaned up and rewritten.

### DIFF
--- a/drivers/watchdog/wdt_nrfx.c
+++ b/drivers/watchdog/wdt_nrfx.c
@@ -181,13 +181,13 @@ static void wdt_event_handler(struct device *dev)
 			.reload_value  = 2000,				       \
 		}							       \
 	};								       \
-	DEVICE_DEFINE(wdt_##idx, DT_NORDIC_NRF_WATCHDOG_WDT_##idx##_LABEL,     \
-		      wdt_##idx##_init,					       \
-		      NULL,						       \
-		      &wdt_##idx##_data,				       \
-		      &wdt_##idx##z_config,				       \
-		      PRE_KERNEL_1, CONFIG_KERNEL_INIT_PRIORITY_DEVICE,	       \
-		      &wdt_nrfx_driver_api)
+	DEVICE_AND_API_INIT(wdt_##idx,					       \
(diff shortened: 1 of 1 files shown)


mbolivar:
> current manifest-rev ([9a0a85c](https://github.com/NordicPlayground/fw-nrfconnect-zephyr/commit/9a0a85cf67602e4fa5b0c6d7cef2c5e58ca4ac34)),

@SebastianBoe Please note that manifest-rev does not include your latest commit 695ce53b78, so that will be removed from history since it was never added to nrf/west.yml. We will need to add it again, and hopefully this time add it to west.yml, to get it back into use.

thst-nordic:
CI doesn't like that there is no CODEOWNER for Jenkinsfile.  I will add myself.

mbolivar:
> CI doesn't like that there is no CODEOWNER for Jenkinsfile. I will add myself.

Please don't do that in this PR. We need the code to be 100% the same, which it is. We need to force push anyway so this will be put into master by hand. We can fix CODEOWNERS later.

oyvindronningstad:
> We need to force push anyway so this will be put into master by hand.

Side note: If you first reset master to zephyr-v2.1.0 and force-push, you could merge this PR the regular way right?

mbolivar:
> Side note: If you first reset master to zephyr-v2.1.0 and force-push, you could merge this PR the regular way right?

In terms of getting this PR in without a force-push, yes, we could. That would break the rules in our development model, though.

When we rewrite history, we tag the old commit (in this case, 9a0a85c) and then force push the same code with different history (this PR) into master. That's as close as we can get to atomically giving users and developers a new commit to do conflict-free rebases onto across the change, and is why we block merges into repositories whose histories are being rewritten.

mbolivar:
Thanks for the review @oyvindronningstad -- would anyone else like to review before I go ahead and push this in?

I will confirm at our next regular meeting before pushing.

PR #257, opened 2020-01-14
[nrf noup] ci: adding common cron configuration

thst-nordic:
- new cron features
- separate logs for sub-jobs
- more options for downstream jobs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,12 +1,25 @@
+//
+// Copyright (c) 2020 Nordic Semiconductor ASA. All Rights Reserved.
+//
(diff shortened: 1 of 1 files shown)


thst-nordic:
the Jenkinsfile changes are the same as NordicPlayground/fw-nrfconnect-nrf#1711 and NordicPlayground/fw-nrfconnect-mcuboot#75

and
adding this change to the NRF manifest immediately isn't needed.

mbolivar:
I'm not sure if @anangl has locked the repository for merges yet during the mergeup periods, so I think I should let you two decide whether we can merge this now.

PR #259, opened 2020-01-15
Revert "[nrf fromlist] tests/settings/functional/file: fix nRF targets"

mbolivar:
This reverts commit 171d34b1a4eb69f9bfe1cfc71a7455b923256a7f,
which is no longer needed since NFFS is pending removal.

@nvlsianpu I assume you are OK with just reverting this, since you've dropped https://github.com/zephyrproject-rtos/zephyr/pull/20208#issuecomment-574612765?


mbolivar:
@anangl Please feel free to simply close this if you're handling this in the mergeup already.

mbolivar:
> I belief it should be kept as [nrf temhack] as NFFS will be removed soon, but this patch was aimed for fix CI results.

OK, thanks for replying. I'll close this and we can revisit after NFFS is gone from our downstream tree.

PR #270, opened 2020-01-31
[nrf temphack] dfu: fix partition manager / mcuboot interaction

mbolivar:
The removal of the scratch partition has broken the build for asset
tracker. There's probably a better way to fix it than this, but let's
get the build working again and figure it out.


mbolivar:
> Looks good. This is needed since the default for NRF is to use a swap operation which does not require a scratch partition.

@hakonfam thanks for review (and sorry we merged it so quickly, but CI was down and I was just trying to get the build unbroken).

Would you be OK if I squash this into the tree-wide partition manager support commit when I rebase zephyr for the 1.2 release?

hakonfam:
> Would you be OK if I squash this into the tree-wide partition manager support commit when I rebase zephyr for the 1.2 release?

Sure :) As this was inherited from upstream there is no proper downstream commit to append it to, so your proposal makes sense.

PR #271, opened 2020-02-07
[nrf noup] ci: python requirement updates

thst-nordic:
- Install ci-tools requirementx.txt before check_compliance.py
-- missing python-magic

- edit python requirement versions
-- setting sphinx!=2.4.0
-- (sphinx:2.4.0 released yesterday with a bug that breaks the build)

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -16,12 +16,12 @@ pykwalify
 pyocd>=0.24.0
 pyserial
 pytest
-sphinx>=1.7.5
+sphinx>=1.7.5, != 2.4.0
 sphinx_rtd_theme
 sphinx-tabs
 sphinxcontrib-svg2pdfconverter
 tabulate
-west>=0.6.2
+west>=0.6.3, !=0.7.*
 wheel>=0.30.0
 # "win32" is used for 64-bit Windows as well
 windows-curses; sys_platform == "win32"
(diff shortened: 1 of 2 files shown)


thst-nordic:
tested here: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1874

scripts/requirements.txt
```diff
@@ -1,27 +1,31 @@
...
 sphinxcontrib-svg2pdfconverter
 tabulate
-west>=0.6.2
-wheel>=0.30.0
+<<<<<<< HEAD
+west==0.6.3
+=======
+west==0.6.2
+>>>>>>> 1962abe7e1... [nrf noup] ci: lock python requirement versions
+wheel==0.34.2
```
tejlmand: This looks like a broken rebase / merge to me.

thst-nordic: should be fixed now

scripts/requirements.txt
```diff
@@ -1,27 +1,27 @@
 Pillow
-PyYAML>=5.1
+PyYAML==5.3
```
mbolivar: Can you please do this in a separate file (requirements-nrf-ci.txt or something) instead?

mbolivar: I mean not just this line, but the entire commit, just to be explicit.

thst-nordic: I understand this will cause conflicts in the upmerge.    But these changes are not just for CI.  The floating versions cause an issue for customers also.

thst-nordic: removed this change, we will look at other tools to solve this issue after the release.   (pipenv?)

PR #272, opened 2020-02-10
[nrf noup] ci: lock python requirement versions

thst-nordic:
sphinx==2.4.0 released yesterday with a bug that breaks the build

setting sphinx==2.3.0


thst-nordic:
moved this patch into #271

PR #274, opened 2020-02-12
[RFC] Doc/smp svr ncs building 

nvlsianpu:
Building and running smp_svr is much easier in NCS infrastructure
than in vanilla zephyr.
Added documentation for shows how to build with NCS.


samples/subsys/mgmt/mcumgr/smp_svr/README.rst
```diff
@@ -47,6 +47,14 @@ Bluetooth Low Energy (BLE) and do not have a built-in or pluggable BLE radio,
...
 
+Building and Running using NCS
+******************************
+Building the sammple as a part off nRF Connect SDK is simplified in comparasion
+to the vanilla zephyr sample.
+
+.. include:: /includes/build_and_run.txt
+
+
 Building and Running
```
nvlsianpu: I would remove original zephyr build and Running section.
Or maybe It would be better to rename it to Building and Running (vanilla zephyr) or something which reflect that this is zephyr base build manual.
What are yours opinion?

samples/subsys/mgmt/mcumgr/smp_svr/README.rst
```diff
@@ -47,6 +47,14 @@ Bluetooth Low Energy (BLE) and do not have a built-in or pluggable BLE radio,
 you can build one and use it following the instructions in
 :ref:`bluetooth-hci-uart-bluez`.
 
+Building and Running using NCS
+******************************
+Building the sammple as a part off nRF Connect SDK is simplified in comparasion
```
mbolivar:
```suggestion
Building the sample using the nRF Connect SDK is simpler than using the
mainline Zephyr procedure.
```

samples/subsys/mgmt/mcumgr/smp_svr/README.rst
mbolivar: I think we should keep it, at the very least to avoid merge conflicts. @ru-fu ?

samples/subsys/mgmt/mcumgr/smp_svr/README.rst
```diff
@@ -47,6 +47,14 @@ Bluetooth Low Energy (BLE) and do not have a built-in or pluggable BLE radio,
...
+to the vanilla zephyr sample.
+
+.. include:: /includes/build_and_run.txt
```
hakonfam: Is this file missing, or is it in the nrf repo?

ru-fu: This file is in the nrf repository, so this won't work unfortunately, and you'll need to include the text manually. :(
We also don't have automatic linking set up this way around (you can link from the NCS docs to the Zephyr docs, but not from Zephyr to NCS), so the only way to link will be to use an absolute link ( https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/gs_programming.html#gs-programming ) - which will point to master and not to a specific release, and it'll point to the online doc even when users build the docs locally. I think we can live with those problems though.

samples/subsys/mgmt/mcumgr/smp_svr/README.rst
ru-fu: No strong opinion here. If we want people to use the NCS way of building, we should remove this section. But if that leads to constant merge conflicts, I don't mind keeping it.

samples/subsys/mgmt/mcumgr/smp_svr/README.rst
```diff
@@ -47,6 +47,14 @@ Bluetooth Low Energy (BLE) and do not have a built-in or pluggable BLE radio,
 you can build one and use it following the instructions in
 :ref:`bluetooth-hci-uart-bluez`.
 
+Building and Running using NCS
```
ru-fu:
```suggestion
Building using the nRF Connect SDK
```
We shouldn't use the NCS abbreviation in the docs.

PR #275, opened 2020-02-12
[nrf noup] linker: use partition manager to set RAM

hakonfam:
Partition manager now supports partitioning RAM as well as FLASH.

Use the RAM_ADDR and RAM_SIZE as defined by the partition manager.

### DIFF
--- a/include/arch/arm/aarch32/cortex_m/scripts/linker.ld
+++ b/include/arch/arm/aarch32/cortex_m/scripts/linker.ld
@@ -34,6 +34,7 @@
 #endif
 
 #if USE_PARTITION_MANAGER
+
 #include <pm_config.h>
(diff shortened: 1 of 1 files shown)


SebastianBoe:
DNM until code freeze is over

hakonfam:
Let the DNM stand after code freeze is over as well, we might need to re-work this, checking.

hakonfam:
Closing this, a more generic proposal will be used, where RAM and FLASH are treated as "regions", allowing external flash, weird internal flash (UICR, OTP) to be mapped using the same mechanism.

sigvartmh (approved):

hakonfam:
@tejlmand can we merge this?

PR #276, opened 2020-02-20
[nrf noup] linker: generalize linking into alternate slot

hakonfam:
Since any image (not only mcuboot) can be linked into S1, update the linker
script so that its naming is generic.

nrf pr https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1939

### DIFF
--- a/include/arch/arm/aarch32/cortex_m/scripts/linker.ld
+++ b/include/arch/arm/aarch32/cortex_m/scripts/linker.ld
@@ -35,19 +35,26 @@
 
 #if USE_PARTITION_MANAGER
 #include <pm_config.h>
-#ifdef LINK_MCUBOOT_INTO_S1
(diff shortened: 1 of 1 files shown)


include/arch/arm/aarch32/cortex_m/scripts/linker.ld
```diff
@@ -35,14 +35,22 @@
 
 #if USE_PARTITION_MANAGER
 #include <pm_config.h>
-#ifdef LINK_MCUBOOT_INTO_S1
-/* We are linking mcuboot against S1, S0 is Image 1 primary */
+
+#ifdef LINK_INTO_S1
+#define OTHER_SLOT_FLASH_ID PM_S0_ID
```
SebastianBoe: What is this define for?

hakonfam: removed

PR #277, opened 2020-02-20
Erratum 19 - SPU flash regions are 32 kB instead of 16 kB

oyvindronningstad:
This PR adds a kconfig for nRF53 erratum 19 (https://infocenter.nordicsemi.com/topic/errata_nRF5340_EngA/ERR/nRF5340/EngineeringA/latest/anomaly_340_19.html?cp=3_0_1_0_1_15).

Review the solution conceptually, and if you see any problems in putting this upstream as is.

@bjorn-tore-taraldsen: Please review the wording on the config in particular.

### DIFF
--- a/arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
+++ b/arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
@@ -4,18 +4,23 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+/* nRF-specific defines. */
+#ifdef CONFIG_CPU_HAS_NRF_IDAU
+	/* This SOC needs the NSC region to be at the end of an SPU region. */
+	#define NSC_ALIGN \
+		. = ALIGN(CONFIG_NRF_SPU_FLASH_REGION_SIZE) \
+			- (1 << LOG2CEIL(__sg_size))
+	#define NSC_ALIGN_END . = ALIGN(CONFIG_NRF_SPU_FLASH_REGION_SIZE)
+#endif /* CONFIG_CPU_HAS_NRF_IDAU */
+
+
 #if CONFIG_ARM_NSC_REGION_BASE_ADDRESS != 0
 	#define NSC_ALIGN . = ABSOLUTE(CONFIG_ARM_NSC_REGION_BASE_ADDRESS)
-#elif defined(CONFIG_CPU_HAS_NRF_IDAU)
-	/* The nRF9160 needs the NSC region to be at the end of a 32 kB region. */
-	#define NSC_ALIGN . = ALIGN(0x8000) - (1 << LOG2CEIL(__sg_size))
-#else
+#elif !defined(NSC_ALIGN)
 	#define NSC_ALIGN . = ALIGN(4)
-#endif
+#endif /* CONFIG_ARM_NSC_REGION_BASE_ADDRESS */
 
-#ifdef CONFIG_CPU_HAS_NRF_IDAU
-	#define NSC_ALIGN_END . = ALIGN(0x8000)
-#else
+#ifndef NSC_ALIGN_END
 	#define NSC_ALIGN_END . = ALIGN(4)
 #endif
 
@@ -31,11 +36,14 @@ __sg_size = __sg_end - __sg_start;
 NSC_ALIGN_END;
 __nsc_size = . - __sg_start;
 
+/* nRF-specific ASSERT. */
 #ifdef CONFIG_CPU_HAS_NRF_IDAU
-	ASSERT(1 << LOG2CEIL(0x8000 - (__sg_start % 0x8000))
-			 == (0x8000 - (__sg_start % 0x8000))
-		&& (0x8000 - (__sg_start % 0x8000)) >= 32
-		&& (0x8000 - (__sg_start % 0x8000)) <= 4096,
+	#define NRF_SG_START (__sg_start % CONFIG_NRF_SPU_FLASH_REGION_SIZE)
+	#define NRF_SG_SIZE (CONFIG_NRF_SPU_FLASH_REGION_SIZE - NRF_SG_START)
+	ASSERT((__sg_size == 0)
+		|| (((1 << LOG2CEIL(NRF_SG_SIZE)) == NRF_SG_SIZE) /* Pow of 2 */
+			&& (NRF_SG_SIZE >= 32)
+			&& (NRF_SG_SIZE <= 4096)),
 		"The Non-Secure Callable region size must be a power of 2 \
 between 32 and 4096 bytes.")
 #endif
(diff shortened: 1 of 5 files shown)


arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
```diff
@@ -7,14 +7,21 @@
 #if CONFIG_ARM_NSC_REGION_BASE_ADDRESS != 0
 	#define NSC_ALIGN . = ABSOLUTE(CONFIG_ARM_NSC_REGION_BASE_ADDRESS)
 #elif defined(CONFIG_CPU_HAS_NRF_IDAU)
-	/* The nRF9160 needs the NSC region to be at the end of a 32 kB region. */
-	#define NSC_ALIGN . = ALIGN(0x8000) - (1 << LOG2CEIL(__sg_size))
+#ifdef CONFIG_SOC_NRF9160
+#define REGION_SIZE 0x8000
```
hakonfam: What about using (32*1024) instead of 0x8000 so that its easier to see that it is 32kB? Some elsewher, but if hex values are used otherwise in this file I understand if you want to keep it consistent.

arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
```diff
@@ -7,14 +7,21 @@
...
+#elif defined(CONFIG_NRF5340_CPUAPP_ERRATUM19)
```
hakonfam: So the logic is:

if 90
else if 53_erratum
else if 53
endif

Would it make more sense to have
if 91
else if 53
if 53_errata
endif 53/91

Not I got by point through, but the errata should only be checked if its a 53

bjorn-tore-taraldsen: Agree. It make sense to only check 53 errata if it's a 53 device (in case it's mixed up)

soc/arm/nordic_nrf/nrf53/Kconfig.defconfig.nrf5340_CPUAPP_QKAA
```diff
@@ -11,4 +11,12 @@ config SOC
 config NUM_IRQS
 	default 69
 
+config NRF5340_CPUAPP_ERRATUM19
+	bool "Include workarounds for Erratum 19"
+	default y
+	help
+	  This MUST be enabled if there is a chance the code will be run on
```
bjorn-tore-taraldsen: Should we mention that this errata workaround also will work on devices that has fixed this errata, with the cost of a bigger flash size?

soc/arm/nordic_nrf/nrf53/Kconfig.defconfig.nrf5340_CPUAPP_QKAA
```diff
@@ -11,4 +11,12 @@ config SOC
...
+	  nRF5340 Engineering A (Variant AAAB).
```
bjorn-tore-taraldsen: Reuse text from the Errata text: "nRF5340 Engineering A, build codes QKAA-AB0"

oyvindronningstad:
> If you have code doing special handling of this errata when writing to FLASHREGION[n].PERM (n=0..31). Would it be possible to add a check for nRF53 / page size to catch if this is nRF53 - Eng A but page size set to 16kB and not 32kB and give an error?

Not 100% sure what you mean. Do you mean the inline functions that read FICR?

soc/arm/nordic_nrf/nrf53/Kconfig.defconfig.nrf5340_CPUAPP_QKAA
oyvindronningstad: Done

soc/arm/nordic_nrf/nrf53/Kconfig.defconfig.nrf5340_CPUAPP_QKAA
oyvindronningstad: Done

arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
oyvindronningstad: Done

arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
oyvindronningstad: I created a kconfig for this.

oyvindronningstad:
Thanks @barsok. Could you also take a look at the new commit that adds SPU region sizes to Kconfig?

arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
```diff
@@ -4,18 +4,24 @@
...
+#ifdef CONFIG_CPU_HAS_NRF_IDAU
+	/* This SOC needs the NSC region to be at the end of an SPU region. */
+	#define NSC_ALIGN \
+		. = ALIGN(CONFIG_NRF_SPU_FLASH_REGION_SIZE) \
+			- (1 << LOG2CEIL(__sg_size))
+	#define NSC_ALIGN_END . = ALIGN(CONFIG_NRF_SPU_FLASH_REGION_SIZE)
+#endif /* CONFIG_CPU_HAS_NRF_IDAU */
+
+
+/* Define NSC_ALIGN */
```
nordic-krch: what's the point of this comment?

oyvindronningstad: To make it clearer that the next block just defines NSC_ALIGN. If it's not clear, I'll remove it.

oyvindronningstad:
Added assert on bootup.

oyvindronningstad:
> Can you send an upstream PR for these changes?

Sure, will open PR within the next day or two.

arch/arm/core/aarch32/cortex_m/tz/secure_entry_functions.ld
oyvindronningstad: Removed

oyvindronningstad:
> > Can you send an upstream PR for these changes?
> 
> Sure, will open PR within the next day or two.

Opened PR https://github.com/zephyrproject-rtos/zephyr/pull/23058

hakonfam:
Please update commit message to [nrf fromlist] (or what the PR equivalent is) and then point to the PR, this so that the upmerge can easily see if these commits can be dropped :)

oyvindronningstad:
> Please update commit message to [nrf fromlist] (or what the PR equivalent is) and then point to the PR, this so that the upmerge can easily see if these commits can be dropped :)

Thanks, done

oyvindronningstad:
@ioannisg If you can, I'd like your approval on this before we merge.

ioannisg:
> @ioannisg If you can, I'd like your approval on this before we merge.

Yeap, I'll review this today

ioannisg:
I added my reviews on the upstream PR - I wonder if we can conclude on that one first and then bring here the changes.

oyvindronningstad:
Upstream is merge. @ioannisg can we merge this?

ioannisg:
> Upstream is merged. @ioannisg can we merge this?

It is identical, right?

oyvindronningstad:
> > Upstream is merged. @ioannisg can we merge this?
> 
> It is identical, right?

Yes

PR #279, opened 2020-02-25
[nrf noup] cmake: move patch from extensions.cmake to nrf repo

hakonfam:
This is done to reduce the diff from upstream, and to
facilitate downstream patches.


SebastianBoe:
Missing tag on the deps commit

hakonfam:
> Missing tag on the deps commit

that was not supposed to be here, my bad.

samples/hello_world/prj.conf
```diff
@@ -1 +1,2 @@
 # nothing here
+CONFIG_SECURE_BOOT=y
```
mbolivar: I think this was unintentional; right?

hakonfam: Correct, I have to start reviewing my amends a bit better :D

mbolivar:
It looks like the latest version added checks for whether ZEPHYR_NRF_MODULE_DIR is defined. Why is that necessary? It doesn't seem like it should be.

cmake/extensions.cmake
```diff
@@ -1579,16 +1579,6 @@ function(generate_unique_target_name_from_filename filename target_name)
...
-  get_filename_component(pm_path ${config_file} REALPATH)
-  set_property(
-    GLOBAL APPEND PROPERTY
-    PM_SUBSYS
-    ${pm_path}
-    )
-endfunction()
+if (ZEPHYR_NRF_MODULE_DIR)
+  include(${ZEPHYR_NRF_MODULE_DIR}/cmake/extensions.cmake)
+endif()
```
tejlmand: Is it possible to use `fw-nrfconnect-zephyr` standalone (that is outside NCS scope) ?

If not, and if it should not be so, then we could skip the `if(ZEPHYR_NRF_MODULE_DIR)`

subsys/fs/CMakeLists.txt
```diff
@@ -15,7 +15,10 @@ if(CONFIG_FILE_SYSTEM)
   target_link_libraries_ifdef(CONFIG_FAT_FILESYSTEM_ELM   FS INTERFACE ELMFAT)
   target_link_libraries_ifdef(CONFIG_FILE_SYSTEM_LITTLEFS FS INTERFACE LITTLEFS)
 
-  add_partition_manager_config(pm.yml)
+  if (ZEPHYR_NRF_MODULE_DIR)
```
tejlmand: Although we know that `ZEPHYR_NRF_MODULE_DIR` contains this function, then I think it would look nicer to test as:
```suggestion
  if(COMMAND add_partition_manager_config)
```
as that will have no opinion on where this command is defined.

subsys/settings/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ zephyr_include_directories(
   include
   )
 
-if (CONFIG_SETTINGS_FCB OR CONFIG_SETTINGS_NVS)
+if (ZEPHYR_NRF_MODULE_DIR AND (CONFIG_SETTINGS_FCB OR CONFIG_SETTINGS_NVS))
```
tejlmand: Same reason as above:
```suggestion
if((COMMAND add_partition_manager_config) AND (CONFIG_SETTINGS_FCB OR CONFIG_SETTINGS_NVS))
```

hakonfam:
> It looks like the latest version added checks for whether ZEPHYR_NRF_MODULE_DIR is defined. Why is that necessary? It doesn't seem like it should be.

Simply to pass CI I'm afraid. The sanitycheck seems to be running without the modules installed, so the ${ZEPHYR_NRF_MODULE_DIR} is not defined when the tests are running.

I have done the same fix before, and seem to remember it was concluded that this was the best way of fixing the issue. I will check if this is still the case.

cmake/extensions.cmake
hakonfam: See https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/279#issuecomment-591828282

subsys/fs/CMakeLists.txt
hakonfam: If you consider https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/279#issuecomment-591828282 and the conclusion is that we don't need the check anymore, would this comment still be valid?

subsys/settings/CMakeLists.txt
hakonfam: If you consider #279 (comment) and the conclusion is that we don't need the check anymore, would this comment still be valid?

tejlmand:
> > It looks like the latest version added checks for whether ZEPHYR_NRF_MODULE_DIR is defined. Why is that necessary? It doesn't seem like it should be.
> 
> Simply to pass CI I'm afraid. The sanitycheck seems to be running without the modules installed, so the ${ZEPHYR_NRF_MODULE_DIR} is not defined when the tests are running.
> 
> I have done the same fix before, and seem to remember it was concluded that this was the best way of fixing the issue. I will check if this is still the case.

Thanks for the explanation.
I think it's fine to test for `ZEPHYR_NRF_MODULE_DIR` in this case.
https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/279#discussion_r384711094

cmake/extensions.cmake
tejlmand: accepted to keep the test here, for the reason given below.

subsys/settings/CMakeLists.txt
tejlmand: Based on your explanation, I think it is still important to safeguard the call to `add_partition_manager_config`, but a personal preference would be to test that the function exist (as suggested above) instead of testing for nRF module.

In principle code doesn't care if `add_partition_manager_config` is defined in `zephyr/cmake/extensions.cmake`, `nrf/cmake/extensions.cmake`, or a 3rd place.
It only requires this function to be defined.

But will not block PR on this, as both solutions are acceptable.

subsys/fs/CMakeLists.txt
tejlmand: see: https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/279/files/f23c709b956fc782298c494fb77453aa82b5e1f5#r385039160

SebastianBoe:
@mbolivar : Hi, about your feedback

> It looks like the latest version added checks for whether ZEPHYR_NRF_MODULE_DIR is defined. Why is that necessary? It doesn't seem like it should be.

Please see

https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/279#issuecomment-591828282

and re-asses the review

mbolivar:
> > It looks like the latest version added checks for whether ZEPHYR_NRF_MODULE_DIR is defined. Why is that necessary? It doesn't seem like it should be.
> 
> Simply to pass CI I'm afraid. The sanitycheck seems to be running without the modules installed, so the ${ZEPHYR_NRF_MODULE_DIR} is not defined when the tests are running.

How can that be? That would mean sanitycheck isn't using our HAL, which I don't understand. @thst-nordic ?

mbolivar:
> @mbolivar : Hi, about your feedback
> 
> > It looks like the latest version added checks for whether ZEPHYR_NRF_MODULE_DIR is defined. Why is that necessary? It doesn't seem like it should be.
> 
> Please see
> 
> [#279 (comment)](https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/279#issuecomment-591828282)
> 
> and re-asses the review

I'm not blocking this PR, but that seems kind of fishy to me and sounds like it should be addressed in CI.

tejlmand:
> > > It looks like the latest version added checks for whether ZEPHYR_NRF_MODULE_DIR is defined. Why is that necessary? It doesn't seem like it should be.
> > 
> > 
> > Simply to pass CI I'm afraid. The sanitycheck seems to be running without the modules installed, so the ${ZEPHYR_NRF_MODULE_DIR} is not defined when the tests are running.
> 
> How can that be? That would mean sanitycheck isn't using our HAL, which I don't understand. @thst-nordic ?

I'm no expert on this, but could it be that CI is actually using the manifest file from Zephyr and not the nrf manifest, when running on this repo ?
That could explain why the nRF repo would be available.

@thst-nordic Could you provide more info on this ?
and also why this happens: https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/279#issuecomment-591828282

nvlsianpu:

cmake/extensions.cmake
tejlmand: @hakonfam don't know if this PR is still valid, but with the introduction of: https://github.com/nrfconnect/sdk-nrf/commit/c59d4df5c53d14f7bc96a588361ef0f14c989906
then we actually have the possibility of loading the nrf extensions BEFORE Zephyr extensions.cmake.
Guess that would be much nicer :smile:

hakonfam:
No longer needed

PR #281, opened 2020-02-26
[nrf noup] linker: use partition manager to set RAM

hakonfam:
Partition manager now supports partitioning RAM as well as FLASH.

Use the RAM_ADDR and RAM_SIZE as defined by the partition manager.

### DIFF
--- a/include/arch/arm/aarch32/cortex_m/scripts/linker.ld
+++ b/include/arch/arm/aarch32/cortex_m/scripts/linker.ld
@@ -34,6 +34,7 @@
 #endif
 
 #if USE_PARTITION_MANAGER
+
 #include <pm_config.h>
 #ifdef LINK_MCUBOOT_INTO_S1
 /* We are linking mcuboot against S1, S0 is Image 1 primary */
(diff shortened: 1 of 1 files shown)


carlescufi:
@ioannisg and @tejlmand we are setting up a lot of variables from the partition manager now, I just want to make sure that this is not going to be an issue down the line.

mbolivar:
I am a bit concerned about NCS partition manager's increasing scope, since it is a downstream-only feature with tree-wide impact. I don't want to get in the way, but I would hope that eventually, PM's job can be done in an upstream-friendly manner.

hakonfam:
I share your concerns, and we wanted to avoid this, but we are at a point now where this has been deemed necessary.

As you probably understand; the issue we are solving is how RAM is shared between the images and subsystems, much like what is currently done for flash. This way, subsystems and images can reserve blocks of ram, using information from Kconfig to dynamically deduce the size of these blocks. As with flash, the configured information is the size reserved, and not the address.

Achieving the same with devicetree would be possible, but not user friendly IMO. This because image/subsystem X has to take into account what other images/subsystems has already reserved in order to know what addresses to use. Hence one would need a large set of configurations to match all the possible permutations.

Please let me know if there is a way to get the same functionality from devicetree, as I'm no expert in that area.

PR #284, opened 2020-03-02
[nrf fromlist] drivers: flash: Fix nrf QSPI NOR driver to behave like flash

sigvartmh:
Add the possibility to use heap to emulate internal flash driver more
closely. 

@kamlaz(Couldn't find you on the reviewer list) 

This change needs a bit more explanation here goes:

While trying to implement the use of external flash into MCUBoot the biggest issue was that the **qspi_nor_flash driver** does not behave the same as **NVMC flash driver**(internal flash) with only subtle differences such as reading/writing needs to be in **RAM** and they need to be 4 bytes aligned(this PR does not currently try to address this issue).

My suggestion is to create a compatibility layer with using allocated memory slabs to either put the data in RAM or return `-ENOMEM` if there are no available memory slabs.

This change should maybe be hidden behind an additional Kconfig as it will enable the use of Zephyr's heap mechanism to allocate the memory in memory slabs for the sizes under 4 or data residing in flash.

### DIFF
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -476,17 +476,42 @@ static inline int qspi_nor_read_id(struct device *dev,
 	return 0;
 }
 
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+u8_t *rx_ram_buf;
+u8_t *tx_ram_buf;
+#endif
+
 static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 			 size_t size)
 {
+	int ret;
+
 	if (!dest) {
 		return -EINVAL;
 	}
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	rx_ram_buf = NULL;
+	void *tmp_dest = dest;
+	size_t var_size = size;
+
+	if (((uint32_t)dest < CONFIG_SRAM_BASE_ADDRESS)
+	   || (size < sizeof(uint32_t))) {
+		if (size < sizeof(uint32_t)) {
+			size = 4;
+		}
+		rx_ram_buf = k_malloc(size);
+		if (rx_ram_buf == NULL) {
+			return -ENOMEM;
+		}
+		dest = (void *)rx_ram_buf;
+	}
+#endif
 
 	/* read size must be multiple of 4 bytes */
 	if (size % sizeof(uint32_t) ||
 	    !(size > 0)) {
-		return -EINVAL;
+		ret = -EINVAL;
+		goto out;
 	}
 
 	const struct qspi_nor_config *params = dev->config->config_info;
@@ -499,35 +524,70 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 		LOG_ERR("read error: address or size "
 			"exceeds expected values."
 			"Addr: 0x%lx size %zu", (long)addr, size);
-		return -EINVAL;
+		ret = -EINVAL;
+		goto out;
 	}
 
 	qspi_lock(dev);
 
-	int ret = nrfx_qspi_read(dest, size, addr);
+	ret = nrfx_qspi_read(dest, size, addr);
 
 	qspi_wait_for_completion(dev);
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	if (rx_ram_buf != NULL) {
+		memcpy(tmp_dest, dest, var_size);
+		k_free(rx_ram_buf);
+	}
+#endif
 	return qspi_get_zephyr_ret_code(ret);
+out:
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	k_free(rx_ram_buf);
+#endif
+	return ret;
 }
 
 static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 			  size_t size)
 {
+	int ret;
+
 	if (!src) {
 		return -EINVAL;
 	}
 
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	tx_ram_buf = NULL;
+	if (((uint32_t)src < CONFIG_SRAM_BASE_ADDRESS)
+	   || (size < sizeof(uint32_t))) {
+		size_t var_size = size;
+
+		if (size < sizeof(uint32_t)) {
+			size = 4;
+		}
+		tx_ram_buf = k_malloc(size);
+		if (tx_ram_buf == NULL) {
+			ret = -ENOMEM;
+		}
+		qspi_nor_read(dev, addr, tx_ram_buf, size);
+		memcpy(tx_ram_buf, src, var_size);
+		src = tx_ram_buf;
+	}
+#endif
+
 	/* write size must be multiple of 4 bytes */
 	if (size % sizeof(uint32_t) ||
 	    !(size > 0)) {
-		return -EINVAL;
+		ret = -EINVAL;
+		goto out;
 	}
 
 	struct qspi_nor_data *const driver_data = dev->driver_data;
 	const struct qspi_nor_config *params = dev->config->config_info;
 
 	if (driver_data->write_protection) {
-		return -EACCES;
+		ret = -EACCES;
+		goto out;
 	}
 
 	/* should be between 0 and flash size */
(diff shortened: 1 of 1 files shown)


sigvartmh:
This PR is aim is to be up streamed.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -476,12 +476,34 @@ static inline int qspi_nor_read_id(struct device *dev,
 	return 0;
 }
 
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+u8_t *rx_ram_buff;
```
hakonfam:
```suggestion
u8_t *rx_ram_buf;
```

hakonfam: AFAIK this is a more used abbreviation.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -507,6 +529,12 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 	int ret = nrfx_qspi_read(dest, size, addr);
 
 	qspi_wait_for_completion(dev);
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
```
hakonfam: We are missing calls to `k_free` on some returns. Please consider using the `goto out` pattern.

sigvartmh: Added

drivers/flash/nrf_qspi_nor.c
sigvartmh: Changed

sigvartmh:
ping @nvlsianpu since you are the code owner of this subsystem maybe you could have a look too and see if this change makes sense to do.

nvlsianpu:
> This PR is aim is to be up streamed.
good!

kamlaz:
Thank you for adding me to the review. Sorry for my late response, but I was ill all previous week.
First of all I am not sure if I got correctly what you are going to do. Your aim was to read/write data that are not 4-byte aligned, am I right? If I was right, please find some comments below. If not, ignore them and correct me.
I have looked into your proposition and I think that it will be better idea to add one layer **above** low layer `qspi_nor_flash driver` and this layer should handle read/write of the files that are no 4 byte aligned. In that case you can allocate memory in upper layer and pass the pointer to allocated memory to i.e `qspi_nor_write`.  This will save from using `memcpy`.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -476,17 +476,42 @@ static inline int qspi_nor_read_id(struct device *dev,
...
 	if (!dest) {
 		return -EINVAL;
 	}
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	rx_ram_buf = NULL;
+	void *tmp_dest = dest;
+	size_t var_size = size;
+
+	if (((uint32_t)dest < CONFIG_SRAM_BASE_ADDRESS)
+	   || (size < sizeof(uint32_t))) {
```
kamlaz: if (((uint32_t)dest < CONFIG_SRAM_BASE_ADDRESS) ||
	     (size < sizeof(uint32_t))) {
Maybe move logic operator? I find it more readable.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -476,17 +476,42 @@ static inline int qspi_nor_read_id(struct device *dev,
...
+		if (size < sizeof(uint32_t)) {
+			size = 4;
+		}
+		rx_ram_buf = k_malloc(size);
```
kamlaz: I think that memory allocation should be done after `qspi_lock(dev)'. In case of few calls to `qspi`, only one call can be handled at a time. The rest are blocked at the semaphore - so there is no need to allocate memory for them at that moment.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -476,17 +476,42 @@ static inline int qspi_nor_read_id(struct device *dev,
...
 	}
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	rx_ram_buf = NULL;
+	void *tmp_dest = dest;
+	size_t var_size = size;
+
+	if (((uint32_t)dest < CONFIG_SRAM_BASE_ADDRESS)
+	   || (size < sizeof(uint32_t))) {
+		if (size < sizeof(uint32_t)) {
+			size = 4;
```
kamlaz: Maybe `size = sizeof(uint32_t);` to avoid magic number?

sigvartmh:
> First of all I am not sure if I got correctly what you are going to do. Your aim was to read/write data that are not 4-byte aligned, am I right? If I was right, please find some comments below. If not, ignore them and correct me.

You are partially correct so the idea here is to have a compatibility mode that ensures that the `qspi_no flash_driver` behaves the same way as the NVMC `flash_driver`. In such a way that any code which uses the `flash_write/read` calls will not have to do any special handling of the data just because the current device changed from NVMC to external flash mapped through QSPI.

kamlaz:
Ok, now I understand. Thank you for the explanation. 
I have compared NVMC `flash_driver` with QSPI `flash_driver`. Their functionality is >almost< equal with exception of `read` function. `Write` and `Erase` behave similarly.
My proposition is to not introduce any changes to `QSPI` driver, but to create one layer above `flash_driver`, let's name it `qspi_nvmc`. If you want to read a file of size `size` that is not word alligned, just : 
1. Call `qspi_nvmc_read(dev, buf, size);`. `qspi_nvmc` layer will calculate how much memory to allocate to fit `qspi` alignment requirements (i.e if you requested to read 3 bytes, it will allocate 4). `buf` is the pointer to memory region. At the input it will be NULL, but function will change its value to point to allocated memory region.
2. Next it will call `flash_read` and store data in allocated region. 
3. At the end function will return pointer to the memory region with saved data.

As a result you obtain memory region in heap filled with data, but its size could be larger than you requested up to three bytes (they are just wasted memory space). I am aware that this solution is not memory-efficient, but is faster since there are no `memcpy` calls. 

Maybe this solution satisfies your needs?

sigvartmh:
I'll see if I can fit in such an abstraction layer. I doubt we can get away with not `memcpy` data if the data is in flash but at least I guess it would rid 1 memcpy for alignment.

sigvartmh:
What I need is for `flash_write` and `flash_read` to behave the same way as NVMC. So the idea behind this patch is to have a `COMPATABILITY_MODE` in the QSPI driver.

drivers/flash/nrf_qspi_nor.c
sigvartmh: That's a good idea thanks :)

drivers/flash/nrf_qspi_nor.c
```diff
@@ -499,35 +524,70 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
 	qspi_lock(dev);
 
-	int ret = nrfx_qspi_read(dest, size, addr);
+	ret = nrfx_qspi_read(dest, size, addr);
 
 	qspi_wait_for_completion(dev);
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	if (rx_ram_buf != NULL) {
+		memcpy(tmp_dest, dest, var_size);
+		k_free(rx_ram_buf);
```
oyvindronningstad:
```suggestion
		k_free(rx_ram_buf);
		rx_ram_buf = NULL;
```

drivers/flash/nrf_qspi_nor.c
```diff
@@ -499,35 +524,70 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
+	}
+#endif
 	return qspi_get_zephyr_ret_code(ret);
+out:
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+	k_free(rx_ram_buf);
```
oyvindronningstad:
```suggestion
	k_free(rx_ram_buf);
	rx_ram_buf = NULL;
```

oyvindronningstad: Should this also check for `rx_ram_buf != NULL`?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -476,17 +476,42 @@ static inline int qspi_nor_read_id(struct device *dev,
 	return 0;
 }
 
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
+u8_t *rx_ram_buf;
+u8_t *tx_ram_buf;
```
oyvindronningstad:
```suggestion
static u8_t *rx_ram_buf = NULL;
static u8_t *tx_ram_buf = NULL;
```

drivers/flash/nrf_qspi_nor.c
oyvindronningstad: BTW could you avoid duplicating this `k_free` snippet?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -499,35 +524,70 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
 static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 			  size_t size)
 {
+	int ret;
+
 	if (!src) {
 		return -EINVAL;
 	}
 
+#if (CONFIG_HEAP_MEM_POOL_SIZE > 0)
```
oyvindronningstad: This snippet could also be deferred until right before `src` is used.

drivers/flash/nrf_qspi_nor.c
oyvindronningstad: AFACT these vars are only used in one function each, why not declare them inside of those functions?

nvlsianpu:

sigvartmh:
Closing this one have a new implementation.

PR #286, opened 2020-03-20
Cherrypick upstream USB fixes

bjda:
Gets rid of some USB issues for connectivity_bridge application

### DIFF
--- a/drivers/usb/device/usb_dc_nrfx.c
+++ b/drivers/usb/device/usb_dc_nrfx.c
@@ -1526,6 +1526,7 @@ int usb_dc_ep_clear_stall(const u8_t ep)
 		return -EINVAL;
 	}
 
+	nrfx_usbd_ep_dtoggle_clear(ep_addr_to_nrfx(ep));
 	nrfx_usbd_ep_stall_clear(ep_addr_to_nrfx(ep));
 	LOG_DBG("Unstall on EP 0x%02x", ep);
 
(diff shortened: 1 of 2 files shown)


emob-nordic:
Those commits later become a part of [#23307](https://github.com/zephyrproject-rtos/zephyr/pull/23307). Don't You want all those fixes?

bjda:
Since I'm not sure if there could be dependencies or interactions between the other commits in that PR and any other changes to Zephyr since the last upmerge, I limited it to those that are known to fix some issues with connectivity_bridge. @nordic-auko , feel free to chime in on whether the whole PR should be included.

nordic-auko:
> 
> 
> Since I'm not sure if there could be dependencies or interactions between the other commits in that PR and any other changes to Zephyr since the last upmerge, I limited it to those that are known to fix some issues with connectivity_bridge. @nordic-auko , feel free to chime in on whether the whole PR should be included.

These two commits fixed the issues I saw with USB MSC (USB device not responding properly to unsupported SCSI commands), so they should be included here. The three other commits I believe are part of an ongoing revamp of the USB MSC implementation. 

I tested both the set of 2 commits and the set of all 5 commits separately and have not seen any issues. I don't believe there are any inter dependencies, but I can't say if the set of 3 commits fixes separate problems not seen by me

bjda:
Rebased on ncs master

PR #291, opened 2020-04-06
Ensure buffer passed to nrfx_qspi_read is word aligned

ball-hayden:
`nrfx_qspi_read` requires that the buffer passed is word aligned.

The Zephyr flash API doesn't require this. As such, it may be that the passed buffer is not word aligned (as is the case when combined with the FS subsystem).

Submitting this here for feedback before submitting to upstream Zephyr.

### DIFF
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -502,11 +502,16 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 		return -EINVAL;
 	}
 
+	u8_t buffer[size] __attribute__ ((aligned (32)));
+
 	qspi_lock(dev);
 
-	int ret = nrfx_qspi_read(dest, size, addr);
+	int ret = nrfx_qspi_read(buffer, size, addr);
 
 	qspi_wait_for_completion(dev);
+
+	memcpy(dest, buffer, size);
+
 	return qspi_get_zephyr_ret_code(ret);
 }
 


ball-hayden:
This is somewhat difficult to reproduce - I've tried to do so using `samples/subsys/fs/littlefs`, but I haven't managed to persuade the compiler to put the read buffer in the "right" place such that it isn't word aligned in that simple example.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -502,11 +502,16 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 		return -EINVAL;
 	}
 
+	u8_t buffer[size] __attribute__ ((aligned (32)));
```
pabigot, line 505: This isn't going to work, because it could overrun the stack.  It's also very inefficient.

ball-hayden: Thank you @pabigot.
I'm very new to low level programming - are you able to explain why this could cause a stack overrun?

pabigot: You're dynamically allocating a buffer of `size` bytes on the stack to hold the aligned copy.  Since `size` is an input to the function there's no bound on how large it is: the caller has guaranteed to provide that much data, but the function can't affort to make a copy of it.

ball-hayden:
Thank you for your feedback @pabigot - I appreciate your time.

I have filed an upstream issue at https://github.com/zephyrproject-rtos/zephyr/issues/24122

PR #292, opened 2020-04-13
[nrf noup] cmake: change how path to subsys pm.yml is set

hakonfam:
The relative path from the build dir to the binary dir of a subsys
folder is not static. In a zephyr subsys it is

build/zephyr/subsys/path

whereas in a nrf subsys it is

build/modules/nrf/subsys/path

The previous way of setting the path to preprocessed pm.yml files
for subsystems only worked with the first of these.

This commit introduces the use of a cmake variable for the binary dir,
and a mechanism for setting the build dir path for each pm.yml
explicitly.

See https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/2122

### DIFF
--- a/cmake/extensions.cmake
+++ b/cmake/extensions.cmake
@@ -1586,9 +1586,14 @@ endfunction()
 # manager configurations.
 function(add_partition_manager_config config_file)
   get_filename_component(pm_path ${config_file} REALPATH)
+
   set_property(
     GLOBAL APPEND PROPERTY
-    PM_SUBSYS
-    ${pm_path}
+    PM_SUBSYS_PATH ${pm_path}
+    )
+
+  set_property(
+    GLOBAL APPEND PROPERTY
+    PM_SUBSYS_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}
     )
 endfunction()


cmake/extensions.cmake
```diff
@@ -1586,9 +1586,12 @@ endfunction()
...
 function(add_partition_manager_config config_file)
   get_filename_component(pm_path ${config_file} REALPATH)
+
+  # Add entry to PM_SUBSYS on the format
+  # "<abs path to pm.yml> <build dir outut folder>"
   set_property(
     GLOBAL APPEND PROPERTY
     PM_SUBSYS
-    ${pm_path}
+    "${pm_path} ${CMAKE_CURRENT_BINARY_DIR}"
```
tejlmand, line 1595: I noticed, this will be split using a regex later.
I haven't been through all parts of existing code, but is the values on property PM_SUBSYS alsways double like here ?
Cause in that case, we could considered to do double lists.

hakonfam: We can do a double list (do you mean a map/dict? Because we need to look up on a per path basis). But the problem is that the path includes symbols not suited for a cmake variable.

tejlmand: I simply meant two lists, similar to:
```
set_property(
    GLOBAL APPEND PROPERTY
    PM_SUBSYS_PATH ${pm_path}
)
set_property(
    GLOBAL APPEND PROPERTY
    PM_SUBSYS_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}
)
```
see following comment here: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/2122#discussion_r409436940

hakonfam: Got it! Will update

hakonfam:
I don't have merge rights here, so feel free to merge, and I will update the corresponding nrf SHA asap.

tejlmand:
@carlescufi seems i'm still lacking merge rights in this repo.

PR #293, opened 2020-04-15
[nrf noup] subsys/fs: move FS partition to the end back

nvlsianpu:
In case of the DFU capability enabled:
Up to NCS v1.1.0 any storage partition lands at the end of the flash memory.
Since v1.2.0 any storage partition lands right after
the executable application.
So it is impossible to build proper update application out-of-the box.

This patch moves storage partitions at the flash end back.
User needs only to correct the partition size (via Kconfig) to
the same as in previous release for ensure proper upgrade build.

### DIFF
--- a/subsys/fs/pm.yml
+++ b/subsys/fs/pm.yml
@@ -2,6 +2,6 @@
 
 #ifdef CONFIG_FILE_SYSTEM_LITTLEFS
 littlefs_storage:
-  placement: {after: [mcuboot_storage, app]}
+  placement: {before: [end]}
   size: CONFIG_PM_PARTITION_SIZE_LITTLEFS
 #endif
--- a/subsys/settings/pm.yml
+++ b/subsys/settings/pm.yml
@@ -1,5 +1,5 @@
 #include <autoconf.h>
 
 settings_storage:
-  placement: {after: [mcuboot_storage, app]}
+  placement: {before: [end]}
   size: CONFIG_PM_PARTITION_SIZE_SETTINGS_STORAGE


nvlsianpu:
btw, `mcuboot_storage` pm partition doesn't exist anymore

mbolivar-nordic:
How can we avoid similar problems affecting users during the 1.2 -> 1.3 transition?

hakonfam:
This stuff should have been picked up by our DFU tests, not sure why they were not. 

@mbolivar I will make sure that we add backwards compatible tests to ensure that this does not happen again. As for products in field, they will use static configuration, so there is no risk of bricking a configuration so that you are not able to produce valid updates once a device is out.

subsys/fs/pm.yml
```diff
@@ -2,6 +2,6 @@
 
 #ifdef CONFIG_FILE_SYSTEM_LITTLEFS
 littlefs_storage:
-  placement: {after: [mcuboot_storage, app]}
+  placement: {after: [mcuboot_scratch, mcuboot_secondary, app]}
```
hakonfam, line 5: Have you tried {before: end} ? It could give the same result, and might look prettier.

nvlsianpu: Indeed `{before: [end]}` does the same

nvlsianpu:
ping @mbolivar @sigvartmh 
cc @carlescufi

nvlsianpu:

nvlsianpu:
moved to https://github.com/nrfconnect/sdk-nrf/pull/2382

PR #297, opened 2020-05-06
[nrf fromtree] storage: add stream flash library

hakonfam:
SHA a9676831cb

This library supports stream writes to flash with
optinal progressive erase.

This module is a direct copy of the functionality found in
subsys/dfu/img_util/flash_img.c


hakonfam:
included in upmerge

PR #299, opened 2020-05-13
[nrf fromlist] cmake: find python3 on windows when multiarch are inst…

tejlmand:
cmake: find python3 on windows when multiarch are installed

Fixes: #24692

This commit fixes an issue where FindPython3 could wrongly select the
Python version not in environment path.

If user installed both 32 and 64 bit versions of the same Python, for
example 3.7.x, the current search would only specify that 3.7 is needed.

In some cases, FindPython3 could select the 32 bit version, even if the
64 bit version is the one on %PATH%.

This is fixed, by setting Python3_ROOT_DIR to point to the tested Python
in %PATH%.

---- 
See upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/25001

----
Manifest update: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/2240

### DIFF
--- a/cmake/python.cmake
+++ b/cmake/python.cmake
@@ -24,10 +24,18 @@ foreach(PYTHON_PREFER ${PYTHON_PREFER} "python" "python3")
                        OUTPUT_STRIP_TRAILING_WHITESPACE)
 
      if(version VERSION_LESS PYTHON_MINIMUM_REQUIRED)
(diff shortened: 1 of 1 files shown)


carlescufi:
You can close this now that the upstream PR is merged. @joerchan will begin an upmerge later today.

PR #303, opened 2020-05-16
[nrf noup] cmake: move pm patches from zephyr to nrf

hakonfam:
This commit should be squashed with the other partition manager
commits so that only the linker script patch remains of
partition manager patches.

Reduce size of downstream patches by moving as much as possible
of the partition manager (pm) functionality to the nrf repo.


hakonfam:
Parent PR https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/2274

PR #305, opened 2020-05-18
drivers: flash: Add one byte read write emulation to qspi nor nrf

sigvartmh:
Adds the functionality of doing one byte read and write to QSPI nor
driver. This is achived through the use of stack in the function calls.

It also adds the functionality of using stack if the data passed to the
driver are stored in flash and not in RAM. So that the DMA will still
work if you refrence data in flash by copying the data onto the stack.

### DIFF
--- a/drivers/flash/Kconfig.nordic_qspi_nor
+++ b/drivers/flash/Kconfig.nordic_qspi_nor
@@ -29,4 +29,16 @@ config NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE
 	  that API.  By default the page size corresponds to the block
 	  size (65536).  Other option include the sector size (4096).
 
+config NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH
+	bool "Use stack for writing data stored in flash"
+	help
+	  When this option is enabled writing data from FLASH is possible.
+	  However the driver will use stack to copy the data from flash onto
+	  the stack then start the DMA transfer. If you get an -EINVALID error
+	  this could be due to the data being in flash. Either copy the data
+	  locally or enable this option. This uses a 4 byte buffer to copy the
+	  data which has a performance penalty for large chunks of data. If the
+	  data is less than a word the function will already copy the data into
+	  ram.
+
 endif # NORDIC_QSPI_NOR
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -491,16 +491,22 @@ static inline int qspi_nor_read_id(struct device *dev,
 static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 			 size_t size)
 {
-	if (!dest) {
+	void *dptr = dest;
+	size_t dlen = size;
+
+	u8_t __aligned(4) buf[4];
+
+	if (!dest || (size == 0)) {
 		return -EINVAL;
 	}
 
-	/* read size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
-	/* address must be 4-byte aligned */
-	if ((addr % 4U) != 0) {
+	/* Since the QSPI driver requires data to be at least 4 bytes we need
+	 * to use a 4 byte buffer for reads smaller than 4 bytes.
+	 */
+	if (size < 4U) {
+		dest = buf;
+		size = sizeof(buf);
+	} else if ((size % 4U) != 0) {
 		return -EINVAL;
 	}
 
@@ -523,7 +529,13 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 
 	qspi_wait_for_completion(dev, res);
 
-	return qspi_get_zephyr_ret_code(res);
+	int rc = qspi_get_zephyr_ret_code(res);
+
+	if ((rc == 0) && (dest != dptr)) {
+		memcpy(dptr, dest, dlen);
+	}
+
+	return rc;
 }
 
 static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -574,6 +574,7 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 		return -EINVAL;
 	}
 
+
 	struct qspi_nor_data *const driver_data = dev->driver_data;
 	const struct qspi_nor_config *params = dev->config_info;
 
@@ -592,11 +593,33 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 		return -EINVAL;
 	}
 
-	qspi_lock(dev);
 
-	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
+	nrfx_err_t res;
 
-	qspi_wait_for_completion(dev, res);
+	/* if the data smaller than a whole word the function already copies
+	 * the data in ram
+	 */
+	if (res != NRFX_SUCCESS) {
+		/* pre-read failed */
+	} else if (IS_ENABLED(CONFIG_NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH) &&
+	    ((u32_t)sptr < CONFIG_SRAM_BASE_ADDRESS) && !(size < 4U)) {
+		for (size_t bytes_written = 0; bytes_written < dlen;) {
+			memcpy(buf, ((u8_t *)sptr + bytes_written),
+			       sizeof(buf));
+			qspi_lock(dev);
+			res = nrfx_qspi_write(buf, sizeof(buf),
+					      addr + bytes_written);
+			qspi_wait_for_completion(dev, res);
+			if (res != NRFX_SUCCESS) {
+				return qspi_get_zephyr_ret_code(res);
+			}
+			bytes_written += sizeof(buf);
+		}
+	} else {
+		qspi_lock(dev);
+		res = nrfx_qspi_write(src, size, addr);
+		qspi_wait_for_completion(dev, res);
+	}
 
 	return qspi_get_zephyr_ret_code(res);
 }
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -541,16 +541,36 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 			  size_t size)
 {
-	if (!src) {
+	u8_t __aligned(4) buf[4];
+
+	const void *sptr = src;
+	size_t dlen = size;
+
+	if (!src || (size == 0)) {
 		return -EINVAL;
 	}
 
-	/* write size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
 	/* address must be 4-byte aligned */
 	if ((addr % 4U) != 0) {
+		return -EINVAL;
+	}
+
+	/* Since the QSPI driver requires data to be at least 4 bytes we need
+	 * to use a 4 byte buffer for writes smaller than 4 bytes.
+	 */
+	if (size < 4U) {
+		src = buf;
+		size = sizeof(buf);
+		/* read out the whole word so that unchanged data can be
+		 * written back
+		 */
+		nrfx_err_t res = nrfx_qspi_read(buf, size, addr);
+
+		if (res != 0) {
+			return qspi_get_zephyr_ret_code(res);
+		}
+		memcpy(buf, sptr, dlen);
+	} else if ((size % 4U) != 0) {
 		return -EINVAL;
 	}
 
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -562,15 +562,6 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 	if (size < 4U) {
 		src = buf;
 		size = sizeof(buf);
-		/* read out the whole word so that unchanged data can be
-		 * written back
-		 */
-		nrfx_err_t res = nrfx_qspi_read(buf, size, addr);
-
-		if (res != 0) {
-			return qspi_get_zephyr_ret_code(res);
-		}
-		memcpy(buf, sptr, dlen);
 	} else if ((size % 4U) != 0) {
 		return -EINVAL;
 	}
@@ -595,7 +586,20 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 	}
 
 
-	nrfx_err_t res;
+	nrfx_err_t res = NRFX_SUCCESS;
+
+	qspi_lock(dev);
+
+	if (sptr != src) {
+		/* read out the whole word so that unchanged data can be
+		 * written back
+		 */
+		res = nrfx_qspi_read(buf, size, addr);
+		qspi_wait_for_completion(dev, res);
+		if (res == NRFX_SUCCESS) {
+			memcpy(buf, sptr, dlen);
+		}
+	}
 
 	/* if the data smaller than a whole word the function already copies
 	 * the data in ram
@@ -604,20 +608,20 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 		/* pre-read failed */
 	} else if (IS_ENABLED(CONFIG_NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH) &&
 	    ((u32_t)sptr < CONFIG_SRAM_BASE_ADDRESS) && !(size < 4U)) {
-		for (size_t bytes_written = 0; bytes_written < dlen;) {
+		size_t bytes_written = 0;
+
+		while ((res == NRFX_SUCCESS)
+		       && (bytes_written < dlen)) {
 			memcpy(buf, ((u8_t *)sptr + bytes_written),
 			       sizeof(buf));
-			qspi_lock(dev);
 			res = nrfx_qspi_write(buf, sizeof(buf),
 					      addr + bytes_written);
 			qspi_wait_for_completion(dev, res);
-			if (res != NRFX_SUCCESS) {
-				return qspi_get_zephyr_ret_code(res);
+			if (res == NRFX_SUCCESS) {
+				bytes_written += sizeof(buf);
 			}
-			bytes_written += sizeof(buf);
 		}
 	} else {
-		qspi_lock(dev);
 		res = nrfx_qspi_write(src, size, addr);
 		qspi_wait_for_completion(dev, res);
 	}
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -503,7 +503,7 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 	/* Since the QSPI driver requires data to be at least 4 bytes we need
 	 * to use a 4 byte buffer for reads smaller than 4 bytes.
 	 */
-	if (size < 4U) {
+	if ((size > 0) && (size < 4U)) {
 		dest = buf;
 		size = sizeof(buf);
 	} else if ((size % 4U) != 0) {
--- a/drivers/flash/nrf_qspi_nor.c
+++ b/drivers/flash/nrf_qspi_nor.c
@@ -248,8 +248,6 @@ static inline void qspi_wait_for_completion(struct device *dev,
 
 	if (res == NRFX_SUCCESS) {
 		k_sem_take(&dev_data->sync, K_FOREVER);
-	} else {
-		qspi_unlock(dev);
 	}
 }
 
@@ -273,7 +271,6 @@ static void qspi_handler(nrfx_qspi_evt_t event, void *p_context)
 
 	if (event == NRFX_QSPI_EVENT_DONE) {
 		qspi_complete(dev);
-		qspi_unlock(dev);
 	}
 }
 
@@ -319,14 +316,14 @@ static int qspi_erase(struct device *dev, u32_t addr, u32_t size)
 		return -EINVAL;
 	}
 
-	int rv = -EIO;
+	int rv = 0;
 	const struct qspi_nor_config *params = dev->config_info;
 
-	while (size) {
+	qspi_lock(dev);
+	while ((rv == 0) && (size > 0)) {
 		nrfx_err_t res = !NRFX_SUCCESS;
 		u32_t adj = 0;
 
-		qspi_lock(dev);
 		if (size == params->size) {
 			/* chip erase */
 			res = nrfx_qspi_chip_erase();
@@ -353,11 +350,13 @@ static int qspi_erase(struct device *dev, u32_t addr, u32_t size)
 			size -= adj;
 		} else {
 			LOG_ERR("erase error at 0x%lx size %zu", (long)addr, size);
-			return rv;
+			rv = qspi_get_zephyr_ret_code(res);
 		}
 	}
 
-	return 0;
+	qspi_unlock(dev);
+
+	return rv;
 }
 
 /**
@@ -529,6 +528,8 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 
 	qspi_wait_for_completion(dev, res);
 
+	qspi_unlock(dev);
+
 	int rc = qspi_get_zephyr_ret_code(res);
 
 	if ((rc == 0) && (dest != dptr)) {
@@ -620,6 +621,8 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 		res = nrfx_qspi_write(src, size, addr);
 		qspi_wait_for_completion(dev, res);
 	}
+
+	qspi_unlock(dev);
 
 	return qspi_get_zephyr_ret_code(res);
 }
--- a/cmake/app/boilerplate.cmake
+++ b/cmake/app/boilerplate.cmake
@@ -89,6 +89,8 @@ if((NOT DEFINED ZEPHYR_BASE) AND (DEFINED ENV_ZEPHYR_BASE))
   set(ZEPHYR_BASE ${ENV_ZEPHYR_BASE} CACHE PATH "Zephyr base")
 endif()
 
+find_package(ZephyrBuildConfiguration NAMES ZephyrBuild PATHS ${ZEPHYR_BASE}/../* QUIET NO_DEFAULT_PATH NO_POLICY_SCOPE)
+
 # Note any later project() resets PROJECT_SOURCE_DIR
 file(TO_CMAKE_PATH "${ZEPHYR_BASE}" PROJECT_SOURCE_DIR)
 
--- a/doc/guides/env_vars.rst
+++ b/doc/guides/env_vars.rst
@@ -106,6 +106,18 @@ To get this value back into your current terminal environment, **you must run**
 The value will be lost if you close the window, etc.; run ``zephyr-env.cmd``
 again to get it back.
 
+Option 4: Using Zephyr Build Configuration CMake package
+--------------------------------------------------------
+
+Choose this option if you want to make those variable settings shared among all
+users of your project.
+
+Using a :ref:`cmake_build_config_package` allows you to commit the shared
+settings into the repository, so that all users can share them.
+
+It also removes the need for running ``source zephyr-env.sh`` or
+``zephyr-env.cmd`` when opening a new terminal.
+
 .. _zephyr-env:
 
 Zephyr Environment Scripts
--- a/doc/guides/zephyr_cmake_package.rst
+++ b/doc/guides/zephyr_cmake_package.rst
@@ -382,6 +382,72 @@ Such a CMakeLists.txt could look as:
 
    project(my_first_app)
 
+.. _cmake_build_config_package:
+
+Zephyr Build Configuration CMake package
+****************************************
+
+The Zephyr Build Configuration CMake package provides a possibility for a Zephyr based project to
+control Zephyr build settings in a generic way.
+
+It is similar to the use of ``.zephyrrc`` but with the possibility to automatically allow all users
+to share the build configuration through the project repository.
+But it also allows more advanced use cases than a ``.zephyrrc``-file, such as loading of additional
+CMake boilerplate code.
+
+The Zephyr Build Configuration CMake package will be loaded in the Zephyr boilerplate code after
+initial properties and ``ZEPHYR_BASE`` has been defined, but before CMake code execution.
+This allows the Zephyr Build Configuration CMake package to setup or extend properties such as:
+``DTS_ROOT``, ``BOARD_ROOT``, ``TOOLCHAIN_ROOT`` / other toolchain setup, fixed overlays, and any
+other property that can be controlled. It also allows inclusion of additional boilerplate code.
+
+To provide a Zephyr Build Configuration CMake package, create ``ZephyrBuildConfig.cmake`` and place
+it in a Zephyr workspace top-level folder as:
+
+.. code-block:: none
+
+   <projects>/zephyr-workspace
+   ├── zephyr
+   ├── ...
+   └── zephyr application (can be named anything)
+        └── share/zephyrbuild-package/cmake/ZephyrBuildConfig.cmake
+
+The Zephyr Build Configuration CMake package will not search in any CMake default search paths, and
+thus cannot be installed in the CMake package registry. There will be no version checking on the
+Zephyr Build Configuration package.
+
+.. note:: ``share/zephyrbuild-package/cmake/ZephyrBuildConfig.cmake`` follows the same folder
+          structure as the Zephyr CMake package.
+
+          It is possible to place ``ZephyrBuildConfig.cmake`` directly in a
+	  ``<zephyr application>/cmake`` folder or another folder, as long as that folder is
+	  honoring the `CMake package search`_ algorithm.
+
+A sample ``ZephyrBuildConfig.cmake`` can be seen below.
+
+.. code-block:: cmake
+
+   # ZephyrBuildConfig.cmake sample code
+
+   # To ensure final path is absolute and does not contain ../.. in variable.
+   get_filename_component(APPLICATION_PROJECT_DIR
+                          ${CMAKE_CURRENT_LIST_DIR}/../../..
+                          ABSOLUTE
+   )
+
+   # Add this project to list of board roots
+   list(APPEND BOARD_ROOT ${APPLICATION_PROJECT_DIR})
+
+   # Default to GNU Arm Embedded toolchain if no toolchain is set
+   if(NOT ENV{ZEPHYR_TOOLCHAIN_VARIANT})
+       set(ZEPHYR_TOOLCHAIN_VARIANT gnuarmemb)
+       find_program(GNU_ARM_GCC arm-none-eabi-gcc)
+       if(NOT ${GNU_ARM_GCC} STREQUAL GNU_ARM_GCC-NOTFOUND)
+           # The toolchain root is located above the path to the compiler.
+           get_filename_component(GNUARMEMB_TOOLCHAIN_PATH ${GNU_ARM_GCC}/../.. ABSOLUTE)
+       endif()
+   endif()
+
 Zephyr CMake package source code
 ********************************
 
@@ -428,3 +494,4 @@ The following is an overview of those files
 .. _CMake package: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html
 .. _CMake user package registry: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#user-package-registry
 .. _CMake package version: https://cmake.org/cmake/help/latest/command/find_package.html#version-selection
+.. _CMake package search: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure


sigvartmh:
Couldn't do it the same way as the nrf soc flash driver as nrx does not expose a `nrfx_nvmc_bytes_write` function for `QSPI` changed the implementation to use stack instead of heap.

sigvartmh:
@kamlaz

drivers/flash/nrf_qspi_nor.c
```diff
@@ -491,15 +491,36 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
+	if ((size < sizeof(u32_t)) && (size != 0)) {
+		/* Minimal read size for QSPI is 4 bytes */
+		size = sizeof(u32_t);
+	}
+
+	if (((size % 4U) != 0) || (size == 0)) {
+		return -EINVAL;
+	}
+
+	u8_t tmp_buf[size];
```
sigvartmh, line 512: Should this be after `qspi_lock` @kamlaz ?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -491,15 +491,36 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
 	/* address must be 4-byte aligned */
 	if ((addr % 4U) != 0) {
 		return -EINVAL;
 	}
 
+#if IS_ENABLED(CONFIG_NORDIC_QSPI_NOR_FLASH_ONE_BYTE_RW_ACCESS)
+	void *tmp_dest = dest;
+	size_t var_size = size;
+
+	if ((size < sizeof(u32_t)) && (size != 0)) {
```
sigvartmh, line 503: Should there also be a check for max size which is Kconfigurable to ensure that the function won't use infinite stack?

kamlaz:
I have looked into your code and I have one question. Your solution provides possibility of writting less than four bytes, but in fact it writes four bytes, but vacants are filled with `0xFF` pattern. It is OK when we assume that the data overwritten with mentioned pattern is not needed anymore. What about this scenario: You would like to write 1 byte to some address. 
1. Read four bytes from desired address and copy them to the buffer.
2. Copy data to write to the buffer that contains data that is present in flash. Using this we overwrite only first byte, the rest (three bytes) remains the same.
3. Write this 4 byte buffer to the memory.
I am not sure if this case is useful. What do you think about it?

drivers/flash/nrf_qspi_nor.c
pabigot: I think it's reasonable for a hack like this to say it only works for sizes less than, say, 8 bytes, which would require a 16-byte buffer on the stack.  Misaligned/badly-sized transfers larger than 8 bytes can be rejected.

sigvartmh: Okei, I'll add it as a Kconfig. I'll also see if I can get it working to only allocate the array on the stack if < 4 or not in RAM.

sigvartmh:
Sure I can add back the read, copy and write version.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -491,15 +491,36 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 		return -EINVAL;
 	}
 
-	/* read size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
 	/* address must be 4-byte aligned */
 	if ((addr % 4U) != 0) {
```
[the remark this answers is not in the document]
sigvartmh: So what you are suggesting have been solved here https://github.com/NordicSemiconductor/nrfx/blob/59ad34410bf79b9b50b86c8befb9d71024f04c0b/drivers/src/nrfx_nvmc.c#L385-L446

So the question is should this be solved in the nrfx driver in the future or should it be done here?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -642,9 +687,9 @@ static int qspi_nor_configure(struct device *dev)
  */
 static int qspi_nor_init(struct device *dev)
 {
-	IRQ_CONNECT(DT_NORDIC_NRF_QSPI_QSPI_0_IRQ_0,
+	IRQ_DIRECT_CONNECT(DT_NORDIC_NRF_QSPI_QSPI_0_IRQ_0,
```
sigvartmh, line 690: This is wrong sorry I'll change it back.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -529,14 +555,34 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
...
 	}
+#if IS_ENABLED(CONFIG_NORDIC_QSPI_NOR_FLASH_ONE_BYTE_RW_ACCESS)
+	size_t tmp_size = size;
+
+	if (size < sizeof(u32_t)) {
+		tmp_size = 4;
+	}
+
+	/* initialize buffer to 0xFF to ensure no-OP for values outside size */
+	u8_t tmp_buf[tmp_size] = {0xFF};
```
sigvartmh, line 570: @kamlaz So what you are suggesting is
```
qspi_nor_read(tmp_buf, addr, tmp_size)
memcpy(tmp_buf, src, size)
```
Right?

drivers/flash/nrf_qspi_nor.c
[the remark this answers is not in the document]
pabigot: I don't know what nrfx does, or whether the QSPI wrapper in it should try to work around hardware alignment limitations as it apparently did for NVMC.

The case I was concerned about was reading a single byte at, say, address 0x03, which was not supported by zephyrproject-rtos/zephyr#25292 which I'd originally hoped to get into Zephyr 2.3.0 to fix the known mcuboot problem.   But it didn't get reviews, and even though it fixes mcuboot it's not as general as it could be.

Your approach is more complete, but it's not going into upstream 2.3.0 either, and if you think it's temporary I don't know what's going on or what the schedule is.  If the intent is to remove the alignment requirements at the nrfx level, why not start there?

pabigot:
> I have looked into your code and I have one question. Your solution provides possibility of writting less than four bytes, but in fact it writes four bytes, but vacants are filled with `0xFF` pattern. It is OK when we assume that the data overwritten with mentioned pattern is not needed anymore.

Is that really what happens?  If the underlying device is a standard SPI NOR, then doesn't writing 0xFF to a location leave it unchanged (because write only clears 1s to 0s, not sets 0s to 1s)?  In that case you could write 0x01FFFFFF, 0xFF02FFFF, 0xFFFF03FF, 0xFFFFFF04 to the same address in four separate transactions and read back 0x01020304.

I guess if the erase value of the device could be 0x00 then it'd be wrong and you'd need a read, set, write.

sigvartmh:
> I guess if the erase value of the device could be 0x00 then it'd be wrong and you'd need a read, set, write.

Yeah, it could be that I made a wrong assumption there expecting no-op for `0xFF` it could be that the SPI controller will interpret it differently.

The one on board the DK works the way you would expect. By having `0xFF` as no-op and `0xFFFFFFFF` as the erase word

pabigot:
The question is whether the QSPI peripheral supports devices with an erase value of zero.  I don't see any reason in principle why it wouldn't.

I've re-opened my upstream PR which would make mcuboot work in 2.3.0 if it got merged; it doesn't try to support sub-word writes, only reads, and only at word boundaries (which could be easily fixed).

drivers/flash/nrf_qspi_nor.c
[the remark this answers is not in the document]
sigvartmh: I don't know if the intent is to remove the alignment requirement at nrfx. I'll ask around :) 

We need to be able to write data from flash to fix MCUboot problem completely. Since writing the BOOT_TEST word uses data in flash.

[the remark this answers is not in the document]
sigvartmh: But I could add that on what you already have upstream.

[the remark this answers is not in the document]
pabigot: Could you comment on that issue explaining what you think is missing (i.e. where a sub-word write could happen)?  I'm pretty sure I did test image swapping with image-1 on the external flash and it worked.

sigvartmh:
My suggestion is to rework it a bit. I think I've identified 1 issue I need to fix and have this as a [nrf temphack] since I require more than what is suggested upstream. Then rework the QSPI NOR so that it fulfills the requirements of https://github.com/zephyrproject-rtos/zephyr/pull/23628

Then the [nrf temphack] can be removed.

sigvartmh:
Since https://github.com/zephyrproject-rtos/zephyr/pull/25292 has been merged I'll add that commit I assume that it will come in the upmerge and just focus on the write part.

drivers/flash/nrf_qspi_nor.c
sigvartmh: Added this now

sigvartmh:
> The question is whether the QSPI peripheral supports devices with an erase value of zero. I don't see any reason in principle why it wouldn't.
> 
> I've re-opened my upstream PR which would make mcuboot work in 2.3.0 if it got merged; it doesn't try to support sub-word writes, only reads, and only at word boundaries (which could be easily fixed).

So I changed the write to copy data read into a buffer then change the buffer before writing to ensure that whatever the device supports 0 or 0xFF it will work at least. Even if it's a bit slower this is a temporary hack then we should find the best solution when doing the word boundaries fix upstream to the driver.

sigvartmh:
@pabigot do you have time to have a look and see if you think this is okei?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -534,19 +534,47 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
+
+		if (res != 0) {
+			return qspi_get_zephyr_ret_code(res);
+		}
+		memcpy(buf, sptr, dlen);
+	} else if (((size % 4U) != 0) || (size == 0)) {
+		return -EINVAL;
+	}
+
+#if IS_ENABLED(NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH)
```
hakonfam, line 564: Do we need to have this configurable? Is there situations when you don't want this feature?

Also, is it possible to avoid CPP here? `if (IS_ENABLED) ...` etc.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -570,7 +598,6 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
 
 	qspi_wait_for_completion(dev, res);
-
```
hakonfam, line 564: nit: intentional?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -487,16 +487,19 @@ static inline int qspi_nor_read_id(struct device *dev,
 static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 			 size_t size)
 {
+	void *dptr = dest;
```
hakonfam, line 490: If this commit is cherry-picked from upstream then it should have a [nrf fromlist] tag.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -534,19 +534,47 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
-	/* write size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
 	/* address must be 4-byte aligned */
 	if ((addr % 4U) != 0) {
 		return -EINVAL;
 	}
 
+	/* read size must be non-zero multiple of 4 bytes */
```
hakonfam, line 550:
```suggestion
	/* write size must be non-zero multiple of 4 bytes */
```

Does this comment reflect that 1-3 byte writes is also allowed?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -534,19 +534,47 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
+	if (size < 4U) {
+		src = buf;
+		size = sizeof(buf);
+		nrfx_err_t res = nrfx_qspi_read(buf, size, addr);
```
hakonfam, line 554: Should we add a comment here stating why we perform a read in the write-function?

drivers/flash/nrf_qspi_nor.c
sigvartmh: Needed to ensure that we don't change the driver to suddenly use tons of stack instead of faulting.

drivers/flash/nrf_qspi_nor.c
sigvartmh: Will add

drivers/flash/nrf_qspi_nor.c
sigvartmh: No :/

drivers/flash/nrf_qspi_nor.c
sigvartmh: Added comment

sigvartmh:
> I have looked into your code and I have one question. Your solution provides possibility of writting less than four bytes, but in fact it writes four bytes, but vacants are filled with `0xFF` pattern. It is OK when we assume that the data overwritten with mentioned pattern is not needed anymore. What about this scenario: You would like to write 1 byte to some address.
> 
>     1. Read four bytes from desired address and copy them to the buffer.
> 
>     2. Copy data to write to the buffer that contains data that is present in flash. Using this we overwrite only first byte, the rest (three bytes) remains the same.
> 
>     3. Write this 4 byte buffer to the memory.
>        I am not sure if this case is useful. What do you think about it?

I've addressed this now right? @kamlaz

drivers/flash/nrf_qspi_nor.c
```diff
@@ -487,16 +487,19 @@ static inline int qspi_nor_read_id(struct device *dev,
...
 	}
 
 	/* read size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
-	/* address must be 4-byte aligned */
-	if ((addr % 4U) != 0) {
+	if (size < 4U) {
+		dest = buf;
```
de-nordic, line 500: I do not know. I find it counterintuitive that `*dptr` local variable gets defined and assigned `dest` function parameter, so that the parameter could be modified.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -519,25 +522,62 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
+	} else if (((size % 4U) != 0) || (size == 0)) {
+		return -EINVAL;
+	}
+
+#if IS_ENABLED(NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH)
+	if ((u32_t)src < CONFIG_SRAM_BASE_ADDRESS) {
+		/* if the value is in flash we need to copy it into RAM since
+		 * the DMA won't be able to transport data from flash to QSPI
+		 */
+		volatile u8_t rambuf[dlen];
```
de-nordic, line 572: What is the change of killing stack with this? Is there a limit for `dlen`?

sigvartmh: No limit, I can add a Kconfig to set a limit if that sounds better?

drivers/flash/nrf_qspi_nor.c
sigvartmh: That change if from upstream so I can't really do anything about that so I just copied the way it is upstream for write also.

drivers/flash/nrf_qspi_nor.c
de-nordic: If you could, that would be nice.

drivers/flash/nrf_qspi_nor.c
de-nordic: Ok then. My bad.

de-nordic: Actually if you have some spare time, could you echange the `dest` and `dptr`, you are working on the code anyway.

drivers/flash/nrf_qspi_nor.c
oyvindronningstad: Since this function is blocking, could we do this in a loop with a preset buffer length? I.e. copy n bytes into buffer, then write to QSPI, repeat.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -487,16 +487,19 @@ static inline int qspi_nor_read_id(struct device *dev,
...
 	/* read size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
-	/* address must be 4-byte aligned */
-	if ((addr % 4U) != 0) {
+	if (size < 4U) {
+		dest = buf;
+		size = sizeof(buf);
+	} else if (((size % 4U) != 0) || (size == 0)) {
```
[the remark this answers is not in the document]
sigvartmh: This is upstream can't really change it here :+1:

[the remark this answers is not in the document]
oyvindronningstad: What happens at line 499 if size is 0? :D

drivers/flash/nrf_qspi_nor.c
```diff
@@ -519,25 +522,62 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
-	/* write size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
 	/* address must be 4-byte aligned */
 	if ((addr % 4U) != 0) {
 		return -EINVAL;
 	}
 
+	/* write size must be non-zero multiple of 4 bytes */
```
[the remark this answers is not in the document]
sigvartmh: Copy paste from the one from upstream above.

drivers/flash/nrf_qspi_nor.c
de-nordic: Good point! Limits the risk that app devs accidentally let slit some big dlen value and crash the stack.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -487,16 +487,19 @@ static inline int qspi_nor_read_id(struct device *dev,
...
 		return -EINVAL;
 	}
 
 	/* read size must be non-zero multiple of 4 bytes */
-	if (((size % 4U) != 0) || (size == 0)) {
-		return -EINVAL;
-	}
-	/* address must be 4-byte aligned */
-	if ((addr % 4U) != 0) {
+	if (size < 4U) {
```
de-nordic, line 499: Can you throw away the comment from line above or change it to something that describes the operation. I know it is not yours but it is basically type of "Assigning 0 to i" comment.

drivers/flash/nrf_qspi_nor.c
sigvartmh: I would assume that could work.

sigvartmh: @de-nordic @oyvindronningstad  could you recheck this part now.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,9 +583,29 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
...
 
-	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
+	nrfx_err_t res;
+#if IS_ENABLED(NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH)
+	if (((u32_t)src < CONFIG_SRAM_BASE_ADDRESS) && !(size < 4U)) {
+		/* if the value is in flash we need to copy it into RAM since
+		 * the DMA won't be able to transport data from flash to QSPI
+		 */
+		src = buf;
+		for (size_t i = 0; i < size ; i += sizeof(buf)) {
```
sigvartmh, line 594: Due to the check above we can be sure that the data is aligned by 4 bytes.

drivers/flash/nrf_qspi_nor.c
sigvartmh: Updated comment now. Hope it's better.

drivers/flash/nrf_qspi_nor.c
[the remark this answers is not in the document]
sigvartmh: Updated with the same as the change to the comment above.

drivers/flash/nrf_qspi_nor.c
de-nordic: Thank!

drivers/flash/nrf_qspi_nor.c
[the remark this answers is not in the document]
oyvindronningstad: @sigvartmh ^^

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,9 +587,29 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 		return -EINVAL;
 	}
 
-	qspi_lock(dev);
 
-	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
+	nrfx_err_t res;
+#if IS_ENABLED(NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH)
```
oyvindronningstad, line 592: Not sure this needs to be configurable now.

drivers/flash/nrf_qspi_nor.c
oyvindronningstad: Did you consider having a larger buffer? (In case it's significant performance-wise - I don't know :))

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,9 +587,29 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
...
+			res = nrfx_qspi_write(src, size, addr);
+			qspi_wait_for_completion(dev, res);
+			if (res != NRFX_SUCCESS) {
+				return qspi_get_zephyr_ret_code(res);
+			}
+		}
+	} else
+#endif
+	{
+	qspi_lock(dev);
```
oyvindronningstad, line 610: Indentation

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,9 +587,29 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
...
+	res = nrfx_qspi_write(src, size, addr);
+	}
 
 	qspi_wait_for_completion(dev, res);
```
oyvindronningstad, line 614: Move inside `else`?

sigvartmh: ah yes.

drivers/flash/nrf_qspi_nor.c
sigvartmh: As you commented below it has a performance penalty. This allows the user to either use this workaround or copy their data in RAM in a local function doing it more efficiently.

drivers/flash/nrf_qspi_nor.c
sigvartmh: I was thinking just to reuse the buffer you already have. Since this is an edge case where you could easily work around it by just doing local copy of the data I think it's okei.

If you want speed it's much faster to copy all of your Flash data into a Ram buffer then pass it to the function than relaying on this.

drivers/flash/nrf_qspi_nor.c
sigvartmh: Fixed.

drivers/flash/nrf_qspi_nor.c
de-nordic: You can use `IS_ENABLED` in C if condition, it will just evaluate to 0 or 1, compiler will throw away the branch that will never get executed so the result will be the same but code will look more clear. Just use it in `if` in line 593.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,11 +587,30 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
...
-	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
 
-	qspi_wait_for_completion(dev, res);
+	nrfx_err_t res;
+#if IS_ENABLED(NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH)
+	if (((u32_t)src < CONFIG_SRAM_BASE_ADDRESS) && !(size < 4U)) {
+		/* if the value is in flash we need to copy it into RAM since
+		 * the DMA won't be able to transport data from flash to QSPI
+		 */
+		src = buf;
```
de-nordic, line 597: Can we get rid of `src` within this `if`, `buf` is the source of data here anyway, and it becomes confusing when in 599 there is copy to `buf` done, but in 601 write is done from `src` and between these lines there is no magic that would change `buf` to `src`.

drivers/flash/nrf_qspi_nor.c
de-nordic: Move  `IS_ENABLED` into if at line 593.

drivers/flash/nrf_qspi_nor.c
sigvartmh: Done

drivers/flash/nrf_qspi_nor.c
sigvartmh: Moved.

sigvartmh:
@oyvindronningstad What you want me to look at?

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,11 +587,30 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 		return -EINVAL;
 	}
 
-	qspi_lock(dev);
-
-	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
 
-	qspi_wait_for_completion(dev, res);
+	nrfx_err_t res;
+	if (IS_ENABLED(
```
de-nordic, line 592: Nearly, I was rather thinking about something like this:
```
if (IS_ENABLED(CONFIG_NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH) &&
    (((u32_t)src < CONFIG_SRAM_BASE_ADDRESS) && !(size < 4U)) {
...
} else {
....
}
```
This way you do not have to indent the body of if..else that deeply.

sigvartmh: Ah okei yeah that's better sorry for misunderstanding.

de-nordic: No worries.

oyvindronningstad:
> @oyvindronningstad What you want me to look at?

Sorry, "copy link" didn't work as I expected, take a look at my comment on line 504
https://github.com/nrfconnect/sdk-zephyr/pull/305/files#diff-4b9e934d450cd9a81c3bd43f4298f4c6R504

"What happens at line 501 if size is 0?"

drivers/flash/nrf_qspi_nor.c
sigvartmh: Hmm this doesn't seem to work :/

de-nordic:
> 
> 
> > @oyvindronningstad What you want me to look at?
> 
> Sorry, "copy link" didn't work as I expected, take a look at my comment on line 504
> https://github.com/nrfconnect/sdk-zephyr/pull/305/files#diff-4b9e934d450cd9a81c3bd43f4298f4c6R504
> 
> "What happens at line 501 if size is 0?"



> 
> 
> > @oyvindronningstad What you want me to look at?
> 
> Sorry, "copy link" didn't work as I expected, take a look at my comment on line 504
> https://github.com/nrfconnect/sdk-zephyr/pull/305/files#diff-4b9e934d450cd9a81c3bd43f4298f4c6R504
> 
> "What happens at line 501 if size is 0?"

I have also missed that :D.

drivers/flash/nrf_qspi_nor.c
de-nordic: Hm, give me a minute I need to checkout your code to see what is happening.

drivers/flash/nrf_qspi_nor.c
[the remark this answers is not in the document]
sigvartmh: Sorry I don't really know what you are getting at. Seems like this check should still be there from what you are commenting. It's invalid use of the driver and you should fault.

sigvartmh:
@oyvindronningstad @de-nordic So from what I can see is that it's invalid usage and you should return -EINVAL as it does.

oyvindronningstad:
> So from what I can see is that it's invalid usage and you should return -EINVAL as it does.

What should happen when `size` is 0 and what actually happens in the code now? It will hit the first `if` and will not return -EINVAL.

sigvartmh:
Sorry was looking at the line not the one above. Yes you are correct, it will do something weird.

drivers/flash/nrf_qspi_nor.c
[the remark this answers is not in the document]
pabigot: I see the point.  The first check should be `(size > 0) && (size < 4U)`.  I can do an upstream fix for that, might not make 2.3.0.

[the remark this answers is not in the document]
pabigot: Fix proposed in zephyrproject-rtos/zephyr#25661

drivers/flash/nrf_qspi_nor.c
```diff
@@ -536,19 +536,38 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
-	}
 	/* address must be 4-byte aligned */
 	if ((addr % 4U) != 0) {
 		return -EINVAL;
 	}
 
+	/* Since the QSPI driver requires data to be at least 4 bytes we need
+	 * to use a 4 byte buffer for writes smaller than 4 bytes.
+	 */
+	if (size < 4U) {
```
pabigot, line 555: This also needs a `size > 0` check.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -536,19 +536,38 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
 static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 			  size_t size)
 {
+	const void *sptr = src;
+	size_t dlen = size;
+	u8_t buf[4];
```
pabigot, line 541: There's a risk this buffer might not be aligned as required.  I'm updating the upstream PR to include a fix.

sigvartmh: Link?

drivers/flash/nrf_qspi_nor.c
sigvartmh: Moved the size check above. Since `size` is `typeof size_t` it's safe to assume that `size` can't be less than 0 right.

de-nordic: I think he has done it at line 543.

drivers/flash/nrf_qspi_nor.c
sigvartmh: Fixed.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,11 +587,28 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
...
 	}
 
-	qspi_lock(dev);
 
-	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
+	nrfx_err_t res;
 
-	qspi_wait_for_completion(dev, res);
+	if (IS_ENABLED(CONFIG_NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH) &&
+	   ((u32_t)sptr < CONFIG_SRAM_BASE_ADDRESS) && !(size < 4U)) {
```
de-nordic, line 594: Just one more space right ;)

sigvartmh: Okei :)

drivers/flash/nrf_qspi_nor.c
oyvindronningstad: > This allows the user to either use this workaround or copy their data in RAM in a local function doing it more efficiently.

They can still do that if there's no config. If they use a RAM buffer, this buffer will not be used.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -556,11 +587,28 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
 		return -EINVAL;
 	}
 
-	qspi_lock(dev);
 
-	nrfx_err_t res = nrfx_qspi_write(src, size, addr);
+	nrfx_err_t res;
 
-	qspi_wait_for_completion(dev, res);
+	if (IS_ENABLED(CONFIG_NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH) &&
```
oyvindronningstad, line 593: Shouldn't this fail when the config is not enabled and the source is in FLASH? Now it will go into the `else`.

sigvartmh: that's fine, then the NRFX driver will fail and the error is reported back to the user.

drivers/flash/nrf_qspi_nor.c
sigvartmh: Added the align fix.

drivers/flash/nrf_qspi_nor.c
oyvindronningstad: Ok, that's not quite clear to the. I still think the config is not needed.

You said:

> This allows the user to either use this workaround or copy their data in RAM in a local function doing it more efficiently.

They can still do that if there's no config. If they use a RAM buffer, this buffer will not be used.

sigvartmh:
> I'd prefer we remove the config, but I can live with this as it is. Thanks for all the changes.

I think it's best not to give users any surprise behavior they haven't opted in for.

oyvindronningstad:
> > I'd prefer we remove the config, but I can live with this as it is. Thanks for all the changes.
> 
> I think it's best not to give users any surprise behavior they haven't opted in for.

I'm not familiar with the existing QSPI driver, what is the surprise?

sigvartmh:
> > > I'd prefer we remove the config, but I can live with this as it is. Thanks for all the changes.
> > 
> > 
> > I think it's best not to give users any surprise behavior they haven't opted in for.
> 
> I'm not familiar with the existing QSPI driver, what is the surprise?

The speed penalty of using data in flash. Due to the double copy.

oyvindronningstad:
> > > > I'd prefer we remove the config, but I can live with this as it is. Thanks for all the changes.
> > > 
> > > 
> > > I think it's best not to give users any surprise behavior they haven't opted in for.
> > 
> > 
> > I'm not familiar with the existing QSPI driver, what is the surprise?
> 
> The speed penalty of using data in flash. Due to the small size copy.

Is that less surprising than an error?

sigvartmh:
> > > > > I'd prefer we remove the config, but I can live with this as it is. Thanks for all the changes.
> > > > 
> > > > 
> > > > I think it's best not to give users any surprise behavior they haven't opted in for.
> > > 
> > > 
> > > I'm not familiar with the existing QSPI driver, what is the surprise?
> > 
> > 
> > The speed penalty of using data in flash. Due to the small size copy.
> 
> Is that less surprising than an error?

Yes, if you get the error you can chose either apply workaround or copy your data locally and pass it to the drive.

drivers/flash/nrf_qspi_nor.c
```diff
@@ -519,25 +524,51 @@ static int qspi_nor_read(struct device *dev, off_t addr, void *dest,
...
+	/* Since the QSPI driver requires data to be at least 4 bytes we need
+	 * to use a 4 byte buffer for writes smaller than 4 bytes.
+	 */
+	if (size < 4U) {
+		src = buf;
+		size = sizeof(buf);
+		/* read out the whole word so that unchanged data can be
+		 * written back
+		 */
+		nrfx_err_t res = nrfx_qspi_read(buf, size, addr);
```
pabigot, line 561: This read is outside the lock.  For correctness it probably should be inside the same lock that's used for the write: a context switch could allow another thread to come in and change the value that was read into `buf` so the following write contains different data in areas that weren't supposed to change.

https://github.com/pabigot/fw-nrfconnect-zephyr/commits/pr/305 has a possible fix.

pabigot: Which isn't correct because it doesn't properly handle the potential for multiple writes in the loop below.

The locking infrastructure doesn't seem prepared to handle multiple nrfx_spi invocations for a single Zephyr-level transaction.  I'm not sure who should tackle this problem, though I can give it a shot.

pabigot: I believe the branch should work.  It passes a simple test, but `samples/subsys/fs/littlefs` doesn't work on the master branch so I can't confirm the changes work.
```
===================================================================
starting test - test_lfs_basic
clearing partition /sml
Error accessing flash area 5 [-2]

    Assertion failed at WEST_TOPDIR/zephyr/tests/subsys/fs/littlefs/src/test_lfs_basic.c:44: )
failed to wipe partition
FAIL - test_lfs_basic
```
This may be my environment; I don't use ncs generally.

sigvartmh: Oh I think my intention was to use the qspi read

pabigot: OK, updated again at https://github.com/pabigot/fw-nrfconnect-zephyr/commits/pr/305.  There're multiple upstream fixes in zephyrproject-rtos/zephyr#25661 that are part of that branch.

sigvartmh: So added your fix and most of it works fine. Just found out that MCUBoot also has some buffers it passes in that are not word aligned. So I must have been lucky and got word aligned buffers before.

I'll add a temp hack to MCUBoot

drivers/flash/nrf_qspi_nor.c
```diff
@@ -586,11 +587,28 @@ static int qspi_nor_write(struct device *dev, off_t addr, const void *src,
...
+					      addr + bytes_written);
+			qspi_wait_for_completion(dev, res);
+			if (res != NRFX_SUCCESS) {
+				return qspi_get_zephyr_ret_code(res);
+			}
+			bytes_written += sizeof(buf);
+		}
+	} else {
+		qspi_lock(dev);
+		res = nrfx_qspi_write(src, size, addr);
```
nvlsianpu, line 616: would be nice to comment that for data size < 4 this function make  the ram copy on its own.

oyvindronningstad:
> > > > > > I'd prefer we remove the config, but I can live with this as it is. Thanks for all the changes.
> > > > > 
> > > > > 
> > > > > I think it's best not to give users any surprise behavior they haven't opted in for.
> > > > 
> > > > 
> > > > I'm not familiar with the existing QSPI driver, what is the surprise?
> > > 
> > > 
> > > The speed penalty of using data in flash. Due to the small size copy.
> > 
> > 
> > Is that less surprising than an error?
> 
> Yes, if you get the error you can chose either apply workaround or copy your data locally and pass it to the drive.

Ok, make sure the error is well documented, and mentions the config, so the user knows the choice they have.

drivers/flash/nrf_qspi_nor.c
sigvartmh: I'll add a comment :+1:

drivers/flash/Kconfig.nordic_qspi_nor
```diff
@@ -29,4 +29,16 @@ config NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE
...
 
+config NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH
+	bool "Use stack for writing data stored in flash"
+	help
+	  When this option is enabled writing data from FLASH is possible.
+	  However the driver will use stack to copy the data from flash onto
+	  the stack then start the DMA transfer. If you get an -EINVALID error
+	  this could be due to the data being in flash. Either copy the data
+	  locally or enable this option. This uses a 4 byte buffer to copy the
+	  data which has a preformance penalty for large chunks of data. If the
```
torsteingrindvik, line 40: preformance -> performance

drivers/flash/Kconfig.nordic_qspi_nor
```diff
@@ -29,4 +29,16 @@ config NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE
 	  that API.  By default the page size corresponds to the block
 	  size (65536).  Other option include the sector size (4096).
 
+config NORDIC_QSPI_NOR_FLASH_ALLOW_STACK_USAGE_FOR_DATA_IN_FLASH
+	bool "Use stack for writing data stored in flash"
+	help
+	  When this option is enabled writing data from FLASH is possible.
+	  However the driver will use stack to copy the data from flash onto
+	  the stack then start the DMA transfer. If you get an -EINVALID error
```
oyvindronningstad, line 37: Thanks for adding. If you have the time, it would be better to have this documentation in the header file for qspi_nor_write, since I think that's where the user will look if they get the error, but it's not a must.

sigvartmh: Okei I'll try to move the information there.

drivers/flash/Kconfig.nordic_qspi_nor
sigvartmh: fixed. Thanks for spotting.

nvlsianpu:
@sigvartmh Will be this upstreamed?

sigvartmh:
@nvlsianpu all of it? or just sub-word @pabigot https://github.com/zephyrproject-rtos/zephyr/pull/25971 made an upstream PR it seems.

PR #316, opened 2020-06-04
KRKNWK-6261 fix

lmaciejonczyk:


rlubos:
For a6316a757e9d6483598dc12042ac4108886d1c28, please add `[nrf fromlist]` prefix to the commit title (also would be good to mention PR number in the commit body.
For a62cd8d66bb883d0a2ca4bcdb07b8078e889f0ad I'm not sure, is it targeting upstream? In any case it needs a downstream prefix as well, as described in the docs:
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/ug_dev_model.html#oss-repositories-downstream-project-history

lmaciejonczyk:
> For [a6316a7](https://github.com/nrfconnect/sdk-zephyr/commit/a6316a757e9d6483598dc12042ac4108886d1c28), please add `[nrf fromlist]` prefix to the commit title (also would be good to mention PR number in the commit body.
> For [a62cd8d](https://github.com/nrfconnect/sdk-zephyr/commit/a62cd8d66bb883d0a2ca4bcdb07b8078e889f0ad) I'm not sure, is it targeting upstream? In any case it needs a downstream prefix as well, as described in the docs:
> https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/ug_dev_model.html#oss-repositories-downstream-project-history

@rlubos do you think we can target [a62cd8d](https://github.com/nrfconnect/sdk-zephyr/commit/a62cd8d66bb883d0a2ca4bcdb07b8078e889f0ad) upstream? It definitely would be easier for us.

rlubos:
@lmaciejonczyk I think that we should try at least. If it's not allowed in the current form, we'll need to find acceptable solution (make a Kconfig switch between preemptive/cooparative thread). Worst case, we'll keep it downstream.

rlubos:
Shouldn't 0ecddee be a `[nrf toup]` instead of `[nrf noup]`?

PR #317, opened 2020-06-07
[nrf noup] ci: move jenkinsfile to test-ci-nrfconnect-cfg

thst-nordic:
- The stub loads a similar pipeline from the jenkins standard library. 
- This will keep restricted platforms private.
- Sends 'PENDING' status to github faster.


Jenkinsfile
```diff
@@ -1,354 +1,5 @@
...
-              }
-              // If not a PR, it's a non-PR-branch or master build. Compare against the origin.
-              else if (BUILD_TYPE == "BRANCH") {
-                COMMIT_RANGE = "origin/${env.BRANCH_NAME}..HEAD"
-                println "Building a Branch: " + COMMIT_RANGE
-              }
-              else {
-                  assert condition : "Build fails because it is not a PR/Tag/Branch"
-              }
+def pipeline = new ncs.sdk_zephyr.Main()
```
tejlmand, line 3: Hard to evaluate what happens here, so I guess I just have to trust that code will do what we want.

But regarding this PR itself, I was wondering if it would be better to make a revert of this PR:
https://github.com/nrfconnect/sdk-zephyr/commit/b52cb0e67fac97823a53766813692fdee211aea0

and then re-add the ` .github/workflows/backport.yml` file to this PR.
that should minimize the patchset during upmerge.
@mbolivar-nordic thoughts ?

mbolivar-nordic: I agree that would be cleaner, but I don't think it's a blocker. I'll notice in time for v1.4.0 and rewrite the history to make it so if this suggestion is not applied during this PR.

thst-nordic: I am on leave for next 3 months.  

@asle-nordic can you follow-up on this PR?

mbolivar-nordic:
Let's just get this in to avoid difficulties while @thst-nordic is on leave. I will clean it up later.

PR #321, opened 2020-06-11
[nrf fromlist] cmake: remove call to cmake and use ${CMAKE_COMMAND}

tejlmand:
This commit removes an occurence of `cmake` and instead uses the correct
form ${CMAKE_COMMAND}.

This fixes issues where CMake is invoked without being present in the
system PATH.

----- 
Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/26114

---- 
Fixes: NCSIDB-107

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1114,7 +1114,7 @@ set(post_build_byproducts "")
 list(APPEND
   post_build_commands
   COMMAND
-  cmake -E rename ${logical_target_for_zephyr_elf}.map ${KERNEL_MAP_NAME}
+  ${CMAKE_COMMAND} -E rename ${logical_target_for_zephyr_elf}.map ${KERNEL_MAP_NAME}
 )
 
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
I don't see an nRF PR for this and I don't see it tagged for v1.3.0 so I will not merge.

tejlmand:
> I don't see an nRF PR for this and I don't see it tagged for v1.3.0 so I will not merge.

and the reason for that, is because it will not make it into v1.3.0.
But having a PR prepared is useful here: https://devzone.nordicsemi.com/f/nordic-q-a/60815/linux-cmake-not-found/254400#254400

Manifest PR will be made after v1.3.0.

mbolivar-nordic:
> and the reason for that, is because it will not make it into v1.3.0.

Great, thanks for confirming.

tejlmand:
Closed as part of last upmerge

PR #322, opened 2020-06-12
West flash - erase chip when programming UICR

oyvindronningstad:
west flash hasn't worked for a while when using CONFIG_SECURE_BOOT since the bootloader places contents in the UICR, and on 53 and 91 --sectoranduicrerase is not available, so those have always been using --sectorerase. This patch changes that to use --chiperase when the hex file has contents in UICR.

### DIFF
--- a/scripts/west_commands/runners/nrfjprog.py
+++ b/scripts/west_commands/runners/nrfjprog.py
@@ -8,8 +8,26 @@
 import os
 import shlex
 import sys
+import re
 
 from runners.core import ZephyrBinaryRunner, RunnerCaps
+
+# Helper function for inspecting hex files.
+# has_region returns True if hex file has any contents in a specific region
+# region_filter is a callable that takes an address as argument and
+# returns True if that address is in the region in question
+def has_region(region_filter, hex_file):
+    with open (hex_file, 'r') as f:
+        offset = 0
+        for line in f:
+            offset_re = re.match(r":02[0-9a-fA-F]{4}(02|04)([0-9a-fA-F]{4})", line)
+            data_re = re.match(r":([0-9a-fA-F]{2})([0-9a-fA-F]{4})00", line)
+            if offset_re:
+                offset = {'02': 0x10, '04': 0x10000}[offset_re.group(1)] * int(offset_re.group(2), 16)
+            elif data_re:
+                if region_filter(int(data_re.group(2), 16) + offset):
+                    return True
+    return False
 
 
 class NrfJprogBinaryRunner(ZephyrBinaryRunner):
@@ -132,10 +150,21 @@ class NrfJprogBinaryRunner(ZephyrBinaryRunner):
                 program_cmd
             ])
         else:
-            if self.family == 'NRF52':
+            if self.family == 'NRF51':
+                commands.append(program_cmd + ['--sectorerase'])
+            elif self.family == 'NRF52':
                 commands.append(program_cmd + ['--sectoranduicrerase'])
             else:
-                commands.append(program_cmd + ['--sectorerase'])
+                is_uicr = {
+                    'NRF53': lambda addr: (0x00FF8000 <= addr < 0x00FF9000) or (0x01FF8000 <= addr < 0x01FF8800),
+                    'NRF91': lambda addr: (0x00FF8000 <= addr < 0x00FF9000),
+                }[self.family]
+
+                if has_region(is_uicr, self.hex_):
+                    # Hex file has UICR contents.
+                    commands.append(program_cmd + ['--chiperase'])
+                else:
+                    commands.append(program_cmd + ['--sectorerase'])
 
         if self.family == 'NRF52' and not self.softreset:
             commands.extend([


hakonfam:
FYI https://github.com/zephyrproject-rtos/zephyr/pull/21928

oyvindronningstad:
@hakonfam Why was it closed?

scripts/west_commands/runners/nrfjprog.py
```diff
@@ -8,9 +8,17 @@
 import os
 import shlex
 import sys
+from intelhex import IntelHex
```
mbolivar-nordic, line 11: Needs to be wrapped in a try; please see https://github.com/zephyrproject-rtos/zephyr/blob/master/scripts/west_commands/runners/core.py#L318 for the policy and here for examples:

https://github.com/zephyrproject-rtos/zephyr/blob/master/scripts/west_commands/runners/openocd.py#L9
https://github.com/zephyrproject-rtos/zephyr/blob/master/scripts/west_commands/runners/openocd.py#L113

scripts/west_commands/runners/nrfjprog.py
```diff
@@ -19,6 +19,38 @@
...
+# Helper function for nRF53
+def has_cpuapp_region(hex_file):
+    return has_region(lambda addr: not is_cpunet(addr), hex_file)
+
+# Helper function for nRF53
+# Generates 2 new hex files where cpunet flash contents are in one,
+# and all other flash contents are in the other.
+def split_hex_file(hex_file):
+    hex_file_dir = os.path.dirname(hex_file)
+    cpuapp_hex = os.path.join(hex_file_dir, "nrf53_cpuapp.hex")
```
mbolivar-nordic, line 39: I am definitely OK with this since it fixes a bug, but I'm not sure if we should be doing this in the build system instead longer-term to avoid potentially expensive hex file recreation even when dependencies are up to date.

@tejlmand  what do you think about this in the longer term?

scripts/west_commands/runners/nrfjprog.py
```diff
@@ -119,18 +151,43 @@ def do_run(self, command, **kwargs):
...
 
-        program_cmd = ['nrfjprog', '--program', self.hex_, '-f', self.family,
-                    '--snr', board_snr] + self.tool_opt
+        if self.family != "NRF53":
+            self.do_flash(self.hex_, self.tool_opt, board_snr)
+        else:
+            # self.family == "NRF53"
+
+            def flash_app(hex_file, reset=True):
+                self.do_flash(hex_file, self.tool_opt + ['--coprocessor', 'CP_APPLICATION'], board_snr, reset)
```
mbolivar-nordic, line 167:
```suggestion
                self.do_flash(hex_file, self.tool_opt + ['--coprocessor', 'CP_APPLICATION'], board_snr, reset=reset)
```

Just a personal preference to use keyword arguments explicitly if you don't mind.

scripts/west_commands/runners/nrfjprog.py
```diff
@@ -119,18 +151,43 @@ def do_run(self, command, **kwargs):
...
+
+            def flash_net(hex_file, reset=True):
+                self.do_flash(hex_file, self.tool_opt + ['--coprocessor', 'CP_NETWORK'], board_snr, reset)
```
mbolivar-nordic, line 170:
```suggestion
                self.do_flash(hex_file, self.tool_opt + ['--coprocessor', 'CP_NETWORK'], board_snr, reset=reset)
```

scripts/west_commands/runners/nrfjprog.py
```diff
@@ -119,18 +151,43 @@ def do_run(self, command, **kwargs):
...
+
+            if has_cpuapp_region(self.hex_) and has_cpunet_region(self.hex_):
+                (cpuapp_hex, cpunet_hex) = split_hex_file(self.hex_)
+                self.logger.info(f'hex file ({self.hex_}) was split into app ({cpuapp_hex}) and net ({cpunet_hex})')
```
mbolivar-nordic, line 174: nice

hakonfam:
> @hakonfam Why was it closed?

@tejlmand came up with the solution where you add a custom command performing `ninja flash` and add this as a dependency to the `flash` target.

```
    add_custom_target(${ACI_NAME}_flash
                      COMMAND
                      ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR}/${ACI_NAME}
                      --target flash
                      DEPENDS
                      ${ACI_NAME}_subimage

    set_property(TARGET zephyr_property_target
                 APPEND PROPERTY FLASH_DEPENDENCIES
                 ${ACI_NAME}_flash
```

We see, however, that this solution does not fix all use cases, as `west flash` does not invoke `ninja flash`. This approach would solve that use case as well.

scripts/west_commands/runners/nrfjprog.py
oyvindronningstad: I tested it now, `split_hex_file()` takes `0.4274630546569824` seconds for my example. I added hashing and a depfile, and brought it down to `0.0014848709106445312`. I'll add it as a separate commit.

oyvindronningstad:
> One question though, why not just send this upstream?

I wanted to see the response first. I'll open a PR now.

scripts/west_commands/runners/nrfjprog.py
tejlmand: to be honest, we have all the information we need already, so we don't need to first merge, and then split the hex file again.

I would prefer we make a more generic approach so the the knowledge of multiple hex files can be passed to other parts of the system.

tejlmand:
As @hakonfam comment shows, there is a discrepancy on how `west flash` and `ninja flash` works.
Ideally, using `west flash` and `ninja flash` should always produce same result, so we should look into how the knowledge can be passed from CMake into `west flash`

oyvindronningstad:
> As @hakonfam comment shows, there is a discrepancy on how `west flash` and `ninja flash` works.
> Ideally, using `west flash` and `ninja flash` should always produce same result, so we should look into how the knowledge can be passed from CMake into `west flash`

With this patch, `west flash` and `ninja flash` produce the same result. They achieve it in slightly different ways, since ninja naturally has more insight into the build. What do you see as the advantage of passing more knowledge from CMake to west flash compared to reading the hex file directly?

oyvindronningstad:
To me the advantages of this fix are:

- It will "always" work. No hex file will make it fail, unless that hex file is properly invalid (contains data in addresses not in flash)
- It allows `west flash` to be work without too much coupling with the cmake (i.e. running west flash on a hex file manually from the command line works without having to know a lot of internal stuff.
- It's contained in just one file (nrfjprog.py) with one exception which is a trivial simplification.

I think it gives `west flash` a cleaner interface.

oyvindronningstad:
If we want `west flash` to reuse the files generated from, I propose keeping the automatic detection of CP_APPLICATION/CP_NETWORK. We probably also need to modify the west flash CLI to accept multiple hex files.

One consideration is if somehow, cpuapp samples start containing cpunet contents or vice versa (after all, samples are built for a _core_ not a flash region per se). Not a problem now, but keeping the whole patch from this PR will allow that.

tejlmand:
@oyvindronningstad Please take a look here:
https://github.com/nrfconnect/sdk-zephyr/pull/324
https://github.com/nrfconnect/sdk-nrf/pull/2487

as simple as just supporting multiple `runners.yml` files, one for each image that we need to flash.
We already fetch the CMake cache, so just supply a list instead of a single file is much cleaner.
https://github.com/nrfconnect/sdk-zephyr/blob/master/scripts/west_commands/run_common.py#L272

oyvindronningstad:
That's very neat @tejlmand, thanks a lot for looking into it!

Your solution solves the original problem, so we can go ahead and get it in.

I will rework this PR to have just the UICR stuff and the CP_NETWORK/CP_APPLICATION detection.

I will also reconsider the file splitting, maybe I'll open a separate PR since I still think it has merit for working with more unusual hex files.

scripts/west_commands/runners/nrfjprog.py
tejlmand: also note, the proposed solution here only works for `nrfjprog` runner, and not openocd.

tejlmand:
> I will also reconsider the file splitting, maybe I'll open a separate PR since I still think it has merit for working with more unusual hex files.

I was wondering if such a feature should actually make it into `nrfjprog` instead ?
For example: `nrfjprog --program <hex-file> --from-addr <addr> --to-addr <addr>`

that way the feature would be a bit more generic and useful outside `west`

and it would remove the need for splitting the hex-file into multiple files.

oyvindronningstad:
> I was wondering if such a feature should actually make it into `nrfjprog` instead ?
> For example: `nrfjprog --program <hex-file> --from-addr <addr> --to-addr <addr>`

I like the idea. It could even be implicit from the `--coprocessor`, so doing

    nrfjprog --program <hex-file> --coprocessor CP_APPLICATION
    nrfjprog --program <hex-file> --coprocessor CP_NETWORK

would flash the whole hex file. I'll create an issue to propose both ideas.

scripts/west_commands/runners/nrfjprog.py
```diff
@@ -132,10 +150,21 @@ def do_run(self, command, **kwargs):
...
             else:
-                commands.append(program_cmd + ['--sectorerase'])
+                is_uicr = {
+                    'NRF53': lambda addr: (0x00FF8000 <= addr < 0x00FF9000) or (0x01FF8000 <= addr < 0x01FF8800),
+                    'NRF91': lambda addr: (0x00FF8000 <= addr < 0x00FF9000),
+                }[self.family]
+
+                if has_region(is_uicr, self.hex_):
+                    # Hex file has UICR contents.
+                    commands.append(program_cmd + ['--chiperase'])
```
tejlmand, line 165: I assume the build system (CMake) knows that data is being placed in UICR.
If part of the build system decides to place data into UICR, shouldn't it update `runners.yml` section with `--chiperase` information instead ?

tejlmand: If you provide a bit of info where UICR is added to hex file, then I can take a look.

oyvindronningstad: Search for `provision_PM_HEX_FILE`

carlescufi:
@mbolivar-nordic please revisit

mbolivar-nordic:
> @mbolivar-nordic please revisit

Looks like this is waiting for a response from @tejlmand. I am confused about all the open PRs and don't feel like I understand what's going on well enough. I am trying to schedule a call for next week to try to get clued in.

PR #327, opened 2020-06-22
[nrf fromlist] soc: arm: Kconfig: Add more nrf flash protection block sizes

oyvindronningstad:
Add
- NRF_MPU_FLASH_REGION_SIZE
- NRF_BPROT_FLASH_REGION_SIZE
- NRF_ACL_FLASH_REGION_SIZE

NRF_SPU_FLASH_REGION_SIZE is already available.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/26335

NRF PR: https://github.com/nrfconnect/sdk-nrf/pull/2515

### DIFF
--- a/soc/arm/Kconfig
+++ b/soc/arm/Kconfig
@@ -57,6 +57,30 @@ config NRF_SPU_RAM_REGION_SIZE
 	  RAM region size for the NRF_SPU peripheral
 endif
 
+if HAS_HW_NRF_MPU
+config NRF_MPU_FLASH_REGION_SIZE
+	hex
+	default 0x1000
+	help
+	  FLASH region size for the NRF_MPU peripheral (nRF51).
+endif
+
+if HAS_HW_NRF_BPROT
+config NRF_BPROT_FLASH_REGION_SIZE
+	hex
+	default $(dt_node_int_prop_hex,$(DT_CHOSEN_ZEPHYR_FLASH),erase-block-size)
+	help
+	  FLASH region size for the NRF_BPROT peripheral (nRF52).
+endif
+
+if HAS_HW_NRF_ACL
+config NRF_ACL_FLASH_REGION_SIZE
+	hex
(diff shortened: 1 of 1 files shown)


soc/arm/Kconfig
```diff
@@ -57,6 +57,30 @@ config NRF_SPU_RAM_REGION_SIZE
 	  RAM region size for the NRF_SPU peripheral
 endif
 
+if HAS_HW_NRF_MPU
+config NRF_MPU_FLASH_REGION_SIZE
+	hex
+	default 0x1000
```
hakonfam, line 63: Just curious, why is this hard coded while the others are extracted from DT?

hakonfam:
I think we need a corresponding sdk-nrf PR to get nrf-CI on this.

soc/arm/Kconfig
oyvindronningstad: Because it's different from the flash page size. NRF_MPU is for nrf51, which has flash page size of 0x400.

oyvindronningstad:
> I think we need a corresponding sdk-nrf PR to get nrf-CI on this.
> 
> Edit: [nrfconnect/sdk-nrf#2515](https://github.com/nrfconnect/sdk-nrf/pull/2515) pulls this in.

Thanks, sorry for not linking.

oyvindronningstad:
I think the DNM can be removed since this does no harm going in. It doesn't have to be synchronized with the nrf PR.

tejlmand:
@oyvindronningstad is this still valid after latest upmerge ?

oyvindronningstad:
> 
> 
> @oyvindronningstad is this still valid after latest upmerge ?

This will not be needed after the next upmerge.

oyvindronningstad:
> 
> 
> > @oyvindronningstad is this still valid after latest upmerge ?
> 
> This will not be needed after the next upmerge.

Merged in https://github.com/zephyrproject-rtos/zephyr/pull/26335

PR #329, opened 2020-07-07
BOARD, SOC, DTS, ARCH ROOT support in Zephyr modules.

tejlmand:
This PR is a series of commits for supporting ROOT definitions in Zephyr modules, `zephyr/module.yml`.

Supported with this commit:
- BOARD_ROOT
- SOC_ROOT
- DTS_ROOT
- ARCH_ROOT

---------
Upstream PR: 

--------
Manifest PR:


tejlmand:
@carlescufi no, not needed.
This was simply for making a x_ROOT test in ncs as an additional verification of this enhancment, and to be prepared on NCS adoption on later upmerge.

So this should not be merged.

tejlmand:
Thinking about it, I could just have updated the manifest to point directly to my own branch on `sdk-zephyr`.
No need for this PR actually. Guess i'm just to used to create PRs :wink:

PR #331, opened 2020-07-14
Upmerge 2020 jul 13

mbolivar-nordic:
Bring in upstream to 4ef29b34e3a29d7b77fddf7baf09990145ab1fed.
nrf: https://github.com/nrfconnect/sdk-nrf/pull/2585

PR #332, opened 2020-07-15
From tree changes required by low power UART (#2591)

nordic-krch:
Note that current upmerge (#331) contains those changes so PR will probably be closed but it is needed to run https://github.com/nrfconnect/sdk-nrf/pull/2591


carlescufi:
@mbolivar is this superseeded by the upmerge?

PR #334, opened 2020-07-23
[nrf fromlist] drivers: serial: nrfx: disable before setting pin config

plskeggs:
Per the nrf9160 datasheet, UARTE must be disabled prior to configuring
PSEL.TXD, PSEL.RXD, PSEL.CTS, and/or PSEL.RTS.  This also ensures that
any prior running code, such as a bootloader, which leaves the UART
enabled will not prevent the Zephyr application from changing the
configuration.

Fixes 27080


mbolivar-nordic:
This looks stale; I think I just got a ping about it because master rebased.

@plskeggs I'm going to close for now -- could you please rebase and reopen if still relevant?

PR #335, opened 2020-07-23
[nrf fromlist] lib/os/work_q: sanity check work_q handler prior to calling it

plskeggs:
Just as NULL pointers should not be dereferenced, they should
not be called either.

Fixes 26723

### DIFF
--- a/lib/os/work_q.c
+++ b/lib/os/work_q.c
@@ -25,6 +25,7 @@ void z_work_q_main(void *work_q_ptr, void *p2, void *p3)
 		}
 
 		handler = work->handler;
+		__ASSERT(handler != NULL, "handler must be provided");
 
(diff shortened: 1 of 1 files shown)


plskeggs:
cc @mbolivar-nordic

mbolivar-nordic:
Do you really need this downstream now, and can't wait for the next mergeup?

I assumed the incoming patch we discussed earlier today was going to fix a bug or something like that if it was being targeted for sdk-zephyr as a fromlist instead of waiting for a mergeup. I'm not sure it has to go in now instead of waiting. As mentioned, every out of tree patch is tech debt that must be reverted and tracked manually by the maintainers, so we try to keep their numbers down.

plskeggs:
Actually, you're right, it can wait.

PR #337, opened 2020-07-24
Openthread cherrypick build sources and custom mbedtls

MarekPorwisz:
Needed for finishing building thread from libraries and c310 support before I go on PTO

### DIFF
--- a/subsys/net/l2/openthread/Kconfig
+++ b/subsys/net/l2/openthread/Kconfig
@@ -22,6 +22,15 @@ menuconfig NET_L2_OPENTHREAD
 	select REBOOT
 	select ENTROPY_GENERATOR
 
+	imply NET_UDP
+	imply NET_IPV6
+	imply NET_CONFIG_NEED_IPV6
+
+if NET_L2_OPENTHREAD
+
+config OPENTHREAD_MBEDTLS
+	bool "Enable built-in mbedtls for use with OpenThread"
+	default y
 	select MBEDTLS
 	select MBEDTLS_ENABLE_HEAP
 	select MBEDTLS_CIPHER_AES_ENABLED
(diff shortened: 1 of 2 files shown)


subsys/net/l2/openthread/Kconfig
```diff
@@ -31,24 +40,35 @@ menuconfig NET_L2_OPENTHREAD
...
-	select MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED
-	select MBEDTLS_ECP_DP_SECP256R1_ENABLED
-	select MBEDTLS_ECP_NIST_OPTIM
-	select MBEDTLS_SSL_EXPORT_KEYS
-	select MBEDTLS_CTR_DRBG_ENABLED
-	select MBEDTLS_HMAC_DRBG_ENABLED
+config OPENTHREAD_SOURCES
+	bool "OpenThread from sources"
+	help
+	  Build Zephyrs OpenThread port from sources.
```
lmaciejonczyk, line 69: Isn't the _Zephyrs OpenThread **copy**_  more precise than _Zephyrs OpenThread **port**_ in this case? What do you mean by port? I would just name the things what they are.

subsys/net/l2/openthread/Kconfig
```diff
@@ -31,24 +40,35 @@ menuconfig NET_L2_OPENTHREAD
...
 
-	imply NET_UDP
-	imply NET_IPV6
-	imply NET_CONFIG_NEED_IPV6
+config OPENTHREAD_MBEDTLS_TARGET
+	string "mbedtls target name"
+	default "mbedTLS"
+	help
+	  This option allows to specify custom mbedtls CMake target name for
+	  openthread. This is sometimes needed to utilize hardware cryptography
```
lmaciejonczyk, line 57: It is too general, try be more specific.

PR #340, opened 2020-07-30
OT nrf53 support

lmaciejonczyk:


subsys/net/lib/openthread/platform/entropy.c
```diff
@@ -15,12 +16,9 @@
 
 LOG_MODULE_REGISTER(net_otPlat_entropy, CONFIG_OPENTHREAD_L2_LOG_LEVEL);
 
-#if !defined(CONFIG_ENTROPY_HAS_DRIVER)
-#error OpenThread requires an entropy source for a TRNG
-#endif
-
 otError otPlatEntropyGet(uint8_t *aOutput, uint16_t aOutputLength)
 {
+#if defined(CONFIG_ENTROPY_HAS_DRIVER)
```
rlubos, line 21: How come nRF53 has no entropy device?

lmaciejonczyk: It has in the network core domain (RNG), which by default is not accessible by app core. It also has CC312 but what I understand is not supported yet, right?

rlubos: I see, I guess then it's just a temporary condition that we don't have an entropy device, but I think it's ok to adapt the code for that.

subsys/net/lib/openthread/platform/entropy.c
```diff
@@ -15,12 +16,9 @@
 
 LOG_MODULE_REGISTER(net_otPlat_entropy, CONFIG_OPENTHREAD_L2_LOG_LEVEL);
 
-#if !defined(CONFIG_ENTROPY_HAS_DRIVER)
-#error OpenThread requires an entropy source for a TRNG
```
rlubos, line 19: I'd degrade it to a warning level, but still print out a warning in case we don't have a valid entorpy device in the system, something like:
```
"No entropy device in the system, cryptographic operations may be insecure!"
```
Just that the user is aware it's using a fake entropy source.

nvlsianpu:
@lmaciejonczyk update?

PR #341, opened 2020-08-04
west: Provide path hint for manifest project

tomchy:
If the Manifest object is created without providing the `topdir` argument and the manifest file does not set `self.path`, the resulting `ManifestProject` has path set to None.
As a result, all commands defined in `zephyr_module.py` fail.

Signed-off-by: Tomasz Chyrowicz <tomasz.chyrowicz@nordicsemi.no>

### DIFF
--- a/scripts/zephyr_module.py
+++ b/scripts/zephyr_module.py
@@ -188,9 +188,9 @@ def main():
         # West is imported here, as it is optional (and thus maybe not installed)
         # if user is providing a specific modules list.
         from west.manifest import Manifest
-        from west.util import WestNotFound
+        from west.util import WestNotFound, west_topdir
         try:
-            manifest = Manifest.from_file()
+            manifest = Manifest.from_file(topdir=west_topdir())
             projects = [p.posixpath for p in manifest.get_projects([])]
         except WestNotFound:
             # Only accept WestNotFound, meaning we are not in a west


tomchy:
Alternative solutions:
 - Make `self.path` obligatory in all manifest files.
 - Set the `manifest_path` in `Manifest.from_file(..)` method if none of the arguments are provided.

tomchy:
Issue introduced by: https://github.com/nrfconnect/sdk-zephyr/commit/ef3c5e551676278ccc1f42ecb572838144c5e33a

carlescufi:
@tejlmand can you please take a look?

tejlmand:
@tomchy could you check if this also solves your issue: https://github.com/zephyrproject-rtos/zephyr/commit/02dea9279bc9cfb4a2944cfbb67a091f6dda97fe

That code was accidentally deleted in the code you refer to.

tejlmand:
could you provide a bit more info on how you reproduce this issue, cause I have not been able to reproduce it.

tomchy:
The commit that you have linked does not solve the issue.
If you create an empty repository with the following manifest file:
```
manifest:
  defaults:
    remote: ncs

  remotes:
      # nRF Connect SDK base URL.
    - name: ncs
      url-base: https://github.com/nrfconnect

  projects:
    - name: nrf
      repo-path: sdk-nrf
      path: nrf
      remote: ncs
      revision: master
      import: true
```
You will be able to download all repositories via `west update`, but you will not be able to build examples.

If you add the following entry at the end of the manifest file, the issue is not reproduced:
```
  self:
    path: <relative-path-to-the-created-repo>
```

scripts/zephyr_module.py
```diff
@@ -188,9 +188,9 @@ def main():
         # West is imported here, as it is optional (and thus maybe not installed)
         # if user is providing a specific modules list.
         from west.manifest import Manifest
-        from west.util import WestNotFound
+        from west.util import WestNotFound, west_topdir
         try:
-            manifest = Manifest.from_file()
+            manifest = Manifest.from_file(topdir=west_topdir())
```
tejlmand, line 193: This is interesting.
I have just run a pdb on west, and when called as `Manifest.from_file()` it looks up `topdir` by itself, and that is working, so currently investigating what is actually happening.

tejlmand:
Fixed here: https://github.com/zephyrproject-rtos/west/pull/422

Feel free to try creating a new local `west` package for testing, see how to do so here:
https://github.com/zephyrproject-rtos/west/#installing-from-source

PR #345, opened 2020-08-20
Upmerge 20.08.2020

rlubos:


rlubos:
@mbolivar-nordic Thanks, nice catch! Those two commits were missing in `west ncs-compare`, which I based on. We should really emphasize in reviews that contributors **should not** trim or modify  the commit title in any way.

PR #347, opened 2020-08-24
Conf file include

tejlmand:
This PR is WIP regarding suggestion to support conf file split.

See: https://github.com/nrfconnect/sdk-nrf/pull/2707#discussion_r474166368

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -455,38 +455,18 @@ add_subdirectory(subsys)
 add_subdirectory(drivers)
 
 # Include zephyr modules generated CMake file.
-if(EXISTS ${CMAKE_BINARY_DIR}/zephyr_modules.txt)
-	file(STRINGS ${CMAKE_BINARY_DIR}/zephyr_modules.txt ZEPHYR_MODULES_TXT
-		ENCODING UTF-8)
-  set(module_names)
-
-  foreach(module ${ZEPHYR_MODULES_TXT})
-    # Match "<name>":"<path>" for each line of file, each corresponding to
-    # one module. The use of quotes is required due to CMake not supporting
-    # lazy regexes (it supports greedy only).
-    string(REGEX REPLACE "\"(.*)\":\".*\"" "\\1" module_name ${module})
-    string(REGEX REPLACE "\".*\":\"(.*)\"" "\\1" module_path ${module})
-
-    list(APPEND module_names ${module_name})
-
-    string(TOUPPER ${module_name} MODULE_NAME_UPPER)
-    set(ZEPHYR_${MODULE_NAME_UPPER}_MODULE_DIR ${module_path})
-    set(ZEPHYR_${MODULE_NAME_UPPER}_MODULE_DIR ${module_path} PARENT_SCOPE)
-  endforeach()
-
-  foreach(module_name ${module_names})
-    # Note the second, binary_dir parameter requires the added
-    # subdirectory to have its own, local cmake target(s). If not then
-    # this binary_dir is created but stays empty. Object files land in
-    # the main binary dir instead.
-    # https://cmake.org/pipermail/cmake/2019-June/069547.html
-    string(TOUPPER ${module_name} MODULE_NAME_UPPER)
-    set(ZEPHYR_CURRENT_MODULE_DIR ${ZEPHYR_${MODULE_NAME_UPPER}_MODULE_DIR})
-    add_subdirectory(${ZEPHYR_CURRENT_MODULE_DIR} ${CMAKE_BINARY_DIR}/modules/${module_name})
-  endforeach()
-  # Done processing modules, clear ZEPHYR_CURRENT_MODULE_DIR.
-  set(ZEPHYR_CURRENT_MODULE_DIR)
-endif()
+foreach(module_name ${ZEPHYR_MODULE_NAMES})
+  # Note the second, binary_dir parameter requires the added
+  # subdirectory to have its own, local cmake target(s). If not then
+  # this binary_dir is created but stays empty. Object files land in
+  # the main binary dir instead.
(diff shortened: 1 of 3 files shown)


tejlmand:
@mbolivar-nordic would like to have some early feedback on whether or not you believe such approach could be accepted upstream.

I did considere to extend the Kconfig conf-format to allow for `source "file.conf"`, but as that would change the conf format itself, when comparing to Linux kernel, then I believe that will be hard to get acceptance for.

Whereas, the `#include` will simply be seen as a comment when processed by Kconfig.

edmont (approved):
Thanks @tejlmand, this looks great!

The only drawback I can see is the users being confused by a commented line with actual functionality.  Maybe some update of the Visual Studio Code Kconfig extension by @trond-snekvik would be beneficial in that sense.

mbolivar-nordic:
> would like to have some early feedback on whether or not you believe such approach could be accepted upstream.

@tejlmand I really doubt that preprocessing the files with a regular expression will be acceptable to upstream.

The discussion thread in #2707 is very long and the link in the PR description is to a comment in the middle which relies on context from earlier in the thread.

Can you explain why simply appending the overlay onto the extra conf files from CMakeLists.txt, before pulling in boilerplate.cmake, is not acceptable?

tejlmand:
> > would like to have some early feedback on whether or not you believe such approach could be accepted upstream.
> 
> @tejlmand I really doubt that preprocessing the files with a regular expression will be acceptable to upstream.
> 
> The discussion thread in #2707 is very long and the link in the PR description is to a comment in the middle which relies on context from earlier in the thread.
> 
> Can you explain why simply appending the overlay onto the extra conf files from CMakeLists.txt, before pulling in boilerplate.cmake, is not acceptable?

short story is that @edmont (and others I believe) want the possibility to share 3 different overlay config among OpenThread projects, as example an `openthread-all-overlay.conf`, `openthread-ftd.conf`, and `openthread-mtd.conf`.

As those settings are common, and may be used for certification by inheritance, they would like an easy approach.

Of course the `-DOVERLAY_CONF=openthread-ftd.conf` is possible, and I already proposed the `OVERLAY_CONF`.
But seems they would like even simpler end-user approach where it is possible to simply include another conf file.

And I do see the benefit of being able to do so.
Another option is of course to do this directly in the CMakeLists.txt, but that would mean that users should change CMakeLists.txt file when they want to use `openthread-ftd.conf` instead of `openthread-mtd.conf`.
Being able to change the conf file itself is a little cleaner, compared to changing CMakeLists.txt for this, imho.

mbolivar-nordic:
>  Being able to change the conf file itself is a little cleaner, compared to changing CMakeLists.txt for this, imho.

This feature as proposed emulates the C preprocesor, without actually being the C preprocessor. I think that's not as clean as a single line of CMake and no new mechanism, personally.

I also don't really think this is "simple". It looks simple, but it's not.

edmont:
Hi @tejlmand @mbolivar 

The current use case is not only about choosing the OpenThread features set (Master, FTD, MTD) but also about sharing the common OpenThread stack configuration among the samples. This means at least two overlay files being used, but some samples come with more overlays for extra functionality.

So, with the existing approach the user would have to type, for instance:

```
cd nrf/samples/openthread/cli/
west build -b nrf52833dk_nrf52833 -- -DCONF_FILE="../common/overlay-ot-defaults.conf ../common/overlay-ot-features-master.conf" -DOVERLAY_CONFIG="harness/overlay-cert.conf"
```

Every time building is needed. Versus having these lines in the `prj.conf`:

```
# Include OpenThread defaults
#include "${ZEPHYR_NRF_MODULE_DIR}/samples/openthread/common/overlay-ot-defaults.conf"

# Enable OpenThread master features set
#include "${ZEPHYR_NRF_MODULE_DIR}/samples/openthread/common/overlay-ot-features-master.conf"
```

And typing:

`west build -b nrf52833dk_nrf52833 -- -DOVERLAY_CONFIG="harness/overlay-cert.conf"`

Or just using the sample defaults and type:

`west build -b nrf52833dk_nrf52833`

I also find the new approach easier for the user and less confusing.

trond-snekvik:
Sorry for gatecrashing, I just wanted to voice support for @mbolivar-nordic's point of view. I think it would be much cleaner to find a solution in the kconfig files instead of creating a new dialect of a file format that hasn't changed in over 30 years. 

If the OpenThread feature sets are immutable, the `imply`s can be changed to `select`s (or `depends on`), and the implied entries can be promptless. If they're not, this can either be solved with `imply`s, or with the conventional `depends on` and `default y if OPENTHREAD_FEATURES_MASTER` approaches we use in the Bluetooth stack. See `BT_CENTRAL`, `BT_MESH` and `BT_PERIPHERAL` for instance, they seem like matching concepts.

tejlmand:
> If the OpenThread feature sets are immutable, the `imply`s can be changed to `select`s (or `depends on`), and the implied entries can be promptless. If they're not, this can either be solved with `imply`s, or with the conventional `depends on` and `default y if OPENTHREAD_FEATURES_MASTER` approaches we use in the Bluetooth stack. See `BT_CENTRAL`, `BT_MESH` and `BT_PERIPHERAL` for instance, they seem like matching concepts.

Based on the discussion here in #2707 this seems to be an overlay, just hidden inside Kconfig.
And if this is done, simply to avoid an overlay file, then that is a wrong direction imho, cause it makes it even harder to understand why/where a given setting originates:
https://github.com/nrfconnect/sdk-nrf/pull/2707#discussion_r473876415

See this comment also:
> The overlay files approach was discarded since there was some opposition within the Thread Team to the user having to type the whole -DCONF_FILE="prj.conf overlay-openthread-ftd.conf" every time they want to build a sample.

tejlmand:
Thanks @mbolivar-nordic @edmont @trond-snekvik for good discussion.

PR #348, opened 2020-08-27
Update to Sphinx>=3.2.0 using C domain

utzig:
Cherry-picks from https://github.com/zephyrproject-rtos/zephyr/pull/27759


utzig:
Not need anymore, can use #354 now...

PR #351, opened 2020-09-08
Clock control from tree and updated CLOCK_CONTROL_NRF_FORCE_ALT configuration

nordic-krch:
Clock control in zephyr has been updated to use nrfx_clock. In order to avoid yet another re-implementation of `clock_control_mpsl` it seems that better approach is to add mpsl_clock-nrfx_clock shim which would be used instead of nrfx_clock. This requires that when `CLOCK_CONTROL_NRF_FORCE_ALT` is set clock_control should be included but without enabling nrfx_clock.


tests/bluetooth/ll_settings/prj.conf
```diff
@@ -13,3 +13,7 @@ CONFIG_BT_CTLR_SETTINGS=y
 CONFIG_BT_CTLR_VERSION_SETTINGS=y
 CONFIG_BT_CTLR_COMPANY_ID=0x5F1
 CONFIG_BT_CTLR_SUBVERSION_NUMBER=0xFFFF
+
+# Needed only to prevent the switch-unreachable warning from being generated
+# for the `__fallthrough;` statement in function ll_rx_dequeue() in ull.c.
+CONFIG_BT_OBSERVER=y
```
joerchan, line 19: I'm getting bad object for 741ad17a37807a5396e73c10cf3e5fb425bc87bb in both zephyr and sdk-zephyr

joerchan (changes requested):
Looks OK to me, although no nrfx expert.
Am I understanding it right that this will lead us to having one clock control driver for both controllers?
It would be nice to have the original commit SHAs in the commit message. `git cherry-pick -x` would do that.

drivers/clock_control/clock_control_nrf.c
```diff
@@ -497,38 +466,68 @@ void z_nrf_clock_control_lf_on(enum nrf_lfclk_start_mode start_mode)
...
+		break;
+	}
+	case NRFX_CLOCK_EVT_LFCLK_STARTED:
+		if (IS_ENABLED(
+			CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC_CALIBRATION)) {
+			z_nrf_clock_calibration_lfclk_started();
+		}
+		clkstarted_handle(dev, CLOCK_CONTROL_NRF_TYPE_LFCLK);
+		break;
+	case NRFX_CLOCK_EVT_CAL_DONE:
```
rugeGerritsen, line 495: `NRFX_CLOCK_EVT_CAL_DONE` should not be raised when `CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC_CALIBRATION` is not enabled. We should raise an ASSERT in that case

nordic-krch: added

nordic-krch:
@joerchan this is now based on ongoing upmerge so no cherry-picking. There are 3 commits based on top of upmerge.

nvlsianpu:
rebase?

tejlmand:
> rebase?

This is included in #354

joerchan:
@nordic-krch Since it is included, then please close this if everything is correct.
Is the sdk-nrf PR also included? If so please close that one as well.

nordic-krch:
closing since it is included in upmerge PR.

PR #356, opened 2020-09-17
OpenThread RCP mode integration into Zephyr

markus-becker-tridonic-com:
* Change RCP Kconfig option
* Limit OpenThread initialisation for RCP

Relates to https://github.com/zephyrproject-rtos/zephyr/issues/26383.

### DIFF
--- a/subsys/net/l2/openthread/Kconfig
+++ b/subsys/net/l2/openthread/Kconfig
@@ -187,12 +187,15 @@ menuconfig OPENTHREAD_NCP
 	help
 	  Enable NCP in OpenThread stack.
 
-if OPENTHREAD_NCP
-
-config OPENTHREAD_NCP_RADIO
-	bool "Network Co-Processor as Radio"
+menuconfig OPENTHREAD_RCP
+	bool "Radio Co-Processor"
+	select OPENTHREAD_MANUAL_START
+	select RING_BUFFER
+	select UART_INTERRUPT_DRIVEN
 	help
-	  Enable NCP in OpenThread stack as radio-only.
+	  Enable RCP in OpenThread stack.
+
+if OPENTHREAD_NCP || OPENTHREAD_RCP
 
 config OPENTHREAD_NCP_SPINEL_ON_UART_DEV_NAME
 	string "UART device to use for NCP SPINEL"
@@ -217,7 +220,7 @@ config OPENTHREAD_NCP_VENDOR_HOOK_SOURCE
 	help
 	  Provides path to compile ncp vendor hook file inside NCP component.
 
-endif # OPENTHREAD_NCP
+endif # OPENTHREAD_NCP || OPENTHREAD_RCP
 
 config OPENTHREAD_PLATFORM_INFO
 	string "Platform information for OpenThread"
--- a/subsys/net/l2/openthread/openthread.c
+++ b/subsys/net/l2/openthread/openthread.c
@@ -326,7 +326,7 @@ int openthread_send(struct net_if *iface, struct net_pkt *pkt)
 int openthread_start(struct openthread_context *ot_context)
 {
 	otInstance *ot_instance = ot_context->instance;
-	otError error;
+	otError error = OT_ERROR_NONE;
 
 	openthread_api_mutex_lock(ot_context);
 
@@ -343,36 +343,38 @@ int openthread_start(struct openthread_context *ot_context)
 		otLinkSetPollPeriod(ot_context->instance, OT_POLL_PERIOD);
 	}
 
-	if (otDatasetIsCommissioned(ot_instance)) {
-		/* OpenThread already has dataset stored - skip the
-		 * configuration.
-		 */
-		NET_DBG("OpenThread already commissioned.");
-	} else if (IS_ENABLED(CONFIG_OPENTHREAD_JOINER_AUTOSTART)) {
-		/* No dataset - initiate network join procedure. */
-		NET_DBG("Starting OpenThread join procedure.");
+	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
+		if (otDatasetIsCommissioned(ot_instance)) {
+			/* OpenThread already has dataset stored - skip the
+			 * configuration.
+		 	*/
+			NET_DBG("OpenThread already commissioned.");
+		} else if (IS_ENABLED(CONFIG_OPENTHREAD_JOINER_AUTOSTART)) {
+			/* No dataset - initiate network join procedure. */
+			NET_DBG("Starting OpenThread join procedure.");
 
-		error = otJoinerStart(ot_instance, OT_JOINER_PSKD, NULL,
-				      PACKAGE_NAME, OT_PLATFORM_INFO,
-				      PACKAGE_VERSION, NULL,
-				      &ot_joiner_start_handler, ot_context);
+			error = otJoinerStart(ot_instance, OT_JOINER_PSKD, NULL,
+					      PACKAGE_NAME, OT_PLATFORM_INFO,
+					      PACKAGE_VERSION, NULL,
+					      &ot_joiner_start_handler, ot_context);
 
-		if (error != OT_ERROR_NONE) {
-			NET_ERR("Failed to start joiner [%d]", error);
+			if (error != OT_ERROR_NONE) {
+				NET_ERR("Failed to start joiner [%d]", error);
+			}
+
+			goto exit;
+		} else {
+			/* No dataset - load the default configuration. */
+			NET_DBG("Loading OpenThread default configuration.");
+
+			otExtendedPanId xpanid;
+
+			otThreadSetNetworkName(ot_instance, OT_NETWORK_NAME);
+			otLinkSetChannel(ot_instance, OT_CHANNEL);
+			otLinkSetPanId(ot_instance, OT_PANID);
+			net_bytes_from_str(xpanid.m8, 8, (char *)OT_XPANID);
+			otThreadSetExtendedPanId(ot_instance, &xpanid);
 		}
-
-		goto exit;
-	} else {
-		/* No dataset - load the default configuration. */
-		NET_DBG("Loading OpenThread default configuration.");
-
-		otExtendedPanId xpanid;
-
-		otThreadSetNetworkName(ot_instance, OT_NETWORK_NAME);
-		otLinkSetChannel(ot_instance, OT_CHANNEL);
-		otLinkSetPanId(ot_instance, OT_PANID);
-		net_bytes_from_str(xpanid.m8, 8, (char *)OT_XPANID);
-		otThreadSetExtendedPanId(ot_instance, &xpanid);
 	}
 
 	NET_INFO("OpenThread version: %s", otGetVersionString());
@@ -380,9 +382,11 @@ int openthread_start(struct openthread_context *ot_context)
 		 log_strdup(otThreadGetNetworkName(ot_instance)));
 
 	/* Start the network. */
-	error = otThreadSetEnabled(ot_instance, true);
-	if (error != OT_ERROR_NONE) {
-		NET_ERR("Failed to start the OpenThread network [%d]", error);
+	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
+		error = otThreadSetEnabled(ot_instance, true);
+		if (error != OT_ERROR_NONE) {
+			NET_ERR("Failed to start the OpenThread network [%d]", error);
+		}
 	}
 
 exit:
@@ -397,9 +401,11 @@ int openthread_stop(struct openthread_context *ot_context)
 
 	openthread_api_mutex_lock(ot_context);
 
-	error = otThreadSetEnabled(ot_context->instance, false);
-	if (error == OT_ERROR_INVALID_STATE) {
-		NET_DBG("Openthread interface was not up [%d]", error);
+	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
+		error = otThreadSetEnabled(ot_context->instance, false);
+		if (error == OT_ERROR_INVALID_STATE) {
+			NET_DBG("Openthread interface was not up [%d]", error);
+		}
 	}
 
 	openthread_api_mutex_unlock(ot_context);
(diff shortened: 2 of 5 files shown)


rlubos:
@markus-becker-tridonic-com Please open this PR towards [upstream zephyr](https://github.com/zephyrproject-rtos/zephyr/) instead, we don't want to deviate the code base from upstream here. We'll pull it in on the next upmerge.

markus-becker-tridonic-com:
Please have a quick review before I take it to upstream?

subsys/net/l2/openthread/Kconfig
```diff
@@ -187,12 +187,15 @@ menuconfig OPENTHREAD_NCP
 	help
 	  Enable NCP in OpenThread stack.
 
-if OPENTHREAD_NCP
-
-config OPENTHREAD_NCP_RADIO
-	bool "Network Co-Processor as Radio"
+menuconfig OPENTHREAD_RCP
+	bool "Radio Co-Processor"
+	select OPENTHREAD_MANUAL_START
```
rlubos, line 192: I'd add `depends on !OPENTHREAD_NCP`, since the options are exclusive. And vice versa in the `OPENTHREAD_NCP` options, `depends on !OPENTHREAD_RCP`. This way, if one option is selected, the other one will be disabled.

subsys/net/lib/openthread/platform/platform.c
```diff
@@ -30,7 +30,8 @@ void otSysProcessDrivers(otInstance *aInstance)
 	platformRadioProcess(aInstance);
 	platformAlarmProcess(aInstance);
 
-	if (IS_ENABLED(CONFIG_OPENTHREAD_NCP)) {
+	if (IS_ENABLED(CONFIG_OPENTHREAD_NCP) ||
```
rlubos, line 33: Please fix the alignment
```
	if (IS_ENABLED(CONFIG_OPENTHREAD_NCP) ||
	    IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
```

markus-becker-tridonic-com:
Nevermind, I created https://github.com/zephyrproject-rtos/zephyr/pull/28458.

subsys/net/l2/openthread/openthread.c
```diff
@@ -424,32 +430,35 @@ static int openthread_init(struct net_if *iface)
 
 	__ASSERT(ot_context->instance, "OT instance is NULL");
 
+//	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) { // TODO: check whether required
```
rlubos, line 433: So far we've just disabled SHELL in the project file manually. I think that a nice improvement would be to actually make `OPENTHREAD_SHELL` dependant on `!OPENTHREAD_NCS & !OPENTHREAD_RCP`, it makes little sense to enable shell for these.

subsys/net/l2/openthread/openthread.c
```diff
@@ -424,32 +430,35 @@ static int openthread_init(struct net_if *iface)
...
 	if (IS_ENABLED(CONFIG_OPENTHREAD_SHELL)) {
 		platformShellInit(ot_context->instance);
 	}
 
-	if (IS_ENABLED(CONFIG_OPENTHREAD_NCP)) {
+	if (IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
```
rlubos, line 438: It should be `IS_ENABLED(CONFIG_OPENTHREAD_NCP) || IS_ENABLED(CONFIG_OPENTHREAD_RCP)`.

subsys/net/l2/openthread/openthread.c
```diff
@@ -424,32 +430,35 @@ static int openthread_init(struct net_if *iface)
...
 
-	if (!IS_ENABLED(CONFIG_OPENTHREAD_RAW)) {
-		otIp6SetEnabled(ot_context->instance, true);
-	}
+	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
+		if (!IS_ENABLED(CONFIG_OPENTHREAD_RAW)) {
+			otIp6SetEnabled(ot_context->instance, true);
+		}
 
-	if (!IS_ENABLED(CONFIG_OPENTHREAD_NCP)) {
```
rlubos, line 439: Same here, I'd rather make it just `!IS_ENABLED(CONFIG_OPENTHREAD_NCP) && !IS_ENABLED(CONFIG_OPENTHREAD_RCP)`, w/o further changes.

subsys/net/l2/openthread/openthread.c
```diff
@@ -343,46 +343,50 @@ int openthread_start(struct openthread_context *ot_context)
...
-
-		error = otJoinerStart(ot_instance, OT_JOINER_PSKD, NULL,
-				      PACKAGE_NAME, OT_PLATFORM_INFO,
-				      PACKAGE_VERSION, NULL,
-				      &ot_joiner_start_handler, ot_context);
-
-		if (error != OT_ERROR_NONE) {
-			NET_ERR("Failed to start joiner [%d]", error);
-		}
+	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
```
rlubos, line 346: I'd rather exit early on `CONFIG_OPENTHREAD_RCP` instead of adding extra indentation level.

That also makes me wonder, how did it work with NCP, as all these configurations are supposed to be done by the Border Router and not the NCP itself. I think that a good option would be to add the following at the beginning of the function:
```
if (IS_ENABLED(CONFIG_OPENTHREAD_NCP) || IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
        return 0;
}
```

Going even further, I wonder if it actually makes sense to create a Zephyr network interface for the NCP/RCP application at all (since it's not supposed to use it directly), but this would need more investigation so let's leave it for now.

subsys/net/l2/openthread/openthread.c
```diff
@@ -397,9 +401,11 @@ int openthread_stop(struct openthread_context *ot_context)
 
 	openthread_api_mutex_lock(ot_context);
 
-	error = otThreadSetEnabled(ot_context->instance, false);
-	if (error == OT_ERROR_INVALID_STATE) {
-		NET_DBG("Openthread interface was not up [%d]", error);
+	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
```
rlubos, line 404: Same here, I'd rather exit early with the same condition as in `openthread_start()`.

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -384,7 +384,9 @@ void platformRadioProcess(otInstance *aInstance)
 		while ((tx_pkt = (struct net_pkt *)k_fifo_get(&tx_pkt_fifo,
 							      K_NO_WAIT))
 		      != NULL) {
-			openthread_handle_frame_to_send(aInstance, tx_pkt);
+			if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
```
rlubos, line 387: I wonder if it's actually needed? The application is not supposed to send anything on its own in the case of NCP/RCP?

rlubos:
@markus-becker-tridonic-com I've already put a review here, please consider it in the upstream PR.

markus-becker-tridonic-com:
> @markus-becker-tridonic-com I've already put a review here, please consider it in the upstream PR.

will do.

subsys/net/l2/openthread/openthread.c
```diff
@@ -424,32 +430,35 @@ static int openthread_init(struct net_if *iface)
...
 	__ASSERT(ot_context->instance, "OT instance is NULL");
 
+//	if (!IS_ENABLED(CONFIG_OPENTHREAD_RCP)) { // TODO: check whether required
 	if (IS_ENABLED(CONFIG_OPENTHREAD_SHELL)) {
 		platformShellInit(ot_context->instance);
 	}
 
-	if (IS_ENABLED(CONFIG_OPENTHREAD_NCP)) {
+	if (IS_ENABLED(CONFIG_OPENTHREAD_RCP)) {
 		otNcpInit(ot_context->instance);
```
kkasperczyk-no, line 439: Same remark as concerning Kconfig options' names. Please find some common naming convention to avoid misunderstanding.

kkasperczyk-no (commented):
I think that @rlubos pointed all important things to change, but please also consider my remarks about naming. In my opinion it is not the best idea to select only RCP architecture option and use other methods and options for NCP, as it's confusing which architecture is actually used. Duplicating everything might not be the best idea, but I think that some common names for common settings should be enough.

subsys/net/l2/openthread/Kconfig
```diff
@@ -187,12 +187,15 @@ menuconfig OPENTHREAD_NCP
...
+	select RING_BUFFER
+	select UART_INTERRUPT_DRIVEN
 	help
-	  Enable NCP in OpenThread stack as radio-only.
+	  Enable RCP in OpenThread stack.
+
+if OPENTHREAD_NCP || OPENTHREAD_RCP
```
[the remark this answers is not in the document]
markus-becker-tridonic-com: how about the following?

```
config OPENTHREAD_SHELL
	bool "Enable OpenThread shell"
	depends on !OPENTHREAD_COPROCESSOR
	select SHELL
	default y
```

```
menuconfig OPENTHREAD_COPROCESSOR
	bool "OpenThread Co-Processor"
	select OPENTHREAD_MANUAL_START
	select RING_BUFFER
	select UART_INTERRUPT_DRIVEN
	help
	  Enable Co-Processor in OpenThread stack.

choice
	prompt "OpenThread Co-Processor type"
	help
	  This option selects Thread network co-processor type

config OPENTHREAD_COPROCESSOR_NCP
	bool "NCP - Network Co-Processor"
config OPENTHREAD_COPROCESSOR_RCP
	bool "RCP - Radio Co-Processor"
endchoice

if OPENTHREAD_COPROCESSOR

config OPENTHREAD_COPROCESSOR_SPINEL_ON_UART_DEV_NAME
	string "UART device to use for Co-Processor SPINEL"
	default "UART_0"
	help
	  UART device to use for Co-Processor SPINEL.

config OPENTHREAD_COPROCESSOR_UART_RING_BUFFER_SIZE
	int "Set Co-Processor UART ring buffer size"
	default 4096
	help
	  TX buffer size for the OpenThread Co-Processor UART.

config OPENTHREAD_COPROCESSOR_SPINEL_ON_UART_ACM
	bool "Run SPINEL over USB-ACM"
	depends on USB_CDC_ACM
	help
	  Is the SPINEL device a USB-CDC-ACM device.

config OPENTHREAD_COPROCESSOR_VENDOR_HOOK_SOURCE
	string "Path to vendor hook source file"
	help
	  Provides path to compile vendor hook file.

endif # OPENTHREAD_COPROCESSOR
```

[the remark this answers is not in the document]
kkasperczyk-no: This one looks fine for me.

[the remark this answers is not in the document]
rlubos: I agree, looks good.

subsys/net/l2/openthread/openthread.c
markus-becker-tridonic-com: If I exit early, I get linker errors : `undefined reference to `otThreadSetEnabled'. AFACT `IS_ENABLED(false)` is actually removed at compile-time.

subsys/net/l2/openthread/openthread.c
markus-becker-tridonic-com: Same as above.

subsys/net/l2/openthread/openthread.c
markus-becker-tridonic-com: Will do that on the upstream PR.

subsys/net/l2/openthread/openthread.c
markus-becker-tridonic-com: Yes, correct.

subsys/net/lib/openthread/platform/platform.c
markus-becker-tridonic-com: Will be replaced with
```
	if (IS_ENABLED(CONFIG_OPENTHREAD_COPROCESSOR)) {
```

subsys/net/lib/openthread/platform/radio.c
markus-becker-tridonic-com: Will test.

subsys/net/l2/openthread/openthread.c
markus-becker-tridonic-com: As discussed will go for `CONFIG_OPENTHREAD_COPROCESSOR`.

subsys/net/l2/openthread/Kconfig
markus-becker-tridonic-com: Will go for the `choice` OPENTHREAD_COPROCESSOR_NCP / OPENTHREAD_COPROCESSOR_RCP.

subsys/net/l2/openthread/Kconfig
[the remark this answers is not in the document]
markus-becker-tridonic-com: OK, will apply this.

subsys/net/lib/openthread/platform/radio.c
markus-becker-tridonic-com: Same problem here, this will pull in `openthread_handle_frame_to_send()` which has references to `otIp6NewMessage()` etc. which are not available. So either the call to `openthread_handle_frame_to_send()` is `if IS_ENABLED` checked or the interior of the function has that check.

subsys/net/l2/openthread/openthread.c
kkasperczyk-no: Ok, but in this place my remark applies to the otNcpInit() method name. I would like to stay consistent and change it to otCoProcessorInit(), but as I guess it would also require contributing to the OpenThread. What do you think about this @rlubos ?

rlubos: Given that `otNcpInit()` is part of the OT API, I don't think there's anything we can do about it in this PR (but I don't find it to be a block either). Of course, it would be nice to have a more generic name, but I'd leave it to upstream community to sort this out (they have the same issue in the `ncp` app, which could be configured either as NCP or RCP).

kkasperczyk-no: Yeah, looking further, there are a lot of places in OpenThread, where NCP name is used and it also applies to RCP, but I think you are right. It's not good idea to change everything at once and for now it is not a problem at all.

markus-becker-tridonic-com:
Rarely I do get build errors:

```
In file included from /home/markus/src/zephyrproject/zephyr/lib/libc/minimal/include/errno.h:20,
                 from /home/markus/src/zephyrproject/modules/lib/openthread/src/lib/spinel/spinel.c:48:
/home/markus/src/zephyrproject/zephyr/include/sys/errno_private.h:34:10: fatal error: syscalls/errno_private.h: No such file or directory
   34 | #include <syscalls/errno_private.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Seems like a concurrency issue in the build system, where the generated file `./build/zephyr/include/generated/syscalls/errno_private.h` is not created by the time spinel.c is being built.

Any chance that you could check with someone who know the build system very well?

markus-becker-tridonic-com:
Pushed the updated files into the upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/28458, please continue the review there.

rlubos:
I've never seen build failures like this before, I think it would be best to open a GitHub issue describing the problem to get more attention.

rlubos:
Discussion moved to zephyrproject-rtos/zephyr#28458

markus-becker-tridonic-com:
Created https://github.com/zephyrproject-rtos/zephyr/issues/28465 about the build issue.

PR #357, opened 2020-09-21
Add zigbee kconfig option for zigbee l2

maciekfabia:
PR adds necessary changes that allow to move zigbee to 802.15.4 radio api instead of using radio driver directly.
It consists of two commits:

- commit adding more error codes to tx function of 802.15.4 nrf5 driver. They provide more details on transmission fail cause that can be passed to ZBOSS zigbee stack.
It also has a PR in upstream: https://github.com/zephyrproject-rtos/zephyr/pull/28546

- commit adding Kconfig option that enables Zigbee L2 layer. It's not supposed to go upstream as zigbee is located in NCS repos.


rlubos (approved):
LGTM

MarekPorwisz (commented):
This change makes me wonder ... Zigbee is not going upstram as I understand, and you're making changes in files that come from upstream. Won't this make upmerges painfull? Was this solution accepted?
When it comes to changes affecting OpenThread - we are currently verifying if the result is 0 (at least that's what I've found) so there should be no issue there.

rlubos:
> Won't this make upmerges painfull?

Yes, this might cause some extra conflicts during an upmerge (but the changes here are really minimal so it should be easy to resolve). But I don't see a way around, the radio driver needs to be aware of the new L2, and it makes little sense to upstream these changes, given we won't provide the actual Zigbee L2 in there.

MarekPorwisz (approved):

maciekfabia:
> When it comes to changes affecting OpenThread - we are currently verifying if the result is 0 (at least that's what I've found) so there should be no issue there.

It's also what I've found in Thread sources, it is checked if any error occured without comparing to specific errno codes

subsys/net/l2/Kconfig
```diff
@@ -76,6 +76,11 @@ source "subsys/net/l2/openthread/Kconfig"
 
 source "subsys/net/l2/canbus/Kconfig"
 
+config NET_L2_ZIGBEE
+	bool "Zigbee L2"
+	depends on NETWORKING
+	imply NET_PKT_TIMESTAMP
```
MarekPorwisz, line 82: since zigbee needs timestamps this should be select or depends on

maciekfabia: changed to `SELECT`

maciekfabia:
@rlubos I think it's ready for merge.
Merging this PR will unblock https://github.com/nrfconnect/sdk-nrf/pull/2948

PR #359, opened 2020-10-01
Bluetooth: controller: Fix qualification issues

joerchan:
Fix regression in the disconnect complete event encoding where the
disconnect reason was accidentally deleted.

Regression introduced during conflict resolution of uint8_t and u8_t
differences in cherry-pick of:
736b79c

Change the LL version back to 5.1, to be able to re-use existing
Bluetooth conformance qualification.


cvinayak (approved):

joerchan:
Manifest update: https://github.com/nrfconnect/sdk-nrf/pull/3009

PR #365, opened 2020-10-07
samples: bluetooth: uart_hci: Enable low power uart in nrf9160dk_nrf52840 configuration

nordic-krch:
Enabled low power uart to be used for bluetooth communication in hci_uart.
Added nrf9160dk_nrf9160 configuration for `samples/bluetooth/peripheral_hr`


tejlmand (changes requested):
Those being `[noup]` commits in this PR means they might need to be reverted and reapplied on upmerges, so would be better for upmerge handling if they are squashed.

joerchan:
> Those being `[noup]` commits in this PR means they might need to be reverted and reapplied on upmerges, so would be better for upmerge handling if they are squashed.

Which brings up the question, why should it be noup? It would reduce our maintaince cost if it wasn't noup.

nordic-krch:
> Which brings up the question, why should it be noup

low power uart exists only in ncs so configuration in upstream does not make sense.

anangl:
> > Which brings up the question, why should it be noup
> 
> low power uart exists only in ncs so configuration in upstream does not make sense.

That's true. But I think it would be a bit weird to have those Zephyr samples modified in our fork to use something that is available only in the sdk-nrf repository.
What about adding instead those samples to the sdk-nrf repository (with slightly modified names, e.g. hci_lpuart and peripheral_hr_lpuart) and basically provide in them only those modified configurations and point (in CMakeLists.txt) to the source code from the original samples?

nordic-krch:
@anangl that's an option, we can even skip peripheral_hr as this is not essential and lte_ble_gateway will already contain example how to setup low power uart on host side. @joerchan, @tejlmand  what do you think?

nordic-krch:
Added something like that: https://github.com/nrfconnect/sdk-nrf/pull/3064/commits/6f60c26f3de6ef8e4c4527c474afcc0fe5c60b60
However, I don't like that `prj.conf` and `debug.conf` are direct copies from zephyr. Any ideas how to not copy them?

nordic-krch:
closing this one as i will proceed with sample in ncs

PR #368, opened 2020-10-12
drivers: ieee802154: nrf5: Add support for tx in the future

maciekfabia:
Upstream PR #29123

Add support for transmission modes that send a packet
at a specific time in the future.

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -52,6 +52,7 @@ static struct nrf5_802154_data nrf5_data;
 #define ACK_REQUEST_BIT (1 << 5)
 #define FRAME_PENDING_BYTE 1
 #define FRAME_PENDING_BIT (1 << 4)
+#define TXTIME_OFFSET_US  (5 * USEC_PER_MSEC)
 
 /* Convenience defines for RADIO */
 #define NRF5_802154_DATA(dev) \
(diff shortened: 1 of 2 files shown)


wbober (approved):
LGTM

tomchy (approved):

rlubos (approved):

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -405,6 +423,7 @@ static int nrf5_tx(const struct device *dev,
 		return -EIO;
 	}
 
+	k_sem_take(&nrf5_radio->tx_started, K_FOREVER);
```
hubertmis, line 426: I'm not sure about blocking on a semaphore here. It is possible that there is a requested transmission in long future (like 30 seconds). It would block possibility to transit other frames meantime. But I don't have any better ideas right. I think we can leave it as it is, possibly refactor later.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -634,6 +654,13 @@ void nrf_802154_receive_failed(nrf_802154_rx_error_t error)
 	nrf5_data.last_frame_ack_fpb = false;
 }
 
+void nrf_802154_tx_started(const uint8_t *frame)
```
hubertmis, line 657: This function will be removed when we add support for nRF53. Bus we can move `k_sem_give()` to `transmitted_raw()` and to `transmit_failed()`.

tomchy: In such case the semaphore is pointless - it will duplicate the `nrf5_radio->tx_wait` semaphore.

Generally - good point, it would be great to check if there is a specific requirement on the `nrf5_tx_started(dev, pkt, frag);` callback in other protocols. This callback is the only reason for adding the new semaphore.

drivers/ieee802154/ieee802154_nrf5.c
tomchy: It would also break the rule that the `tx` operation is blocking. Maybe the `net_pkt_set_timestamp` API should be changed to something else (e.g. `radio->tx_scheduled(..)`)?

PR #369, opened 2020-10-13
drivers: ieee802154: nrf5: Remove nrf_802154_tx_started callback

maciekfabia:
Remove the implementation of nrf_802154_tx_started
radio driver callback. Remove the semaphore dedicated
to this callback.

PR updating the manifest:
https://github.com/nrfconnect/sdk-nrf/pull/3108

Jira: KRKNWK-7789

PR #373, opened 2020-10-15
[nrf fromlist] net: openthread: fix unwanted 802.15.4 radio up

edmont:
This commit moves IPv6 initialization from OT init to OT start to avoid unwanted bringing 802.15.4 radio up.

Previously, even when OT manual start was enabled, the radio would be receiving frames resulting in unnecessary power consumption and causing issues for instance when the device just wants to use Bluetooth for provisioning before moving to Thread.

Signed-off-by: Eduardo Montoya <eduardo.montoya@nordicsemi.no>

KRKNWK-7786

Upstream: https://github.com/zephyrproject-rtos/zephyr/pull/29262

Pending on:
- [x] https://github.com/nrfconnect/sdk-zephyr/pull/369

### DIFF
--- a/subsys/net/l2/openthread/openthread.c
+++ b/subsys/net/l2/openthread/openthread.c
@@ -330,6 +330,8 @@ int openthread_start(struct openthread_context *ot_context)
 
 	openthread_api_mutex_lock(ot_context);
 
+	otIp6SetEnabled(ot_context->instance, true);
+
 	/* Sleepy End Device specific configuration. */
 	if (IS_ENABLED(CONFIG_OPENTHREAD_MTD_SED)) {
 		otLinkModeConfig ot_mode = otThreadGetLinkMode(ot_instance);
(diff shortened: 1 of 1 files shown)


MarekPorwisz (approved):
After a quick look into it seems that otIp6Set... functions can safely be called before SetEnabled so the change should be fine.

rlubos (approved):

edmont:
> After a quick look into it seems that otIp6Set... functions can safely be called before SetEnabled so the change should be fine.

Yes, previously I had issues because I was testing without the required radio driver fix.

LuDuda (approved):

shantha-14 (approved):

carlescufi:
I'd much rather see a `[nrf fromlist]` here. Are you going to post a PR to upstream zephyr?

edmont:
@carlescufi can we merge this?

shantha-14:
Covered by #378

rlubos:
Already in

PR #375, opened 2020-10-15
[nrf fromlist] net: openthread: Increase the number of allowed children

MarekPorwisz:
Another team reported that current default values for number of allowed
IP addresses per child (4) and and max number of children (10) are too
small for some customers.
Increased the values allowed configuring child count.

Jira
https://projecttools.nordicsemi.no/jira/browse/KRKNWK-7796

manifest: 
- https://github.com/nrfconnect/sdk-nrf/pull/3139

Signed-off-by: Marek Porwisz <marek.porwisz@nordicsemi.no>

### DIFF
--- a/subsys/net/l2/openthread/Kconfig.thread
+++ b/subsys/net/l2/openthread/Kconfig.thread
@@ -52,3 +52,13 @@ config OPENTHREAD_POLL_PERIOD
 	int "Poll period for sleepy end devices [ms]"
 	default 236000
 	depends on OPENTHREAD_MTD_SED
+
+config OPENTHREAD_MAX_CHILDREN
+	int "The maximum number of children"
+	range 10 512
+	default 32
+
+config OPENTHREAD_MAX_IP_ADDR_PER_CHILD
+	int "The maximum number of IPv6 address registrations per child."
+	range 4 255
+	default 6
--- a/subsys/net/lib/openthread/platform/openthread-core-zephyr-config.h
+++ b/subsys/net/lib/openthread/platform/openthread-core-zephyr-config.h
@@ -131,6 +131,27 @@
 #define RADIO_CONFIG_SRC_MATCH_SHORT_ENTRY_NUM 0
 
 /**
+ * @def OPENTHREAD_CONFIG_MLE_MAX_CHILDREN
+ *
+ * The maximum number of children.
+ *
+ */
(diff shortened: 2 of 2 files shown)


rlubos (commented):
Please prefix the commit title with `[nrf toup] net: openthread: `.

Is this change targeting 1.4? If not, I"d suggest opening an upstream PR first.

MarekPorwisz:
NOTE: This increases ram usage by 4KB. We are still fitting into 833 ram, but there is not much space left. I've created Kconfig switch for the customers that require more RAM for theirs app.

edmont:
Minimum required addresses per child in Thread is 4. It would be beneficial to have a config switch for that as well. Also there should be a check for both values not to go below the minimum requirements.

edmont (approved):
LGTM, thanks.

subsys/net/l2/openthread/Kconfig.thread
```diff
@@ -52,3 +52,13 @@ config OPENTHREAD_POLL_PERIOD
 	int "Poll period for sleepy end devices [ms]"
 	default 236000
 	depends on OPENTHREAD_MTD_SED
+
+config OPENTHREAD_MAX_CHILDREN
+	int "The maximum number of children"
+	range 10 512
```
rlubos, line 58: Not sure about adding a lower limit on the maximum number of children supported (is it specified in spec that router must support at least 10 children?).

rlubos (approved):

subsys/net/l2/openthread/Kconfig.thread
MarekPorwisz: I think 10 is minimum to pass the certification.

edmont: The requirement comes from the Thread Conformance document.

shantha-14 (approved):

carlescufi:
I'd much rather see a `[nrf fromlist]` here. Are you going to post a PR to upstream zephyr?

MarekPorwisz:
> I'd much rather see a `[nrf fromlist]` here. Are you going to post a PR to upstream zephyr?

Yes, I will, but that won't get into upstream on time, and I'm involved in some other tasks as well, so I'll do that probably next week or the week after.

MarekPorwisz:
I've created upstream pr https://github.com/zephyrproject-rtos/zephyr/pull/29264

PR #379, opened 2020-10-20
[nrf temphack] soc: use PM when configuring MPU regions

hakonfam:
Temporary solution for using partition manager defines when configuring
the MPU regions.

This fixes assert seen when configuring dynamic regions.

### DIFF
--- a/soc/arm/common/cortex_m/arm_mpu_regions.c
+++ b/soc/arm/common/cortex_m/arm_mpu_regions.c
@@ -8,6 +8,9 @@
 #include <arch/arm/aarch32/cortex_m/mpu/arm_mpu.h>
 
 #include "arm_mpu_mem_cfg.h"
+#if USE_PARTITION_MANAGER
+#include <pm_config.h>
+#endif
 
 static const struct arm_mpu_region mpu_regions[] = {
 	/* Region 0 */
@@ -21,6 +24,14 @@ static const struct arm_mpu_region mpu_regions[] = {
 #endif
 	/* Region 1 */
 	MPU_REGION_ENTRY("SRAM_0",
+#if USE_PARTITION_MANAGER
+			 PM_SRAM_ADDRESS,
+#if defined(CONFIG_ARMV8_M_BASELINE) || defined(CONFIG_ARMV8_M_MAINLINE)
+			 REGION_RAM_ATTR(PM_SRAM_ADDRESS, PM_SRAM_SIZE)),
+#else
+			 REGION_RAM_ATTR(REGION_SRAM_SIZE)),
+#endif
+#else
 			 CONFIG_SRAM_BASE_ADDRESS,
 #if defined(CONFIG_ARMV8_M_BASELINE) || defined(CONFIG_ARMV8_M_MAINLINE)
 			 REGION_RAM_ATTR(CONFIG_SRAM_BASE_ADDRESS, \
@@ -28,6 +39,9 @@ static const struct arm_mpu_region mpu_regions[] = {
 #else
 			 REGION_RAM_ATTR(REGION_SRAM_SIZE)),
 #endif
+
+#endif /* USE_PARTITION_MANAGER */
+
 };
 
 const struct arm_mpu_config mpu_config = {


tejlmand (changes requested):
Commit title should be:
```
[nrf temphack] soc: use PM when configuring MPU regions
```

and is this really a `temphack` ?
What will be the cprrect solution.

ioannisg (commented):
This looks really "hacky".
@hakonfam is there a link to the observed issue, this PR attempts to solve?

soc/arm/common/cortex_m/arm_mpu_regions.c
```diff
@@ -21,13 +24,20 @@ static const struct arm_mpu_region mpu_regions[] = {
 #endif
 	/* Region 1 */
 	MPU_REGION_ENTRY("SRAM_0",
+#ifdef USE_PARTITION_MANAGER
```
ioannisg, line 27: A sort of generic solution would be to use linker symbols here, since we, effectively, want to configure the SRAM used by Zephyr image. This could work upstream as well. Please investigate.

andvib (approved):
Tested, app boots successfully for _both_ nrf5340pdk_nrf5340_cpuapp and nrf5340pdk_nrf5340_cpuappns

soc/arm/common/cortex_m/arm_mpu_regions.c
hakonfam: AFAICS we cannot use linker symbols as value is required by the preprocessor.

hakonfam:
> Commit title should be:
> 
> ```
> [nrf temphack] soc: use PM when configuring MPU regions
> ```
> 
> and is this really a `temphack` ?
> What will be the cprrect solution.

I never remember if its with or without the hyphen, fixed now.

This feels like a hack to me. Firstly since its a diff from upstream, secondly since it should be possible to solve this in a generic way where we can "hook in" without having downstream patches. Thirdly since the use of these kconfig symbols for SRAM values other places would still be incorrect.

I was thinking in the lines of adding an overlay kconfig file setting the Kconfig symbols for SRAM_START and SRAM_SIZE if PM is used for RAM partitioning. This way we don't need an upstream patch, and the change will be traceable by the user.

hakonfam:
> This looks really "hacky".
> @hakonfam is there a link to the observed issue, this PR attempts to solve?

I added link to the NCSDK ticket describing the issue in the commit message. NCSDK-6718

soc/arm/common/cortex_m/arm_mpu_regions.c
oyvindronningstad: Maybe you could change the usage of CONFIG_SRAM_* to use linker values instead, or do they need to be macros? That way you don't need the new ifdef.

hakonfam:
@ioannisg @tejlmand an alternative approach would be to include a headerfile which does #undef of `CONFIG_SRAM_BASE_ADDRESS` and `CONFIG_SRAM_SIZE` before it redefines them based on the values of `PM_SRAM_SIZE / 1024` and `PM_SRAM_ADDRESS`. What I really dislike about this solution is that it is not visible for the user, and I actually prefer the more hacky, but visible way it is currently done in this PR.

To clean up the hack I could do the #undef pattern inside this file as well, but then AFAIK values for CONFIG_SRAM_* would be changed for the rest of the preprocessor run (not 100% sure about this though).

tejlmand (approved):
> I was thinking in the lines of adding an overlay kconfig file setting the Kconfig symbols for SRAM_START and SRAM_SIZE if PM is used for RAM partitioning. This way we don't need an upstream patch, and the change will be traceable by the user.

Agree, then this is a temp hack.
I think this is acceptable for the coming release, and I like that it is very visible to the reader what is happening. (although the fix itself is not so nice).
Especially as this is something where we want a better solution.

hakonfam:
For future reference:

I have verified that there are no other uses of `CONFIG_SRAM_BASE_ADDRESS` which could be affected by this issue.

There is one use in `nrf_qspi_nor.c`, but that should be changed to use the symbol for the base of RAM from DT instead.

oyvindronningstad (approved):

ioannisg (changes requested):
I think we need to solve this problem in a nice way:

`config SRAM_BASE_ADDRESS` and `config SRAM_SIZE` determine the **image SRAM** start address and size; I think this is very problematic if this is not the case. Application users may rely on SRAM_SIZE and SRAM_BASE_ADDRESS reflecting the image ram.

The partition manager can overwrite the SRAM_BASE_ADDRESS and SRAM_SIZE, so their values are consistent with the build.

In this case you can avoiding editing this core zephyr ARCH source file.

> I have verified that there are no other uses of CONFIG_SRAM_BASE_ADDRESS which could be affected by this issue.

This is a very risky assumption, considering how generic these Kconfig symbols are. They may well be used in upstream samples, which we are supposed to build successfully in NCS.

hakonfam:
> I think we need to solve this problem in a nice way:
> 
> `config SRAM_BASE_ADDRESS` and `config SRAM_SIZE` determine the **image SRAM** start address and size; I think this is very problematic if this is not the case. Application users may rely on SRAM_SIZE and SRAM_BASE_ADDRESS reflecting the image ram.

Yes they may, but I don't think this is something which is done a lot. When do you need to operate on the borders of RAM? Its not even the physical RAM, which could be used for checking if a pointer is in RAM or FLASH.. The SRAM_*  does also not represent the RAM the application has available, since that is found in linker symbols.

IMO it is OK for us to change this value here, but of course not optimal.

> The partition manager can overwrite the SRAM_BASE_ADDRESS and SRAM_SIZE, so their values are consistent with the build.
> 
> In this case you can avoiding editing this core zephyr ARCH source file.

I agree that the best solution would be to override the kconfig values for SRAM_*, but this ends in a dependency loop; PM depends on Kconfig, and should if PM sets kconfig values, then Kconfig needs to re-evaluted and so on.

> > I have verified that there are no other uses of CONFIG_SRAM_BASE_ADDRESS which could be affected by this issue.
> 
> This is a very risky assumption, considering how generic these Kconfig symbols are. They may well be used in upstream samples, which we are supposed to build successfully in NCS.

They are not used in upstream samples, this is what I have verified.

soc/arm/common/cortex_m/arm_mpu_regions.c
```diff
@@ -21,13 +24,24 @@ static const struct arm_mpu_region mpu_regions[] = {
 #endif
 	/* Region 1 */
 	MPU_REGION_ENTRY("SRAM_0",
+#if USE_PARTITION_MANAGER
```
tejlmand, line 27: If doing this, then please ensure users are unable to set the `SRAM_SIZE` and `SRAM_BASE_ADDRESS` when `USE_PARTITION_MANAGER=y`.
https://github.com/nrfconnect/sdk-zephyr/blob/41ce8e165a9af4171ef52a73aede89da6ce60920/arch/Kconfig#L130-L144
by adding:
```
  depends on !USE_PARTITION_MANAGER
```

tejlmand (changes requested):

soc/arm/common/cortex_m/arm_mpu_regions.c
hakonfam: Will try it!

hakonfam: Discussed offline, posting the conclusion here for others to see: It won't work since `USE_PARTITION_MANAGER` is not available as a kconfig symbol.

tejlmand (approved):
Accepted for 1.4.0.

mbolivar-nordic (approved):
I agree with the discussion that it's an acceptable temphack.

I continue to worry and fret over the fork between NCS and Zephyr with regards to partition handling being confusing and hoping that one of these years we can find a harmonious solution that works without extra patches. But that's not for today.

hakonfam:
@shantha-14 this needs your approval

carlescufi:
Replaced by https://github.com/nrfconnect/sdk-zephyr/pull/384

PR #380, opened 2020-10-20
Bluetooth: Mesh: Fix uninitialized Model Extension fields

michal-narajowski:
These uninitialized fields were causing undefined behavior in
bt_mesh_model_tree_walk().

Signed-off-by: Michał Narajowski <michal.narajowski@codecoup.pl>


michal-narajowski:
This PR does not fix the underlying issue.

PR #382, opened 2020-10-21
[nrf fromlist] Bluetooth: shell: Fix shell build with out-of-tree ...

joerchan:
controller

Fix shell build errors when building with combined host and controller
but the selected controller is not the in-tree zephyr controller, i.e
CONFIG_BT_CTLR=y, CONFIG_BT_LL_SW_SPLIT=n and cmd_scanx and others
are not defined.

Manifest PR: https://github.com/nrfconnect/sdk-nrf/pull/3201

### DIFF
--- a/subsys/bluetooth/shell/bt.c
+++ b/subsys/bluetooth/shell/bt.c
@@ -2863,6 +2863,8 @@ SHELL_STATIC_SUBCMD_SET_CREATE(bt_cmds,
 #if defined(CONFIG_BT_HCI_MESH_EXT)
 	SHELL_CMD(mesh_adv, NULL, "<on, off>", cmd_mesh_adv),
 #endif /* CONFIG_BT_HCI_MESH_EXT */
+
+#if defined(CONFIG_BT_LL_SW_SPLIT)
 #if defined(CONFIG_BT_CTLR_ADV_EXT)
 #if defined(CONFIG_BT_BROADCASTER)
 	SHELL_CMD_ARG(advx, NULL,
@@ -2881,6 +2883,8 @@ SHELL_STATIC_SUBCMD_SET_CREATE(bt_cmds,
 		      4, 0),
 	SHELL_CMD_ARG(test_end, NULL, HELP_NONE, cmd_test_end, 1, 0),
 #endif /* CONFIG_BT_CTLR_DTM */
+#endif /* CONFIG_BT_LL_SW_SPLIT */
+
 	SHELL_SUBCMD_SET_END
 );
 


tejlmand:
Manifest PR ?

joerchan:
> Manifest PR ?

Sorry, updated

subsys/bluetooth/shell/bt.c
```diff
@@ -2881,6 +2882,7 @@ SHELL_STATIC_SUBCMD_SET_CREATE(bt_cmds,
 		      4, 0),
 	SHELL_CMD_ARG(test_end, NULL, HELP_NONE, cmd_test_end, 1, 0),
 #endif /* CONFIG_BT_CTLR_DTM */
+#endif /* CONFIG_BT_LL_SW_SPLIT */
```
tejlmand, line 2885: This doesn't seem correct, looking at the whole code snippet:
```
#if defined(CONFIG_BT_LL_SW_SPLIT)
#if defined(CONFIG_BT_BROADCASTER)
	SHELL_CMD_ARG(advx, NULL,
		      "<on hdcd ldcd off> [coded] [anon] [txp] [ad]",
		      cmd_advx, 2, 4),
#endif /* CONFIG_BT_BROADCASTER */
#if defined(CONFIG_BT_OBSERVER)
	SHELL_CMD_ARG(scanx, NULL, "<on passive off> [coded]", cmd_scanx,
		      2, 1),
#endif /* CONFIG_BT_OBSERVER */
#endif /* CONFIG_BT_CTLR_ADV_EXT */    <------ This seems to end defined(CONFIG_BT_LL_SW_SPLIT)
#if defined(CONFIG_BT_CTLR_DTM)
	SHELL_CMD_ARG(test_tx, NULL, "<chan> <len> <type> <phy>", cmd_test_tx,
		      5, 0),
	SHELL_CMD_ARG(test_rx, NULL, "<chan> <phy> <mod_idx>", cmd_test_rx,
		      4, 0),
	SHELL_CMD_ARG(test_end, NULL, HELP_NONE, cmd_test_end, 1, 0),
#endif /* CONFIG_BT_CTLR_DTM */
#endif /* CONFIG_BT_LL_SW_SPLIT */  <------ Wonder which if this ends ?
defined(CONFIG_BT_LL_SW_SPLIT)
	SHELL_SUBCMD_SET_END
```

tejlmand (changes requested):
`#endif /* CONFIG_BT_LL_SW_SPLIT */` seems to be placed wrongly, or another `#endif` should have been moved.

subsys/bluetooth/shell/bt.c
tejlmand: is there an upstream PR ?

joerchan: Thanks @tejlmand.
Submittet follow up PR to zephyr: https://github.com/zephyrproject-rtos/zephyr/pull/29447
The first one already got merged

tejlmand: So I guess the followup commit should be included in this PR, right ?

tejlmand: I guess if upstream is already merged, then title should be `[nrf fromtree]`.

joerchan: @tejlmand I squashed them together. The follow up one is not merged so I have it as fromlist still.

tejlmand: ok. it will be reverted on next upmerge anyway, but I generally think `fromtree` and `fromlist` should be as close to their upstream counterparts as possible.

tejlmand (approved):

subsys/bluetooth/shell/bt.c
carlescufi: We should probably have both commits, but I guess it's not important

carlescufi (approved):

subsys/bluetooth/shell/bt.c
joerchan: Both PRs are listed in the commit message. I though it would be less clutter with less commits, only one that needs to be reverted now.

shantha-14 (approved):

PR #400, opened 2020-12-01
[nrf fromlist] doc: script: Default to empty string in DT docset env var

carlescufi:
Windows does not default to an empty string when an env var is not
declared, like Linux does. Instead, explicitly default to one in order
for the .strip() call to work properly.


utzig (approved):

carlescufi:
Closing in favor of https://github.com/nrfconnect/sdk-nrf/pull/3443

PR #403, opened 2020-12-03
[nrf fromlist] settings: fix possible unaligned write while a line coping

nvlsianpu:
It was possible that settings_line_entry_copy() did unaligned
flash write.
This patch introduce respecting the flash write-block-size.

upstream patch:
zephyrproject-rtos/zephyr#30421

### DIFF
--- a/subsys/settings/src/settings_line.c
+++ b/subsys/settings/src/settings_line.c
@@ -295,7 +295,14 @@ int settings_line_entry_copy(void *dst_ctx, off_t dst_off, void *src_ctx,
 			break;
 		}
 
-		rc = settings_io_cb.write_cb(dst_ctx, dst_off, buf, chunk_size);
+		size_t write_size = chunk_size;
+
+		if (chunk_size % settings_io_cb.rwbs) {
+			write_size += settings_io_cb.rwbs -
+				      chunk_size % settings_io_cb.rwbs;
+		}
+
+		rc = settings_io_cb.write_cb(dst_ctx, dst_off, buf, write_size);
 		if (rc) {
 			break;
 		}


subsys/settings/src/settings_line.c
```diff
@@ -295,7 +295,14 @@ int settings_line_entry_copy(void *dst_ctx, off_t dst_off, void *src_ctx,
 			break;
 		}
 
-		rc = settings_io_cb.write_cb(dst_ctx, dst_off, buf, chunk_size);
+		size_t write_size = chunk_size;
+
+		if (chunk_size % settings_io_cb.rwbs) {
+			write_size = settings_io_cb.rwbs -
+				     chunk_size % settings_io_cb.rwbs;
+		}
```
de-nordic, line 303: I don;t get the math and need help here.
So the `chunk_size` is size what we have read to `buf`? And it might be anything in range [1-16].
To write to flash that has non-1 alignment we would have to write enough bytes to get that alignment?
Then the size of the write_size should then be `(chunk_size + settings_io_cb.rwsb - 1) & ~(settings_io_cb.wrsb - 1)`
Because here `write_size` would still be a fraction of `settings_io_cb.rwbs` if my calculations are ok.
Example given: let rwsb be 4 and len be 3
 write_size = 4 - 3 % 4 => write_size = 1
let rswb be 4 and len be 5
 write_size = 4 - 5 % 4 => write_size = 4 - 1 = 3

What am I missing?

This would only happen in last run of while loop? 

And the FCB, if it is used, does the alignment during writes with `fcb_len_in_flash`.

nvlsianpu: missed +=

de-nordic: OK probably the `dst_off`, in line 311, should be increased by `write_size` although it differs from `chunk_size` only in the last run so the current code does not break anything.

nvlsianpu: It is the last run so no sense to touch.

ioannisg:
@carlescufi @shantha-14 this PR does not have a milestone set, btw.

jhn-nordic (approved):

nvlsianpu:
went in within #404

PR #408, opened 2020-12-08
[nrf noup] ci: update compliance test workflow

utzig:
Update GH compliance test workflow to be compatible with NCS.

### DIFF
--- a/.github/workflows/compliance-ncs.yml
+++ b/.github/workflows/compliance-ncs.yml
@@ -1,0 +1,63 @@
+name: Compliance NCS
+
+on: pull_request
+
+jobs:
+  compliance_job:
+    runs-on: ubuntu-latest
+    name: Run compliance checks on patch series (PR)
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v2
+        with:
+          path: ncs/zephyr
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: cache-pip
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-doc-pip
+
+      - name: Run Compliance Tests
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        working-directory: ncs
+        run: |
+          pip3 install -U pip
+          pip3 install -U setuptools
+          export PATH="$HOME/.local/bin:$PATH"
+          pip3 install -U west
+          west init -m https://github.com/nrfconnect/sdk-nrf
+          west update -k
+          pip3 install -U gitlint
+          pip3 install -r zephyr/scripts/requirements-compliance.txt
+          cd zephyr
+          source zephyr-env.sh
+          ./scripts/ci/check_compliance.py -m Codeowners -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m Kconfig -c origin/${BASE_REF}.. || true
+
+      - name: Upload Results
+        uses: actions/upload-artifact@master
+        continue-on-error: True
+        with:
+          name: compliance.xml
+          path: ncs/zephyr/compliance.xml
+
+      - name: Check Warnings
+        run: |
+          for file in Codeowners.txt Gitlint.txt Identity.txt Nits.txt pylint.txt checkpatch.txt Kconfig.txt; do
+            path="ncs/zephyr/${file}"
+            if [[ -s $path ]]; then
+              errors=$(cat $path)
+              errors="${errors//'%'/'%25'}"
+              errors="${errors//$'\n'/'%0A'}"
+              errors="${errors//$'\r'/'%0D'}"
+              echo "::error file=${file}::$errors"
+              exit=1
+            fi
+          done
+
+          if [[ $exit == 1 ]]; then
+            exit 1
+          fi


thst-nordic:
This has the git lint issue that other PR's have.

utzig:
> This has the git lint issue that other PR's have.

But it comes from Jenkins, no?

thst-nordic:
yes, I found the problem.  testing it now.

thst-nordic (approved):

carlescufi (changes requested):
I'd rather you leave the upstream one untouched and add a new workflow `compliance-ncs.yml`

utzig:
> I'd rather you leave the upstream one untouched and add a new workflow `compliance-ncs.yml`

Done.

.github/workflows/compliance-ncs.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          pip3 install -U pip
+          pip3 install -U setuptools
+          export PATH="$HOME/.local/bin:$PATH"
+          pip3 install -U west
+          west init -m https://github.com/nrfconnect/sdk-nrf
+          west update zephyr nrfxlib
+          pip3 install -U gitlint
+          pip3 install -r zephyr/scripts/requirements-compliance.txt
+          cd sdk-zephyr
```
carlescufi, line 36: This is checked out in `ncs/zephyr`, not `ncs/sdk-zephyr`

.github/workflows/compliance-ncs.yml
```diff
@@ -0,0 +1,62 @@
...
+          ZEPHYR_BASE=$(pwd) ../zephyr/scripts/ci/check_compliance.py -m Codeowners -m Devicetree -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m Kconfig -c origin/${BASE_REF}.. || true
```
carlescufi, line 37: no need for the `$(pwd) ../zephyr/` since you are already in `zephyr/`

carlescufi (changes requested):

carlescufi:
> > I'd rather you leave the upstream one untouched and add a new workflow `compliance-ncs.yml`
> 
> Done.

Thank you, a couple of comments from me.

.github/workflows/compliance-ncs.yml
utzig: No, it is not, check the `checkout` step.

.github/workflows/compliance-ncs.yml
utzig: No, I am in `sdk-zephyr`. The reason for both is that `zephyr` is fetched by `west` and `sdk-zephyr` is the clone of the current repo. If I don't set the variable on the run, it will try to execute in `zephyr` which is a whatever `west` brought in and not what the contents of the current PR are.

carlescufi:
The commit should have a `[nrf noup]` sauce tag too.

carlescufi:
Also, both in `sdk-zephyr` and `sdk-nrfxlib` CI, you should do a `west update` of everything and then check out the PR in the corresponding folder (`zephyr` here and `nrfxlib` there). Please link this PR to one in nrfxlib making the same changes so we can compare them.

hakonfam:
https://github.com/nrfconnect/sdk-nrf/pull/3232#discussion_r558298590

FYI

Would it be possible to get the ball rolling on this again?

utzig:
> Also, both in `sdk-zephyr` and `sdk-nrfxlib` CI, you should do a `west update -k` or `git checkout -b pr` of everything and then check out the PR in the corresponding folder (`zephyr` here and `nrfxlib` there). Please link this PR to one in nrfxlib making the same changes so we can compare them.

https://github.com/nrfconnect/sdk-nrfxlib/pull/367

.github/workflows/compliance-ncs.yml
```diff
@@ -0,0 +1,63 @@
...
+          pip3 install -U setuptools
+          export PATH="$HOME/.local/bin:$PATH"
+          pip3 install -U west
+          west init -m https://github.com/nrfconnect/sdk-nrf
+          west update zephyr nrfxlib
+          pip3 install -U gitlint
+          pip3 install -r zephyr/scripts/requirements-compliance.txt
+          export ZEPHYR_BASE=$(pwd)/zephyr
+          cd zephyr
+          ./scripts/ci/check_compliance.py -m Codeowners -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m Kconfig -c origin/${BASE_REF}.. || true
```
carlescufi, line 38: You will need to remove `-m Kconfig` here. If you try locally you will see it fails because of missing Kconfig options that are defined in sdk-nrf

.github/workflows/compliance-ncs.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          pip3 install -U pip
+          pip3 install -U setuptools
+          export PATH="$HOME/.local/bin:$PATH"
+          pip3 install -U west
+          west init -m https://github.com/nrfconnect/sdk-nrf
+          west update zephyr nrfxlib
+          pip3 install -U gitlint
+          pip3 install -r zephyr/scripts/requirements-compliance.txt
+          export ZEPHYR_BASE=$(pwd)/zephyr
```
carlescufi, line 36: Why not set `working-directory: ncs/zephyr` and then:

```
source zephyr-env.sh
```

.github/workflows/compliance-ncs.yml
utzig: Hmmm, I tried locally and had no issue. The Kconfig compliance test sets `KCONFIG_BINARY_DIR` to `/tmp` and writes `Kconfig.modules` with the following contents:

```
osource "/home/utzig/src/Nordic/ncs/nrf/Kconfig.nrf"
osource "/home/utzig/src/Nordic/ncs/bootloader/mcuboot/zephyr/Kconfig"
osource "/home/utzig/src/Nordic/ncs/nrfxlib/Kconfig.nrfxlib"
```

I assumed this is how all external modules were "glued" together. What error did you see running locally?

.github/workflows/compliance-ncs.yml
utzig: I don't really know if running `west init -m https://github.com/nrfconnect/sdk-nrf` inside `zephyr` works OK but if it does I can update it. The only requirement for running `Kconfig` compliance is to have `ZEPHYR_BASE` pointing to the correct place, so I don't know what purpose the extra `zephyr-env.sh` lines would add here (and the other compliance checks are all local). Also I messed up my conflict resolution because there was a `west update -k` in before I pushed, will fix now.

utzig: `source zephyr-env.sh` <- Done

.github/workflows/compliance-ncs.yml
utzig: Well, I can see how `sdk-nrf` and `sdk-zephyr` could get out of sync in their config options. Given that `west.yml` from `sdk-nrf` is used, it might have options that would only match some other `sdk-zephyr` branch, but I think this would also affect `sdk-nrfxlib`. The only reliable place to test `-m Kconfig` is `sdk-nrf` since it needs to bring in the appropriate modules. Also the whole point of using `west update -k` was for running `-m Kconfig`, since every other compliance check is kinda self-contained, so only Zephyr itself is needed because the scripts are there. So I could as well just grab `sdk-zephyr`, what do you think?

.github/workflows/compliance-ncs.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Run Compliance Tests
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        working-directory: ncs
+        run: |
+          pip3 install -U pip
+          pip3 install -U setuptools
+          export PATH="$HOME/.local/bin:$PATH"
+          pip3 install -U west
+          west init -m https://github.com/nrfconnect/sdk-nrf
```
carlescufi, line 32: This is wrong. Compliance in sdk-zephyr must use sdk-zephyr/west.yml

carlescufi: Use `west init -l zephyr/` instead.

thst-nordic:
The CI passed. the only failure is the lwm2m_client flash overflow which is being fixed in #424 

We should merge this if it is ready.

carlescufi (commented):
Please include the changes in here:
https://github.com/zephyrproject-rtos/zephyr/pull/31276/files

carlescufi:
Also the ones here:
https://github.com/zephyrproject-rtos/zephyr/pull/31452/files

carlescufi:
@utzig I will take over this.

utzig:
> @utzig I will take over this.

Thanks, do you prefer me to close this PR?

PR #410, opened 2020-12-11
[nrf fromlist] runners: mdb: fix imports

mbolivar-nordic:
Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/30661/

Runner implementations are only allowed to unconditionally import
modules in the python standard library. They are not allowed to import
anything that comes from pip or other third party sources unless they
catch the ImportError and gracefully do nothing.

Fix the imports in the mdb runner accordingly, sorting the imports
into sections to make it clearer what's stdlib, what's runners, and
what's third party while we're here.

### DIFF
--- a/scripts/west_commands/runners/mdb.py
+++ b/scripts/west_commands/runners/mdb.py
@@ -5,12 +5,18 @@
 '''Runners for Synopsys Metaware Debugger(mdb).'''
 
 
-import time
 import shutil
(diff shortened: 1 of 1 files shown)


carlescufi (approved):

mbolivar-nordic:
nrf CI has passed; merging as hotfix

PR #412, opened 2020-12-14
[nrf toup]: Improve existing clock control toup

rugeGerritsen:
Improves d875660a86418497a90e12185914eae902105e4c

With this change it is possible to share all configurations
related to CLOCK_CONTROL_NRF_K32SRC_RC_CALIBRATION without
including the clock calibration configurations.

Alternative temperature driver will be introduced with https://github.com/nrfconnect/sdk-nrf/pull/3022/

Ref: NCSIDB-180


nordic-krch (approved):

de-nordic:
Regarding our discussion on teams:
As I understand that the most desired situation is that the  d875660a86418497a90e12185914eae902105e4c would be squashed with this PR, but that would require reverting the d875660a86418497a90e12185914eae902105e4c  and bringing the changes that it introduces here.
The other possible solution would be to add information, into the commit message of this PR, that  commit from this PR, when passes to upstream, should be reverted with from NRF fork together with the d875660a86418497a90e12185914eae902105e4c .
I am in favor of not reverting and and on adding info to commit message.

@mbolivar What is your insight on this?

drivers/clock_control/clock_control_nrf.c
```diff
@@ -210,7 +210,8 @@ static void lfclk_start(void)
 
 static void lfclk_stop(void)
 {
-	if (IS_ENABLED(CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC_CALIBRATION)) {
+	if (IS_ENABLED(CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC_CALIBRATION)
+	    && !IS_ENABLED(CONFIG_CLOCK_CONTROL_NRF_FORCE_ALT)) {
```
joerchan, line 214: Everywhere this style is used:
```suggestion
	if (IS_ENABLED(CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC_CALIBRATION) &&
	    !IS_ENABLED(CONFIG_CLOCK_CONTROL_NRF_FORCE_ALT)) {
```

joerchan (approved):
Changes looks good, what is the way forward with these patches now?

drivers/clock_control/Kconfig.nrf
```diff
@@ -56,16 +56,15 @@ endchoice
 
 config CLOCK_CONTROL_NRF_K32SRC_RC_CALIBRATION
 	bool "Enable LF clock calibration"
-	depends on !SOC_SERIES_NRF91X
-	depends on !CLOCK_CONTROL_NRF_FORCE_ALT
+	depends on !SOC_SERIES_NRF91X && CLOCK_CONTROL_NRF_K32SRC_RC
 	default n if BOARD_NRF5340PDK_NRF5340_CPUAPP || BOARD_NRF5340PDK_NRF5340_CPUAPPNS
 	default y if CLOCK_CONTROL_NRF_K32SRC_RC
```
anangl, line 61:
```suggestion
	default y
```
This condition here became redundant.

drivers/clock_control/clock_control_nrf.c
rugeGerritsen: Fixed

drivers/clock_control/Kconfig.nrf
rugeGerritsen: Fixed

anangl (approved):

rugeGerritsen:
> Regarding our discussion on teams:
> As I understand that the most desired situation is that the [d875660](https://github.com/nrfconnect/sdk-zephyr/commit/d875660a86418497a90e12185914eae902105e4c) would be squashed with this PR, but that would require reverting the [d875660](https://github.com/nrfconnect/sdk-zephyr/commit/d875660a86418497a90e12185914eae902105e4c) and bringing the changes that it introduces here.
> The other possible solution would be to add information, into the commit message of this PR, that commit from this PR, when passes to upstream, should be reverted with from NRF fork together with the [d875660](https://github.com/nrfconnect/sdk-zephyr/commit/d875660a86418497a90e12185914eae902105e4c) .
> I am in favor of not reverting and and on adding info to commit message.
> 
> @mbolivar What is your insight on this?

@mbolivar-nordic , can you provide some insight here?

mbolivar-nordic:
@de-nordic 's summary is correct in the two options.

It's just a matter of when you want the job done, and by whom. If you revert d875660 and merge its changes into this PR, then tools like `west ncs-loot` will know that d875660 is "gone" from the logical set of out of tree patches we are currently carrying.

I am the one who rebases our out of tree NCS patches into a linear history when NCS heads towards a release (the fact that we rebase is documented in https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/dm_code_base.html#oss-repositories-downstream-project-history).

The way I do this uses `west ncs-loot`, so if you revert+reapply, it means less work for me and less work to do during release time.

If you just stack it on top and add commit messages, I will read them and do the squashing manually at release time.

Whatever you choose is fine with me; if you don't revert, please do write clear instructions.

rugeGerritsen:
Reverted the commit this PR intends to improve and rebased on master

rugeGerritsen:
Added sign-off in the revert-commit

mbolivar-nordic:
Hi @rugeGerritsen ,

> Reverted the commit this PR intends to improve and rebased on master

This was not a clean revert; please don't do it like this in the future.

Any revert needs to be conflict-free; do not resolve conflicts in the revert commit. If you manually resolve conflicts, it breaks our scripts. That happened in this case.

Details:

- the original commit being "reverted" is d875660a86418497a90e12185914eae902105e4c
- the commit that claims to be a revert is 490354229a2a4fd65a010ecc51c9bdfbbb95a41a; its parent commit is f2a69816da9ad047c40c0d2eb364e00755ec33e6

I should therefore be able to `git checkout f2a6981 && git revert d875660` and end up with a commit that has no diff with 490354229a2a4fd65a010ecc51c9bdfbbb95a41a.

But what I actually get is a conflict:

```
$ git revert d875660a86418497a90e12185914eae902105e4c
Auto-merging drivers/clock_control/Kconfig.nrf
CONFLICT (content): Merge conflict in drivers/clock_control/Kconfig.nrf
error: could not revert d875660a86... [nrf toup] drivers: clock_control: nrf: Change CLOCK_CONTROL_NRF_FORCE_A
```

The manual resolution of the conflict is tricking `west ncs-loot` into giving misleading output, which in turn requires extra manual steps in preparing a clean rebase of our out of tree patches.

TL;DR: next time if the revert is not clean, please don't resolve conflicts manually. Feel free to PM me off GitHub if you want more details on what to do instead and other rationale.

carlescufi:
@rugeGerritsen Could you please send this PR upstream?

PR #419, opened 2021-01-11
TFM: configurable memory layout

oyvindronningstad:
Mix of toup and noup (and temphack).

### DIFF
--- a/boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
+++ b/boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
@@ -46,13 +46,10 @@ if (CONFIG_BUILD_WITH_TFM)
 		# Merge tfm_s and zephyr (NS) image to a single binary.
 		set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
 			COMMAND ${SREC_CAT}
-			ARGS ${CMAKE_BINARY_DIR}/tfm/bin/tfm_s.bin -Binary
+			ARGS $<TARGET_PROPERTY:tfm,TFM_S_BIN_FILE> -Binary
 				${CMAKE_BINARY_DIR}/zephyr/${KERNEL_BIN_NAME} -Binary
 				-offset ${CONFIG_FLASH_LOAD_OFFSET}
 				-o ${CMAKE_BINARY_DIR}/tfm_merged.hex -intel
-
-			# Copy tfm_sign.hex to zephyr
-			COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/tfm_merged.hex ${CMAKE_BINARY_DIR}/zephyr/${KERNEL_HEX_NAME}
 		)
 	else()
 		#Create and sign for concatenated binary image, should align with the TF-M BL2
@@ -102,9 +99,13 @@ if (CONFIG_BUILD_WITH_TFM)
 				-offset 0x10000
 				-o ${CMAKE_BINARY_DIR}/tfm_sign.hex
 				-intel
+		)
 
-			# Copy tfm_sign.hex to zephyr
-			COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/tfm_sign.hex ${CMAKE_BINARY_DIR}/zephyr/${KERNEL_HEX_NAME}
+		set_property(GLOBAL APPEND PROPERTY extra_post_build_byproducts
+			${CMAKE_BINARY_DIR}/tfm_s_signed.bin
+			${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
+			${CMAKE_BINARY_DIR}/tfm_sign.bin
+			${CMAKE_BINARY_DIR}/tfm_sign.hex
 		)
 	endif()
 endif()
--- a/boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
+++ b/boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
@@ -94,7 +94,7 @@ if (CONFIG_BUILD_WITH_TFM)
 			COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_PROPERTY:tfm,BL2_BIN_FILE> ${CMAKE_BINARY_DIR}/mcuboot.bin
 
 			# Generate an intel hex file from the signed output binary
-			COMMAND srec_cat ${CMAKE_BINARY_DIR}/tfm_sign.bin
+			COMMAND ${SREC_CAT} ${CMAKE_BINARY_DIR}/tfm_sign.bin
 				-binary
 				-offset 0x10000
 				-o ${CMAKE_BINARY_DIR}/tfm_sign.hex
--- a/boards/arm/nrf9160dk_nrf9160/CMakeLists.txt
+++ b/boards/arm/nrf9160dk_nrf9160/CMakeLists.txt
@@ -65,7 +65,7 @@ if (CONFIG_BUILD_WITH_TFM)
 
 		#Create concatenated binary image from the two independently signed binary files
 		COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/assemble.py
-		     --layout ${PREPROCESSED_FILE_S}
+			 --layout ${PREPROCESSED_FILE_S}
 			 -s ${CMAKE_BINARY_DIR}/tfm_s_signed.bin
 			 -n ${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
 			 -o ${CMAKE_BINARY_DIR}/tfm_sign.bin
@@ -79,8 +79,12 @@ if (CONFIG_BUILD_WITH_TFM)
 		-offset 0x10000
 		-o ${CMAKE_BINARY_DIR}/tfm_sign.hex
 		-intel
+	)
 
-		# Copy tfm_sign.hex to zephyr
-		COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/tfm_sign.hex ${CMAKE_BINARY_DIR}/zephyr/${KERNEL_HEX_NAME}
+	set_property(GLOBAL APPEND PROPERTY extra_post_build_byproducts
+		${CMAKE_BINARY_DIR}/tfm_s_signed.bin
+		${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
+		${CMAKE_BINARY_DIR}/tfm_sign.bin
+		${CMAKE_BINARY_DIR}/tfm_sign.hex
 	)
 endif()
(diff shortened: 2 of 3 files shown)


modules/tfm/src/zephyr_tfm_log.c
```diff
@@ -0,0 +1,18 @@
+/*
```
oyvindronningstad, line 1: @ioannisg @tejlmand Please comment on the choice of location for this file. It's used here: https://github.com/oyvindronningstad/trusted-firmware-m-zephyr-module/blob/plat-ext-2/CMakeLists.txt#L178

modules/tfm/src/zephyr_tfm_ns_interface.c
```diff
@@ -0,0 +1,36 @@
+/*
```
oyvindronningstad, line 1: @ioannisg @tejlmand Please comment on the choice of location for this file. It's used here: https://github.com/oyvindronningstad/trusted-firmware-m-zephyr-module/blob/plat-ext-2/CMakeLists.txt#L245

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
```diff
@@ -70,6 +70,13 @@ if (CONFIG_BUILD_WITH_TFM)
 				$<TARGET_PROPERTY:tfm,TFM_S_BIN_FILE>
 				${CMAKE_BINARY_DIR}/tfm_s_signed.bin
 
+			# Generate an intel hex file from the signed output binary
```
hakonfam, line 73: Can you append these commands to the `extra_post_build_commands` in a separate cmake statement?

To avoid merge conflicts when upmerging.

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
```diff
@@ -103,7 +117,9 @@ if (CONFIG_BUILD_WITH_TFM)
 
 		set_property(GLOBAL APPEND PROPERTY extra_post_build_byproducts
 			${CMAKE_BINARY_DIR}/tfm_s_signed.bin
+			${CMAKE_BINARY_DIR}/tfm_s_signed.hex
```
hakonfam, line 120: Can you append these commands to the `extra_post_build_byproducts` in a separate cmake statement?

To avoid merge conflicts when upmerging.

boards/arm/nrf9160dk_nrf9160/CMakeLists.txt
```diff
@@ -50,6 +50,13 @@ if (CONFIG_BUILD_WITH_TFM)
 			 $<TARGET_PROPERTY:tfm,TFM_S_BIN_FILE>
 			 ${CMAKE_BINARY_DIR}/tfm_s_signed.bin
 
+		# Generate an intel hex file from the signed output binary
```
hakonfam, line 53: Same comment as for nrf53, try doing this in separate statements to avoid merge-conflicts.

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
oyvindronningstad: I considered that, but since the commands depend on other post_build_commands, I thought we couldn't guarantee that they are run in the right order unless they are part of the same entry like this. Maybe @tejlmand can chime in on this ordering?

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
```diff
@@ -46,13 +46,10 @@ if (CONFIG_BUILD_WITH_TFM)
 		# Merge tfm_s and zephyr (NS) image to a single binary.
 		set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
 			COMMAND ${SREC_CAT}
-			ARGS ${CMAKE_BINARY_DIR}/tfm/bin/tfm_s.bin -Binary
+			ARGS $<TARGET_PROPERTY:tfm,TFM_S_BIN_FILE> -Binary
```
tejlmand, line 49: Nothing wrong here, comment relates to this commit message: f394554937510bd90300eabf98f88f8a24bdd217
Please provide reference to upstream PR.

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
```diff
@@ -70,6 +70,13 @@ if (CONFIG_BUILD_WITH_TFM)
 				$<TARGET_PROPERTY:tfm,TFM_S_BIN_FILE>
 				${CMAKE_BINARY_DIR}/tfm_s_signed.bin
 
+			# Generate an intel hex file from the signed output binary
+			COMMAND srec_cat ${CMAKE_BINARY_DIR}/tfm_s_signed.bin
```
tejlmand, line 74: Never call a command directly, always use `find_program()` and ensure this code is not activated if `srec_cat` is missing.

It's already looked up here: https://github.com/nrfconnect/sdk-zephyr/blob/21046b8cdb4eac989c6b17bb21ffc8196be3d5e4/boards/arm/nrf5340dk_nrf5340/CMakeLists.txt#L40

tejlmand (changes requested):
Why do we need a Zephyr `[noup]` patch here (85707880b1108c8fa19f73b4c0e509eb099e1eda) ? 

Those:
```
extra_post_build_commands
extra_post_build_byproducts
```
are global properties, so we can do all of this in `sdk-nrf` directly.

Secondly, why is the conversion from bin to hex-files suddenly needed ?
I don't see any references to what this `[noup]` fixes or why it is needed.
I don't see the relation to configurable memory layout.

oyvindronningstad:
> 
> 
> Why do we need a Zephyr `[noup]` patch here ([8570788](https://github.com/nrfconnect/sdk-zephyr/commit/85707880b1108c8fa19f73b4c0e509eb099e1eda)) ?
> 
> Those:
> 
> ```
> extra_post_build_commands
> extra_post_build_byproducts
> ```
> 
> are global properties, so we can do all of this in `sdk-nrf` directly.

I think this is the same as @hakonfam's question: https://github.com/nrfconnect/sdk-zephyr/pull/419#discussion_r567724123

> 
> Secondly, why is the conversion from bin to hex-files suddenly needed ?
> I don't see any references to what this `[noup]` fixes or why it is needed.
> I don't see the relation to configurable memory layout.

It's because we want Partition Manager to handle the merging of hex files. Originally, we only had a hex file for the signed combination of tfm+app, but the logic for creating that hex file no longer works now that the sizes are controlled from outside, since the bin->hex conversion needs the correct offset.

tejlmand:
> I think this is the same as @hakonfam's question: #419 (comment)

I understood that comment as placing it in a separate statement like:
```
	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
			... existing cmake statement with commands ...
        )

	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
			... new cmake statement with bin2hex commands ...
        )
```
in the file you already update.

I'm asking to put this code into `sdk-nrf` repo here: https://github.com/nrfconnect/sdk-nrf/
to completely avoid a `[noup]` patch.

oyvindronningstad:
> 
> 
> > I think this is the same as @hakonfam's question: #419 (comment)
> 
> I understood that comment as placing it in a separate statement like:
> 
> ```
> 	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
> 			... existing cmake statement with commands ...
>         )
> 
> 	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
> 			... new cmake statement with bin2hex commands ...
>         )
> ```
> 
> in the file you already update.
> 
> I'm asking to put this code into `sdk-nrf` repo here: https://github.com/nrfconnect/sdk-nrf/
> to completely avoid a `[noup]` patch.

But my reply about ordering is the same. Will the order be correct?

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
oyvindronningstad: Fixed

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
oyvindronningstad: Fixed

tejlmand:
> > Secondly, why is the conversion from bin to hex-files suddenly needed ?
> > I don't see any references to what this `[noup]` fixes or why it is needed.
> > I don't see the relation to configurable memory layout.
> 
> It's because we want Partition Manager to handle the merging of hex files. Originally, we only had a hex file for the signed combination of tfm+app, but the logic for creating that hex file no longer works now that the sizes are controlled from outside, since the bin->hex conversion needs the correct offset.

Thanks for explaining this. Thisd wasn't clear to me from PR text or commit messages.

As partition manager is a sdk-nrf specific addition, then I even more consider the extra byproducts commit to be something that deserves to live in `sdk-nrf` repo.

I even think you could do this commonly, asI don't see why this conversion / extra byproduct needs to be in board specific CMakeLists.txt file.

tejlmand:
> 
> But my reply about ordering is the same. Will the order be correct?

If I understand your patch correctly, then your patch depends on the bin files being created before your bin2hex conversion, but nothing in your code depends regarding the extra by products with the bin2hex conversion to have happened, from what I can tell.

Then the order is guaranteed.
The board CMakeLists.txt with existing extra commands are appended to the build here:
https://github.com/nrfconnect/sdk-zephyr/blob/master/CMakeLists.txt#L446
and the Zephyr modules (including `sdk-nrf`) is processed afterwards here:
https://github.com/nrfconnect/sdk-zephyr/blob/master/CMakeLists.txt#L461

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
oyvindronningstad: Done

oyvindronningstad:
> Then the order is guaranteed.
> The board CMakeLists.txt with existing extra commands are appended to the build here:
> https://github.com/nrfconnect/sdk-zephyr/blob/master/CMakeLists.txt#L446
> and the Zephyr modules (including `sdk-nrf`) is processed afterwards here:
> https://github.com/nrfconnect/sdk-zephyr/blob/master/CMakeLists.txt#L461

Thanks a lot for confirming. I made the change now.

boards/arm/nrf5340dk_nrf5340/CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ if (CONFIG_BUILD_WITH_TFM)
 			COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_PROPERTY:tfm,BL2_BIN_FILE> ${CMAKE_BINARY_DIR}/mcuboot.bin
 
 			# Generate an intel hex file from the signed output binary
-			COMMAND srec_cat ${CMAKE_BINARY_DIR}/tfm_sign.bin
+			COMMAND ${SREC_CAT} ${CMAKE_BINARY_DIR}/tfm_sign.bin
```
tejlmand, line 97: nothing wrong here, but 55d587fed3583c758f7f0fe2d31e48a6205e11cc should now be a `[nrf fromtree]` as the upstream PR has been merged.

oyvindronningstad: Done

oyvindronningstad: Fixed

tejlmand (approved):
lgtm, thanks for the patience and the cleanup.

mbolivar-nordic:

hakonfam (approved):

ioannisg (approved):
Looks fine to me. 
fromtree is also approved (did not make it due to stabilization freeze)

PR #426, opened 2021-01-20
[nrf fromlist] drivers/flash/soc_flash_nrf: FTPAN-242 workaround

nvlsianpu:
- Introduce support for situation when synchronization back-end aborts operation before it is done. synchronization API will transfer operation return code to the driver shim back.

- Disable POFWARN by writing POFCON before a write or erase operation. Do not attempt to write or erase if EVENTS_POFWARN is already asserted.

This is upstream PR copy https://github.com/zephyrproject-rtos/zephyr/pull/31759

### DIFF
--- a/drivers/flash/soc_flash_nrf.c
+++ b/drivers/flash/soc_flash_nrf.c
@@ -15,6 +15,7 @@
 #include <drivers/flash.h>
 #include <string.h>
 #include <nrfx_nvmc.h>
+#include <nrf_erratas.h>
 
 #include "soc_flash_nrf.h"
 
@@ -80,6 +81,17 @@ static struct k_sem sem_lock;
 #define SYNC_UNLOCK()
 #endif
 
+#if NRF52_ERRATA_242_PRESENT
+#include <hal/nrf_power.h>
+static int suspend_pofwarn(void);
+static void restore_pofwarn(void);
+
+#define SUSPEND_POFWARN() suspend_pofwarn()
+#define RESUME_POFWARN()  restore_pofwarn()
+#else
+#define SUSPEND_POFWARN() 0
+#define RESUME_POFWARN()
+#endif /* NRF52_ERRATA_242_PRESENT */
 
 static int write(off_t addr, const void *data, size_t len);
 static int erase(uint32_t addr, uint32_t size);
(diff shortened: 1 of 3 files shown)


nvlsianpu (commented):
need to extend to cover nRF52840 as well (just Kconfig)

drivers/flash/soc_flash_nrf.c
```diff
@@ -80,6 +80,18 @@ static struct k_sem sem_lock;
 #define SYNC_UNLOCK()
 #endif
 
+#ifdef CONFIG_NRF52_ANOMALY_242_WORKAROUND
+#include <hal/nrf_power.h>
+static int suspend_pofwarn(void);
+static void restore_pofwarn(void);
+
+#define SUSPEND_COLIDING_SYSTEMS() suspend_pofwarn()
```
eugmes, line 88: That are strange names those macros have. It should be "colliding" probably, but that too is weird. Maybe "conflicting" or something.

drivers/flash/soc_flash_nrf_ticker.c
```diff
@@ -58,18 +58,20 @@ static void time_slot_callback_work(uint32_t ticks_at_expire,
...
-	if (op_desc->handler(op_desc->context) == FLASH_OP_DONE) {
+	rc = op_desc->handler(op_desc->context);
+	if (rc != FLASH_OP_ONGOING) {
 		ll_timeslice_ticker_id_get(&instance_index, &ticker_id);
 
 		/* Stop the time slot ticker */
 		_ticker_stop(instance_index, 0, ticker_id);
 
-		_ticker_sync_context.result = 0;
+		_ticker_sync_context.result = (rc == FLASH_OP_DONE) ? 0 : rc;
```
eugmes, line 74: This should be one of constants from `errno.h` negated.

eugmes (changes requested):

nika-nordic (approved):
LGTM, just enable workaround for nRF52840 as well

drivers/flash/soc_flash_nrf_ticker.c
nvlsianpu: @eugmes
OK, I would use `-ECANCELED`.
What about using -ECANCELED as the `handler()` return code instead of `FLASH_OP_ABORTED`?

eugmes: Yes, that's one I was thinking about.

nvlsianpu: now  `-ECANCELED` is transferred from the handler error code

mbolivar-nordic:

nvlsianpu:
PR was rebased and synchronized with the upstream PR.

eugmes (approved):
Looks ok to me (the cancel logic, I didn't evaluate the FTPAN workaround).

tejlmand (changes requested):
Included in #459

nvlsianpu:
Already Merged via #459

PR #428, opened 2021-01-21
[DNM] Int/nrf5340 multiprotocol until upmerge

jciupis:
This PR serves as a placeholder for changes that are supposed to be cherry-picked from upstream Zephyr when the upmerge is finished. Until then all the changes needed in Zephyr to enable multiprotocol operation on nRF5340 are going to be kept here.


jciupis:
This PR was never intended to be merged. I'll close it when #435 is merged

PR #438, opened 2021-02-01
Compliance downstream patches

carlescufi:
- Cherry picks the latest changes to the compliance workflow
- Adds a new `noup` commit with a couple of tweaks

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -36,6 +36,7 @@ jobs:
         west update
 
     - name: Run Compliance Tests
+      continue-on-error: true
       id: compliance
       env:
         BASE_REF: ${{ github.base_ref }}
@@ -48,7 +49,7 @@ jobs:
         # debug
         ls -la
         git log  --pretty=oneline | head -n 10
-        ./scripts/ci/check_compliance.py -m Codeowners -m Devicetree -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m Kconfig -c origin/${BASE_REF}.. || true
+        ./scripts/ci/check_compliance.py -m Codeowners -m Devicetree -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m Kconfig -c origin/${BASE_REF}..
 
     - name: upload-results
       uses: actions/upload-artifact@master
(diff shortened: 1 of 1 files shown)


.github/workflows/compliance.yml
```diff
@@ -59,6 +60,11 @@ jobs:
 
     - name: check-warns
       run: |
+        git log --oneline --first-parent origin/${BASE_REF}..HEAD | grep -E "\[nrf (mergeup|fromlist|toup|noup|temphack|fromtree)\]" || { echo 'Sauce tag missing'; exit 1; }
```
carlescufi, line 63: The `--first-parent` ensures that, in the case of an upmerge, only the merge commit and any other reverts or out-of-tree commits require the sauce tag

mbolivar-nordic:

.github/workflows/compliance.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         # debug
         ls -la
         git log  --pretty=oneline | head -n 10
-        ./scripts/ci/check_compliance.py -m Codeowners -m Devicetree -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m Kconfig -c origin/${BASE_REF}..
+        ./scripts/ci/check_compliance.py -m Codeowners -m Devicetree -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m KconfigBasic -c origin/${BASE_REF}..
```
tejlmand, line 52: Why is `KconfigBasic` needed here ?

When we run compliance in `sdk-zephyr`, then we use the `sdk-zephyr/west.yml` and thus have all needed repos available for running full Kconfig check, which should pass without errors.

When running compliance in `sdk-nrf` repo, we already uses `KconfigBasic` here:
https://github.com/nrfconnect/sdk-nrf/blob/4680001fd63b066eb688af3909bc31bbde99d7cf/.github/workflows/compliance.yml#L60
```
        $ZEPHYR_BASE/scripts/ci/check_compliance.py -m Codeowners -m Devicetree -m Gitlint -m Identity -m Nits -m pylint -m checkpatch -m KconfigBasic -c origin/${BASE_REF}..
```

carlescufi: Because we have references to `CONFIG_` symbols in the code that are only present in sdk-nrf. IIRC it's the partition manager.

tejlmand: right, had not taken into consideration that we have `[noup]` patches that would impact.
Thanks for clarifying.

tejlmand (approved):
lgtm

PR #448, opened 2021-02-04
[nrf fromlist] drivers: serial: nrfx_uarte: Fix power management

nordic-krch:
Power consumption was still high after putting uarte device into off
state. It was caused by ENDRX interrupt that was triggered after
calling STOPRX. ENDRX event was called with 0 amount but interrupt
got triggered and fifo_read was starting RX again.

Added disabling RX interrupt before disabling UARTE and reenabling at
device activation.

Signed-off-by: Krzysztof Chruscinski <krzysztof.chruscinski@nordicsemi.no>

### DIFF
--- a/drivers/serial/uart_nrfx_uarte.c
+++ b/drivers/serial/uart_nrfx_uarte.c
@@ -424,6 +424,33 @@ static bool is_tx_ready(const struct device *dev)
 		       nrf_uarte_event_check(uarte, NRF_UARTE_EVENT_ENDTX) : 0);
 }
 
+/* Wait until the transmitter is in the idle state. When this function returns,
+ * IRQ's are locked with the returned key.
+ */
+static int wait_tx_ready(const struct device *dev)
+{
+	int key;
+
+	do {
+		/* wait arbitrary time before back off. */
+		bool res;
+
+		NRFX_WAIT_FOR(is_tx_ready(dev), 100, 1, res);
+
+		if (res) {
+			key = irq_lock();
+			if (is_tx_ready(dev)) {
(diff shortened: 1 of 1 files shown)


anangl (approved):

anangl:
I think that UARTE driver fixes from https://github.com/zephyrproject-rtos/zephyr/pull/31988 should also be included in NCS 1.5. Maybe those commits could be picked into this PR?

nordic-krch:
@anangl added.

carlescufi (approved):

MarekPorwisz (approved):

pszkotak:
@nordic-krch, could you rebase this?
It seems that we need commits bumping up the hal to pass CI tests.

carlescufi:
> @nordic-krch, could you rebase this?
> It seems that we need commits bumping up the hal to pass CI tests.

Tests seem green now, what is the issue?

pszkotak:
> > @nordic-krch, could you rebase this?
> > It seems that we need commits bumping up the hal to pass CI tests.
> 
> Tests seem green now, what is the issue?

PR with the manifest is red on the Zigbee sample build (https://github.com/nrfconnect/sdk-nrf/pull/3853) - https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-3853/5/console.

The https://github.com/nrfconnect/sdk-zephyr/commit/85b46fbe054f6fb35e6a6becaf9b23c18161ef08 should fix it.

tejlmand (changes requested):
Included in #459

PR #452, opened 2021-02-08
drivers: ieee802154: load EUI64 from UICR

edmont:
Add `IEEE802154_NRF5_UICR_EUI64_ENABLE` option to allow loading EUI64 from UICR registers.

KRKNWK-8932

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -65,13 +65,21 @@ static struct nrf5_802154_data nrf5_data;
 
 #if defined(CONFIG_IEEE802154_NRF5_UICR_EUI64_ENABLE)
 #if defined(CONFIG_SOC_NRF5340_CPUAPP)
+#if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+#define EUI64_ADDR (NRF_UICR_S->OTP)
+#else
 #define EUI64_ADDR (NRF_UICR->OTP)
+#endif /* CONFIG_TRUSTED_EXECUTION_NONSECURE */
 #else
 #define EUI64_ADDR (NRF_UICR->CUSTOMER)
 #endif /* CONFIG_SOC_NRF5340_CPUAPP */
 #else
 #if defined(CONFIG_SOC_NRF5340_CPUAPP) || defined(CONFIG_SOC_NRF5340_CPUNET)
+#if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+#define EUI64_ADDR (NRF_FICR_S->INFO.DEVICEID)
+#else
 #define EUI64_ADDR (NRF_FICR->INFO.DEVICEID)
+#endif /* CONFIG_TRUSTED_EXECUTION_NONSECURE */
 #else
 #define EUI64_ADDR (NRF_FICR->DEVICEID)
 #endif /* CONFIG_SOC_NRF5340_CPUAPP || CONFIG_SOC_NRF5340_CPUNET */
@@ -114,8 +122,9 @@ static void nrf5_get_eui64(uint8_t *mac)
 	defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
 	int ret = -EPERM;
 #if defined(CONFIG_SPM)
-	ret = spm_request_read(&factoryAddress, (uint32_t)EUI64_ADDR,
-				   sizeof(factoryAddress));
+	ret = spm_request_read(&factoryAddress,
+			       (uint32_t)&EUI64_ADDR[EUI64_ADDR_HIGH],
+			       sizeof(factoryAddress));
 #endif
 	if (ret != 0) {
 		LOG_ERR("Unable to read EUI64 from the secure zone.");
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -63,15 +63,27 @@ static struct nrf5_802154_data nrf5_data;
 #define FRAME_PENDING_BIT (1 << 4)
 #define TXTIME_OFFSET_US  (5 * USEC_PER_MSEC)
 
-#if defined(CONFIG_SOC_NRF5340_CPUAPP) || defined(CONFIG_SOC_NRF5340_CPUNET)
-#if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
-#define EUI64_ADDR (NRF_FICR_S->INFO.DEVICEID)
+#if defined(CONFIG_IEEE802154_NRF5_UICR_EUI64_ENABLE)
+#if defined(CONFIG_SOC_NRF5340_CPUAPP)
+#define EUI64_ADDR (NRF_UICR->OTP)
 #else
+#define EUI64_ADDR (NRF_UICR->CUSTOMER)
+#endif /* CONFIG_SOC_NRF5340_CPUAPP */
+#else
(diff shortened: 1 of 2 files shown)


carlescufi (changes requested):
Why is this a `noup`? Is it a problem to have this upstream?

edmont:
> Why is this a `noup`? Is it a problem to have this upstream?

For me it's not a problem, but this is somehow extending this other commit which was also merged as `noup`: https://github.com/nrfconnect/sdk-zephyr/commit/0ab7df74b22af2f06be7142741975add6a482b42

Also we want this to be part of NCS 1.5.0. If you agree I could change it to `toup` and handle both in another PR upstream.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -63,6 +63,19 @@ static struct nrf5_802154_data nrf5_data;
...
+#define EUI64_ADDR (NRF_UICR_S->OTP)
+#else
+#define EUI64_ADDR (NRF_UICR->OTP)
+#endif
+#else
+#define EUI64_ADDR (NRF_UICR->CUSTOMER)
+#endif
+#define EUI64_ADDR_HIGH CONFIG_IEEE802154_UICR_EUI64_REG
+#define EUI64_ADDR_LOW (CONFIG_IEEE802154_UICR_EUI64_REG + 1)
+#else
```
hubertmis, line 78: Could you add comments to each `#else` and `#endif`? It's very hard to follow now

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -110,8 +128,8 @@ static void nrf5_get_eui64(uint8_t *mac)
 	}
```
MarekPorwisz, line 128: I can not add comment to the line that is outside of change scope so adding here:
in line 121: you pass EUI64_ADDR directly that points to the start of the UICR/OTP array. Shouldn't an offcest be added there to point to the register specified by the CONFIG_IEEE802154_UICR_EUI64_REG?

hubertmis (commented):
I think this patch should be `fromlist`. The only part unavailable upstream is `CONFIG_SPM` (BTW, we should create a matching config upstream to remove deltas between upstream and sdk-zehyr), which is not modified in this PR.

drivers/ieee802154/Kconfig
```diff
@@ -57,6 +57,27 @@ config IEEE802154_VENDOR_OUI
...
+	  This option enables setting custom vendor EUI64 value
+	  stored in User information configuration registers (UICR).
+	  Notice that this disables the default setting of EUI64
+	  value from Factory information configuration registers
+	  (FICR).
+
+if IEEE802154_UICR_EUI64_ENABLE
+
+config IEEE802154_UICR_EUI64_REG
+	int "UICR base register for the EUI64 value"
```
MarekPorwisz, line 73: lets add range here not to exceed the registers count

drivers/ieee802154/ieee802154_nrf5.c
edmont: You are right, will fix.

drivers/ieee802154/ieee802154_nrf5.c
edmont: Indeed.

drivers/ieee802154/Kconfig
edmont: Sure.

edmont:
I changed this PR to be `toup`. Will handle upstream updates once this is approved.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -91,17 +111,20 @@ static void nrf5_get_eui64(uint8_t *mac)
...
+#endif
 
 #if defined(CONFIG_SOC_NRF5340_CPUAPP) && \
 	defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
 	int ret = -EPERM;
 #if defined(CONFIG_SPM)
-	ret = spm_request_read(&factoryAddress, (uint32_t)EUI64_ADDR,
-				   sizeof(factoryAddress));
+	ret = spm_request_read(&factoryAddress,
+			       (uint32_t)EUI64_ADDR[EUI64_ADDR_HIGH],
```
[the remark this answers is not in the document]
edmont: Yes, missed this.

hubertmis (approved):

carlescufi:
> > Why is this a `noup`? Is it a problem to have this upstream?
> 
> For me it's not a problem, but this is somehow extending this other commit which was also merged as `noup`: [0ab7df7](https://github.com/nrfconnect/sdk-zephyr/commit/0ab7df74b22af2f06be7142741975add6a482b42)
> 
> Also we want this to be part of NCS 1.5.0. If you agree I could change it to `toup` and handle both in another PR upstream.

Yes, let's do that. I have added this patch and 0ab7df7 in the list to revert.

MarekPorwisz (approved):

jfischer-no (approved):

mariuszpos (approved):

carlescufi:
@edmont please link back to the upstream PR from here once you've posted it.

edmont:
@carlescufi this is the upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/32177

Actually https://github.com/nrfconnect/sdk-zephyr/commit/0ab7df74b22af2f06be7142741975add6a482b42 can't be posted upstream since there is no SPM support there.

ioannisg:
> @carlescufi this is the upstream PR: [zephyrproject-rtos/zephyr#32177](https://github.com/zephyrproject-rtos/zephyr/pull/32177)
> 
> Actually [0ab7df7](https://github.com/nrfconnect/sdk-zephyr/commit/0ab7df74b22af2f06be7142741975add6a482b42) can't be posted upstream since there is no SPM support there.

Yes, I think this was pointed out earlier in the thread. 
@carlescufi , I suppose anything that relates to SPM is "no-upstream" - the problem is that SPM and its APIs are defined and implemented in sdk-nrf, but the driver that needs the service(s) is in the zephyr tree.

ioannisg (changes requested):
Cannot be "toup"

ioannisg:
Sorry to complain again @edmont, pls clarify for me: can this really be "fromlist", since it contains the spm stuff? The idea with "fromlist" is that we could skip the commit in following upmerges (should the upstream corresponding commit get merged). But this is missleading since this commit contains the 'no-upstream' SPM-related patch. I wonder if the best is to split this commit in 2 commits (one fromlist and one noup). Unless i am missing sthg of course.

edmont:
@ioannisg thanks for the suggestions, I proceeded with splitting the commit.

ioannisg:
> > > Why is this a `noup`? Is it a problem to have this upstream?
> > 
> > 
> > For me it's not a problem, but this is somehow extending this other commit which was also merged as `noup`: [0ab7df7](https://github.com/nrfconnect/sdk-zephyr/commit/0ab7df74b22af2f06be7142741975add6a482b42)
> > Also we want this to be part of NCS 1.5.0. If you agree I could change it to `toup` and handle both in another PR upstream.
> 
> Yes, let's do that. I have added this patch and [0ab7df7](https://github.com/nrfconnect/sdk-zephyr/commit/0ab7df74b22af2f06be7142741975add6a482b42) in the list to revert.

@carlescufi I don't think you can revert the last commit - it contains downstream-only patches.

ioannisg (approved):

edmont:
Closing since it's covered by https://github.com/nrfconnect/sdk-zephyr/pull/459

PR #462, opened 2021-02-15
[nrf noup] dts: Adding cryptocell_sw as chosen entropy

frkv:
-Adding cryptocell_sw as a chosen entropy source as the non-secure
 does not have access to a entropy peripheral and need to SPM (and
 later PSA APIs through the same driver)
-Not upstream as driver is only downstream

manifest pr: https://github.com/nrfconnect/sdk-nrf/pull/3978

ref: NCSDK-8246

### DIFF
--- a/dts/arm/nordic/nrf5340_cpuappns.dtsi
+++ b/dts/arm/nordic/nrf5340_cpuappns.dtsi
@@ -30,6 +30,15 @@
 
 	chosen {
 		zephyr,flash-controller = &flash_controller;
+
+		/*
+		 * By default, system entropy comes from the entropy_cc310.c
+		 * driver in the nrf repository. This is devicetree glue
+		 * needed to make the system aware of that fact. Individual
+		 * applications can override this by changing this property
+		 * value.
+		 */
+		zephyr,entropy = &cryptocell_sw;
 	};
 
 	soc {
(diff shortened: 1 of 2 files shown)


frkv:
Note that we have not changed the name of the file entropy_cc310.c into entropy_cc3xx.c so the statement in the comment is still pointing to entropy_cc310 driver even though  nrF5340 has a Arm CryptoCell CC312 HW peripheral. The file-name change will happen at a later time

ioannisg (approved):
@frkv this is already supported in the same way for nRF9160 NS, right?

tejlmand (approved):

anangl (approved):

anangl (changes requested):
A binding for "nordic,nrf-cc312-sw" should be also added.

ioannisg:
> A binding for "nordic,nrf-cc312-sw" should be also added.

Right, i guess the existing one is the 310 :)

anangl:
> > A binding for "nordic,nrf-cc312-sw" should be also added.
> 
> Right, i guess the existing one is the 310 :)

Yes, either the compatible property in the added node needs to be changed to "nordic,nrf-cc310-sw" or a new binding matching the "nordic,nrf-cc312-sw" string needs to be provided. I'm not sure what was the original intention of handling this.

ioannisg:
@anangl I added a binding file for the 312 "sw" cryptocell. I thought of taking the second option, as more future-proof, in case the two sw-devices will ever need to support different features.

anangl (approved):

frkv:
LGTM. Pending zephyr testing

PR #463, opened 2021-02-16
[nrf fromtree] samples: boards: nrf: ieee802154: Optimize power on nRF53

pszkotak:
Add an overlay disabling unused peripherals
and UART in the prj.conf.

KRKNWK-8952

https://github.com/zephyrproject-rtos/zephyr/pull/32341

Signed-off-by: Piotr Szkotak <piotr.szkotak@nordicsemi.no>


lmaciejonczyk (approved):

hubertmis (changes requested):
I think using [fromlist] here is far-fetched. There are comments in the upsteram PR suggesting that this approach is not correct.

pszkotak:
> I think using [fromlist] here is far-fetched. There are comments in the upsteram PR suggesting that this approach is not correct.

Adding DNM till it's clarified. Will sync these two if there is a need for it.

MarekPorwisz (approved):

tejlmand:
also remember to point to the upstream PR.

tejlmand (approved):

ioannisg (changes requested):
pls, change this to from-tree, appropriately.

shantha-14 (approved):

ioannisg (approved):

PR #465, opened 2021-02-17
[nrf fromlist] net: openthread: separated variable for rx/tx operations results

edmont:
The use of a single error variable for RX and TX operations has shown the OpenThread code to be asserting in some conditions.

Assert happens here: https://github.com/zephyrproject-rtos/openthread/blob/zephyr/src/core/mac/sub_mac.cpp#L494

KRKNWK-8998


MarekPorwisz (approved):

rlubos (commented):
Looks good, but I'd be good to open an upstream PR rightaway and mark the commits as `[nrf fromlist]` with a reference to the upstream PR - it's easier to track changes to revert during upmerges then.

edmont:
> Looks good, but I'd be good to open an upstream PR rightaway and mark the commits as `[nrf fromlist]` with a reference to the upstream PR - it's easier to track changes to revert during upmerges then.

OK, done here: https://github.com/zephyrproject-rtos/zephyr/pull/32389

rlubos (approved):
Thank you!

lmaciejonczyk (approved):

shantha-14 (approved):

PR #466, opened 2021-02-17
net: add EAI_SOCKTYPE EAI_FAMILY DNS error codes

richardmccrae-nordicsemi:
Added EAI_SOCKTYPE and EAI_FAMILY translations from dns_resolve.h

Signed-off-by: Richard McCrae <richard.mccrae@nordicsemi.no>


richardmccrae-nordicsemi:
Will create new

PR #469, opened 2021-02-18
[nrf draft] twister: prototyping support for CMakePresets.json files.

tejlmand:
Prototyping.

This commit introducing support for CMakePresets.json files.

Supporting CMakePresets.json files allows for keeping a single file
to describe a sample or test, and at the same time introduce support
for CMake preset files.

CMake presets can be used to add describe a given project in more
details in a way that can also be used by IDEs.

Twister will only use the CMakePresets.json file if:
- No testcase.yaml or sample.yaml is used
- The CMakePresets.json contains a vendor::zephyr::twister section.

### DIFF
--- a/scripts/pylib/twister/twisterlib.py
+++ b/scripts/pylib/twister/twisterlib.py
@@ -1348,7 +1348,11 @@ class TwisterConfigParser:
         self.common = {}
 
     def load(self):
-        self.data = scl.yaml_load_verify(self.filename, self.schema)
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
It's interesting to see this prototype, but I don't think we want to try to merge this anytime soon, right? Moving twister to use this is out of scope for this effort IMO.

jfischer-no:
Can it be closed?

PR #471, opened 2021-02-22
[nrf fromtree] kernel/timeout: Schedule zero-time timeouts

carlescufi:
Fix an edge case that snuck in with the recent fix: if timeslicing is
enabled, the CPU's slice_ticks will be zero, and thus match a timeout
object's dticks value of zero, and thus get suppressed (because "we
already have a timeout scheduled for that") incorrectly.

Fixes #31789

Signed-off-by: Andy Ross <andrew.j.ross@intel.com>

### DIFF
--- a/kernel/timeout.c
+++ b/kernel/timeout.c
@@ -134,7 +134,8 @@ void z_add_timeout(struct _timeout *to, _timeout_func_t fn,
 			 */
 			int32_t next_time = next_timeout();
 
-			if (_current_cpu->slice_ticks != next_time) {
+			if (next_time == 0 ||
+			    _current_cpu->slice_ticks != next_time) {
 				z_clock_set_timeout(next_time, false);
 			}
 #else


ioannisg (approved):

PerMac (approved):
Copied from teams:
In the latest master I see new fails, already seen them in the upstream before:
https://github.com/zephyrproject-rtos/zephyr/issues/31822 They are probably introduced by:
https://github.com/nrfconnect/sdk-zephyr/commit/93f083b9afdc4d7fe0cd59c3f1a1bdb460eb75de
The error in the upstream is closed as it got fixed but the fixing commit was not known. I run git bisect with reversed logic and it seems that https://github.com/zephyrproject-rtos/zephyr/commit/544475d8a709f42926bd4059c3a1a4f78d023676 fixes the issue. All of the commits relate to time-slicing so I guess it would make sense

tejlmand (approved):

PerMac:
There is another issue connected with this. This commit also solves the issue of `nrf5340dk_nrf5340_cpuapp  tests/kernel/mem_protect/sys_sem/kernel.memory_protection.sys_sem` failing. What is interesting, that this test only failed when `ZEPHYR_TOOLCHAIN_VARIANT=gnuarmemb`. For sdk-zephyr 0.12.2 the test was passing even without the above fix. @tejlmand Is it something worth looking at?
For gnuarmemb, without this commit, tests/kernel/mem_protect/sys_sem/kernel.memory_protection.sys_sem is failing on nrf5340 because `test_sem_take_no_wait_fails` was hanging.

tejlmand:
> What is interesting, that this test only failed when ZEPHYR_TOOLCHAIN_VARIANT=gnuarmemb. For sdk-zephyr 0.12.2 the test was passing even without the above fix. @tejlmand Is it something worth looking at?

Definitely worth looking into. Initial feeling is that this could be related to changed behavior in newer gcc versions.
Which again means this is an area that requires close attention when introducing broader toolchain support.

PR #472, opened 2021-02-22
[nrf fromtree] twisterlib: checking for classname and testname

DatGizmo:
Currently duplicated testcases are only determined by the classname.
This results in testcases not beeing added to the twister.xml if the
testcases are from the same class (e.g. sample.yaml), but with different
names.

The changed check if the combination of classname and testname
already exists.

Signed-off-by: Sebastian Wezel <sebastian.wezel@nordicsemi.no>

### DIFF
--- a/scripts/pylib/twister/twisterlib.py
+++ b/scripts/pylib/twister/twisterlib.py
@@ -3428,7 +3428,7 @@ class TestSuite(DisablePyTestCollectionMixin):
                         classname = p + ":" + ".".join(instance.testcase.name.split(".")[:2])
 
                     # remove testcases that are being re-run from exiting reports
-                    for tc in eleTestsuite.findall(f'testcase/[@classname="{classname}"]'):
+                    for tc in eleTestsuite.findall(f'testcase/[@classname="{classname}"][@name="{instance.testcase.name}"]'):
                         eleTestsuite.remove(tc)
 
                     eleTestcase = ET.SubElement(eleTestsuite, 'testcase',


PerMac:
I think you might be looking at the wrong place. The twister.xml report was used to show only high-level results, only at the test suite level, without test cases. If you want to get a detailed report with test cases you should look into twister_report.xml.  Or run twister with `--platform-reports` which will generate in addition individual reports per platform.  
I guess the line you are changing was used as a simple solution for what is written in the comment above: "remove testcases that are being re-run from exiting reports" The problem is that twister nomenclature got mixed along the way. Testcase was used as a name for an instance from .yaml files, which is actually a test suite that can have multiple test cases. We are in the process of cleaning up the names of structures.
I guess the entry in twister.xml can be change, so it does not contain test case name (the last name after dot), and only give the test suite name.

PerMac (changes requested):
~~If my comment make sense, this PR is to be closed~~

PerMac (approved):

mbolivar-nordic (approved):

mbolivar-nordic:
@PerMac how are you going to track this upstream? Seems like we should at least file a bug based on the description.

DatGizmo:
I also created the same pull request in the upstream repo, but Thomas Stillwell ment I should also create a PR in the nordic version  of zephyr.
https://github.com/zephyrproject-rtos/zephyr/pull/32540#event-4365626786
I think it already got merged :)

mbolivar-nordic (changes requested):
OK, the sauce tag needs to be `[nrf fromtree]` then, and please include a reference to the upstream commit SHA in the downstream commit log.

DatGizmo:
@mbolivar-nordic just for clarification. I change the commit message tag from "[nrf tmphack]" to "[nrf fromtree]" and also include the commit SHA from the merge with the same change in the upstream repo?

DatGizmo:
I'm putting the link to the commit in upstream here for now.
https://github.com/zephyrproject-rtos/zephyr/commit/148dbdc9fefc53a58c1fe990ebfc9390cb4a09f6

mbolivar-nordic:
@PerMac do we need this as an out of tree commit, or can it wait until the next upmerge?

@DatGizmo responses to your questions:

> @mbolivar-nordic just for clarification. I change the commit message tag from "[nrf tmphack]" to "[nrf fromtree]" 

Yes. Right now, you've changed the pull request title to use `[nrf fromtree]`, but the commit itself still says `[nrf temphack]`. That needs to be changed, e.g. by using `git commit --amend` to change your local commit log, then force pushing into the branch on your fork, e.g. using `git push --force-with-lease`.

These sauce tags are how we categorize our downstream commits. Details in https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/dm_code_base.html#oss-repositories-downstream-project-history.

> and also include the commit SHA from the merge with the same change in the upstream repo?

I'm not exactly sure what you mean by "from the merge" but here are more details:

- "Upstream commit SHA" means "the commit SHA in the https://github.com/zephyrproject-rtos/zephyr repository's master branch which this is a copy of". In this case, that commit SHA is 148dbdc9fefc53a58c1fe990ebfc9390cb4a09f6.
- "Downstream commit log" means the commit log message for the commit being added in this pull request, the one which should have the `[nrf fromtree]` sauce tag.

I'm asking for the upstream commit SHA to be added to the downstream commit log, so adding a line line "Upstream commit: 148dbdc9fefc53a58c1fe990ebfc9390cb4a09f6" or similar to the commit message for this pull request, e.g. by using `git commit --amend` and `git push --force-with-lease`.

PerMac:
>do we need this as an out of tree commit, or can it wait until the next upmerge?
I don't use twister.xml report in any of my plans, the results are based on the more detailed version twister_report.xml and platform.xml. I guess sdk-nrf reporting is based on twister.xml and that's why @DatGizmo found that it is wrong and should be fixed?  I think the question should be to @DatGizmo: do you want to play and switch to twister_report.xml/platform.xml (it is used for results publishing in the upstream and has better maintenance) or use twister.xml and have it fixed now?

DatGizmo:
I know that at least sdk-nrf and sdk-zephyr are using twister.xml. Would need to check if more jobs are using that log.
The main different between twister and twister_report is, that the twister.xml also contains the error log if a test fails.
And from my testing, with my fix in place, the difference between the numbers for success, skipped and failed in both of these logs is, that the twister_report is containing more entries for skipped test.

thst-nordic:
I think we should stick with parsing twister.xml (because it provides the error log) 

I don't know the exact date of the next upmerge but IMO we should merge this since it is ready to go.

PerMac:
>The main different between twister and twister_report is, that the twister.xml also contains the error log if a test fails.
I see, you're right. The difference is that build logs are not dumped into the twister_report.xml in case of build failure, only the logs from execution. Since you're only building it would make sense to be based on twister.xml. And as for the different amount of results in reports I'll contact you on teams, as if this is the real issue then might be worth reporting to the upstream

thst-nordic (approved):

mbolivar-nordic (approved):
OK, sounds good. Is there an sdk-nrf PR updating the manifest so we can see the CI results before merging this?

DatGizmo:
I have a test setup using the fix in my test space on Jenkins (don't know if I should post links to internal jenkins here)

But to outline the difference:
Without the fix, you will only find one of the two testcase entries in the xml file (see below) as the check is only done on the class name.

`
<testcase classname="nrf51_vbluno51:bluetooth/central_bas/samples.bluetooth" name="bluetooth/central_bas/samples.bluetooth.central_bas" time="0.000000">
<skipped type="skipped" message="Skipped"/>
</testcase>
`
`
<testcase classname="nrf51_vbluno51:bluetooth/central_bas/samples.bluetooth" name="bluetooth/central_bas/samples.bluetooth.central_bas.build" time="0.000000">
<skipped type="skipped" message="Skipped"/>
</testcase>
`

PerMac:
I think Marti's point was to add PR to sdk-nrf updating the manifest there (pointing to sdk-zephyr with your fix) and to see the sdk-nrf CI passing on that PR. Will help you with this

DatGizmo:
Created a DNM PR for sdk-nrf: 
https://github.com/nrfconnect/sdk-nrf/pull/4051

mbolivar-nordic:
> Created a DNM PR for sdk-nrf:
> [nrfconnect/sdk-nrf#4051](https://github.com/nrfconnect/sdk-nrf/pull/4051)

Shouldn't be DNM, we want this merged, otherwise nobody who clones NCS will get the change.

mbolivar-nordic (changes requested):
Putting a -1 on this until the sdk-nrf PR is updated to follow the right procedure for getting things through CI.

I know this is a pain, but we're conservative about sdk-zephyr changes. Upmerges are hard enough already, so we have to test and track any out of tree patches carefully.

mbolivar-nordic (approved):

PR #489, opened 2021-03-26
 [nrf fromlist] runners: remove dependencies from runners

tejlmand:
Note, the upstream PR was originally taken in as a `fromlist` during the upmerge.
However, the upstream PR has been reworked because it only works with CMake >= 3.16.

Therefore the original `fromlist` has been reverted and a new `fromlist` been created to minimize conflicts on next upmerge.

----------

Today, there is a build target is added for each runner: flash, debug,
debugserver, attach.

And those runners will have a dependency to Zephyr logical target that
is built before invoking `west <runner>`.

This design has some flaws, mainly that additional dependencies directly
on the target will not be built when running `west <runner>` directly.
That generator expressions cannot be used for the DEPENDS argument.

Instead, the build target `<runner>` will not have any dependencies, and
will raise a build error if a dependency is added to the target.
Due to how `add_dependencies()` work, this must be done as a build time
check, and not configure time check.

`west <runner>` will invoke a build before executing the runner, and
this way ensure the build target is up-to-date, which again removes the
need for a dedicated `west_<runner>_target`.

It also minimizes the risk of developer errors, as developers no longer
need to consider the need for adding additional dependencies.
If a custom target is part of the default `all` build, then it's ensured
to be up-to-date.

Fixes: Issue reported on slack.

Upstream PR: zephyrproject-rtos/zephyr#33425

------

Manifest PR: https://github.com/nrfconnect/sdk-nrf/pull/4227

### DIFF
--- a/cmake/app/boilerplate.cmake
+++ b/cmake/app/boilerplate.cmake
@@ -636,8 +636,6 @@ endif()
 #
 # Currently used properties:
 # - COMPILES_OPTIONS: Used by application memory partition feature
-# - ${TARGET}_DEPENDENCIES: additional dependencies for targets that need them
-#   like flash (FLASH_DEPENDENCIES), debug (DEBUG_DEPENDENCIES), etc.
 add_custom_target(zephyr_property_target)
 
 # "app" is a CMake library containing all the application code and is
--- a/cmake/app/boilerplate.cmake
+++ b/cmake/app/boilerplate.cmake
@@ -636,6 +636,8 @@ endif()
 #
 # Currently used properties:
 # - COMPILES_OPTIONS: Used by application memory partition feature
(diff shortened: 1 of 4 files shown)


lemrey (approved):
Thank you, this fixes `ninja flash` with `cmake version 3.13.2`.

mbolivar-nordic (approved):

mbolivar-nordic:
Looks good, but CI is red in the manifest PR; any idea what is wrong?

tejlmand:
> Looks good, but CI is red in the manifest PR; any idea what is wrong?

Regular instability in CI :disappointed:

mbolivar-nordic:
> Regular instability in CI disappointed

OK, merging then

PR #492, opened 2021-03-30
[nrf fromtree] Support for low power mode in UARTE driver

nordic-krch:
From https://github.com/zephyrproject-rtos/zephyr/pull/30693 and https://github.com/zephyrproject-rtos/zephyr/pull/33708


nordic-krch:
fyi @jhn-nordic

jfischer-no (approved):

nordic-krch:
Upmerge went first.

PR #494, opened 2021-03-31
[nrf fromlist] Add initial Thingy53 support.

emob-nordic:
There is a thing I'm concerned about.
Setting `CONFIG_CONSOLE=y` in `*_defconfig` files even tho there is nothing connected to the UART by default.
I left this for now as its done similarly for Thingy52 but I'm open for suggestions.

My thinking is not to enable it by default, or enable it together with Segger RTT.

### DIFF
--- a/boards/arm/thingy53_nrf5340/board.c
+++ b/boards/arm/thingy53_nrf5340/board.c
@@ -1,0 +1,146 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+ */
+
+#include <init.h>
+#include <hal/nrf_gpio.h>
+#include <nrfx.h>
+#include <device.h>
+#include <drivers/gpio.h>
+
+#include <logging/log.h>
+#define LOG_LEVEL LOG_LEVEL_ERR
+LOG_MODULE_REGISTER(thingy53_board_init, LOG_LEVEL);
+
+#define NRF21540FEM_CTRL_LABEL		 DT_NODELABEL(nrf21540fem_ctrl)
+#define NRF21540FEM_CTRL_RX_EN_PIN	 DT_GPIO_PIN(NRF21540FEM_CTRL_LABEL, rx_en_gpios)
+#define NRF21540FEM_CTRL_MODE_PIN	 DT_GPIO_PIN(NRF21540FEM_CTRL_LABEL, mode_gpios)
+#define NRF21540FEM_CTRL_TX_EN_PIN	 DT_GPIO_PIN(NRF21540FEM_CTRL_LABEL, tx_en_gpios)
+#define NRF21540FEM_CTRL_PDN_GPIO_NODE   DT_GPIO_CTLR(NRF21540FEM_CTRL_LABEL, pdn_gpios)
+#define NRF21540FEM_CTRL_PDN_PIN	 DT_GPIO_PIN(NRF21540FEM_CTRL_LABEL, pdn_gpios)
+#define NRF21540FEM_CTRL_PDN_FLAGS	 DT_GPIO_FLAGS(NRF21540FEM_CTRL_LABEL, pdn_gpios)
+
+#define REG_3V3_LABEL			 DT_NODELABEL(regulator_3v3)
+#define REG_3V3_GPIO_NODE		 DT_GPIO_CTLR(REG_3V3_LABEL, enable_gpios)
+#define REG_3V3_ENABLE_PIN		 DT_GPIO_PIN(REG_3V3_LABEL, enable_gpios)
+#define REG_3V3_ENABLE_FLAGS		 DT_GPIO_FLAGS(REG_3V3_LABEL, enable_gpios)
+
+#define SENSOR_PWR_CTRL_LABEL		 DT_NODELABEL(sensor_pwr_ctrl)
+#define SENSOR_PWR_CTRL_GPIO_NODE	 DT_GPIO_CTLR(SENSOR_PWR_CTRL_LABEL, enable_gpios)
+#define SENSOR_PWR_CTRL_PIN		 DT_GPIO_PIN(SENSOR_PWR_CTRL_LABEL, enable_gpios)
+#define SENSOR_PWR_CTRL_FLAGS		 DT_GPIO_FLAGS(SENSOR_PWR_CTRL_LABEL, enable_gpios)
+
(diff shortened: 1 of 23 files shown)


emob-nordic:
@oyvindronningstad I just want You too look at the last commit.

kkasperczyk-no:
I verified that button, RGB led, BLE and NFC runs on Thingy:53 in Connected Home over IP sample. I also didn't have much problems with running logging via RTT. As we discussed on private chat, please move IPM section from `Kconfig.defconfig` to `Kconfig` to fix the building process. After longer reflection I think that RTT should not be enabled by default in the board files, but user should do it individually in the prj.conf files of specific application.

emob-nordic:
> I verified that button, RGB led, BLE and NFC runs on Thingy:53 in Connected Home over IP sample. I also didn't have much problems with running logging via RTT. As we discussed on private chat, please move IPM section from `Kconfig.defconfig` to `Kconfig` to fix the building process. After longer reflection I think that RTT should not be enabled by default in the board files, but user should do it individually in the prj.conf files of specific application.

Thanks for the input. I will put IPM section configs back to `Kconfig` instead of `Kconfig.defconfig`.

modules/trusted-firmware-m/Kconfig
```diff
@@ -11,6 +11,7 @@ config TFM_BOARD
 	string
 	default "nordic_nrf/nrf9160dk_nrf9160" if BOARD_NRF9160DK_NRF9160NS
 	default "nordic_nrf/nrf5340dk_nrf5340_cpuapp" if BOARD_NRF5340DK_NRF5340_CPUAPPNS
+	default "nordic_nrf/thingy53_nrf5340_cpuapp" if BOARD_THINGY53_NRF5340_CPUAPP
```
oyvindronningstad, line 14: This won't work as it is referencing platform folders inside the TFM repo. This should work automatically in sdk-nrf, though because of this:

https://github.com/nrfconnect/sdk-nrf/blob/master/modules/tfm/Kconfig#L10

oyvindronningstad (changes requested):

carlescufi (changes requested):
Why isn't this being sent upstream directly and then taken here as `[fromlist]` or `[fromtree]`?

emob-nordic:
> Why isn't this being sent upstream directly and then taken here as `[fromlist]` or `[fromtree]`?

I thought this is a compromise agreed offline (mails).

carlescufi:
@ioannisg and @anangl please review here first and then @emob-nordic will send a PR upstream and then rework this one as `[fromlist]`. I will keep my -1 to avoid merging by mistake.

carlescufi:
@jfischer-no would appreciate a review from you too.

modules/trusted-firmware-m/Kconfig
oyvindronningstad: What I mean is, you should remove the tfm commit. You will get tfm support automatically in ncs without having to add the board anywhere.

emob-nordic:
I've renamed the commit messaged to [nrf fromlist] + removed tfm commit (thanks @oyvindronningstad). I will post upstream PR once this is reviewed.

boards/arm/thingy53_nrf5340/board.c
```diff
@@ -0,0 +1,146 @@
...
+	if (!device_is_ready(gpio)) {
+		LOG_ERR("Gpio device not ready");
+		return -ENODEV;
+	}
+	err = gpio_pin_configure(gpio, SENSOR_PWR_CTRL_PIN,
+				 SENSOR_PWR_CTRL_FLAGS | GPIO_OUTPUT_ACTIVE);
+	if (err < 0) {
+		LOG_ERR("Failed to configure pin %d", SENSOR_PWR_CTRL_PIN);
+		return err;
+	}
```
geotitas, line 79: We have spotted that two sensors (BME688 and BMM150) fail to initialize with their respective Zephyr drivers in NCS v.1.5.0.
Those two drivers do not accommodate for the sensors' STARTUP time requirement (2ms and 1ms respectively according to the datasheets).
bme688_timing

Hence a delay is needed in the board.c file after SENS_PWR_CTRL is toggled. Otherwise we could add the delay to the drivers directly.

oyvindronningstad (approved):
Ok from me since TFM changes are removed.

emob-nordic:
@geotitas I've added a commit with start up delays, could you please test with that?

drivers/sensor/bme680/bme680.c
```diff
@@ -414,6 +414,8 @@ static int bme680_init(const struct device *dev)
 
 	data->i2c_slave_addr = DT_INST_REG_ADDR(0);
 
+	k_msleep(2); /* Start up delay */
```
pdunaj, line 417: just wondering is there is no better way to handle such delays
it will delay application start which in some cases may be critical issue

@anangl is there any async model for drivers bringup now? when we done motion sensors I found none such thing soin the end we return ebusy from api calls until driver is up (application tracks that).

geotitas:
> @geotitas I've added a commit with start up delays, could you please test with that?

Works as expected.

pdunaj:
@carlescufi who else should have a look before we migrate the PR?

jfischer-no (commented):
Please move this PR upstream.

boards/arm/thingy53_nrf5340/board.c
```diff
@@ -0,0 +1,146 @@
...
+
+	const struct device *gpio;
+	int err;
+
+	gpio = DEVICE_DT_GET(REG_3V3_GPIO_NODE);
+	if (!device_is_ready(gpio)) {
+		LOG_ERR("Gpio device not ready");
+		return -ENODEV;
+	}
+	err = gpio_pin_configure(gpio, REG_3V3_ENABLE_PIN,
```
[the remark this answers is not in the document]
emob-nordic: With the PR you posted just before review ([#34232](https://github.com/zephyrproject-rtos/zephyr/pull/34232)) it will, but currently no. I wanted to avoid the overhead the regulator driver gives as this is just pin switching :P

Good you pointed it out as i was thinking if i should enable `CONFIG_REGULATOR` and `CONFIG_REGULATOR_FIXED` by default and make it automatic, at that point i didn't spot the issue within the `regulator/Kconfig.fixed`. With your PR it has to be done that way :)

P.S.
tbh it's a bit overuse for that purpose but i assume its valuable for the future.

emob-nordic:
Closing as I will post this PR upstream with regulators fixed.

PR #500, opened 2021-04-12
Pin breathe rev to avoid issues with documentation build

carlescufi:


tejlmand (approved):

gmarull (approved):
Thanks for taking care @carlescufi

thst-nordic (approved):

scripts/requirements-doc.txt
```diff
@@ -1,9 +1,8 @@
 # DOC: used to generate docs
 
-breathe>=4.23.0
-docutils>=0.16
+breathe>=4.23.0,<4.29.0
```
mbolivar-nordic, line 3:
```suggestion
breathe>=4.23.0,!=4.29.0
```

mbolivar-nordic (commented):
github is making me put something here

scripts/requirements-doc.txt
carlescufi: That assumes that 4.30 will fix the issue, but that is not certain at this point? Shouldn't we be more conservative? Anyway this is a fromtree so I rather not change it, and let @gmarull deal with this upstream and then get it as part of an upmerge

gmarull: As @mbolivar-nordic pointed upstream a lock file would allow us to be more permissive when it comes to version specifiers and still have a stable CI. Unless we have something like that in place, I'd tend to be more conservative. For example, block major versions until they get verified (e.g. `>0.5,<1.0`).

gmarull: I didn't want to add `!=4.29.0` for breathe because I do not trust its maintenance, it's a personal feeling though. I reported the bug, and it seems to be affecting more people now. So I hope 4.30.0 will fix the problem...

gmarull: I guess @mbolivar-nordic is more in favor of https://en.wikipedia.org/wiki/Linus%27s_law :smile: 

> The law was formulated by Eric S. Raymond in his essay and book The Cathedral and the Bazaar (1999), and was named in honor of Linus Torvalds.[1][2]

> A more formal statement is: "Given a large enough beta-tester and co-developer base, almost every problem will be characterized quickly and the fix obvious to someone." Presenting the code to multiple developers with the purpose of reaching consensus about its acceptance is a simple form of software reviewing. Researchers and practitioners have repeatedly shown the effectiveness of reviewing processes in finding bugs and security issues.[3]

It makes me think it's not a bad idea after all.

mbolivar-nordic: I'm just providing a comment, neither blocking nor approving. I've made my perspective clear and you can do what you think is best here.

PR #504, opened 2021-04-19
[nrf toup] net: openthread: add CSL transmitter functionality

edmont:
Draft PR for discussion about the implementation.


edmont:
The actual transmission security processing is missing in this PR. Current vanilla OT implementation relies greatly on OT API for frame handling: https://github.com/openthread/openthread/blob/main/examples/platforms/nrf528xx/src/radio.c#L255-L302

It seems this should be implemented in a different manner within the IEEE 802.15.4 driver.

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -878,3 +901,46 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
...
+    ARG_UNUSED(aInstance);
+
+    struct ieee802154_config config = {
+		.frame_counter = aMacFrameCounter
+	};
+
+	(void)radio_api->configure(radio_dev, IEEE802154_CONFIG_FRAME_COUNTER,
+							   &config);
+}
+#endif // CONFIG_OPENTHREAD_THREAD_VERSION_1_2
```
rlubos, line 946: /**/  Please

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -878,3 +901,46 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
...
+		.mac_keys.prev_key = aPrevKey->m8,
+		.mac_keys.curr_key = aCurrKey->m8,
+		.mac_keys.next_key = aNextKey->m8,
+	};
+
+	(void)radio_api->configure(radio_dev, IEEE802154_CONFIG_MAC_KEYS, &config);
+}
+
+void otPlatRadioSetMacFrameCounter(otInstance *aInstance,
+								   uint32_t aMacFrameCounter)
```
rlubos, line 935: This line seems to be misaligned.

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -878,3 +901,46 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
...
+{
+    ARG_UNUSED(aInstance);
+
+    struct ieee802154_config config = {
+		.frame_counter = aMacFrameCounter
+	};
+
+	(void)radio_api->configure(radio_dev, IEEE802154_CONFIG_FRAME_COUNTER,
+							   &config);
```
rlubos, line 944: Alignment

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -878,3 +901,46 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
...
+		.mac_keys.next_key = aNextKey->m8,
+	};
+
+	(void)radio_api->configure(radio_dev, IEEE802154_CONFIG_MAC_KEYS, &config);
+}
+
+void otPlatRadioSetMacFrameCounter(otInstance *aInstance,
+								   uint32_t aMacFrameCounter)
+{
+    ARG_UNUSED(aInstance);
```
rlubos, line 937: Tab please.

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -878,3 +901,46 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
...
+
+    struct ieee802154_config config = {
```
rlubos, line 939: Tab please.

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -878,3 +901,46 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
...
+
+#if defined(CONFIG_OPENTHREAD_THREAD_VERSION_1_2)
+uint64_t otPlatRadioGetNow(otInstance *aInstance)
+{
+	ARG_UNUSED(aInstance);
+
+	return radio_api->get_time(radio_dev);
+}
+
+void otPlatRadioSetMacKey(otInstance *    aInstance,
```
rlubos, line 913:
```suggestion
void otPlatRadioSetMacKey(otInstance     *aInstance,
```

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -878,3 +901,46 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
...
+                          uint8_t         aKeyIdMode,
+                          uint8_t         aKeyId,
+                          const otMacKey *aPrevKey,
+                          const otMacKey *aCurrKey,
+                          const otMacKey *aNextKey)
+{
+	ARG_UNUSED(aInstance);
+    __ASSERT_NO_MSG(aPrevKey != NULL && aCurrKey != NULL && aNextKey != NULL);
```
rlubos, line 921: Tab here and below.

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -257,6 +261,19 @@ void transmit_message(struct k_work *tx_job)
 	radio_api->set_channel(radio_dev, sTransmitFrame.mChannel);
 	radio_api->set_txpower(radio_dev, tx_power);
 
+#if defined(CONFIG_OPENTHREAD_THREAD_VERSION_1_2)
+    if (sTransmitFrame.mInfo.mTxInfo.mTxDelay != 0)
```
rlubos, line 265: Tab here and below.

subsys/net/lib/openthread/platform/radio.c
```diff
@@ -257,6 +261,19 @@ void transmit_message(struct k_work *tx_job)
...
+    {
+		uint64_t tx_at = sTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime +
+						 sTransmitFrame.mInfo.mTxInfo.mTxDelay;
+		net_pkt_set_txtime(tx_pkt, NSEC_PER_USEC * tx_at);
+		if (radio_api->tx(radio_dev, IEEE802154_TX_MODE_TXTIME_CCA, tx_pkt,
+						  tx_payload)) {
```
rlubos, line 271: Alignment

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -836,6 +861,16 @@ void nrf_802154_serialization_error(const nrf_802154_ser_err_data_t *p_err)
 }
 #endif
 
+#if defined(CONFIG_OPENTHREAD_THREAD_VERSION_1_2)
+static K_MUTEX_DEFINE(sec_mutex);
+
+static uint8_t     key_id;
+static uint8_t[16] prev_key;
```
rlubos, line 868: Would be good to add some #define for the key size.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -690,6 +697,21 @@ static int nrf5_configure(const struct device *dev,
 	case IEEE802154_CONFIG_EVENT_HANDLER:
 		nrf5_data.event_handler = config->event_handler;
 
+#if defined(CONFIG_OPENTHREAD_THREAD_VERSION_1_2)
```
rlubos, line 700: I think it would make sense to introduce a new Kconfig at the shim-layer level, enabled by default if OpenThread 1.2 is used.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -836,6 +861,16 @@ void nrf_802154_serialization_error(const nrf_802154_ser_err_data_t *p_err)
 }
 #endif
 
+#if defined(CONFIG_OPENTHREAD_THREAD_VERSION_1_2)
+static K_MUTEX_DEFINE(sec_mutex);
```
rlubos, line 865: I think all these variables should rather belong to `struct nrf5_802154_data`.

drivers/ieee802154/ieee802154_nrf5.c
edmont: Added `NET_PKT_TXSEC`, but probably it should go somewhere else.

rlubos: I was actually thinking to add it [here](https://github.com/nrfconnect/sdk-zephyr/blob/master/drivers/ieee802154/Kconfig.nrf5). It's more of a driver feature than net_pkt one.

edmont: Given that it's used in the platform implementation (`radio.c`) I guess it would be better in https://github.com/nrfconnect/sdk-zephyr/blob/master/drivers/ieee802154/Kconfig.

rlubos: For the `radio.c` it's fine to use `CONFIG_OPENTHREAD_THREAD_VERSION_1_2` since it's a part of the OT implementation. I was thinking of introducing a separate config just for the radio driver (which could default to `y` if OT 1.2 is used). In case it's disabled in the radio driver, the configuration would simply fail. I see though the error is ignored right now and since it cannot be propagated it would make sense IMO to add an assert in `radio.c` in case the configuration fails. This would also be a clear indication for the other drivers that they need to implement this feature in order to use OT 1.2.

edmont: Actually now I think the new Kconfig to be introduced by Radio Driver to support this functionality should be used here.

PR #505, opened 2021-04-21
[nrf fromlist] soc: nrf52805: fix enabled instance 0 check

mbolivar-nordic:
Upstream:
https://github.com/zephyrproject-rtos/zephyr/pull/34433/

UART0, TWI0, and SPI0 have different peripheral IDs (2, 3, and 4
respectively) on this SoC and therefore should not be mutually
exclusive to use.


anangl (approved):

tejlmand (approved):

ioannisg:
These can be from-tree, now

ioannisg (changes requested):
from-tree, otherwise, fine.

mbolivar-nordic:
brought in by the current upmerge

PR #507, opened 2021-04-21
Upmerge 21.04.2021

trond-snekvik:
Merges in upstream up to commit [8e1cfe9](https://github.com/zephyrproject-rtos/zephyr/commit/8e1cfe9a46844041544d1efcc55c59d069bd2432).


mbolivar-nordic (approved):

boards/arm/mps2_an521/Kconfig.defconfig
```diff
@@ -17,10 +17,8 @@ config BOARD
 
 # By default, if we build for a Non-Secure version of the board,
 # force building with TF-M as the Secure Execution Environment.
-# Openamp samples assume the non-secure build is the remote
-# core image, so do not build with TF-M in this case.
 config BUILD_WITH_TFM
-	default y if TRUSTED_EXECUTION_NONSECURE && !OPENAMP
+	default y if TRUSTED_EXECUTION_NONSECURE
```
ioannisg, line 21: So, for now, I think this is fine, however, I believe CI won't pass, because it will try to build TFM for mps2_an521_nonsecure , in the context of the openamp samples.

trond-snekvik: AFAICT, OPENAMP is being used as a heuristic here to determine whether this is a nonsecure app that talks to a secure app. Is this correct, and if it is, is there some other way to solve this?

oyvindronningstad: This is now an upstream patch (e04a57de046e0e1fba25d3ecc11bd6d7611e3664). It's part of a set of 4, but the other 3 should only be relevant for users of the mps2_an521 board. PR: https://github.com/zephyrproject-rtos/zephyr/pull/34576. cc @ioannisg

carlescufi (approved):

PR #508, opened 2021-04-26
TF-M integration branch

frkv:
Draft PR giving temporary fixes to items to integrate nrf_security and
PSA Crypto APIs with Zephyr + TF-M project

ref: NCSDK-7907


samples/net/sockets/echo_server/prj.conf
```diff
@@ -52,3 +52,13 @@ CONFIG_POSIX_MAX_FDS=12
 
 # How many client can connect to echo-server simultaneously
 CONFIG_NET_SAMPLE_NUM_HANDLERS=1
+
+CONFIG_BUILD_WITH_TFM=y
```
ioannisg, line 56: Better add a sample variant in sample.yaml, which enables the extra config CONFIG_BUILD_WITH_TFM=y
This should be a temp-hack.

tejlmand (approved):

ioannisg:
@rlubos can you look at the temphack commits, please?

samples/net/sockets/echo_server/prj.conf
ioannisg: What will we finally do here, @frkv ? Will you force BUILD_WITH_TFM by default, or somehow introduce a Kconfig "overlay"?

tejlmand: I tried to do what we discussed at first, but we need to make a dedicated Kconfig fragment that must be manually applied, cause there are Kconfig symbols unknown without the corresponding nrfxlib patches.
```
/projects/github/ncs/zephyr/samples/net/sockets/echo_server/prj.conf:58: warning: attempt to assign the value 'y' to the undefined symbol MBEDTLS_PSA_CRYPTO_C

error: Aborting due to Kconfig warnings
```

Also removed the dts overlay.

PR #519, opened 2021-05-12
[nrf noup] net: sockets: Add new DTLS and RAI socket options 

rlubos:
Introduce new nRF91 specific TLS and RAI socket options. 

Additionally, rework how local extensions to the networking headers are handled downstream. Instead of modifying the upstream socket.h header, introduce a new one, specifically for NCS extensions, and keep all of the local changes in there. The only change needed in the original header is an inclusion of the new header. This should help to keep the local changes in order and facilitate the future upmerge process.

### DIFF
--- a/include/net/net_ip.h
+++ b/include/net/net_ip.h
@@ -47,7 +47,6 @@ extern "C" {
 #define PF_NET_MGMT     5          /**< Network management info.      */
 #define PF_LOCAL        6          /**< Inter-process communication   */
 #define PF_UNIX         PF_LOCAL   /**< Inter-process communication   */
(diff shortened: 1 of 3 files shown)


lemrey (approved):
LGTM, one thing is that some of the relative `nrf` symbols we have here are deprecated, for example everything related to the PDN socket. I wonder if it would make sense to deprecate them here too?

lemrey:
I would like to get this in because we will introduce a new socket option with `libmodem` v1.2.0 which we'd want to keep in the separate file that is being created in this PR.

rlubos:
We need to wait for the upmerge...

PR #521, opened 2021-05-19
[nrf toup] modules: tfm: add missing CMAKE_ARGS multi value arg

hakonfam:
CMAKE_ARGS is documented and used, but not parsed.


ioannisg:
> Already in place upstream

Yeap, it needs to be cherry-picked, btw, as it was merged only yesterday

PR #523, opened 2021-05-19
coredump_log: coredump log in panic mode

RobertGalatNordic:
Current implementation assumes that CONFIG_LOG_MODE_IMMEDIATE=y guarantees the complete transfer, and it is not true.
In my opinion core dump should always be printed in panic mode.

Signed-off-by: Robert Gałat <robert.galat@nordicsemi.no>

### DIFF
--- a/subsys/debug/coredump/coredump_backend_logging.c
+++ b/subsys/debug/coredump/coredump_backend_logging.c
@@ -30,9 +30,7 @@ static void coredump_logging_backend_start(void)
 	/* Reset error */
 	error = 0;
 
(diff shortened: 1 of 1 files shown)


RobertGalatNordic:
I had CONFIG_LOG_MODE_IMMEDIATE set to 'y', and still got issues with getting a complete core dump. Therefore, I think that core dump should always be printed in panic mode

RobertGalatNordic:
moved to upstream https://github.com/zephyrproject-rtos/zephyr/pull/35598

PR #524, opened 2021-05-24
[nrf noup] Adapt ieee802154_nrf5 driver to TFM

Damian-Nordic:
* When CONFIG_BUILD_WITH_TFM=y, use tfm_platform_mem_read() to read EUI64 from the non-secure world.
* Include secure_services.h header only if SPM=y.

Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -39,9 +39,12 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 
 #include <net/ieee802154_radio.h>
 
-#if defined(CONFIG_SOC_NRF5340_CPUAPP) && \
-	defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+#if defined(CONFIG_SOC_NRF5340_CPUAPP) && defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+#if defined(CONFIG_SPM_SERVICE_READ)
 #include <secure_services.h>
+#elif defined(CONFIG_BUILD_WITH_TFM)
+#include <tfm/tfm_ioctl_api.h>
+#endif
 #endif
 
 #include "ieee802154_nrf5.h"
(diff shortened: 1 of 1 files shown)


drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -118,13 +121,22 @@ static void nrf5_get_eui64(uint8_t *mac)
 	mac[index++] = IEEE802154_NRF5_VENDOR_OUI & 0xff;
 #endif
 
-#if defined(CONFIG_SOC_NRF5340_CPUAPP) && \
-	defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+#if defined(CONFIG_SOC_NRF5340_CPUAPP) && defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
 	int ret = -EPERM;
```
frkv, line 125: Is -EPERM a good enough error code for all possible tfm_platform_mem_read error codes?

Damian-Nordic: Right, I guess we can map TFM_PLATFORM_ERR_INVALID_PARAM to -EINVAL.

frkv: Yes. This is the most important error code to provide to the user.

czeslawmakarski (approved):

drivers/ieee802154/ieee802154_nrf5.c
Damian-Nordic: Done.

frkv (approved):
LGTM

ioannisg:
@rlubos I'll take a look into this after the tfm integration review. Is this needed for 1.6.0?

rlubos:
> Is this needed for 1.6.0?

@Damian-Nordic ?

Damian-Nordic:
@ioannisg @rlubos Although we won't manage to switch OpenThread to the PSA crypto I think it would be nice to be able to at least build and run OpenThread samples with TF-M and that is impossible without adding this change. Therefore, we would like to have it merged to 1.6

ioannisg:
> @ioannisg @rlubos Although we won't manage to switch OpenThread to the PSA crypto I think it would be nice to be able to at least build and run OpenThread samples with TF-M and that is impossible without adding this change. Therefore, we would like to have it merged to 1.6

Ack

ioannisg:
Please add  a milestone, so it is at least in the radar for 1.6

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -39,9 +39,12 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 
 #include <net/ieee802154_radio.h>
 
-#if defined(CONFIG_SOC_NRF5340_CPUAPP) && \
-	defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+#if defined(CONFIG_SOC_NRF5340_CPUAPP) && defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+#if defined(CONFIG_SPM_SERVICE_READ)
```
hakonfam, line 43: Would it make sense to check `CONFIG_SPM` instead? As that is the equivalent of `CONFIG_BUILD_WITH_TFM`

hakonfam (approved):

drivers/ieee802154/ieee802154_nrf5.c
Damian-Nordic: On the other hand, `spm_request_read` is only available when `CONFIG_SPM_SERVICE_READ` so I thought it would be better to depend on the setting which directly controls the functionality. I can change, however, if it's against our rules.

hakonfam: Makes sense, no strong opinion :)

ioannisg (approved):
Looks ok to me, but would like @oyvindronningstad to review as well

oyvindronningstad (approved):

tejlmand:
Included in: #549

PR #518, opened 2021-05-12
Upmerge 12 V 2021

nvlsianpu:
Regular upmerge to v2.6.0 tag

PR #528, opened 2021-05-28
[nrf fromtree] Bluetooth: Mesh: Pass correct pointer to publish_sent

michal-narajowski:
This commit fixes a bug where incorrect pointer passed to publish_sent
in access.c caused bus fault.

Signed-off-by: Pavel Vasilyev <pavel.vasilyev@nordicsemi.no>

### DIFF
--- a/subsys/bluetooth/mesh/access.c
+++ b/subsys/bluetooth/mesh/access.c
@@ -215,7 +215,7 @@ static int pub_period_start(struct bt_mesh_model_pub *pub)
 		BT_DBG("Update failed, skipping publish (err: %d)", err);
 		pub->count = 0;
 		pub->period_start = k_uptime_get_32();
-		publish_sent(err, pub);
+		publish_sent(err, pub->mod);
 		return err;
 	}
 


joerchan:
Why is this a noup? Shouldn't it be fixed there as well?

michal-narajowski:
@joerchan That's my bad, I just checked if this issue exists upstream and didn't check if it was fixed there.

joerchan:
@michal-narajowski Thanks, that makes more sense. We like to keep references to commits (fromtree) or pull-requests (fromlist) by having references to what is being cherry-picked. That makes the upmerge procedure much easier.

So please add "cherry-picked from 8717a0f6787421c0adb566a6b20495a34f76f1d5" to your commit message.
FYI you can do `git cherry-pick -s -x <sha>` to do this automatically.

Also since there is an upmerge procedure already in progress you might want to ask @nvlsianpu if he can cherry-pick it into that for you instead.

trond-snekvik:
Included in upmerge (#518)

PR #533, opened 2021-06-01
openthread: Enable support for CSL and Link metrics in NCS

lmaciejonczyk:


nvlsianpu (approved):

edmont (approved):

MarekPorwisz (approved):

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -61,7 +62,12 @@ static struct nrf5_802154_data nrf5_data;
 #define ACK_REQUEST_BIT (1 << 5)
 #define FRAME_PENDING_BYTE 1
 #define FRAME_PENDING_BIT (1 << 4)
-#define TXTIME_OFFSET_US  (5 * USEC_PER_MSEC)
+#define TXTIME_OFFSET_US (1 * USEC_PER_MSEC)
+
+#define DRX_SLOT_PH 0
```
hubertmis, line 67: nit: Short comment would be nice that this one is a slot for placeholder, while the other one is for actual RX window. I know how it works, but for someone getting familiar with this mechanism it would be easier to understand meaning of `PH` and `RX`.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -61,7 +62,12 @@ static struct nrf5_802154_data nrf5_data;
...
+#define DRX_SLOT_RX 1
+#define PH_DURATION 10
+#define XTAL_ACCURACY 20
```
hubertmis, line 70: This value should be configurable. I think it should be taken from devicetree. Alternatively, kconfig could be used.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -577,7 +606,16 @@ static int nrf5_start(const struct device *dev)
 
 static int nrf5_stop(const struct device *dev)
 {
+#if defined(CONFIG_OPENTHREAD_CSL_RECEIVER)
+	if (nrf_802154_sleep_if_idle() != NRF_802154_SLEEP_ERROR_NONE) {
+		if (nrf5_data.event_handler) {
```
hubertmis, line 611: What would happen if there is no `event_handler` set? I think `else` block should be somehow implemented.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -650,6 +688,69 @@ static void nrf5_iface_init(struct net_if *iface)
...
+	nrf_802154_key_t key;
+	uint8_t key_id_to_remove;
+
+	__ASSERT(mac_keys, "Invalid argument.");
+
+	/* Remove old invalid key assuming that its index is first_valid_key_id - 1.
+	 * TODO: This is Thread specific assumption, need to be changed when RD will provided
+	 * API for removing all keys or handling this internally.
+	 */
+	key_id_to_remove = (*mac_keys).key_index == 1 ? 0x80 : (*mac_keys).key_index - 1;
```
hubertmis, line 704: Throughout this function there is `(*mac_keys).sth`. Customary in C we use `mac_keys->sth`. It would be easier to follow the code if `->` was used instead.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -650,6 +688,69 @@ static void nrf5_iface_init(struct net_if *iface)
...
+		err = nrf_802154_security_key_store(&key);
+		__ASSERT(err == NRF_802154_SECURITY_ERROR_NONE ||
+				 err == NRF_802154_SECURITY_ERROR_ALREADY_PRESENT,
+			 "Storing key failed, err: %d", err);
+		mac_keys++;
+	};
+}
+#endif /* CONFIG_NRF_802154_ENCRYPTION */
+
+#if defined(CONFIG_OPENTHREAD_CSL_RECEIVER)
```
hubertmis, line 729: This `#if` - `#endif` block seems OpenThread specific. I think it should be replaced with something more generic like simple `nrf_802154_receive_at` abstraction and move most of this code to OpenThread radio platform implementation.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -712,6 +813,50 @@ static int nrf5_configure(const struct device *dev,
...
+#if defined(CONFIG_NRF_802154_ENCRYPTION)
+	case IEEE802154_CONFIG_MAC_KEYS:
+		nrf5_config_mac_keys(config->mac_keys);
+		break;
+
+	case IEEE802154_CONFIG_FRAME_COUNTER:
+		nrf_802154_security_global_frame_counter_set(config->frame_counter);
+		break;
+
+	case IEEE802154_CONFIG_ENH_ACK_HEADER_IE: {
```
hubertmis, line 827: This `case` should not be placed inside `#if defined(CONFIG_NRF_802154_ENCRYPTION)`

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -712,6 +813,50 @@ static int nrf5_configure(const struct device *dev,
...
+		uint8_t short_addr_le[SHORT_ADDRESS_SIZE];
+		uint8_t ext_addr_le[EXTENDED_ADDRESS_SIZE];
+
+		sys_put_le16(config->ack_ie.short_addr, short_addr_le);
+		sys_memcpy_swap(ext_addr_le, config->ack_ie.ext_addr, EXTENDED_ADDRESS_SIZE);
```
hubertmis, line 832: Why `_swap`? If it is caused by endianness I think both BE and LE should be handled here. Or at least a comment that you expect `config->ack_ie.ext_addr` in big (little?) endian format.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -650,6 +688,69 @@ static void nrf5_iface_init(struct net_if *iface)
...
+	 * API for removing all keys or handling this internally.
+	 */
+	key_id_to_remove = (*mac_keys).key_index == 1 ? 0x80 : (*mac_keys).key_index - 1;
+
+	key.id.mode = (*mac_keys).key_id_mode;
+	key.id.p_key_id = &key_id_to_remove;
+
+	nrf_802154_security_key_remove(&key.id);
+
+	while ((*mac_keys).key_value != NULL) {
```
hubertmis, line 711: I think `for (struct ieee802154_key *keys = mac_keys; keys->key_value; keys++)` would clearer show the intention to iterate over all values in `mac_keys` array.

hubertmis (approved):
I would modify market places in this PR. But I'm OK to keep it for now as it is, take technical debt and pay it later.

include/net/ieee802154_radio.h
```diff
@@ -171,8 +172,8 @@ enum ieee802154_config_type {
 	/** Configure a radio reception slot */
 	IEEE802154_CONFIG_RX_SLOT,
 
-	/** Enable CSL receiver (Endpoint) */
-	IEEE802154_CONFIG_CSL_RECEIVER,
+	/** Configure CSL receiver (Endpoint) period */
+	IEEE802154_CONFIG_CSL_PERIOD,
```
rlubos, line 176: Does that mean there's no runtime configuration for enabling/disabling CLS? Just a compile-time switch?

rlubos (approved):
Looks good, but please don't hesitate with submitting `[nrf toup]` patches upstream.

include/net/ieee802154_radio.h
hubertmis: In current API design, radio drivers do not support CSL. They just expose function that enable higher layers to implement CSL. Because of this, it is not possible to enable/disable CSL runtime.

If higher layer would like to disable CSL, it should just stop calling `config(IEEE802154_CONFIG_CSL_RX_TIME, ...)` and reset Header IEs in Enh-Acks.

lmaciejonczyk: From what I understand, in the case of OpenThread, if we build with the CSL support we can turn it on/off in runtime.

edmont: From IEEE 802.15.4-2-15:

> The CSL mode is turned on when the PIB attribute macCslPeriod is nonzero and turned off when macCslPeriod is zero.

hubertmis: > From IEEE 802.15.4-2015:
> 
> > The CSL mode is turned on when the PIB attribute macCslPeriod is nonzero and turned off when macCslPeriod is zero.

But it does not work for this API. If radio drivers fully supported CSL, you would need this simple function to set `IEEE802154_CONFIG_CSL_PERIOD` to enable or disable CSL. However, in current design, most of CSL is implemented in a higher (MAC) layer that must keep setting `IEEE802154_CONFIG_RX_SLOT`, `IEEE802154_CONFIG_CSL_RX_TIME` sequentially to keep CSL active.

rlubos: > But it does not work for this API. If radio drivers fully supported CSL, you would need this simple function to set IEEE802154_CONFIG_CSL_PERIOD to enable or disable CSL.

I think I'm understanding even less about the feature now. What about the `nrf5_config_csl_period()` implementation then, it seems to call `nrf5_receive_at()` underneath - what's the point if the upper layer is supposed to call `IEEE802154_CONFIG_CSL_RX_TIME `/`IEEE802154_CONFIG_RX_SLOT` anyway? I think we need better documentation for these configuration options given that we're confused about the usage even internally.

edmont: @rlubos do you think this point should be better explained?

```
/* A placeholder reception window is scheduled so that the radio driver is able to inject
 * the proper CSL Phase in the transmitted CSL Information Elements.
 */
```

rlubos: @edmont I wonder if `IEEE802154_CONFIG_CSL_PERIOD` should have a side effect of setting an RX slot, or whether the user should rather call `IEEE802154_CONFIG_CSL_PERIOD`/`IEEE802154_CONFIG_RX_SLOT` slot combo. 

But if it indeed should have configure the RX slot as well (I assume it's a single shot, as from the discussion above it seems the radio driver will not reschedule the next slot) this information should be at very least documented in the `ieee802154_radio.h` option descripiton, not hidden inside Nordic implementation.

hubertmis: I agree we should document it better. Right now implementing CSL in higher layer is enigmatic.

Basically, to implement CSL higher layer must:

1. Set `IEEE802154_CONFIG_CSL_PERIOD`
2. Set following (not the closes one) CSL window time (`IEEE802154_CONFIG_CSL_RX_TIME`)
3. Schedule the closest CSL window (set `IEEE802154_CONFIG_RX_SLOT`)
4. After the closest CSL window is finished, go to step 2

When period is changed, the higher layer shall:

1. Set new `IEEE802154_CONFIG_CSL_PERIOD`
2. Continue iteration from step 2 above

hubertmis: From a driver implementation point of view:

`IEEE802154_CONFIG_CSL_PERIOD` is needed to set period in CSL IE
`IEEE802154_CONFIG_CSL_RX_TIME` is needed to set phase in CSL IE
`IEEE802154_CONFIG_RX_SLOT` is needed to schedule RX window precisely

rlubos: Thanks for explaination. Shouldn't `IEEE802154_CONFIG_CSL_RX_TIME` be confgured to some value before `IEEE802154_CONFIG_CSL_PERIOD` though, given it makes use `nrf5_data.csl_rx_time`?

Anyway, I am aware we're tight on schedule, so don't bother with the documentation now if the solution works for you. But we should really consider improving it when submitting the patches upstream.

hubertmis: > Shouldn't IEEE802154_CONFIG_CSL_RX_TIME be confgured to some value before IEEE802154_CONFIG_CSL_PERIOD though, given it makes use nrf5_data.csl_rx_time?

Right. Fixing in the comment above.

ioannisg:
PArt of #546

PR #535, opened 2021-06-01
[nrf fromlist] tfm: Put saving of FPU context into its own file so ...

oyvindronningstad:
... it can be reused

Also, to ease readability.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/35846


ioannisg (approved):
I've reviewed the upstream PR, so +1 here, since I guess it's a cherry-pick.

arch/Kconfig
```diff
@@ -830,6 +830,7 @@ menu "Floating Point Options"
 config FPU
 	bool "Enable floating point unit (FPU)"
 	depends on CPU_HAS_FPU
+	depends on !IS_SPM
```
ioannisg, line 833: hm, this is really not needed, IMHO.

ioannisg (changes requested):
Honestly, I think this is something we should block in upstream as well. I'll send a patch that you can review @oyvindronningstad .

ioannisg (approved):

shantha-14 (approved):

tejlmand (approved):

PR #536, opened 2021-06-01
[nrf toup] Override the TF-M MBEDCRYPTO_PATH

ioannisg:
The TF-M build is passed a path to the mbedtls project
directory, however, NCS has its own mbedtls variant. When
building with TF-M we use a generator expression to allow
setting the path to mbedtls from nrf_security.

Signed-off-by: Ioannis Glaropoulos <Ioannis.Glaropoulos@nordicsemi.no>

### DIFF
--- a/modules/trusted-firmware-m/CMakeLists.txt
+++ b/modules/trusted-firmware-m/CMakeLists.txt
@@ -183,7 +183,7 @@ function(trusted_firmware_build)
       ${PSA_TEST_ARG}
       ${TFM_CMAKE_ARGS}
       -DTFM_TEST_REPO_PATH=${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/tf-m-tests
-      -DMBEDCRYPTO_PATH=${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/../../crypto/mbedtls/mbedtls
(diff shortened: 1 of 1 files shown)


frkv (approved):
LGTM

oyvindronningstad (approved):

tejlmand (approved):

ioannisg:
Will change this to "toup", as it looks we might want to support this upstream, i.e. a configurable mbedcrypto path.

frkv:
> Will change this to "toup", as it looks we might want to support this upstream, i.e. a configurable mbedcrypto path.

I would like to have a cleaner thing that this. It is very generator-expression heavy as it stands right now... This was the reason for the nrf_temphack tag. @tejlmand: We discussed splitting it up into more statements to make it more readable. Is that possible to do quickly?

magnev (approved):

ioannisg:
> > Will change this to "toup", as it looks we might want to support this upstream, i.e. a configurable mbedcrypto path.
> 
> I would like to have a cleaner thing that this. It is very generator-expression heavy as it stands right now... This was the reason for the nrf_temphack tag. @tejlmand: We discussed splitting it up into more statements to make it more readable. Is that possible to do quickly?

That's why I set it to "toup", signaling our intentions here. Does not have to be submitted as is.

tejlmand:
> That's why I set it to "toup", signaling our intentions here. Does not have to be submitted as is.

and which is exactly why I approved as-is :wink: 
Already discussed outside of this PR.

anangl (approved):

PR #540, opened 2021-06-04
[nrf fromlist] soc: Make capacitor for LFXO configurable for nRF53 and Host security Kconfig

jostn:
Switch from using internal 6pf to internal 7pf load capacitor for LFXO.
Use a default clock accuracy of 50PPM as this matches lab results.

50PPM should be sufficient for common operating temperatures of
25degC +- 15.

Signed-off-by: Johan Stridkvist <johan.stridkvist@nordicsemi.no>


soc/arm/nordic_nrf/nrf53/Kconfig.soc
```diff
@@ -142,6 +142,30 @@ config SOC_ENABLE_LFXO
 	  to use the LFXO. Otherwise, XL1 and XL2 pins will behave as regular
 	  GPIOs.
 
+choice SOC_ENABLE_LFXO_CAP
```
eugmes, line 145: Maybe this can be done in the `soc` node of the device tree? Clock node is probably appropriate:

```
clock: clock@xxxx {
   lfco-cap: "external"; // or "6pf", "7pf", "9pf"
}
```

eugmes: That may require a different node because it is not common between net/app cores.

jostn: I suggest to not change the device tree at this point - It will drag in more changes.
For instance, we should move more parts of the configuration into the device tree:
- LFCLK source
- PPM
- Capacitors

in addition we will need to split the device tree for nRF52 and nRF53, as the options are different for the MCU cores and families.

sigurdnev:
Seem to be some overlap here with this PR: https://github.com/zephyrproject-rtos/zephyr/pull/35874/

carlescufi (changes requested):
Cherry-pick fromlist: https://github.com/zephyrproject-rtos/zephyr/pull/35874

anangl (approved):

ioannisg:
@carlescufi if this is to go in for 1.6, you need to complete your review :)

carlescufi:
> @carlescufi if this is to go in for 1.6, you need to complete your review :)

I have just talked to @jostn and I’d like him to make a change in the second commit.

carlescufi (approved):

ioannisg (approved):

PR #542, opened 2021-06-07
Bluetooth: Mesh: Align capitalization from BT Mesh to BT mesh in doc.

Balaklaka:
Align the capitalization of the term "Bluetooth Mesh" to Bluetooth mesh" in the documentation.


Balaklaka:
This was going to wrong project.

PR #545, opened 2021-06-09
[nrf fromtree] modules: tf-m: fix NS image selection for regression t…

ioannisg:
…ests

When running with TFM Regression option set, but without
BL2 (MCUboot) we need to use the TF-M's NS binary, and not
the Zephyr application. This was selected properly for the
case with MCUboot boot already, but not when running the
regression tests without MCUboot.

Signed-off-by: Ioannis Glaropoulos <Ioannis.Glaropoulos@nordicsemi.no>


ioannisg:
This is needed, for fixing the TFM build plan in NCSDK CI

andvib (approved):
Tested, fixes bug

ioannisg:
Part of #546

PR #551, opened 2021-06-10
Ncs 1.6 rebase

mbolivar-nordic:
Empty diff with d0a048cd01edc65e775cbf06a572f4d511cfcc1c.

### DIFF
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -830,6 +830,7 @@ menu "Floating Point Options"
 config FPU
 	bool "Enable floating point unit (FPU)"
 	depends on CPU_HAS_FPU
+	depends on !IS_SPM
 	help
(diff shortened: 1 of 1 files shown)


tejlmand (commented):
Looking good,

You may squash:
https://github.com/nrfconnect/sdk-zephyr/pull/551/commits/e6435eeabb4575cb90025c326b978204dccb44ed
into 
https://github.com/nrfconnect/sdk-zephyr/commit/c5feda0deb63888d33331874b24c69308b11b824

as the commit message says:
> Hopefully, those two commits can be dropped in near future when code
is reworked, but in case both patches are still needed next time the
sdk-zephyr patchset is cleaned up and history is re-written then please
squash this commit with:
c5feda0

This will give us one less `temphack` to maintain.

PR #555, opened 2021-06-15
Bluetooth: Cherry-pick upstream commits

jori-nordic:
[nrf fromtree] Bluetooth: host: Prevent race condition in pairing procedure

Moving the key check after checking that no pairing or encryption
procedure is started fixes a race condition that is seen in some PTS
tests:

GATT/CL/GAR/BI-04-C
GATT/CL/GAR/BI-16-C
GATT/CL/GAR/BI-21-C
GATT/CL/GAW/BI-05-C
GATT/CL/GAW/BI-12-C

[nrf fromlist] Bluetooth: L2CAP: Fix missing net_buf_unref()

In case of non-recoverable errors (e.g. the connection breaks while
transmitting), the l2cap_chan_tx_process deques the net_buf but does
not unreference it. As this is inside a work queue thread, the sending
thread gets no information on this error, relying on the tx_process to
ultimately free the buffer.

Upstream PR: zephyrproject-rtos/zephyr#36031

[nrf fromtree] Bluetooth: controller: Check length field of scan response data

Check the length field of the scan response data.

[nrf fromtree] Bluetooth: controller: Fix initial empty flag for peripheral

Fix to avoid an assertion when processing an acknowledgment
in the first connection event in peripheral role.

Ensure that empty flag reflects the state of the Tx queue,
as a peripheral the first connection event has no prior PDU
transmitted, an incorrect acknowledgment by peer should
not dequeue a PDU that has not been transmitted on air.

Relates to assert added in commit 2bfaadf ("Bluetooth:

### DIFF
--- a/subsys/bluetooth/controller/ll_sw/nordic/lll/lll_slave.c
+++ b/subsys/bluetooth/controller/ll_sw/nordic/lll/lll_slave.c
@@ -165,6 +165,24 @@ static int prepare_cb(struct lll_prepare_param *p)
 		lll->slave.window_size_prepare_us;
 	lll->slave.window_size_prepare_us = 0;
 
+	/* Ensure that empty flag reflects the state of the Tx queue, as a
+	 * peripheral if this is the first connection event and as no prior PDU
+	 * is transmitted, an incorrect acknowledgment by peer should not
+	 * dequeue a PDU that has not been transmitted on air.
+	 */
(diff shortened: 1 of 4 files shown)


joerchan (approved):

jori-nordic:
Rebasing to make CI pass

cvinayak (approved):

PR #557, opened 2021-06-15
[nrf temphack] Revert "[nrf fromlist] drivers: ...

e-rk:
... ieee802154: add option to retransmit a frame"

This reverts commit 24700b84247a0465ea2244268360130609a84ee4.

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -480,34 +480,17 @@ static bool nrf5_tx_at(struct net_pkt *pkt, bool cca)
 	uint32_t tx_at = net_pkt_txtime(pkt) / NSEC_PER_USEC;
 	bool ret;
 
-	if (net_pkt_ieee802154_frame_retry(pkt) == false) {
-		ret = nrf_802154_transmit_raw_at(nrf5_data.tx_psdu,
-						 cca,
-						 tx_at - TXTIME_OFFSET_US,
-						 TXTIME_OFFSET_US,
-						 nrf_802154_channel_get());
-	} else {
-		ret = nrf_802154_retransmit_raw_at(nrf5_data.tx_psdu,
-						   cca,
-						   tx_at - TXTIME_OFFSET_US,
-						   TXTIME_OFFSET_US,
-						   nrf_802154_channel_get());
-	}
+	ret = nrf_802154_transmit_raw_at(nrf5_data.tx_psdu,
+					 cca,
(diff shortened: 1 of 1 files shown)


czeslawmakarski (approved):

edmont (approved):

hubertmis:
What's the reason of reverting this commit?

czeslawmakarski:
@hubertmis It was decided to remove retransmitting functionality due to the fact that it can leak security and introduce the limitation to the released solution.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -567,20 +550,6 @@ static int nrf5_tx(const struct device *dev,
...
-	 * When frame encryption by the radio driver is enabled, the frame stored in
-	 * the tx_psdu buffer is:
-	 * 1) authenticated and encrypted in place which causes that after an unsuccessful
-	 *    TX attempt, this frame must be propagated back to the upper layer for retransmission.
-	 *    The upper layer must ensure that the exact same secured frame is used for
-	 *    retransmission
-	 * 2) frame counters are updated in place and for keeping the link frame counter up to date,
-	 *    this information must be propagated back to the upper layer
-	 */
-	memcpy(payload, nrf5_radio->tx_psdu + 1, payload_len);
```
edmont, line 581: @e-rk I'm afraid we can't remove this `memcpy` because of the second reason.

edmont (changes requested):

drivers/ieee802154/ieee802154_nrf5.c
e-rk: That's not good. This will still break the frames, because it might be encrypted multiple times.

e-rk: I have removed this change for the time being.

edmont: Thanks. We are assuming invalid retransmitted frames anyway, but at least operation without retransmissions won't be compromised.

edmont (approved):

hubertmis:
> @hubertmis It was decided to remove retransmitting functionality due to the fact that it can leak security and introduce the limitation to the released solution.

I understand the reason, but I don't understand the solution. I think the patch should go to OpenThread to prevent retransmissions. Not to shim that simply relays requests from OpenThread to 802154 layer

e-rk:
Closed in favor of an alternative solution.

PR #558, opened 2021-06-15
[nrf temphack]: Limit number of parallel jobs for TF-M build to 1

tejlmand:
Fixes: NCSDK-10008

Set number of parallel jobs for TF-M build to 1.
In some circumstances it has been experienced that building TF-M with
multiple parallel jobs then `permission denied` may occur.
Root cause on Windows has not been identified but current suspicion
is around folder / file lock mechanism. To ensure correct behaviour
in all cases, limit number of parallel jobs to 1.

This issue is especially experienced when building TF-M through SES-NE.

This fix should be reverted when the root cause has been identified and
fixed.

### DIFF
--- a/modules/trusted-firmware-m/CMakeLists.txt
+++ b/modules/trusted-firmware-m/CMakeLists.txt
@@ -224,12 +224,22 @@ function(trusted_firmware_build)
 
   include(ExternalProject)
 
+  if(${CMAKE_HOST_SYSTEM_NAME} STREQUAL Windows)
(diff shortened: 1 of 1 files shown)


frkv (approved):
LGTM and can be fixable later (expecting fixup of the checkpatch)

shantha-14 (approved):

andvib (approved):

tejlmand:
Closed, included in #559

PR #563, opened 2021-06-17
[nrf fromlist] drivers: serial: nrfx_uarte: Add pullups to RXD and CTS

nordic-krch:
Configured UARTE input pins to have pullups. Otherwise when
uart is disconnected pins are floating and generate receiver
errors.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/36366

Signed-off-by: Krzysztof Chruscinski <krzysztof.chruscinski@nordicsemi.no>

### DIFF
--- a/drivers/serial/uart_nrfx_uart.c
+++ b/drivers/serial/uart_nrfx_uart.c
@@ -995,7 +995,7 @@ static int uart_nrfx_init(const struct device *dev)
 	nrf_gpio_cfg_output(TX_PIN);
 
 	if (RX_PIN_USED) {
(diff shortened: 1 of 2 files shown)


MarekPorwisz (changes requested):
similar change needs to be applied in: uarte_nrfx_pins_enable function as it overwrites those settings

MarekPorwisz (approved):

anangl (approved):
I think it would be good to also apply the same corrections in `uart_nrfx_uart.c`.

carlescufi (approved):
This now fixes both uarte and uart

edmont (approved):

PR #565, opened 2021-06-18
[nrf temphack] samples: tfm: Disable SHA-1 for TF-M Regression tests

andvib:
Disable SHA-1 as the regression tests expects SHA-1 to be disabled. Treating this as a temphack as we would prefer to disable the tests instead of disabling SHA-1, but there is currently no mechanism in TF-M to disable the tests and we don't know the scope of the tests well enough to disable them. So temphack to fix the tests for the release, and then I'll do an investigation into how the tests can be disabled after the release.

Ref.: NCSDK-9954

sdk-nrf manifest PR: [PR-4914](https://github.com/nrfconnect/sdk-nrf/pull/4914)


shantha-14 (approved):

andvib:
sdk-nrf manifest PR: [PR-4914](https://github.com/nrfconnect/sdk-nrf/pull/4914)

frkv (approved):
LGTM

ioannisg (approved):
Thanks @andvib for the explanation in the PR description. I think this is OK to do as a temp-hack.

oyvindronningstad (approved):

samples/tfm_integration/tfm_regression_test/prj.conf
```diff
@@ -8,3 +8,6 @@ CONFIG_TEST=y
 CONFIG_BUILD_WITH_TFM=y
 CONFIG_TFM_REGRESSION_S=y
 CONFIG_TFM_REGRESSION_NS=y
+
+# TF-M Regression tests expect SHA-1 to be disabled
+CONFIG_MBEDTLS_SHA1_C=n
```
tejlmand, line 13: This is introducing a downstream symbol in sdk-zephyr which should normally be avoided.

This symbol should default to `n` from here:
https://github.com/nrfconnect/sdk-nrf/blob/4aa68fe9c49604ca66b96d4333dc9dbc3362b734/modules/trusted-firmware-m/Kconfig.mbedtls.defconfig#L7-L8

so it should be investigated why this is not working as expected instead of introducing yet another layer using a temphack.

tejlmand (changes requested):

samples/tfm_integration/tfm_regression_test/prj.conf
tejlmand: I see https://github.com/nrfconnect/sdk-nrf/blob/4aa68fe9c49604ca66b96d4333dc9dbc3362b734/modules/trusted-firmware-m/Kconfig.mbedtls.defconfig is never being sourced in TF-M Kconfig.
@hakonfam I assume that's a mistake.

andvib: Took me some time to realize what was failing, but I started working on fixing the failing tests in a different way. Are the mbdetls symbols expected to be available in sdk-zephyr, or is the changes in this PR a dead-end that will never work?

andvib:
Fixed in [PR-4920](https://github.com/nrfconnect/sdk-nrf/pull/4920)

PR #570, opened 2021-06-24
bluetooth: Bluetooth configuration for GATT serialization

KAGA164:
This PR introduces changes required to run the GATT serialized API. It contains some hooks in bluetooth CMake file which are necessary because Bluetooth header files uses options from Kconfig so we cannot handle this on a Kconfig level.


subsys/bluetooth/CMakeLists.txt
```diff
@@ -5,14 +5,17 @@ add_library(subsys__bluetooth INTERFACE)
 target_include_directories(subsys__bluetooth INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_subdirectory(common)
+add_subdirectory_ifdef(CONFIG_BT_CONN services)
```
tejlmand, line 8: Instead of modifying this patch with a `[noup]` on top of a `[temphack]` commit (https://github.com/nrfconnect/sdk-zephyr/commit/2f39f35b1b7d1fbc1813606a435f410b7f74960b), could you please fix this:
https://projecttools.nordicsemi.no/jira/browse/NCSDK-10028

samples/bluetooth/beacon/Kconfig
```diff
@@ -1,8 +0,0 @@
-# Copyright (c) 2021 Nordic Semiconductor ASA
```
tejlmand, line 1: Instead of adding a `[noup]` patch removing files introduced with a `[temphack]` patch, then please revert the `[temphack]` commit (https://github.com/nrfconnect/sdk-zephyr/commit/2f39f35b1b7d1fbc1813606a435f410b7f74960b).

tejlmand (changes requested):
Do not just try to pull in temphack as a `noup`.

A noup commit should never be dependent on a previous temphack commit.

Also there need to be a good reasoning why a noup patch is required.

subsys/bluetooth/CMakeLists.txt
KAGA164: The part of this task is already fixed under this PR https://github.com/nrfconnect/sdk-nrf/pull/4984/files#diff-9bca706979b311a3bd4662437968b9fcfc854228b8d342154900ded7959a82f6R11. Should I do a revert commit and next add a new one @tejlmand ?

tejlmand: No, when I say revert, I mean this code should be reverted:
```
if (NOT CONFIG_BT_RPC_CLIENT)
  add_subdirectory_ifdef(CONFIG_BT_HCI host)
  add_subdirectory_ifdef(CONFIG_BT_SHELL shell)
  add_subdirectory_ifdef(CONFIG_BT_CONN services)
  add_subdirectory_ifdef(CONFIG_BT_MESH mesh)
  add_subdirectory_ifdef(CONFIG_BT_AUDIO audio)

  if(CONFIG_BT_CTLR AND CONFIG_BT_LL_SW_SPLIT)
    add_subdirectory(controller)
  endif()
```

This PR brings in this addition:
```
else()
  zephyr_library_sources(host/uuid.c)
```
in addition.
This is what I mean by: `modifying this patch with a [noup] on top of a [temphack] commit`.

The `else()` clause you are introducing in this PR is depending on this `if` clause from a `temphack` commit which should be fixed instead of adding more to the mess.

tejlmand: Please look at: https://github.com/nrfconnect/sdk-zephyr/commit/2f39f35b1b7d1fbc1813606a435f410b7f74960b

KAGA164:
@tejlmand I reverted temporary hack commit and used fromlist to cherry-pick my changes from upstream

doki-nordic (approved):

grochu (approved):
LGTM

grochu (approved):

tejlmand (approved):
lgtm, thanks for the cleanup :+1:

PR #571, opened 2021-06-25
Pr/add temperature platform for 802154 driver

pkwiek:
This PR adds temperature platform used by the IEEE 802.15.4 Radio
Driver. The new platform requires an enabled temperature sensor.


rlubos (commented):
Looks good but we need the hal_nordic PR merged first.

PR #583, opened 2021-07-23
Add nrf21540_ek shield definition

KAGA164:
This PR adds the nRF21540 EK Board shield definition. It is a continuation of 
https://github.com/zephyrproject-rtos/zephyr/pull/31523
which was targeted to upstream Zephyr but due to lack of use cases there, it was rejected.


hubertmis:
Would it be possible to move this PR to sdk-nrf to minimize delta between upstream and sdk-zephyr and to place shield description in the same repository FEMs are described: https://github.com/nrfconnect/sdk-nrf/tree/master/dts/bindings/radio_fem

KAGA164:
> Would it be possible to move this PR to sdk-nrf to minimize delta between upstream and sdk-zephyr and to place shield description in the same repository FEMs are described: https://github.com/nrfconnect/sdk-nrf/tree/master/dts/bindings/radio_fem

@hubertmis I don't think that we support a shield extension in sdk-nrf. So we would need to add such possibility probably from cmake level but I might be wrong. @tejlmand Do you have opinion about it ?

hubertmis:
Also, open-source BLE Controller supports nRF21. Perhaps we could use it to justify putting shield description to upstream Zephyr. Thoughts?

hubertmis:
> Also, open-source BLE Controller supports nRF21. Perhaps we could use it to justify putting shield description to upstream Zephyr. Thoughts?

Discussed offline. Opensource controller does not use device tree. It's not justification enough.

hubertmis (approved):
I would prefer to add this shield to sdk-nrf to avoid `noup` patches to sdk-zephyr, but if it would require serious changes in the build system to support shields in that repository, I'm OK to merge this patch to deliver it quicker.

grochu (approved):

KAGA164:
> I would prefer to add this shield to sdk-nrf to avoid `noup` patches to sdk-zephyr, but if it would require serious changes in the build system to support shields in that repository, I'm OK to merge this patch to deliver it quicker.

@tejlmand What is your opinion about this ?

tejlmand (changes requested):
> I would prefer to add this shield to sdk-nrf to avoid `noup` patches to sdk-zephyr, but if it would require serious changes in the build system to support shields in that repository, I'm OK to merge this patch to deliver it quicker.

What changes would be needed ?
Shields outside Zephyr repo is already supported:
https://github.com/zephyrproject-rtos/zephyr/blob/497cb2e58795142fb3b117eecf21ca8497b04e0a/cmake/app/boilerplate.cmake#L332-L333
```
foreach(root ${BOARD_ROOT})
  set(shield_dir ${root}/boards/shields)
```
and sdk-nrf is a board root:
https://github.com/nrfconnect/sdk-nrf/blob/master/zephyr/module.yml#L5

Please move this shield to sdk-nrf.

hubertmis (changes requested):
Following @tejlmand 's comment, I would strongly prefer to move it to sdk-nrf.

KAGA164:
Continued in sdk-nrf https://github.com/nrfconnect/sdk-nrf/pull/5263

PR #588, opened 2021-08-05
[nrf noup] twister: Ignore build_on_all from samples/tests yamls

PerMac:
An option "build_on_all: true" can be set in sample/testcase.yaml.
It tells Twister to add all existing platforms (~400) to the test
scope. This option is abused in (almost?) all Nordic samples and
tests. It is pointless as "platform_allow" is also used there
limiting the building/executing scope to selected platforms. Hovewer,
this causes huge mess in the genereted results reports, where ~400
platforms are added with all test statuses "skipped". This patch
make Twister to ignore "build_on_all: true" in yamls. A proper solution
will follow: all yamls will need to have "build_on_all: true" removed.

### DIFF
--- a/scripts/pylib/twister/twisterlib.py
+++ b/scripts/pylib/twister/twisterlib.py
@@ -2960,7 +2960,9 @@ class TestSuite(DisablePyTestCollectionMixin):
                         if tc.harness == 'console' and not tc.harness_config:
                             raise Exception('Harness config error: console harness defined without a configuration.')
                         tc.build_only = tc_dict["build_only"]
-                        tc.build_on_all = tc_dict["build_on_all"]
+
+                        # TEMPHACK: we have to ignore "build_on_all" until all yamls are fixed
+                        # tc.build_on_all = tc_dict["build_on_all"]
                         tc.slow = tc_dict["slow"]
                         tc.min_ram = tc_dict["min_ram"]
                         tc.depends_on = tc_dict["depends_on"]


PerMac:
This patch is needed before the "integration_mode" is added to the sdk-nrf CI. Without it test results are spammed with skipped platforms. It wasn't observed previously because platform scope was set by adding args to Twister call. If so, "build_on_all" is omitted. We don't use platform args in integration mode, hence this mess.

scripts/pylib/twister/twisterlib.py
```diff
@@ -2960,7 +2960,9 @@ def add_testcases(self, testcase_filter=[]):
                         if tc.harness == 'console' and not tc.harness_config:
                             raise Exception('Harness config error: console harness defined without a configuration.')
                         tc.build_only = tc_dict["build_only"]
-                        tc.build_on_all = tc_dict["build_on_all"]
+
```
jangalda-nsc, line 2963: Shouldn't we add any warning or message?

PerMac: I don't think so. I am pretty sure 99% of sample/testcase.yamls were created as a copy passed based on some wrong examples and some "bad habits" spread. In this case, the problem is with "build_on_all: true". As I pointed out, it makes no sense in any of nrf specific samples/tests/apps. There is no way that any of our tests/samples/apps can be build on all ~400 platforms that exist in zephyr.  Using "build_on_all: true" is clearly a misuse in nrf yamls because there is also platform_allow used. These entries are contradictory.
I am think of creating later on some script that will automatically remove all this  "build_on_all: true" lines from nrf yamls.

thst-nordic: I think removing all the 'build_on_all' lines in sdk-nrf will be faster than applying this temp-hack, doing manifest update in sdk-nrf, and then rolling it back later.

PerMac: making a script removing all the 'build_on_all' lines in sdk-nrf samples/tests/apps yamls is not hard but needs a bit more caution.  It will also require a PR to sdk-nrf which will modify all these files. If you think such temp hack is not the best idea I will start working on the script.

thst-nordic: either way is fine for me, but making two changes in zephyr will be slow, and the current upmerge might cause a problem.

thst-nordic (approved):

PerMac:
I addressed the issue directly, no need for temp hack. https://github.com/nrfconnect/sdk-nrf/pull/5250

PR #591, opened 2021-08-12
modules: nrfx_glue: Reserve PPI channels used by Gazell

leewkb4567:
Mark the PPI channels used by the Gazell Link Layer as occupied and
thus unavailable for allocation through nrfx_ppi.


modules/hal_nordic/nrfx/nrfx_glue.h
```diff
@@ -271,6 +272,13 @@ extern const uint32_t z_mpsl_used_nrf_ppi_groups;
 #define NRFX_PPI_GROUPS_USED_BY_MPSL     0
 #endif
 
+#if defined(CONFIG_GAZELL)
+extern const uint32_t nrf_gzll_used_nrf_ppi_channels;
+#define NRFX_PPI_CHANNELS_USED_BY_GZLL   nrf_gzll_used_nrf_ppi_channels
```
anangl, line 277: Is this necessary to reserve the PPI channels this way. Cannot those be allocated by calling `nrfx_ppi_channel_alloc()`?

leewkb4567: Calling nrfx_ppi_channel_alloc() also works.

The current code saves 32 bytes of RAM using constant variables.

PR #585, opened 2021-07-26
Upmerge 2021 07 26

mbolivar-nordic:
Status
------

Awaiting +1 or -1.

Reverted and reapplied
----------------------

This is a list of reverted patches that were reapplied after the
mergeup commit, in no particular order. Some of them were squashed
into other commits or had their shortlogs changed. Conflict resolution
was done as needed.

* e29515afd9 [nrf noup]: Remove "toup" and "temphack" from sauce tags
* 79c07819c3 [nrf noup] ci: Compliance workflow downstream tweaks
* 6c7c360a9a [nrf noup] dts: Adding cryptocell_sw as chosen entropy
* 44b243cc01 [nrf temphack] boards: arm: nrf5340: disable TF-M on non-secure builds
* 6b81e7cff7 [nrf temphack] soc: nrf: Switch null-pointer detection option to MPU
* 5db7cb24c0 [nrf noup] Bluetoooth: Add experimental for non-qualified options
* 05837f70ac [nrf noup] include: net: add TCP sesstimeo socket option
* 5b2f44fa72 [nrf noup] net: sockets: Add new DTLS and RAI socket options
* 5f87c2967a [nrf noup] net: sockets: Move NCS additions to a separate header
* ec1065c882 [nrf noup] include: net: add AI_PDNSERV flag
* e021755883 [nrf noup] include: net: add NCS extensions
* 938c7d44af [nrf noup] boards: don't set BUILD_WITH_TFM default for 91ns
* 02f76ba0e9 [nrf fromlist] modules: tfm: add configuration for disabling the RNG module
* c1e71016c5 [nrf toup] modules: mbedtls: Provide 3rd option for MBEDTLS configuration
* 71ee2945e9 [nrf toup] modules: tfm: Override the TF-M MBEDCRYPTO_PATH
* e22a7803db [nrf temphack]: Limit number of parallel jobs for TF-M build to 1

Left as-is, in place
--------------------

These patches are left as-is before the mergeup commit, as they did
not conflict with anything upstream.

* 28e2001ef2 [nrf temphack] kconfig: Disallow FPU for SPM
* 2f39f35b1b [nrf temphack] bluetooth: Bluetooth configuration for RPC
* 0c1f3a498a [nrf noup] tfm: adapt ieee802154_nrf5 driver to TFM
* c43ab78056 [nrf noup] Bluetoooth: host: Update experimental for qualification
* e6435eeabb [nrf temphack] samples: echo_server: remove mbed TLS settings for TF-M
[...]

PR #592, opened 2021-08-12
[nrf fromtree] manifest: openthread upmerge up to commit 2e625bfe

edmont:
OpenThread upmerge with new transmit frame header updated flag.

Depends on:
- [ ] https://github.com/nrfconnect/sdk-zephyr/pull/585


mbolivar-nordic:
Now applied to the upmerge branch.

The manifest update conflicted with the cherry-pick of 4f35a4eaa9e7282f1933500de237401091dd60bb, so I reverted that. Also the third fromtree did not apply to the upmerge branch so I recreated it by cherry-picking the original upstream commit. Please recheck my branch to make sure it contains the right commits; thanks.

PR #593, opened 2021-08-17
nRF53 Multi Image DFU

sigvartmh:
Since we are using this for emulating flash in RAM for the nRF53 Dual Core update procedure some changes need to be made to make it more usable. 

*  Removing the stats to save code space
* Make `mock_flash` public so that we can access the address of the RAM region so that it can be passed to `pcd_update`.

This PR also adds some fixes/hacks to make `mcumgr` and `bootutil_public` to play nice with multi-image. Since we don't have the flash simulator available in memory for the application we need to skip it. If not `mcumgr` will try to open a flash area that does not exist and bus fault.

Also commit dc36b19299f48bb56778340d25b4daf7c3e9b2d5 needs to be removed before this is put in. And any changes in the branch committed into the upstream mcumgr repo.


nvlsianpu (changes requested):
Following recent changes to the manual you need to open PR to the upstream with changes to the flash_simulator and make changes here as [fromlist].

sigvartmh:
Changed to fromlist and PR opened upstream

subsys/dfu/boot/CMakeLists.txt
```diff
@@ -4,5 +4,14 @@ if(CONFIG_MCUBOOT_IMG_MANAGER)
 zephyr_library()
 zephyr_library_sources(mcuboot.c)
 zephyr_library_sources_ifdef(CONFIG_MCUBOOT_SHELL mcuboot_shell.c)
+if(CONFIG_BOOT_IMAGE_ACCESS_HOOKS)
```
hakonfam, line 7: Why is this here?

hakonfam:
Superseded by what https://github.com/nrfconnect/sdk-nrf/pull/5439 included

PR #598, opened 2021-08-24
[nrf noup] boards: thingy53_nrf5340: Add static PM configuration

MarekPieta:
Change introduces static configuration of Partition Manager available in NCS. Thingy:53 secure cpuapp configuration in NCS by default uses MCUboot bootloader.


kkasperczyk-no (approved):

LuDuda (approved):

MarekPieta:
Hi @tejlmand, please review so we will be able to merge this one.

boards/arm/thingy53_nrf5340/CMakeLists.txt
```diff
@@ -5,3 +5,8 @@ if (CONFIG_BOARD_THINGY53_NRF5340_CPUAPP OR CONFIG_BOARD_THINGY53_NRF5340_CPUAPP
 	zephyr_library()
 	zephyr_library_sources(board.c)
 endif()
+
+# In NCS Thingy:53 by default uses predefined partition map.
+if(CONFIG_BOARD_THINGY53_NRF5340_CPUAPP)
+	set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/pm_static.yml CACHE INTERNAL "")
```
tejlmand, line 11: when this is being set as a cache variable, then how is the end-user suppose to be able to specify an alternative pm_static file ?

I know we want to provide a common pm_static file for general use, but I don't recall we decided to remove the users ability to have an application specific file.

Are we dependent on this commit ?
https://github.com/nrfconnect/sdk-nrf/pull/5117/commits/429838a39fb1c44c72a9f64de2a060db04ff5213

to support board names in pm_static, maybe with a board default defined static in addition.

MarekPieta: Hi @tejlmand 
I actually tried to introduce it in the same way as it's introduced for Thingy:91 in nrf-sdk repo:
https://github.com/nrfconnect/sdk-nrf/blob/4955f577c617c60c1bb48635a8ab27f209d19dd2/boards/arm/thingy91_nrf9160/CMakeLists.txt#L12-L20
Do you think it should be done in a different way here?

MarekPieta: If board default static PM configuration would be introduced by the linked change and it would be available before the code freeze, we could use it (I think we should use it for Thingy:91 too). Do you know when the change is going to be introduced?

pdunaj: I agree with @tejlmand here. The whole point of having the appstore specific configuration removed from board files was to ensure that thingy53 will remain generic for anybody.

@tejlmand , would it be possible to add feature to zephyr that would allow to inherit from a generic board? That would work by creating a board_xyz that essentially takes dts and basic config from other board but adds its own stuff on top. I was exploring the idea to create a thingy53_appstore board in NCS but I run into problems with netcore board not being able to correctly inherit things. Was suppose to ask you about it but then holiday happened...

pdunaj (changes requested):
I am against having it does thins way.
We essentially add app store specific things to thingy53 rendering it specific to this single configuration. If we do it why not adding all other things?

This was something that was already brought up few months ago. It would be good to have a common place for configuration (for appstore samples) but there is no such place at the moment.

PR #609, opened 2021-09-01
BLE RPC configuration

KAGA164:


carlescufi (changes requested):
You need a sauce tag on the new commit

mbolivar-nordic:
> You need a sauce tag on the new commit

which it looks like should be "fromlist" -- agree @carlescufi ?

KAGA164:
I will close it. I used it only for test purposes. I don't know why you were assigned to it

PR #610, opened 2021-09-02
[nrf fromlist] samples: hci_rpmsg: Introduce Thingy:53 configuration

MarekPieta:
Upstream PR:
https://github.com/zephyrproject-rtos/zephyr/pull/38252

Change introduces default configuration overlay for Thingy:53.
The configuration overlay is needed to speed up DFU over BLE.


MarekPieta:
The configuration is updated for all of the supported samples.
Still, it could cause some problems for new samples/apps on Thingy:53. I wonder if we should mention about this requirements in user guide or rather make the hci_rpmsg configuration specific to selected subset of Thigny:53 samples/apps.

MarekPieta:
The change is going to be introduced as hci_rpmsg Kconfigs defined separately for every sample/application.

PR #618, opened 2021-09-09
[nrf fromlist] modules: Add TF-M SECURE_UART1 config

Vge0rge:
-This adds a Kconfig configuration which allows
 disabling SECURE_UART1.

Upstream PR: [pull/38670](https://github.com/zephyrproject-rtos/zephyr/pull/38670)

### DIFF
--- a/modules/trusted-firmware-m/CMakeLists.txt
+++ b/modules/trusted-firmware-m/CMakeLists.txt
@@ -30,6 +30,11 @@ set(TFM_CRYPTO_MODULES
 
 
 if (CONFIG_BUILD_WITH_TFM)
+  if (CONFIG_TFM_SECURE_UART1)
+    list(APPEND TFM_CMAKE_ARGS -DSECURE_UART1=ON)
+  else()
+    list(APPEND TFM_CMAKE_ARGS -DSECURE_UART1=OFF)
+  endif()
   if (CONFIG_TFM_IPC)
     list(APPEND TFM_CMAKE_ARGS -DTFM_PSA_API=ON)
     # PSA API awareness for the Non-Secure application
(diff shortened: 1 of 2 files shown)


modules/trusted-firmware-m/CMakeLists.txt
```diff
@@ -30,6 +30,9 @@ set(TFM_CRYPTO_MODULES
 
 
 if (CONFIG_BUILD_WITH_TFM)
+  if (CONFIG_TFM_SECURE_UART1)
+	list(APPEND TFM_CMAKE_ARGS -DSECURE_UART1=ON)
```
oyvindronningstad, line 34: Fix alignment

oyvindronningstad (commented):
Good to have this! Please reference the upstream PR in the commit message.

modules/trusted-firmware-m/CMakeLists.txt
Vge0rge: I will, thanks!

Vge0rge:
> Good to have this! Please reference the upstream PR in the commit message.

I will indeed, I just wanted to have the first round of the reviews before I do to avoid multiple revisions upstream.

modules/trusted-firmware-m/CMakeLists.txt
Vge0rge: Should be fixed now :)

tejlmand (changes requested):
upstream Zephyr doesn't provide a custom `tfm_hal_platform_init` that avoids calling `stdio_init()`.

This means that disabling SECURE_UART1 in Zephyr will lead to a bus fault, hence I can't see how the upstream PR can be approved at current state.

See: https://github.com/zephyrproject-rtos/zephyr/pull/38670#pullrequestreview-760915168

Vge0rge:
I will do the change only in the sdk-nrf.

PR #620, opened 2021-09-10
dfu/boot: Kconfig for MCUboot's bootutil

nvlsianpu:
Mcuboot's bootutil libraries has option to use hooks which
allows to customize its behavior while proceeding on images
date. This patch introduces configuration options required that.

upstream https://github.com/zephyrproject-rtos/zephyr/pull/38496

### DIFF
--- a/modules/Kconfig.mcuboot_bootutil
+++ b/modules/Kconfig.mcuboot_bootutil
@@ -21,4 +21,21 @@ module-str = MCUboot bootutil
 source "subsys/logging/Kconfig.template.log_config"
 endif
 
+config BOOT_IMAGE_ACCESS_HOOKS
+	bool "Enable hooks for overriding MCUboot's bootutil native routines"
+	help
+	  Allow to provide procedures for override or extend native
+	  MCUboot's routines required for access the image data.
+
+config BOOT_IMAGE_ACCESS_HOOKS_FILE
+	string "Hooks implementation file path"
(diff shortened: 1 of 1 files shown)


subsys/dfu/Kconfig
```diff
@@ -93,6 +93,23 @@ config UPDATEABLE_IMAGE_NUMBER
 	help
 	  If value is set to 2 or greater then, this enables support needed when
 	  application is combined with MCUboot multi-image boot.
+
+config BOOT_IMAGE_ACCESS_HOOKS
```
hakonfam, line 97: I assume the answer is no, and you don't need to provide any explanation why what I suggest here is not possible:

To me it seems like we could expose these config options to MCUboot, and then replace the ones we currently have in the mcuboot repo (`bootloader/mcuboot/boot/zephyr/Kconfig`) with this one. That way we could also avoid duplicating the cmake code for adding the hooks file since the same cmake code could be executed by both mcuboot and the app.

nvlsianpu:
will do this [nrf fromlist].

subsys/dfu/Kconfig
nvlsianpu: moved configuration to zephyr module specidic option as they are common to the application and the bootloader.

I would keep Mcuboot own Kconfig (as duplication) as well as it doesn't make any problem - just make this configurable from the MCUoot project's own menu as well.

MarekPieta (approved):

PR #621, opened 2021-09-13
[nrf fromlist] Add multi-image for Thingy:53

MarekPieta:
Upstream PR:
https://github.com/zephyrproject-rtos/zephyr/pull/38496

Mcuboot's bootutil libraries has option to use hooks which allows to customize its behavior while proceeding on images date. This patch introduces configuration options required that.


MarekPieta:
The changes are no longer used. I close PR.

PR #624, opened 2021-09-16
[nrf noup] soc: Use DWT for null-pointer detection on nRF5340 net core

anangl:
This is a follow-up to commit 7c9139a300ace41d8f66270b3003684999c01d63
and should be squashed with it on the next upmerge.

The MPU-based null-pointer dereference detection cannot be used on
the network core in nRF5340 because of its flash base address and
this causes the zephyr/tests/arch/arm/arm_interrupt test to fail.
Use the DWT option instead for this core.

Ref. NCSDK-11348

### DIFF
--- a/soc/arm/nordic_nrf/Kconfig.defconfig
+++ b/soc/arm/nordic_nrf/Kconfig.defconfig
@@ -43,8 +43,10 @@ config GPIO
 	depends on SPI
 
 # Switch the default mechanism for null-pointer dereference detection from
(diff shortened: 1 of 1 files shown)


PerMac (approved):
Checked on hw. Tests are passing now.

SebastianBoe:
Bravo

SebastianBoe (approved):

carlescufi (approved):

PR #626, opened 2021-09-17
[nrf fromlist] doc: cmake package: remove stale cmake version lines

mbolivar-nordic:
Upstream PR:
https://github.com/zephyrproject-rtos/zephyr/pull/38636

This page is meant to document the behavior of the Zephyr cmake
package. However, its example CMake snippets contain some stale
references to a minimum version (3.13.1, now 3.20).

These lines are a bit of a distraction anyway, so just remove them.
They matter in general, but they don't matter where the Zephyr cmake
package is concerned.

### DIFF
--- a/doc/guides/zephyr_cmake_package.rst
+++ b/doc/guides/zephyr_cmake_package.rst
@@ -248,7 +248,6 @@ For example:
 
   .. code-block:: cmake
 
-     cmake_minimum_required(VERSION 3.13.1)
      find_package(Zephyr 2.2.0)
      project(app)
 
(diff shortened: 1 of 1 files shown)


shantha-14 (changes requested):
Shouldnt this update cmake_minimum_required to 3.20.5 instead of removing. This bug was actually found by an external who followed the getting started guide and found build errors ?

tejlmand (approved):

tejlmand:
> Shouldnt this update cmake_minimum_required to 3.20.5 instead of removing. This bug was actually found by an external who followed the getting started guide and found build errors ?

This is not from the getting started guide, but from the Zephyr CMake package description.
Users should read the guide to understand and be familiar with the Zephyr CMake package but not use it as a reference / template on how to write a CMakeLists.txt file for an application, as it lacks several steps in that regard.

carlescufi (approved):

mbolivar-nordic:
Not needed

PR #629, opened 2021-09-29
subsys: net: lib: lwm2m: Decouple CoAP message size from CoAP block size

VeijoPesonen:
If blockwise transfer is not in use it should be possible to send bigger
messages than what is the block size. 

To compensate for the increased memory usage number of in-flight packets
allowed could be decreased.

Signed-off-by: Veijo Pesonen <veijo.pesonen@nordicsemi.no>


subsys/net/lib/lwm2m/Kconfig
```diff
@@ -47,7 +47,7 @@ config LWM2M_COAP_BLOCK_SIZE
 	range 64 1024
 	help
 	  CoAP block size used by LWM2M when performing block-wise
-	  transfers. Possible values: 16, 32, 64, 128, 256, 512 and 1024.
```
rlubos, line 50: Why not 16 and 32? These are still valid values for block size, I don't think we need to impose any limits on the LwM2M side.

subsys/net/lib/lwm2m/Kconfig
```diff
@@ -56,6 +56,16 @@ config LWM2M_ENGINE_MESSAGE_HEADER_SIZE
 	help
 	  Extra room allocated to handle CoAP header data
 
+config LWM2M_COAP_MAX_MSG_SIZE
+	int "LWM2M CoAP maximum message size"
+	default LWM2M_COAP_BLOCK_SIZE
+	range 64 8168
```
rlubos, line 62: 8kB is 8192

subsys/net/lib/lwm2m/Kconfig
```diff
@@ -56,6 +56,16 @@ config LWM2M_ENGINE_MESSAGE_HEADER_SIZE
...
+	help
+	  CoAP message size used by LWM2M. Minimum is the block size used
+	  in blockwise transfers. Packet size(msg plus header) must not exceed 8kB.
+	  The 8kB upper limit is arbitrary. Theoretical max would be:
```
rlubos, line 66: I'm not sure, do we need to impose any arbitrary limit on LwM2M layer? Assuming IP-level fragmentation is supported, it should be fine to support UDP limits.

And note, that IP fragmentation is disabled in Zephyr by default and **it's only supported by IPv6**. Therefore I think it should also be worth to mention that the user have to make sure that the network configuration he uses needs to have IP fragmentation support enabled it this value is near or above network MTU.

subsys/net/lib/lwm2m/Kconfig
VeijoPesonen: I decided to go with modifying the _help_ instead of the _range_ itself. I shall do it the other way round.

subsys/net/lib/lwm2m/Kconfig
VeijoPesonen: The header takes at least 24 bytes. But of course this could be left for the build time assert to check.

subsys/net/lib/lwm2m/Kconfig
VeijoPesonen: Good point. I'll set the upper limit to be the same as the theoretical one. There would be need to extract the size of the UDP, IPv4 and IPv6 headers in use but maybe that's a moving target we don't consider in this case.

subsys/net/lib/lwm2m/Kconfig
rlubos: Ah, now I noticed the the range did not correspond to the help string. Nevermind then, please ignore this comment.

subsys/net/lib/lwm2m/Kconfig
SeppoTakalo: Maybe we could have this as big as needed, and then have it in a separate task to implement splitting larger than PMTU packets into CoAP block-wise requests.
Because even if we agree to split messages, we still need at lest one big message buffer where we would start the splitting.

Also, lets not try to guess the underlying transport protocol. It might be UDP or DTLS, in which case we should handle the PMTU limits (to avoid fragmentation), but it could also be stream, like TCP or TLS, in which case fragmentation does not happen, or matter.

subsys/net/lib/lwm2m/Kconfig
VeijoPesonen: Boundaries removed.

subsys/net/lib/lwm2m/Kconfig
```diff
@@ -47,7 +47,7 @@ config LWM2M_COAP_BLOCK_SIZE
 	range 64 1024
```
rlubos, line 47: Regarding this commit, plase note that every commit requires some body in Zephyr/NCS, even if it's trivial.

rlubos (commented):
Looks good, but please submit an upstream PR with this change first.

subsys/net/lib/lwm2m/Kconfig
VeijoPesonen: Time to check all commits separately with gitlint.

```
...:~/ws/ncs2/zephyr$ gitlint 
1: UC3 Title does not follow [subsystem]: [subject] (and should not start with literal subsys:): "subsys: net: lib: lwm2m: Decouple CoAP message..."
```

rlubos: Right, I didn't notice, `subsys:` in the prefix is not allowed :)

VeijoPesonen:
To be replaced with the upstream PR  zephyrproject-rtos/zephyr#39003

PR #635, opened 2021-10-18
[nrf noup] mcumgr: smp: Support enabling low latency for SMP BT

eobalski:
This change adds a posibility to enable low latency connection
parameters for BT when SMP commands are handled.

Support for this functionality is disabled by the default and
can be enabled by CONFIG_MCUMGR_SMP_BT_LOW_LATENCY=y option.

Jira: NCSDK-11180

Signed-off-by: Emil Obalski <emil.obalski@nordicsemi.no>

### DIFF
--- a/subsys/mgmt/mcumgr/Kconfig
+++ b/subsys/mgmt/mcumgr/Kconfig
@@ -262,15 +262,51 @@ config MCUMGR_SMP_BT
 	help
 	  Enables handling of SMP commands received over Bluetooth.
 
+if MCUMGR_SMP_BT
+
 config MCUMGR_SMP_BT_AUTHEN
 	bool "Authenticated requirement for Bluetooth mcumgr SMP transport"
-	depends on MCUMGR_SMP_BT
 	select BT_SMP
 	default y
 	help
 	  Enables encrypted and authenticated connection requirement to
 	  Bluetooth SMP transport.
 
+config MCUMGR_SMP_BT_LOW_LATENCY
+	bool "Request low latency connection when handling SMP commands"
+	depends on SYSTEM_WORKQUEUE_PRIORITY < 0
+	help
+	  Enables support for requesting low latency connection parameter when
+	  SMP commands are handled. This option allows to speed up the command
+	  exchange process.
+	  Its recommended to enable this if SMP is used for DFU.
+
+config MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_TIME
+	int "Connection latency restore time in milliseconds"
+	depends on MCUMGR_SMP_BT_LOW_LATENCY
+	default 5000
+	range 1 65535
+	help
+	  The value is a time after which connection latency is restored
+	  to default value, that is, before requesting low latency. Latency
+	  restoration time could take up to twice as long as specified time.
+	  This is because of limiting CPU time needed to support this feature.
+	  The implementation periodically checks if the low latency is still
+	  needed every MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_TIME. The default
+	  latency is restored during check only if there was no SMP command
+	  exchanged in period before the check
+
+config MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_RETRY_TIME
+	int "Connection latency restore retry time in milliseconds"
+	depends on MCUMGR_SMP_BT_LOW_LATENCY
+	default 1000
+	range 1 5000
+	help
+	  In case connection latency restoration fails due to an error, this
+	  option specifies the time after retry to set the default latency
+	  would be executed.
+
+endif # MCUMGR_SMP_BT
 
 config MCUMGR_SMP_SHELL
 	bool "Shell mcumgr SMP transport"
--- a/subsys/mgmt/mcumgr/smp_bt.c
+++ b/subsys/mgmt/mcumgr/smp_bt.c
@@ -21,13 +21,29 @@
 
 #include <mgmt/mcumgr/smp.h>
 
-struct device;
+#define RESTORE_TIME	   COND_CODE_1(CONFIG_MCUMGR_SMP_BT_LOW_LATENCY, \
+				(CONFIG_MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_TIME), (0))
+#define RESTORE_RETRY_TIME COND_CODE_1(CONFIG_MCUMGR_SMP_BT_LOW_LATENCY, \
+				(CONFIG_MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_RETRY_TIME), (0))
 
 struct smp_bt_user_data {
 	struct bt_conn *conn;
 };
 
+enum {
+	CONN_LOW_LATENCY_ENABLED	= BIT(0),
+	CONN_LOW_LATENCY_REQUESTED	= BIT(1),
+};
+
+struct conn_param_data {
+	struct bt_conn *conn;
+	struct k_work_delayable dwork;
+	uint16_t default_latency;
+	uint8_t latency_state;
+};
+
 static struct zephyr_smp_transport smp_bt_transport;
+static struct conn_param_data conn_data[CONFIG_BT_MAX_CONN];
 
 /* SMP service.
  * {8D53DC1D-1DB7-4CD3-868B-8A527460AA84}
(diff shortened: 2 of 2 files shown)


subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -191,10 +297,67 @@ int smp_bt_unregister(void)
...
+}
+
+/* BT disconnected callback. */
+static void disconnected(struct bt_conn *conn, uint8_t reason)
+{
+	struct conn_param_data *cpd = get_conn_param_data(conn);
+
+	if (cpd->latency_state) {
+		/* Cancel work if ongoing. */
+		(void)k_work_cancel_delayable(&cpd->dwork);
```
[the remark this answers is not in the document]
MarekPieta: Also I think you could always cancel the work here (for simplicity).

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -41,6 +55,94 @@ static struct bt_uuid_128 smp_bt_svc_uuid = BT_UUID_INIT_128(
...
+
+		ret = bt_conn_le_param_update(cpd->conn, &params);
+		if (ret != 0) {
+			return;
+		}
+		cpd->default_latency = info.le.latency;
+		cpd->latency_state |= CONN_LOW_LATENCY_REQUESTED;
+
+	} else if (cpd->latency_state & CONN_LOW_LATENCY_ENABLED) {
+		(void)k_work_reschedule(&cpd->dwork, K_MSEC(WORK_TIMEOUT));
```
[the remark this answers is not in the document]
MarekPieta: We may not need high accuracy for the timeout here I think

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -191,10 +297,67 @@ int smp_bt_unregister(void)
...
+	cpd->conn = NULL;
+}
+
+/* BT LE connection parameters updated callback. */
+static void le_param_updated(struct bt_conn *conn, uint16_t interval,
+			     uint16_t latency, uint16_t timeout)
+{
+	struct conn_param_data *cpd = get_conn_param_data(conn);
+
+	if (cpd->latency_state != 0) {
```
MarekPieta, line 327: Maybe we should mark that low latency was already enabled anyway (also in case we did not request it)?

MarekPieta: You could also check that on connected callback

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -61,6 +163,10 @@ static ssize_t smp_bt_chr_write(struct bt_conn *conn,
 	ud = net_buf_user_data(nb);
 	ud->conn = bt_conn_ref(conn);
 
+	if (IS_ENABLED(CONFIG_MCUMGR_SMP_BT_LOW_LATENCY)) {
```
MarekPieta, line 166: Should we check if we use slave role for given connection ?

MarekPieta:
Hi @joerchan
This is the change that was requested by you some time ago. It could be useful for Bluetooth samples that are going to support DFU. Could you take a look at it?

joerchan:
@MarekPieta Thanks. I just read low latency as LLPM and got confused.

subsys/mgmt/mcumgr/smp_bt.c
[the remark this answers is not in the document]
emob-nordic: Good idea, that will save some CPU time.

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -41,6 +55,94 @@ static struct bt_uuid_128 smp_bt_svc_uuid = BT_UUID_INIT_128(
 static struct bt_uuid_128 smp_bt_chr_uuid = BT_UUID_INIT_128(
 	BT_UUID_128_ENCODE(0xda2e7828, 0xfbce, 0x4e01, 0xae9e, 0x261174997c48));
 
+/* Helper function that allocates conn_param_data for a conn. */
+static struct conn_param_data *alloc_conn_param_data(struct bt_conn *conn)
```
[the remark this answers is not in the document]
emob-nordic: Checked that and it would be problematic to have. Having this argument `const` triggers compilation warning when assigning this pointer to non const `conn_data[i].conn`. I could add const to structure member `conn` but then, `bt_conn_*` API is used inconsistent. I will leave this as is.

subsys/mgmt/mcumgr/Kconfig
```diff
@@ -262,15 +262,33 @@ config MCUMGR_SMP_BT
 	help
 	  Enables handling of SMP commands received over Bluetooth.
 
+if MCUMGR_SMP_BT
```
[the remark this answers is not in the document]
emob-nordic: fwd to @joerchan @de-nordic is this worth upstreaming?

subsys/mgmt/mcumgr/Kconfig
```diff
@@ -262,15 +262,47 @@ config MCUMGR_SMP_BT
...
+config MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_TIME
+	int "Connection latency restore time in milliseconds"
+	depends on MCUMGR_SMP_BT_LOW_LATENCY
+	default 5000
+	range 1 65535
+	help
+	  The value is a time after which connection latency is restored
+	  to default value, that is, before requesting low latency. Latency
+	  restoration time could take up to twice as long as specified time.
+	  This is because of limiting CPU time needed to support this feature.
```
MarekPieta, line 293: Maybe we could explain it further?
`The implementation periodically checks if the low latency is still needed every MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_TIME. The default latency is restored during check only if there was no SMP command exchanged in period before the check`

Or just inform that it's because of implementation limitations without going into details.

subsys/mgmt/mcumgr/Kconfig
```diff
@@ -262,15 +262,47 @@ config MCUMGR_SMP_BT
...
+
+config MCUMGR_SMP_BT_LOW_LATENCY_RESTORE_RETRY_TIME
+	int "Connection latency restore retry time in milliseconds"
+	depends on MCUMGR_SMP_BT_LOW_LATENCY
+	default 1000
+	range 1 5000
+	help
+	  In case connection latency restoration fails due to some internal
```
MarekPieta, line 301: I would say: `due to an error.` (you do not check the error code value).

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -191,10 +305,70 @@ int smp_bt_unregister(void)
...
+			     uint16_t latency, uint16_t timeout)
+{
+	struct conn_param_data *cpd = get_conn_param_data(conn);
+
+	if (latency == 0) {
+		if (cpd->latency_state & CONN_LOW_LATENCY_REQUESTED) {
+			cpd->latency_state &= ~CONN_LOW_LATENCY_REQUESTED;
+			(void)k_work_reschedule(&cpd->dwork, K_MSEC(RESTORE_TIME));
+		}
+		cpd->latency_state = CONN_LOW_LATENCY_ENABLED;
```
MarekPieta, line 341: You never use it anywhere, it's just set here. Also it's not cleared under else here

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -21,13 +21,27 @@
...
+};
+
+struct conn_param_data {
+	struct bt_conn *conn;
+	struct k_work_delayable dwork;
+	uint16_t default_latency;
+	uint8_t latency_state;
+};
+
+struct device;
```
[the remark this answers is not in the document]
emob-nordic: Okay, will remove this.

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -41,6 +56,101 @@ static struct bt_uuid_128 smp_bt_svc_uuid = BT_UUID_INIT_128(
...
+	int ret = 0;
+
+	cpd = CONTAINER_OF(work, struct conn_param_data, dwork);
+
+	if (cpd->latency_state & CONN_LOW_LATENCY_REQUESTED) {
+		(void)k_work_reschedule(&cpd->dwork, K_MSEC(RESTORE_TIME));
+		cpd->latency_state &= ~CONN_LOW_LATENCY_REQUESTED;
+	} else {
+		ret = bt_conn_get_info(cpd->conn, &info);
+		if (ret != 0) {
```
MarekPieta, line 103: Then we would be stuck with low latency ...

MarekPieta: I think we may either explicitly assume this never happens (I think it would actually never happen looking at the implementation) or reschedule work here

subsys/mgmt/mcumgr/Kconfig
```diff
@@ -262,15 +262,47 @@ config MCUMGR_SMP_BT
...
 	bool "Authenticated requirement for Bluetooth mcumgr SMP transport"
-	depends on MCUMGR_SMP_BT
 	select BT_SMP
 	default y
 	help
 	  Enables encrypted and authenticated connection requirement to
 	  Bluetooth SMP transport.
 
+config MCUMGR_SMP_BT_LOW_LATENCY
+	bool "Request low latency connection when handling SMP commands"
```
MarekPieta, line 276: Consider naming it `low peripheral(slave) latency` to avoid confusion with LLPM

subsys/mgmt/mcumgr/Kconfig
MarekPieta: Also the BT SMP subsys triggers the retry, not a Bluetooth driver itself.

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -41,6 +56,101 @@ static struct bt_uuid_128 smp_bt_svc_uuid = BT_UUID_INIT_128(
...
+			return;
+		}
+
+		if (info.le.latency >= cpd->default_latency) {
```
MarekPieta, line 107: Should we accept any non-zero latency here?

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -41,6 +56,101 @@ static struct bt_uuid_128 smp_bt_svc_uuid = BT_UUID_INIT_128(
...
+		}
+
+		params.interval_min = info.le.interval;
+		params.interval_max = info.le.interval;
+		params.latency = cpd->default_latency;
+		params.timeout = info.le.timeout;
+
+		ret = bt_conn_le_param_update(cpd->conn, &params);
+
+		if (ret != 0 && (ret != -EALREADY)) {
```
MarekPieta, line 118: Use brackets consistently, please

subsys/mgmt/mcumgr/smp_bt.c
MarekPieta: The work should actually not be called if latency is not equal to zero, right ?

subsys/mgmt/mcumgr/smp_bt.c
```diff
@@ -41,6 +56,101 @@ static struct bt_uuid_128 smp_bt_svc_uuid = BT_UUID_INIT_128(
...
+/* Enables low latency connection for a given conn. */
+static void enable_low_latency(struct bt_conn *conn)
+{
+	struct bt_le_conn_param params;
+	struct conn_param_data *cpd;
+	struct bt_conn_info info;
+	int ret = 0;
+
+	cpd = get_conn_param_data(conn);
+	if (cpd->latency_state == 0) {
```
MarekPieta, line 133: This would not work properly if connection slave latency is set to 0 and requested flag was cleared by the work (in that case flag needs to be set, but we should not send the request)

subsys/mgmt/mcumgr/smp_bt.c
emob-nordic: What? Which latency? Restored one or read by get_info one?

subsys/mgmt/mcumgr/smp_bt.c
emob-nordic: No.

subsys/mgmt/mcumgr/smp_bt.c
emob-nordic: I forgot to clear the flags in the work, when the latency is restored. In case you are describing the work would exit at first if:
```
		if (ret != 0 || info.le.latency == 0) {
			return;
		}

emob-nordic: ACtually i do it in `le_param_updated` callback so i think there is no need to add this.

emob-nordic: If i understand you correctly the answer is: It would work as there is ENABLED flag that is not cleared by the work.

subsys/mgmt/mcumgr/smp_bt.c
emob-nordic: I do clear this, but not explicitly: `cpd->latency_state &= ~CONN_LOW_LATENCY_ENABLED;`
`cpd->latency_state = 0;`

subsys/mgmt/mcumgr/smp_bt.c
MarekPieta: We could assert or ignore the returned value if we know that given code path would never be taken

subsys/mgmt/mcumgr/smp_bt.c
MarekPieta: The work handler can be called only if the `info.le.latency` equals 0. We could actually assert that I think

PR #637, opened 2021-10-20
Upmerge 2021 10 20

rlubos:


rlubos:
Had to restore https://github.com/rlubos/fw-nrfconnect-zephyr/commit/38a185a7a3023ccc380cc96bf7aadbe27c3bfbc4, Zephyr v2.7 uses older version than we currently use in NCS and `ncs-compare` tricked me as the commit name is usually the same for mcumgr revision bumps.

subsys/net/lib/openthread/platform/shell.c
```diff
@@ -25,8 +25,19 @@ static bool is_shell_initialized;
 
 static int ot_console_cb(void *context, const char *format, va_list arg)
 {
+	char prompt_check[4];
```
tejlmand, line 28: Nothing wrong on this line, but commit message says:
>  [nrf noup] net: openthread: Temporary fix for OT prompt in the shell 

To me this looks like a disguised `[nrf temphack]` which is no longer allowed.

I think a proper fix is required here.
@carlescufi and @mbolivar-nordic thoughts on this ?

tejlmand (dismissed):

subsys/net/lib/openthread/platform/shell.c
rlubos: A proper fix would be needed on the OpenThread side (to allow to disable CLI prompt), I don't think we have time to cover this. Alternatively, I can propose a similar fix to the upstream Zephyr and make it a `[nrf fromlist]`.

rlubos:
There are several failing tetst in CI/Jenkins/ZEPHYR plan (https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-zephyr/job/PR-637/19/). Together with @PerMac we've idenified 3 types of failures:

* Several TFM samples fail with `TypeError: __init__() got an unexpected keyword argument 'rom_fixed'` I see however this was marked as fixed for 2.7 release (there's also an issue covering that https://github.com/zephyrproject-rtos/zephyr/issues/38692). @SebastianBoe Any idea why would this occur for the upemerge? We're pulling the v2.7.0 relase tag this time.

* `tests/ztest/busy_sim` fails with the following:
```
/opt/gnuarmemb/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld.bfd: zephyr/libzephyr.a(busy_sim.c.obj): in function `rng_pool_work_handler':
/home/work/zephyr/subsys/testsuite/busy_sim/busy_sim.c:73: undefined reference to `__device_dts_ord_67'
/opt/gnuarmemb/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld.bfd: zephyr/libzephyr.a(busy_sim.c.obj):/home/work/zephyr/subsys/testsuite/busy_sim/busy_sim.c:37: undefined reference to `__device_dts_ord_67'
```
As far as I understand, this happens because we're switching to `cryptocell` RNG in sdk-zephyr but the driver itself is implemented in sdk-nrf. I don't know if there's a way to address that.

* Several newlib/cmsis_dsp tests fail due to FLASH/RAM overflow on certain nRF platforms - don't know why isn't present upstream, any clues on how to address that?

@PerMac mentioned we could add some test cases to the quarantine, how did we proceed in previous upmerges, were issues like this present? Any opinion @carlescufi @mbolivar-nordic ?

subsys/net/lib/openthread/platform/shell.c
rlubos: I've created an upstream PR with this change and made it a `[nrf fromlist]` commit.

carlescufi (approved):

PR #645, opened 2021-11-08
[nrf noup] fs: littlefs: Fix automount for partition manager

sigvartmh:
This adds a conditional that selects FLASH_AREA_ID if
partition manager is enabled so that littlefs uses the littlefs_storage
partition or the storage parition.
The littlefs_storage partition takes precedence over the storage partition.

Ref. NCSDK-11402

### DIFF
--- a/subsys/fs/littlefs_fs.c
+++ b/subsys/fs/littlefs_fs.c
@@ -767,6 +767,12 @@ static const struct fs_file_system_t littlefs_fs = {
 	.statvfs = littlefs_statvfs,
 };
 
+#ifdef USE_PARTITION_MANAGER
+#define _LFS_USE_NCS_PM 1
+#else
+#define _LFS_USE_NCS_PM 0
+#endif
+
 #define DT_DRV_COMPAT zephyr_fstab_littlefs
 #define FS_PARTITION(inst) DT_PHANDLE_BY_IDX(DT_DRV_INST(inst), partition, 0)
 
@@ -801,7 +807,11 @@ struct fs_mount_t FS_FSTAB_ENTRY(DT_DRV_INST(inst)) = { \
 	.type = FS_LITTLEFS, \
 	.mnt_point = DT_INST_PROP(inst, mount_point), \
 	.fs_data = &fs_data_##inst, \
-	.storage_dev = (void *)DT_FIXED_PARTITION_ID(FS_PARTITION(inst)), \
+	.storage_dev = (void *) (_LFS_USE_NCS_PM ?\
+	  COND_CODE_1(FLASH_AREA_LABEL_EXISTS(littlefs_storage), \
+		      (FLASH_AREA_ID(littlefs_storage)), \
+		      (FLASH_AREA_ID(storage))) :\
+	  DT_FIXED_PARTITION_ID(FS_PARTITION(inst))), \
 	.flags = FSTAB_ENTRY_DT_MOUNT_FLAGS(DT_DRV_INST(inst)), \
 };
 


de-nordic (commented):
Since this is nrf specific and LFS is used here, shouldn't this be directly `littlefs_storage` instead of `storage`? Wouldn't pm rename it anyway?

nvlsianpu (commented):
> Since this is nrf specific and LFS is used here, shouldn't this be directly littlefs_storage instead of storage? Wouldn't pm rename it anyway?

That's good idea (I was going to rise as well). We should make possible to use more than one storage implementation at once, like NVS and littlefs, so distinguishable names are needed.

P.S.

Can we do something which allow application to use the `littlefs_storage` if it is defined, otherwise use `storage` (as legacy bindings)?

de-nordic:
> 
> Can we do something which allow application to use the `littlefs_storage` if it is defined, otherwise use `storage` (as legacy bindings)?

There are *_OR device macros that allow to select property from DTS or set default, maybe they can be used somehow here.

nvlsianpu:
@sigvartmh Can you respond to my suggestion above?

sigvartmh:
@nvlsianpu I'm trying to look into it I would assume it should be possible to do it with `#ifdef` but I tried a bit looking into which *_OR device macros we have. Problem is that I don't think all of these are coming from device tree and some is coming from PM but I'm looking into it.

subsys/fs/littlefs_fs.c
```diff
@@ -801,7 +801,10 @@ struct fs_mount_t FS_FSTAB_ENTRY(DT_DRV_INST(inst)) = { \
 	.type = FS_LITTLEFS, \
 	.mnt_point = DT_INST_PROP(inst, mount_point), \
 	.fs_data = &fs_data_##inst, \
-	.storage_dev = (void *)DT_FIXED_PARTITION_ID(FS_PARTITION(inst)), \
+	.storage_dev = (void *) (DT_NODE_EXISTS(PARTITION_NODE) ?\
+	DT_FIXED_PARTITION_ID(FS_PARTITION(inst)) :\
+	(defined(littlefs_storage) ? FLASH_AREA_ID(littlefs_storage) :\
+	 FLASH_AREA_ID(storage))), \
```
sigvartmh, line 807: @nvlsianpu Here is one suggestion in how to get what you are asking I think?

sigvartmh: This did unfortunately not seem to work for everything.

subsys/fs/littlefs_fs.c
```diff
@@ -801,9 +802,49 @@ struct fs_mount_t FS_FSTAB_ENTRY(DT_DRV_INST(inst)) = { \
...
+		.lookahead_buffer = lookahead_buffer_##inst, \
+	}, \
+}; \
+struct fs_mount_t FS_FSTAB_ENTRY(DT_DRV_INST(inst)) = { \
+	.type = FS_LITTLEFS, \
+	.mnt_point = DT_INST_PROP(inst, mount_point), \
+	.fs_data = &fs_data_##inst, \
+	.storage_dev = (void *) (USE_PARTITION_MANAGER ?\
+	  FLASH_AREA_ID(storage) :\
+	  DT_FIXED_PARTITION_ID(FS_PARTITION(inst))), \
```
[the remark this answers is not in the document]
sigvartmh:
```
fixed-partitions.h:53:9: error: 'DT_COMPAT_fixed_partitions_LABEL_littlefs_storage_PARTITION_ID' undeclared here (not in a function); did you mean 'DT_COMPAT_fixed_partitions_LABEL_storage_EXISTS'?
   53 |  DT_CAT(DT_COMPAT_fixed_partitions_LABEL_, label)
littlefs_fs.c:806:4: note: in expansion of macro 'FLASH_AREA_ID'
  806 |    FLASH_AREA_ID(littlefs_storage) :\
```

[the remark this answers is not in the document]
de-nordic: > 

The `COND_CODE_1` and `COND_CODE_0` could be used here?

tejlmand:
please rebase.

subsys/fs/littlefs_fs.c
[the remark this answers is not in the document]
nvlsianpu: I tried
``` C
.storage_dev = (void *) (USE_PARTITION_MANAGER ?\
	  COND_CODE_1(FLASH_AREA_LABEL_EXISTS(littlefs_storage), (FLASH_AREA_ID(littlefs_storage)), (FLASH_AREA_ID(storage))) :\
	  DT_FIXED_PARTITION_ID(FS_PARTITION(inst))), \
```
@de-nordic Is this something what you expected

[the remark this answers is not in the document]
de-nordic: Yes, something like this if it is possible.

nvlsianpu:
@de-nordic @sigvartmh @hakonfam please revisit

de-nordic (approved):
Looks OK.

sigvartmh (commented):
Looks good to me. Thanks for fixing it up 🥇

subsys/fs/littlefs_fs.c
```diff
@@ -801,7 +805,9 @@ struct fs_mount_t FS_FSTAB_ENTRY(DT_DRV_INST(inst)) = { \
 	.type = FS_LITTLEFS, \
 	.mnt_point = DT_INST_PROP(inst, mount_point), \
 	.fs_data = &fs_data_##inst, \
-	.storage_dev = (void *)DT_FIXED_PARTITION_ID(FS_PARTITION(inst)), \
+	.storage_dev = (void *) (USE_PARTITION_MANAGER ?\
```
hakonfam, line 808: This looks nice!

Is it possible to use `( defined(USE_PARTITION_MANAGER) ? \` instead of having the `#define USE_PARTITION_MANAGER 0` above?

hakonfam (approved):

subsys/fs/littlefs_fs.c
de-nordic: I am not sure whether `defined` is evaluated when not used with `#if`. I am quite sure that up to c99 the `defined` would be left there after pre-processor phase.

nvlsianpu: Think `IS_ENABLED(USE_PARTITION_MANAGE)` will does the job

de-nordic: Maybe `IF_ENABLED` ?

de-nordic: Seems that there is need for `DEFINED()` macro :D

hakonfam:
I created this https://github.com/nrfconnect/sdk-nrf/pull/6479 in a desperate attempt to get this merged before the upmerge starts

nvlsianpu:
^^ I had opened manifest update some time agohttps://github.com/nrfconnect/sdk-nrf/pull/6459 
Actuallly CI looks god enough to merge this one.

de-nordic:
Seems that we should be able to use IS_ENABLED after this https://github.com/zephyrproject-rtos/zephyr/pull/39390 gets upmerged.

PR #659, opened 2021-11-19
ipc: use highest coop priority for IPM workqueue

ankuns:
Signed-off-by: Andrzej Kuroś <andrzej.kuros@nordicsemi.no>


ankuns:
There is newer PR https://github.com/nrfconnect/sdk-zephyr/pull/666

PR #647, opened 2021-11-09
Extend nrf_rtc_timer and nrf_802154_lp_timer

jciupis:
This PR mirrors the upstream PR https://github.com/zephyrproject-rtos/zephyr/pull/39099.

PR #660, opened 2021-11-19
boards: posix: nrf52_bsim: provide missing CMSIS functions

jciupis:
This commit provides missing CMSIS functions for nRF52 babblesim board.

Signed-off-by: Jedrzej Ciupis <jedrzej.ciupis@nordicsemi.no>


tejlmand:
is this PR expected to fix the bable sim failures seen in other PRs like here:
https://github.com/nrfconnect/sdk-zephyr/pull/657
https://github.com/nrfconnect/sdk-zephyr/pull/643

jciupis:
@tejlmand that was the initial idea, but now it's clear that the simple fix suggested in this PR is insufficient to solve all babblesim problems silently introduced in #647. We're missing support for RTC's OVERFLOW event in babblesim and as far as I understand that's not a trivial to implement, so temporarily disabling the babblesim tests might be necessary.

alxelax:
Will it be possible to run BabbleSim tests with only these fixes? Now it seems changes in nrf rtc driver broke compatibility between this driver and the nrf52_bsim platform. Our team (Paladin: BLE Mesh) tries to use Babblesim for Bluetooth Mesh model testing but we are blocked by this incompatibility.

Should related changes be reverted from Downstream and merged in Upstream first (with passing all CI steps, including BabbleSim) and then upmerged into downstream?

@carlescufi  FYI

czeslawmakarski:
@axlelax We (Radio Software) are using the downstream nrf_rtc_timer to provide the Thread 1.2 Functionality on nRF5340 platform. If I'm not mistaken, the issue revolves around the OVRFLOW event in RTC not being handled by the babblesim simulator. Maybe we should aim towards implementing this support in babblesim? Let's synchronize on this.

Adding @jciupis as a developer who was handling this case.

jciupis:
@alxelax as @czeslawmakarski stated above, the changes that this PR introduces are insufficient to solve all babblesim issues that arose after https://github.com/nrfconnect/sdk-zephyr/pull/647 was merged. Although it fixes the problem of missing CMSIS functions, entire implementation of RTC's OVERFLOW event is needed in babblesim to make it possible for the tests to pass, which we were unaware at the time #647 was being reviewed and merged.

AFAIK reverting https://github.com/nrfconnect/sdk-zephyr/pull/647 is hardly an option as features dependent on it have already been developed. However, upstream has not been "contaminated" with that regression - the upstream PR has been marked as `DNM` and not merged yet: https://github.com/zephyrproject-rtos/zephyr/pull/39099. Would it be possible for your team to base on the upstream?

alxelax:
@jciupis BLE Mesh models are located only in sdk-nrf repo. Unfortunately, it is not possible to test them in Upstream. 
If reverting isn't the option then the only way is to add missed functionality in BabbleSim. Not possible to use BabbleSim before that.

carlescufi:
closing, will come as part of the upmerge

PR #671, opened 2021-11-25
[nrf fromtree] usb: remove Kconfig option USB_UART_CONSOLE

jfischer-no:
Adjust init level and priority of CDC ACM driver to regular serial drivers.
Limit scope of CONFIG_USB_UART_CONSOLE and finally remove this option.

https://github.com/zephyrproject-rtos/zephyr/pull/40428


jfischer-no:
Probably we only need the first four commits, WIP.

anangl (approved):

PR #678, opened 2021-12-07
[nrf noup] subsys: lorawan: NVM support and SE implemenation for nRFx

GiulianoFranchetto:
For now, the only available backend is settings.
This could be an issue when using soft-se, as the
keys would also be stored un-encryptedon the flash.

Once more stable we must give the possibility to use
higher security feature, like for example using the
capacity of the nrf9160 to store keys in its own
secure element

Signed-off-by: Giuliano Franchetto <giuliano.franchetto@intellinium.com>


mbolivar-nordic:

PR #679, opened 2021-12-16
Multi instance lwm2m object 5

jarlamsa:
Adding a possibility to create multiple instances of the firmware update object.
This gives a possibility to update multiple parts of the firmware at once and follow the status.
Currently this isn't included in the OMA LWM2M specification, but the discussions are ongoing.


jarlamsa:
@hasnainvirk @jheiskan81 please review

jarlamsa:
Seems that sauce tags are wrong. I was wondering that should we use `mergeup` here. @rlubos do you have opinion about the upstreaming?

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
```diff
@@ -348,31 +379,36 @@ static int lwm2m_firmware_init(const struct device *dev)
...
+		/* Initialize state machine */
+		/* TODO: should be restored from the permanent storage */
+		update_state[idx] = STATE_IDLE;
+		update_result[idx] = RESULT_DEFAULT;
+#ifdef CONFIG_LWM2M_FIRMWARE_UPDATE_PULL_SUPPORT
+		delivery_method[idx] = DELIVERY_METHOD_BOTH;
+#else
+		delivery_method[idx] = DELIVERY_METHOD_PUSH_ONLY;
+#endif
+		LOG_INF("Creating LWM2M FW instance %d", idx);
```
SeppoTakalo, line 405: I think this is too verbose. Either DBG or drop the whole line. Previously it did not print anything for a success case. And I would also prefer not to print, as this can rarely fail, and is part of so normal initialization flow that it should be obvious.

SeppoTakalo (approved):
Looks good.
I left few comments there which I don't consider mandatory, feel free to ignore if you don't like.

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
```diff
@@ -319,27 +345,32 @@ static struct lwm2m_engine_obj_inst *firmware_create(uint16_t obj_inst_id)
...
+		     false, true, package_uri[obj_inst_id], PACKAGE_URI_LEN, NULL, NULL, NULL,
+		     package_uri_write_cb, NULL);
+	INIT_OBJ_RES_EXECUTE(FIRMWARE_UPDATE_ID, res[obj_inst_id], i, firmware_update_cb);
+	INIT_OBJ_RES_DATA(FIRMWARE_STATE_ID, res[obj_inst_id], i, res_inst[obj_inst_id], j,
+			  &(update_state[obj_inst_id]), sizeof(update_state[obj_inst_id]));
+	INIT_OBJ_RES_DATA(FIRMWARE_UPDATE_RESULT_ID, res[obj_inst_id], i, res_inst[obj_inst_id], j,
+			  &(update_result[obj_inst_id]), sizeof(update_result[obj_inst_id]));
+	INIT_OBJ_RES_OPT(FIRMWARE_PACKAGE_NAME_ID, res[obj_inst_id], i, res_inst[obj_inst_id], j, 1,
+			 false, true, NULL, NULL, NULL, NULL, NULL);
+	INIT_OBJ_RES_OPT(FIRMWARE_PACKAGE_VERSION_ID, res[obj_inst_id], i, res_inst[obj_inst_id], j,
```
jheiskan81, line 363: It think those Optional unbuffered resources could use INIT_OBJ_RES_OPTDATA() but that is calling INIT_OBJ_RES_OPT() with NULL like your code.  Correct me If I'm wrong.

jheiskan81 (approved):
Looks good for me.

subsys/net/lib/lwm2m/Kconfig
```diff
@@ -235,6 +235,15 @@ config LWM2M_FIRMWARE_UPDATE_OBJ_SUPPORT
 	help
 	  Include support for LWM2M Firmware Update Object (ID 5)
 
+config LWM2M_FIRMWARE_INSTANCE_COUNT
+	int "Maximum # of LWM2M Firmware object instances"
+	default 1
+	range 1 3
```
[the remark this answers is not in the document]
jarlamsa: True, this is unnecessary limitation.

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
jarlamsa: Yes they could, behaviour would be exactly the same though.

subsys/net/lib/lwm2m/Kconfig
```diff
@@ -235,6 +235,15 @@ config LWM2M_FIRMWARE_UPDATE_OBJ_SUPPORT
 	help
 	  Include support for LWM2M Firmware Update Object (ID 5)
 
+config LWM2M_FIRMWARE_INSTANCE_COUNT
```
VeijoPesonen, line 238: This could be marked as experimental - please see a [similar](https://github.com/zephyrproject-rtos/zephyr/pull/41158#issuecomment-996483962) suggestion from me for further details.

include/net/lwm2m.h
```diff
@@ -136,6 +136,11 @@ struct lwm2m_ctx {
 	 *  copied into the actual resource buffer.
```
rlubos, line 136: Technically, such commit's should contain `[nrf toup]` sauce tag, such approach is now "forbidden due to excessive abuse" (see development model page on confluence). You should rather open an upstream PR first, and then cherrypick the commits with `[nrf fromlist]`.

include/net/lwm2m.h
```diff
@@ -136,6 +136,11 @@ struct lwm2m_ctx {
...
 	 */
 	uint8_t validate_buf[CONFIG_LWM2M_ENGINE_VALIDATION_BUFFER_SIZE];
+
+	/**
+	 * Current FW update object instance.
+	 */
+	uint16_t fw_obj_inst;
```
rlubos, line 143: I'm not sure if this variable belongs in a generic lwm2m_context structure. Anyway, it shall not be needed with the `lwm2m_pull_context` from the upstream PR.

subsys/net/lib/lwm2m/lwm2m_obj_firmware_pull.c
```diff
@@ -27,9 +27,10 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 #define NETWORK_CONNECT_TIMEOUT	K_SECONDS(10)
 #define PACKET_TRANSFER_RETRY_MAX	3
 
-static char firmware_uri[URI_LEN];
+static char firmware_uri[CONFIG_LWM2M_FIRMWARE_INSTANCE_COUNT][URI_LEN];
```
rlubos, line 30: Genrally, changes in this file will be in conflict with https://github.com/zephyrproject-rtos/zephyr/pull/40552, which I believe will be integrated in NCS on the next upmerge (it's near finalization). I think we should rather build our solution for multiple FW instances on top of that PR (specifically the reusable pull context).

jarlamsa: Yeah, I agree. As that PR has now been merged to the upstream-zephyr, I will do some rework over that and make a PR to the upstream.

subsys/net/lib/lwm2m/Kconfig
jarlamsa: Good suggestion. Unfortunately that doesn't apply to `int` configs, so something different is needed here.

jarlamsa:
Thank you for the reviews. I opened a new PR against upstream zephyr. Review-comments have been addressed with that PR.

jarlamsa:
This has been merged to the upstream zephyr, closing this one.

PR #681, opened 2021-12-20
doc improvement:update config option

kittydepa:
Update config option.

Signed-off-by: Katherine Depa <katherine.depa@nordicsemi.no>


b-gent:
@tejlmand  compliance check fails on 'Sauce tag missing' , what is this tag?

tejlmand:
> @tejlmand compliance check fails on 'Sauce tag missing' , what is this tag?

It's missing a:
```
<sauce-tag>: commit message title
```

Right now the commit message tittle says: https://github.com/nrfconnect/sdk-zephyr/pull/681/commits/c5de2c304cb95944a10e930334fb209db579b176
```
 doc improvement:update config option
```
So I propose to change that into:
```
doc: update config option CONFIG_BT_EXT_ADV to CONFIG_BT_PER_ADV
```

kittydepa:
> > @tejlmand compliance check fails on 'Sauce tag missing' , what is this tag?
> 
> It's missing a:
> 
> ```
> <sauce-tag>: commit message title
> ```
> 
> Right now the commit message tittle says: [c5de2c3](https://github.com/nrfconnect/sdk-zephyr/commit/c5de2c304cb95944a10e930334fb209db579b176)
> 
> ```
>  doc improvement:update config option
> ```
> 
> So I propose to change that into:
> 
> ```
> doc: update config option CONFIG_BT_EXT_ADV to CONFIG_BT_PER_ADV
> ```

 Okay, will do.  Where do I need to fill in/edit the <sauce-tag>?

Avalei (approved):

richabp (approved):

mbolivar-nordic:

doc/application/index.rst
```diff
@@ -534,7 +534,7 @@ at CMake configure time if any experimental feature is enabled.
    CONFIG_WARN_EXPERIMENTAL=y
 
 For example, enabling experimental warnings and building a sample which enables
-:kconfig:`CONFIG_BT_EXT_ADV` will print the following warning at CMake
+:kconfig:`CONFIG_BT_PER_ADV` will print the following warning at CMake
```
mbolivar-nordic, line 537: A couple of questions about this one:

1. Why should this be squashed into `nrf noup] Bluetooth: update experimental for qualificatio`? According to `git blame`, the line was originally added in commit f17630ba750, which is upstream.
2. Why is CONFIG_BT_EXT_ADV wrong? I see that option used throughout the zephyr tree
3. Did you forget to update the code block on line 542 also?

mbolivar-nordic (changes requested):
A few questions

doc/application/index.rst
kittydepa: Thanks for the questions, Martin. I will do my best to answer them.

1. This is something I am not entirely sure I understand, nor can answer with confidence... I am open to suggestions though.
2. This was recommended by a PM in the Technical Support Team/tech-pubs Teams channel, on 17th of Dec. In the original suggestion, it was suggested ```CONFIG_BT_EXT_ADV``` maybe still be experimental, and the other config option was recommended.
This is something I did not double-check... My apologies. What are your thoughts about this @mbolivar-nordic? And please feel free to let me know if there are others I should include in this conversation. Thank you!

3. Whoops! Yes, thank you for checking that. I will edit that, if we all agree by then end that this should be changed.

mbolivar-nordic: > This is something I am not entirely sure I understand, nor can answer with confidence... I am open to suggestions though.

@tejlmand can you respond here? This looks suspicious to me but I am not sure what the right way forward is here.

tejlmand: > @tejlmand can you respond here? This looks suspicious to me but I am not sure what the right way forward is here.

@mbolivar-nordic I believe I might have been asking @kittydepa to use the `squash` info.
It is correct that upstream uses `CONFIG_BT_EXT_ADV` when showing how the `EXPERIMENTAL` feature work, if you read the whole part:
> For example, enabling experimental warnings and building a sample which enables
> :kconfig:`CONFIG_BT_EXT_ADV` will print the following warning at CMake
> configure time.
>
> .. code-block:: shell
> warning: Experimental symbol BT_EXT_ADV is enabled.

However, trying this with NCS will not print that warning, because in this `[noup]` commit, we remove `select EXPERIMENTAL` on `BT_EXT_ADV`: https://github.com/nrfconnect/sdk-zephyr/commit/ec6d4e8185295f48e9255128966469240932c4b7

Therefore we must update the docs to show case another `EXPERIMENTAL` symbol.
And therefore I think this commit ought to be squashed into the commit that is the root cause for this symbol no longer being experimental.

tejlmand (changes requested):
@kittydepa you must also update line 542 in this file.
It currently says:
> warning: Experimental symbol BT_EXT_ADV is enabled.

The correct symbol `CONFIG_BT_PER_ADV` must now be used in this location also.

mbolivar-nordic:
@tejlmand but why is this change needed? The commit log doesn't explain and the docs commit is already upstream. The whole thing seems to be just an example for how to enable experimental features. What's going on?

tejlmand:
> The whole thing seems to be just an example for how to enable experimental features. What's going on?

Correct, so upstream if a user follows this guide: https://docs.zephyrproject.org/latest/application/index.html?highlight=config_warn_experimental#experimental-features
> The [CONFIG_WARN_EXPERIMENTAL](https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_WARN_EXPERIMENTAL.html#std-kconfig-CONFIG_WARN_EXPERIMENTAL) setting can be used to enable warnings at CMake configure time if any experimental feature is enabled.
>
> `CONFIG_WARN_EXPERIMENTAL=y`
>
> For example, enabling experimental warnings and building a sample which enables [CONFIG_BT_EXT_ADV](https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_BT_EXT_ADV.html#std-kconfig-CONFIG_BT_EXT_ADV) will print the following warning at CMake configure time.
> `warning: Experimental symbol BT_EXT_ADV is enabled.`

but if a users follow the same procedure in NCS, then **no warning** will be printed, because `BT_EXT_ADV` in **not** an experimental feature in NCS, see https://github.com/nrfconnect/sdk-zephyr/commit/ec6d4e8185295f48e9255128966469240932c4b7.
This means that any user reading the [NCS documentation](http://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/application/index.html#experimental-features) and tries to do exactly this will not see the referred warning, and hence could think this feature is broken.

But in fact the feature is working.

This is why in NCS, we need to refer to a symbol that is still experimental in NCS context (and someone has proposed `BT_PER_ADV`).

Of course, someone can also propose to change the upstream example setting, but unless we know of an always experimental symbol upstream, I think it's hard to justify that upstream must always change their example every time a symbol is not considered experimental in NCS :confused:

kittydepa:
> @kittydepa you must also update line 542 in this file. It currently says:
> 
> > warning: Experimental symbol BT_EXT_ADV is enabled.
> 
> The correct symbol `BT_PER_ADV` must now be used in this location also.

Change has been applied, and I have pushed again. 
If there are any objections from anyone, please feel free to let me know.

mbolivar-nordic:
> I think it's hard to justify that upstream must always change their example every time a symbol is not considered experimental in NCS

That's a fair point, but I still think the fix is needed upstream. I've proposed a patch here: https://github.com/zephyrproject-rtos/zephyr/pull/42876

mbolivar-nordic:
In terms of moving this PR forward, I would suggest taking the upstream patch as an `[nrf fromlist]` into sdk-zephyr. Any objections?

mbolivar-nordic:
https://github.com/nrfconnect/sdk-zephyr/pull/740

PR #683, opened 2021-12-27
Upmerge branch

SebastianBoe:


SebastianBoe:
When this branch is merged it will be force-pushed to main.

This is because release branches were used so merging was deemed to not be the best strategy.

The conflicts reported here are therefore expected and can be ignored.

tejlmand (changes requested):
The upmerge is missing: [9cad422b7e](https://github.com/nrfconnect/sdk-zephyr/commit/9cad422b7e3765f77ce1e4e82b2fda3f32e4c794)

That commit is already on `sdk-zephyr/main` but not included in this PR and hence will be lost.
Remember to cherry-pick that commit also for the upmerge.

It has probably been forgotten as it is on main, but not refered to by the `sdk-nrf/west.yml`.

Secondly, remember to tag `sdk-zephyr/main` so that git will not garbage collect dangling SHAs once the upmerge are forced pushed to main.

SebastianBoe:
@tejlmand : No, it's here, just with a different SHA. And a different location in the history. It's at the tip of the branch.

tejlmand (approved):
Upmerge looks good.

Still remember to tag old SHA before force-pushing the upmerge.

scripts/kconfig/kconfiglib.py
```diff
@@ -556,6 +556,16 @@ def my_other_fn(kconf, name, arg_1, arg_2, ...):
 
 VERSION = (14, 1, 0)
 
+# Bump the recursion limit from 1k to 5k. The limit was observed to be
+# broken in NCS when gen_kconfig_rest.py was executed. Specifically,
```
mbolivar-nordic, line 560: Why not do this in gen_kconfig_rest.py instead, then?

SebastianBoe: Because that would also be a noup commit. I'm trying to figure out a way to make this upstreamable.

But I had to task-switch to crypto.

Perhaps someone else is able to find a way to make this fromlist instead of noup?

Perhaps just a PR will do, maybe upstream will accept this as-is. After moving the patch to gen_kconfig_rest maybe.

de-nordic (approved):
I have done basic test of BT, BT Tiny and UART of mcumgr with smp_svr sample.
Tested areas were image upload/list/test/confirm, file upload/download, basic commands (echo, reset) and shell.
Everything seems to be OK.

PR #684, opened 2022-01-05
ci: Update to Zephyr SDK 0.13.2

SebastianBoe:
Pull in the CI docker image v0.21.0, which contains the Zephyr SDK
0.13.2 release, and use the Zephyr SDK 0.13.2 for building and testing
Zephyr in the CI.

Signed-off-by: Stephanos Ioannidis <root@stephanos.io>

### DIFF
--- a/.buildkite/daily.yml
+++ b/.buildkite/daily.yml
@@ -3,14 +3,14 @@ steps:
     - .buildkite/run.sh
     env:
       ZEPHYR_TOOLCHAIN_VARIANT: "zephyr"
-      ZEPHYR_SDK_INSTALL_DIR: "/opt/toolchains/zephyr-sdk-0.13.1"
+      ZEPHYR_SDK_INSTALL_DIR: "/opt/toolchains/zephyr-sdk-0.13.2"
     parallelism: 475
     timeout_in_minutes: 210
     retry:
       manual: true
     plugins:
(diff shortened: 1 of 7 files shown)


PerMac (approved):

tejlmand (approved):

PerMac:
I think other commits are required as well. I tried following the procedure for environment https://github.com/zephyrproject-rtos/zephyr/blob/main/.github/workflows/clang.yaml#L43 under docker and it seems git rebase is crucial to make it work.

In older version of clang-action (the one we use) we have only west init west update.
This doesn't work due to west submanifests failure.

The newer (upstream) version has more steps added. I found that adding `git rebase origin/main` before west update solves the manifest progblem
https://github.com/zephyrproject-rtos/zephyr/blob/main/.github/workflows/clang.yaml#L50

git rebase fails at one point due to the conflict CONFLICT (content): Merge conflict in subsys/bluetooth/mesh/proxy_srv.c
but then there is no submanifest error

I have another one idea to try. I can open a noup PR to sdk-zephyr which will modify the clang action to use the code from the PR, not from the main. Maybe if we merge it pre upmerge, the upmerge PR will use the action from the PR i.e. with all the modifications the PR is introducing to the action itself

SebastianBoe:
Closed in favour of a different approach.

PR #686, opened 2022-01-10
[nrf noup] action: clang: set the name of checkout repo to zephyr

PerMac:
Fixes isses with downstream forks, where repos names are not zephyr

Will have to be reworked after the upmerge

### DIFF
--- a/.github/workflows/bsim.yaml
+++ b/.github/workflows/bsim.yaml
@@ -42,13 +42,16 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
+          path: ./zephyr
 
       - name: west setup
+        working-directory: ./zephyr
         run: |
           west init -l . || true
           west update
 
       - name: Run Bluetooth Tests with BSIM
+        working-directory: ./zephyr
         run: |
           #source zephyr-env.sh
           export ZEPHYR_BASE=${PWD}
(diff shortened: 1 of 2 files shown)


SebastianBoe:
I feel that it was a mistake to name the fork sdk-zephyr.

Having the fork's name be zephyr when cloned through west, but sdk-zephyr when cloned directly is unnecessarily confusing.

But I guess that ship has sailed.

SebastianBoe (approved):

PerMac:
I discuss this action with Anas and probably he won't allow for this change in the upstream.  The problem is this action was defined to use a specific self-hosted AWS agent (not yet in our port) which uses cache and is not wiped clean before running PRs and changing the structure might screw other PR checks.
Anas advice was to change the path in the manifest itself zephyr->sdk-zephyr, which should solve the problem with submanifest import

PerMac:
@mbolivar  I tried a modification which would change the name of the directory, however other errors started popping out. It looks like it messes the original structure and github actions + docker doesn't like this. I think the following steps were trying to use the initial  structure and producing the error, and i guess it would have to be given explicit for all of them any way:
```
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: chdir to cwd ("/__w/sdk-zephyr/sdk-zephyr") set in config.json failed: no such file or directory: unknown
Error: Process completed with exit code 126.
```

Since, according to Marti, this might get a fix with a following west release and would need a post upmerge rework anyway. I would merge this version and see if it helps for the upmerge PR.

SebastianBoe (approved):

nvlsianpu (approved):

SebastianBoe:
We need this to unlock test coverage in the upmerge.

Will be re-worked after/during the upmerge.

de-nordic (approved):

PR #689, opened 2022-01-13
Revert "[nrf noup] soc: flash: Add tfm read service to nRF SoC Flash …

SebastianBoe:
revert [nrf noup] soc: flash: Add tfm read service to nRF SoC Flash.

It has several problems:

The patch adding tfm_read_service should be in a fromlist commit using DT[0].
The patch adding PM support should be in a noup commit that removes the DT references.
The patch adding SPM support should be in it's own noup commit.

[0] Something like this: https://github.com/zephyrproject-rtos/zephyr/pull/41756


SebastianBoe:
The commit that is being reverted did not have a corresponding manifest update (was closed), so it is safe to revert without any manifest changes.

Revert was suggested by @carlescufi and I agree. I am assuming that nothing in the upmerge requires this commit.

The commit will be removed from the upmerge PR as well.

joerchan:
I don't get it. If it is removed from the upmerge, then why is this needed? sdk-zephyr is blocked from merging by the upmerge.
Reverted in the upmerge and not-reapplied should be enough.

SebastianBoe:
I've deleted it from the upmerge history.

I want a revert commit so it doesn't look like we accidentally dropped a commit.

The history will be re-written in the upmerge.

joerchan (approved):

sigvartmh (approved):

de-nordic (approved):

SebastianBoe:
too late for this.

PR #690, opened 2022-01-13
[nrf noup] cmake: linker: rework s1 linking

hakonfam:
With the latest updates in downstream we don't need to call
s1.cmake. Also the symbol to check in the linker to see if
we are being linked against S1 has changed.

Ref: NCSDK-13188


hakonfam:
Just to get CI result

SebastianBoe:
Pushed to upmerge branch

PR #694, opened 2022-01-24
LwM2M raw CBOR content format support

VeijoPesonen:
This is just s preview for LwM2M CBOR content format support.

### TODO 
* Need to adapt on changes done with PR zephyrproject-rtos/zephyr#41744
* 64-bit value support once next zcbor release is out
* floating-point support once next zcbor release is out
* Adapting to API changes introduced by the next zcbor release

To upstream this, it would require upstreaming [NordicSemiconductor/zcbor](https://github.com/NordicSemiconductor/zcbor) first.


VeijoPesonen:
@rlubos, @SeppoTakalo@, @jheiskan81, @jarlamsa For your information.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,529 @@
...
+	.get_float32fix = NULL,
+	.get_bool = get_bool,
+	.get_opaque = get_opaque,
+	.get_objlnk = get_objlnk,
+};
+
+int do_read_op_cbor(struct lwm2m_message *msg, int content_format)
+{
+	/* Can only return single resource */
+	if (msg->path.level != 3U) {
```
jheiskan81, line 467: This cuold be 4 I think change to >= 3

subsys/net/lib/lwm2m/lwm2m_rw_cbor.h
```diff
@@ -0,0 +1,26 @@
...
+
+size_t cbor_put_format(struct lwm2m_output_context *out,
+			     const char *format, ...);
+
+size_t cbor_put_float32fix(struct lwm2m_output_context *out,
+				 struct lwm2m_obj_path *path,
+				 float32_value_t *value);
+
+
+int do_read_op_cbor(struct lwm2m_message *msg, int content_format);
```
jheiskan81, line 23: Do we need to push content_format when we call lwm2m_perform_read_op() we can give LWM2M_FORMAT_APP_CBOR

VeijoPesonen: Fixed.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: This should be `(msg->path.level < 3U)` once we have support for version 1.1 in place. There is a need to do some changes on the engine side to get this to work first 

```
Extended LwM2M operations to enable Resource Instance level access
``` 
[LightweightM2M; Chapter 1.1. LwM2M version 1.1](https://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.pdf)

subsys/net/lib/lwm2m/lwm2m_object.h
```diff
@@ -138,9 +138,14 @@ BUILD_ASSERT(CONFIG_LWM2M_COAP_BLOCK_SIZE <= CONFIG_LWM2M_COAP_MAX_MSG_SIZE,
 #define CPKT_BUF_WRITE(cpkt)	(cpkt)->data, &(cpkt)->offset, (cpkt)->max_len
 #define CPKT_BUF_READ(cpkt)	(cpkt)->data, (cpkt)->max_len
 
+#define CPKT_BUF_W_PTR(cpkt)	 ((cpkt)->data + (cpkt)->offset)
```
jheiskan81, line 141: Is this same than at my PR?

VeijoPesonen:
Pushed to fix the coding style issues. _checkpatch.pl_ doesn't seem to be able to detect all of those.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,558 @@
...
+#include "cbor_encode.h"
+#include "lwm2m_object.h"
+#include "lwm2m_rw_cbor.h"
+#include "lwm2m_engine.h"
+#include "lwm2m_util.h"
+
+#define CPKT_CBOR_W_SZ(pos, cpkt) \
+	((size_t)(pos) - (size_t)(cpkt)->data - (size_t)(cpkt)->offset)
+
+enum  cbor_tag { /* www.iana.org/assignments/cbor-tags/cbor-tags.xhtml */
```
VeijoPesonen, line 29: Not found from cddl-gen

subsys/net/lib/lwm2m/Kconfig
```diff
@@ -370,6 +370,13 @@ config LWM2M_RW_JSON_SUPPORT
 	help
 	  Include support for writing JSON data
 
+config LWM2M_RW_CBOR_SUPPORT
+	bool "support for CBOR writer [EXPERIMENTAL]"
```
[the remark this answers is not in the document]
VeijoPesonen: I would like to get rid of that last FIXME comment. But yeah, it does follow the spec but it doesn't support all the data types needed for reading&writing resources.

mbolivar-nordic:

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: @oyvindronningstad Worth of adding the to library itself?

VeijoPesonen:
Rebased.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
oyvindronningstad: Good idea

mbolivar-nordic (changes requested):
The reason why this is `[nrf noup]` needs to be justified. Out of tree patches to core functionality like this are discouraged as they cause maintenance burden which slows down upmerges.

Edit: I see it's in the pull request description, but this needs to be in the commit logs as well if you're going to go this route when taking this out of draft

VeijoPesonen:
> Edit: I see it's in the pull request description, but this needs to be in the commit logs as well if you're going to go this route when taking this out of draft

Ok, I updated the commit logs and realized that the first commit can be upstreamed already.

subsys/net/lib/lwm2m/lwm2m_object.h
```diff
@@ -147,6 +156,8 @@ struct lwm2m_message;
 #define LWM2M_PATH_LEVEL_RESOURCE 3
 #define LWM2M_PATH_LEVEL_RESOURCE_INST 4
 
+#define LWM2M_OBJLNK_NULL 65535 /* null link */
```
rlubos, line 159: There's `LWM2M_OBJLNK_MAX_ID` defined in `lwm2m.h`, I think it could be reused here (the spec says NULL objlnk should use the MAX_ID for both values).

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+	char *idp = objlnk;
+
+	errno = 0;
+
+	for (int idx = 0; idx < 2; idx++) {
+		unsigned long id = strtoul(idp, &end, 10);
+
+		idp = end;
+
+		if (errno == ERANGE || id > 65535) {
```
rlubos, line 459: s/65535/LWM2M_OBJLNK_MAX_ID

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+#define LOG_MODULE_NAME net_lwm2m_cbor
+#define LOG_LEVEL CONFIG_LWM2M_LOG_LEVEL
+
+#include <logging/log.h>
+#include <sys/util.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+LOG_MODULE_REGISTER(LOG_MODULE_NAME);
```
rlubos, line 16: I'd move this macro call below all includes.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+
+#include "cbor_common.h"
+#include "cbor_decode.h"
+#include "cbor_encode.h"
```
rlubos, line 20: Are those public headers? Anyway, they should use `<>` includes as compiler should search inlcude paths for those.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+	CBOR_TAG_BIGFLOAT_ARR = 5, /* array		Bigfloat */
+	CBOR_TAG_2BASE64URL =  21, /* (any)		Expected conversion to base64url encoding */
+	CBOR_TAG_2BASE64 =     22, /* (any)		Expected conversion to base64 encoding */
+	CBOR_TAG_2BASE16 =     23, /* (any)		Expected conversion to base16 encoding */
+	CBOR_TAG_BSTR =        24, /* byte string	Encoded CBOR data item */
+	CBOR_TAG_URI_TSTR =    32, /* text string	URI */
+	CBOR_TAG_BASE64URL_TSTR = 33, /* text string	base64url */
+	CBOR_TAG_BASE64_TSTR =    34, /* text string	base64 */
+	CBOR_TAG_MIME_TSTR =      36, /* text string	MIME message */
+	CBOR_TAG_CBOR =        55799, /* (any)	Self-described CBOR */
```
rlubos, line 47: Couldn't all values be aligned?

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+};
+
+static size_t put_time(struct lwm2m_output_context *out,
+		      struct lwm2m_obj_path *path, int64_t value)
+{
+	/* CBOR time output format is unspecified but SenML CBOR uses string format.
+	 * Let's stick into the same format with plain CBOR
+	 */
+	struct tm pTime;
```
rlubos, line 56: We should not use camelCase in Zephyr/NCS.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+
+	if (gmtime_r((time_t *)&value, &pTime) != &pTime) {
+		LOG_WRN("unable to convert from secs since Epoch to a date/time construct");
```
rlubos, line 59: Shouldn't this be LOG_ERR? Same for other logs below.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+	int time_str_len = snprintk(time_str, sizeof(time_str),
+			"%04d-%02d-%02dT%02d:%02d:%02d-00:00",
+			pTime.tm_year+1900,
+			pTime.tm_mon+1,
+			pTime.tm_mday,
+			pTime.tm_hour,
+			pTime.tm_min,
+			pTime.tm_sec);
+
+	if (time_str_len < 0 || time_str_len > sizeof(time_str)-1) {
```
rlubos, line 74:
```suggestion
	if (time_str_len < 0 || time_str_len > sizeof(time_str) - 1) {
```

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+}
+
+static size_t put_s64(struct lwm2m_output_context *out,
+		      struct lwm2m_obj_path *path, int64_t value)
+{
+	cbor_state_t states[1];
+	int payload_len = 0;
+
+	/* TODO: remove this check once the CBOR library supports 64-bit values */
+	if (value != (value & 0xFFFFFFFF)) {
```
rlubos, line 116: Could use `UINT32_MAX` here.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+		LOG_WRN("64-bit values are not supported");
+		return -EINVAL;
+	}
+
```
rlubos, line 120: Unneeded empty line.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+	 */
+	struct tm pTime;
+
+	if (gmtime_r((time_t *)&value, &pTime) != &pTime) {
+		LOG_WRN("unable to convert from secs since Epoch to a date/time construct");
+		return -EINVAL;
+	}
+
+	/* Time construct to a string. Time in UTC, offset to local time not known */
+	char time_str[sizeof("1970-01-01T00:00:00-00:00")] = {0};
```
rlubos, line 64: Not a big deal for me, but sometimes upstream reviewers request to define variables at the beginning of the block, could fix that rightaway.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+	if (time_str_len < 0 || time_str_len > sizeof(time_str)-1) {
+		LOG_WRN("unable to form a date/time string");
+		return -EINVAL;
+	}
+
+	cbor_state_t states[1];
+
+	new_state(states, 1, CPKT_BUF_W_REGION(out->out_cpkt), 1);
+
+	/* Are tags required? Spec V1.1 leaves this unspecified but some servers require tags */
```
rlubos, line 83: I think such comments should be tagged with TODO, otherwise they might get actually in unnoticed.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+	bool ret = tag_decode(states, &tag);
+
+	if (ret) {
+		cbor_tag_sz = ICTX_CBOR_R_SZ(states[0].payload, in);
+
+		in->offset += cbor_tag_sz;
+
+		switch (tag) {
+		case CBOR_TAG_TIME_TSTR:
+			goto tstring;
```
rlubos, line 323: Could we avoid using `goto`s like this? They're good for error handling, but in this case I think it's debatable whether readability is improved.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,574 @@
...
+	.get_float = NULL,
+	.get_bool = get_bool,
+	.get_opaque = get_opaque,
+	.get_objlnk = get_objlnk,
+};
+
+int do_read_op_cbor(struct lwm2m_message *msg)
+{
+	/* Can only return single resource */
+	if (msg->path.level < 3U) {
```
rlubos, line 510: Would be good to use symbols for path levels, i. e. `LWM2M_PATH_LEVEL_RESOURCE` etc.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Removed.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
```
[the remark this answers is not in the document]
VeijoPesonen: Fixed.

subsys/net/lib/lwm2m/lwm2m_object.h
VeijoPesonen: Ok, done.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Done.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Done. Yes, those are cddl-gen (zcbor) public headers.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Done.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: My head was spinning when I tried to use structured constructs here. I'll try address this later.

VeijoPesonen:
Changes requested by Robert were introduced in a separate fixup-commit. I'll need to rework this PR to add 64-bit and floating point support once the next release of cddl-gen (zcbor) is out. Due to that reason this PR will stay as a draft.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: At final stage these enums come from zbor side so I'll leave this as it is.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Done.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Fixed.

mbolivar-nordic:
> Ok, I updated the commit logs and realized that the first commit can be upstreamed already.

Awesome, thanks for doing that! I'm dismissing my stale review. I assume you're going to squash the `fixup!` commit at some point before taking this out of draft.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,559 @@
...
+	return tag_sz + str_sz;
+}
+
+static size_t put_s64(struct lwm2m_output_context *out, struct lwm2m_obj_path *path, int64_t value)
+{
+	cbor_state_t states[1];
+	int payload_len;
+
+	/* TODO: remove this check once the CBOR library supports 64-bit values */
+	if (value != (value & UINT32_MAX)) {
```
SeppoTakalo, line 119: What is this comparison trying to do?
This looks like a check that our value can fit into `uint32_t`, but it used binary AND operation for that.
However, as signed negative values are two's complement, `-1` is same as `0xFFFFFFFFFFFFFFFF` when in `int64_t` so it looks like it is not fitting. but equally `int32_t (-1) = 0xFFFFFFFF` which should fit.

So is it that we only support unsigned 32 bit? or do we have signed 32 bits as well? If we do, we should have
```
if (value > INT32_MAX || value < INT32_MIN) {
    LOG_WRN(...);
    return -EINVAL;
}
/* Encode as signed 32 bit, until the library supports 64-bit values */
return put_s32(out, path, (int32_t) value);
```
We have `intx32_encode()` right?

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
SeppoTakalo: I agree with Robert. Use `goto` for error handling only.
Those `tsring`, `tnumberical`  can be small static functions, so your switch-case becomes

```
int err;
switch (tag) {
case A:
    err = get_time_tstring();
    break;
case B:
    err = get_time_tnumerical();
    break;
default:
     ...
     return -EBADMSG;
}
if (err) {
   // handle error
}
// continue from what is 'advance' section
```

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,559 @@
...
+
+	value->obj_id = LWM2M_OBJLNK_MAX_ID;
+	value->obj_inst = LWM2M_OBJLNK_MAX_ID;
+
+	int len = get_string(in, objlnk, sizeof(objlnk) - 1);
+
+	errno = 0;
+
+	for (int idx = 0; idx < 2; idx++) {
+		unsigned long id = strtoul(idp, &end, 10);
```
SeppoTakalo, line 441: This is nitpicking, but to be obviously pedantic, I would follow See CERT-C [ERR30-C](https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152351)
So set `errno = 0;` just before the call, then check the return value as well.

Now you have to look twice, to see if errno is ever cleared after someone sets it.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: For now I only intend to use this function for handling uint32_t values. This will change once the zbor's 64-bit support is available. And yes, we do have `intx32_encode()`

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,559 @@
...
+	if (value != (value & UINT32_MAX)) {
+		LOG_WRN("64-bit values are not supported");
+		return -EINVAL;
+	}
+
+	new_state(states, 1, CPKT_BUF_W_REGION(out->out_cpkt), 1);
+
+	/* TODO: use intx64_encode once/if available */
+	uint32_t valueu32 = (uint32_t)value;
+	bool ret = uintx32_encode(states, &valueu32);
```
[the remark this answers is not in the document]
VeijoPesonen: This is now only used for handling unsigned 32-bit values but will change in the future.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,559 @@
...
+
+tstring:
+	if (!tstrx_decode(states, &hndl)) {
+		/* Might be numerical */
+		goto tnumerical;
+	}
+
+	unsigned long ival = strtoul(time_str, &end, 0);
+
+	errno = 0;
```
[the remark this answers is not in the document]
VeijoPesonen: Yeah, this is a clear bug. To be fixed.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: You're correct. To be fixed.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: I'll fix this.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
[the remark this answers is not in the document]
VeijoPesonen: Fixed.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Fixed.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
```diff
@@ -0,0 +1,588 @@
...
+	new_state(states, 1, ICTX_BUF_R_REGION(in), 1);
+
+	if (!tstrx_decode(states, &hndl)) {
+		return -EBADMSG;
+	}
+
+	/* TODO: decode date/time string */
+	LOG_DBG("decoding a date/time string not supported");
+
+	return -ENOTSUP;
```
VeijoPesonen, line 315: I realized that the date/time strings are not supported yet. Leshan passes time in a numerical format so this hasn't been an issue yet.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
VeijoPesonen: Fixed.  Leshan uses numerical values and I didn't realize earlier that I'm missing date/time string support.

subsys/net/lib/lwm2m/lwm2m_rw_cbor.c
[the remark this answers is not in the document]
VeijoPesonen: This was a bug from the beginning. If we receive a string it's a date/time string - not a numerical string representing seconds since Epoch.

VeijoPesonen:
The documentation build fails because of an undocumented cddl-gen(zcbor) Kconfig option. That is to be fixed on the next release.

VeijoPesonen:
Replaced with PR #785

PR #699, opened 2022-01-26
[nrf fromtree] test/subsys/storage/steram_flash: fix build size issue

nvlsianpu:
This test uses SoC's flash from offset of 64 KB which might
overlap with executable for a target
(for instance nrf52840dk_nrf52840).
This patch moves this region to 128 KB which solves the issue.

mirror of https://github.com/zephyrproject-rtos/zephyr/pull/42192

fixes NCSDK-13488


PerMac:
The fix works

hakonfam (approved):
There's a typo in the commit message (steram -> stream).

You know this better than me @nvlsianpu , but do you need a PR ref in an `[nrf fromlist]` for the upmerge tools to handle the commit correctly, or will it happen anyway?

Anyway, just nits, LGTM

PR #703, opened 2022-01-27
[nrf fromtree] Lwm2m multi-instanced firmware update object

jarlamsa:
As https://github.com/zephyrproject-rtos/zephyr/pull/41402 didn't make it to the upmerge, this PR brings the commits to sdk-zephyr. 
Also https://github.com/zephyrproject-rtos/zephyr/pull/41598 is brought in this PR.

### DIFF
--- a/subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
+++ b/subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
@@ -116,9 +116,9 @@ void lwm2m_firmware_set_update_state(uint8_t state)
 			update_state, state);
 	}
 
-	update_state = state;
-	NOTIFY_OBSERVER(LWM2M_OBJECT_FIRMWARE_ID, 0, FIRMWARE_STATE_ID);
-	LOG_DBG("Update state = %d", update_state);
+	lwm2m_engine_set_u8("5/0/3", state);
+
+	LOG_DBG("Update state = %d", state);
 }
 
 uint8_t lwm2m_firmware_get_update_result(void)
@@ -185,9 +185,9 @@ void lwm2m_firmware_set_update_result(uint8_t result)
 			result, state);
 	}
 
-	update_result = result;
-	NOTIFY_OBSERVER(LWM2M_OBJECT_FIRMWARE_ID, 0, FIRMWARE_UPDATE_RESULT_ID);
-	LOG_DBG("Update result = %d", update_result);
+	lwm2m_engine_set_u8("5/0/5", result);
+
+	LOG_DBG("Update result = %d", result);
 }
 
 static int package_write_cb(uint16_t obj_inst_id, uint16_t res_id,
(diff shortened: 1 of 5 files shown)


rlubos (approved):

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
```diff
@@ -66,41 +73,47 @@ static struct lwm2m_engine_obj_field fields[] = {
...
-extern int lwm2m_firmware_start_transfer(char *package_uri);
+extern int lwm2m_firmware_start_transfer(uint16_t obj_inst_id, char *package_uri);
 #endif
 
+uint8_t lwm2m_firmware_get_update_state_inst(uint16_t obj_inst_id)
+{
+	return update_state[obj_inst_id];
+}
+
 uint8_t lwm2m_firmware_get_update_state(void)
```
VeijoPesonen, line 92: This function should be removed as it's not needed anymore.

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
```diff
@@ -113,67 +126,80 @@ void lwm2m_firmware_set_update_state(uint8_t state)
...
+{
+	lwm2m_firmware_set_update_state_inst(0, state);
+}
+
+uint8_t lwm2m_firmware_get_update_result_inst(uint16_t obj_inst_id)
+{
+	return update_result[obj_inst_id];
 }
 
 uint8_t lwm2m_firmware_get_update_result(void)
```
VeijoPesonen, line 149: This function should be removed as it's not needed anymore.

VeijoPesonen (changes requested):
Those two functions are something I should have spotted myself earlier, sorry for that.

rlubos:
Just note, we shouldn't modify commit cherry-picked from the upstream, as the changes would be lost anyway on the next upmerge. If there are some shortcomings in the code, we should aim to fix them in the upstream instead, and cherry-pick the fixes if urgent.

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
jarlamsa: It is needed for the backwards compatibility

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
jarlamsa: It is needed for the backwards compatibility

jarlamsa:
Manifest PR: https://github.com/nrfconnect/sdk-nrf/pull/6665
@rlubos thanks for guidance.

subsys/net/lib/lwm2m/lwm2m_obj_firmware.c
VeijoPesonen: Then it should be kept on the header file too.

VeijoPesonen (approved):

jarlamsa:
The missing function declarations were merged to zephyr main, cherry-picked that commit. @rlubos is there something still needed for this and the manifest PR to get these merged?

rlubos:
Let's just wait for the CI here, I see the manifest update is green already so I can pull it in

PR #707, opened 2022-01-28
[nrf fromtree] net: sockets: tls: Fix ZSOCK_POLLHUP detection 

rlubos:


pirun (approved):
Approved.
This PR fixes the regression bug introduced in previous commit.
Confirmed on SLM native TLS.

junqingzou (approved):
In case of any local or mbedTLS error, the logic would return POLLERR as default.
This is OK to SLM at least as it handles POLLHUP, POLLERR, POLLIN sequentially.

PR #709, opened 2022-01-31
[nrf_noup]: Disable loading of Kconfig.tls-generic

frkv:
-Has unsupported TLS configurations (TLS 1.0, 1.1)
-Collides with PSA configurations
-Converted Zephyr configuration MBEDTLS_TLS_VERSION_1_2
 to standardized configuration MBEDTLS_SSL_PROTO_TLS1_2
-Disabled Kconfig.tls-generic specific configurations in zephyr
-Disabled usage of MBEDTLS_BUILTIN

ref: NCSDK-13503

### DIFF
--- a/modules/mbedtls/Kconfig.tls-generic
+++ b/modules/mbedtls/Kconfig.tls-generic
@@ -9,6 +9,8 @@ menu "TLS configuration"
 
 menu "Supported TLS version"
 
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
+
 config MBEDTLS_TLS_VERSION_1_0
 	bool "Enable support for TLS 1.0"
 	select MBEDTLS_CIPHER
@@ -33,6 +35,8 @@ config MBEDTLS_DTLS
 	bool "Enable support for DTLS"
 	depends on MBEDTLS_TLS_VERSION_1_1 || MBEDTLS_TLS_VERSION_1_2
 
+endif
+
 config MBEDTLS_SSL_EXPORT_KEYS
 	bool "Enable support for exporting SSL key block and master secret"
 	depends on MBEDTLS_TLS_VERSION_1_0 || MBEDTLS_TLS_VERSION_1_1 || MBEDTLS_TLS_VERSION_1_2
@@ -46,6 +50,8 @@ endmenu
 menu "Ciphersuite configuration"
 
 comment "Supported key exchange modes"
+
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
 
 config MBEDTLS_KEY_EXCHANGE_ALL_ENABLED
 	bool "Enable all available ciphersuite modes"
@@ -80,12 +86,16 @@ config MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED
 		MBEDTLS_KEY_EXCHANGE_ECDHE_PSK_ENABLED || \
 		MBEDTLS_KEY_EXCHANGE_RSA_PSK_ENABLED
 
+endif
+
 config MBEDTLS_PSK_MAX_LEN
 	int "Max size of TLS pre-shared keys"
 	default 32
 	depends on MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED
 	help
 	  Max size of TLS pre-shared keys, in bytes.
+
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
 
 config MBEDTLS_KEY_EXCHANGE_RSA_ENABLED
 	bool "Enable the RSA-only based ciphersuite modes"
@@ -119,7 +129,11 @@ if MBEDTLS_KEY_EXCHANGE_ECDHE_PSK_ENABLED || \
 	MBEDTLS_KEY_EXCHANGE_ECDH_RSA_ENABLED || \
 	MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED
 
+endif
+
 comment "Supported elliptic curves"
+
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
 
 config MBEDTLS_ECP_ALL_ENABLED
 	bool "Enable all available elliptic curves"
@@ -182,8 +196,11 @@ config MBEDTLS_ECP_NIST_OPTIM
 	bool "Enable NSIT curves optimization"
 
 endif
+endif
 
 comment "Supported cipher modes"
+
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
 
 config MBEDTLS_CIPHER_ALL_ENABLED
 	bool "Enable all available ciphers"
@@ -248,7 +265,11 @@ config MBEDTLS_CHACHAPOLY_AEAD_ENABLED
 	bool "Enable the ChaCha20-Poly1305 AEAD algorithm"
 	depends on MBEDTLS_CIPHER_CHACHA20_ENABLED || MBEDTLS_MAC_POLY1305_ENABLED
 
+endif
+
 comment "Supported message authentication methods"
+
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
 
 config MBEDTLS_MAC_ALL_ENABLED
 	bool "Enable all available MAC methods"
@@ -293,9 +314,13 @@ config MBEDTLS_MAC_CMAC_ENABLED
 	bool "Enable the CMAC (Cipher-based Message Authentication Code) mode for block ciphers."
 	depends on MBEDTLS_CIPHER_AES_ENABLED || MBEDTLS_CIPHER_DES_ENABLED
 
+endif
+
 endmenu
 
 comment "Random number generators"
+
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
 
 config MBEDTLS_CTR_DRBG_ENABLED
 	bool "Enable the CTR_DRBG AES-256-based random generator"
@@ -306,13 +331,19 @@ config MBEDTLS_HMAC_DRBG_ENABLED
 	bool "Enable the HMAC_DRBG random generator"
 	select MBEDTLS_MD
 
+endif
+
 comment "Other configurations"
 
 config MBEDTLS_CIPHER
 	bool "Enable the generic cipher layer."
 
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
+
 config MBEDTLS_MD
 	bool "Enable the generic message digest layer."
+
+endif
 
 config MBEDTLS_GENPRIME_ENABLED
 	bool "Enable the prime-number generation code."
@@ -331,9 +362,13 @@ config MBEDTLS_HAVE_ASM
 	  of asymmetric cryptography, however this might have an impact on the
 	  code size.
 
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
+
 config MBEDTLS_ENTROPY_ENABLED
 	bool "Enable mbedTLS generic entropy pool"
 	depends on MBEDTLS_MAC_SHA256_ENABLED || MBEDTLS_MAC_SHA512_ENABLED
+
+endif
 
 config MBEDTLS_OPENTHREAD_OPTIMIZATIONS_ENABLED
 	bool "Enable mbedTLS optimizations for OpenThread"
--- a/modules/mbedtls/configs/config-tls-generic.h
+++ b/modules/mbedtls/configs/config-tls-generic.h
@@ -438,6 +438,10 @@
 #include CONFIG_MBEDTLS_USER_CONFIG_FILE
 #endif
 
+#if defined(CONFIG_NRF_CC3XX_PLATFORM)
+#define MBEDTLS_PLATFORM_ZEROIZE_ALT
+#endif
+
 #include "mbedtls/check_config.h"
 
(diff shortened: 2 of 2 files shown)


modules/mbedtls/Kconfig
```diff
@@ -58,8 +58,6 @@ config MBEDTLS_CFG_FILE
 	  special TLS usage, use available Kconfig options, or select an
 	  alternative config.
 
-rsource "Kconfig.tls-generic"
```
rlubos, line 61: There is a dependency to `MBEDTLS_BUILTIN` inside the file, why would it be an issue for nrf_security then? If it is, maybe we need to wrap this line within `if MBEDTLS_BUILTIN`.

samples/net/lwm2m_client/overlay-dtls.conf
```diff
@@ -3,7 +3,7 @@ CONFIG_LWM2M_PEER_PORT=5684
 
 # Select Zephyr mbedtls
 CONFIG_MBEDTLS=y
-CONFIG_MBEDTLS_TLS_VERSION_1_2=y
+CONFIG_MBEDTLS_SSL_PROTO_TLS1_2=y
```
rlubos, line 6: If we want to have compatible symbol names, we should rather rename them usptream (justifying by the fact its more consistent with mbed TLS nomenclature).

tests/benchmarks/mbedtls/prj.conf
```diff
@@ -12,15 +12,8 @@ CONFIG_MBEDTLS_DEBUG_LEVEL=4
 
 # TLS configuration
 CONFIG_MBEDTLS=y
-CONFIG_MBEDTLS_BUILTIN=y
```
rlubos, line 15: That should be ok, the sample could use the default configuration used in the system.

tests/benchmarks/mbedtls/prj.conf
```diff
@@ -12,15 +12,8 @@ CONFIG_MBEDTLS_DEBUG_LEVEL=4
...
 CONFIG_MBEDTLS_HEAP_SIZE=64000
 CONFIG_MBEDTLS_SSL_MAX_CONTENT_LEN=2048
 
-CONFIG_MBEDTLS_TLS_VERSION_1_2=y
-CONFIG_MBEDTLS_KEY_EXCHANGE_ALL_ENABLED=y
-CONFIG_MBEDTLS_CIPHER_ALL_ENABLED=y
-CONFIG_MBEDTLS_ECP_ALL_ENABLED=y
-CONFIG_MBEDTLS_MAC_ALL_ENABLED=y
-CONFIG_MBEDTLS_GENPRIME_ENABLED=y
-CONFIG_MBEDTLS_HMAC_DRBG_ENABLED=y
```
rlubos, line 26: Those configs were introduced to make sure all of the algorithms supported by mbedTLS are tested. If you want to use the benchmark from the upstream, perhaps you should provide similar in nrf_security.

rlubos (commented):
I'm not really fond of such invasive `nrf noup` changes, we should rather try to align the builtin mbed TLS with nrf_security, and vice versa.

tests/benchmarks/mbedtls/prj.conf
SebastianBoe: @frkv : For which boards was MBEDTLS_BUILTIN causing trouble? If it was a non-secure board then that should be fixed by the patches I have in PR.

SebastianBoe:
@frkv : Ping me when CI is green and I can review.

tejlmand:
The commit message says:
> -Disabled Kconfig.tls-generic specific configurations in zephyr
> -Disabled usage of MBEDTLS_BUILTIN

and there is a reference to NCSDK-13503.

But from this commit message and the referenced task I still don't understand why this is needed ?

Exactly what is the issue you are trying to solve / what is breaking when `Kconfig,tls-generic` or `MBEDTLS_BUILTIN` is used.

modules/mbedtls/Kconfig
frkv: Comment is outdated, We still require more things inside, so it is still used

samples/net/lwm2m_client/overlay-dtls.conf
frkv: @rlubos: If you have the capacity to do this, then that would be great!

MBEDTLS_TLS_VERSION_1_2 is a zephyr-specific configuration, from what I understand. I haven't seen this configuration elsewhere

tests/benchmarks/mbedtls/prj.conf
frkv: It will cause troubles on all nRF-boards if nrf_security is used. 

If you have another fix for this, then I'm happy to accept that!

SebastianBoe: NRF_SECURITY is doing
	select DISABLE_MBEDTLS_BUILTIN if MBEDTLS

so I don't think this patch is necessary, I suggest we remove this patch and re-test.

SebastianBoe: will look into this

SebastianBoe: this sample is not building with any nrf board I can find.

So I am unable to test the patch.

on NCS main it overflows, in this PR it has a Kconfig issue.

Will debug.

mbolivar-nordic:
@frkv I've already fixed https://github.com/nrfconnect/sdk-zephyr/pull/722 for existing patches, but it's `[nrf noup] subsystem: blah blah blah`, not `[nrf_noup]: subsystem: blah blah blah`; please fix

mbolivar-nordic:
Addendum: that's why compliance is failing:

```

Run git config --global user.email "you@example.com"
origin	https://github.com/nrfconnect/sdk-zephyr (fetch)
origin	https://github.com/nrfconnect/sdk-zephyr (push)
a6068a8cb7 [nrf_noup]: Disable configurations in Kconfig.tls-generic
Sauce tag missing
```

Not sure how the other patches in main were merged in the first place

tests/benchmarks/mbedtls/prj.conf
frkv: > @frkv I've already fixed #722 for existing patches, but it's `[nrf noup] subsystem: blah blah blah`, not `[nrf_noup]: subsystem: blah blah blah`; please fix

Will fix

frkv:
> The commit message says:
> 
> > -Disabled Kconfig.tls-generic specific configurations in zephyr
> > -Disabled usage of MBEDTLS_BUILTIN
> 
> and there is a reference to NCSDK-13503.
> 
> But from this commit message and the referenced task I still don't understand why this is needed ?
> 
> Exactly what is the issue you are trying to solve / what is breaking when `Kconfig,tls-generic` or `MBEDTLS_BUILTIN` is used.

You forgot to put this bullet-point:
-Collides with PSA configurations

I can flesh this out more properly in the commit message as I'm doing some cleanup

samples/net/lwm2m_client/overlay-dtls.conf
frkv: We are leaving this for a later time. Reverted removal

tests/benchmarks/mbedtls/prj.conf
frkv: Fixed and cleaned up the commit substantially. Now it only checks for nrf_security being enabled and adds a minor fix in the config-file to prevent mbedtls_platform_zeroize from being duplicated (present in nrf_cc3xx_platform)

tests/benchmarks/mbedtls/prj.conf
frkv: Removed these. No longer needed to do

joerchan (approved):

tejlmand (changes requested):
If the Zephyr mbedTLS Kconfig is causing conflict with nRF Security, then why don't we completely disable the upstream mbed TLS Kconfig when nRF Security is in use ?

mbedTLS uses `kconfig-ext: True` here:
https://github.com/nrfconnect/sdk-mbedtls/blob/68dfba636301bd618bc9b5c1a946f358ad8714ee/zephyr/module.yml#L3

As `nrf` comes first in the `MODULE_EXT_ROOT`, we can have our own mbedTLS Kconfig, and then source the upstream version when `CONFIG_NRF_SECURITY=n`, siimilar to what we do in TF-M here:
https://github.com/nrfconnect/sdk-nrf/blob/51b5383b2cf3dbf9ab51cca4177a59f0b73bfc71/modules/trusted-firmware-m/Kconfig#L44-L52

This means we are in complete control on when upstream mbed TLS kconfig is available, without adding `[nrf noup]` patches.

Has such strategy been considered ?
If it has, what were the shortcommings in this case ?

modules/mbedtls/Kconfig.tls-generic
```diff
@@ -9,6 +9,8 @@ menu "TLS configuration"
 
 menu "Supported TLS version"
 
+if !(NRF_SECURITY || NORDIC_SECURITY_BACKEND)
```
SebastianBoe, line 12: I thought one of these selected the other meaning we only need to be conditional on one of them?

modules/mbedtls/configs/config-tls-generic.h
```diff
@@ -438,6 +438,10 @@
 #include CONFIG_MBEDTLS_USER_CONFIG_FILE
 #endif
 
+#if defined(CONFIG_NRF_CC3XX_PLATFORM)
+#define MBEDTLS_PLATFORM_ZEROIZE_ALT
+#endif
+
```
SebastianBoe, line 444: Strange. I thought we generated the config.

It's not clear to me when this file is used.

But perhaps we can add the MBEDTLS_PLATFORM_ZEROIZE_ALT define from the nrf build system so we avoid this patch.

It should also be in it's own commit.

SebastianBoe:
@tejlmand : Interesting approach Torsten. But I don't understand how we would 
conditionally disable the mbedtls kconfig.

Could you write a quick patch to demonstrate?

IMHO the patchset here is clean enough for the release, and the noups should be removed
after release. Possibly with Torsten's suggestion.

But that's not really my call.

SebastianBoe (approved):

frkv:
> If the Zephyr mbedTLS Kconfig is causing conflict with nRF Security, then why don't we completely disable the upstream mbed TLS Kconfig when nRF Security is in use ?
> 
> mbedTLS uses `kconfig-ext: True` here: https://github.com/nrfconnect/sdk-mbedtls/blob/68dfba636301bd618bc9b5c1a946f358ad8714ee/zephyr/module.yml#L3
> 
> As `nrf` comes first in the `MODULE_EXT_ROOT`, we can have our own mbedTLS Kconfig, and then source the upstream version when `CONFIG_NRF_SECURITY=n`, siimilar to what we do in TF-M here: https://github.com/nrfconnect/sdk-nrf/blob/51b5383b2cf3dbf9ab51cca4177a59f0b73bfc71/modules/trusted-firmware-m/Kconfig#L44-L52
> 
> This means we are in complete control on when upstream mbed TLS kconfig is available, without adding `[nrf noup]` patches.
> 
> Has such strategy been considered ? If it has, what were the shortcommings in this case ?

Given the short timing of this, and to keep the information going, This is added to a task to be handled after RC-1 tagging
ref: NCSDK-13734

All my efforts trying to do what you suggest gave med dependency loops on colliding configurations

modules/mbedtls/Kconfig.tls-generic
frkv: No. They can be either/or

modules/mbedtls/configs/config-tls-generic.h
frkv: The difference is in the new cc3xx runtime library which has this symbol

SebastianBoe:
> All my efforts trying to do what you suggest gave med dependency loops on colliding configurations

This is probably because with the new approach we would have some mbedtls Kconfig dependent on nrf_security whereas before they were not. For sure, it's a big job to make the whole mbedtls dependent on nrf_security.

But we should still try to remove the noup (after release).

modules/mbedtls/configs/config-tls-generic.h
SebastianBoe: I don't know what that means, but let's add this to the task we already have about removing noups to the Zephyr repo.

tejlmand:
> @tejlmand : Interesting approach Torsten. But I don't understand how we would conditionally disable the mbedtls kconfig.
> 
> Could you write a quick patch to demonstrate?

Sure, like this:
https://github.com/tejlmand/fw-nrfconnect-nrf/commit/8e11261cc5b033f97fb4de41c307b8354afac625

Now, whenever Zephyr mbedTLS module is available, the Kconfig is sourced from `sdk-nrf/modules/mbedtls/Kconfig`.
And in there we can decide:
- If / when Zephyr upstream kconfig should be used.
   the `if` is in practive adding a `depends on !NRF_SECURITY` on all upstream symbols.
- We can decide not to source upstream at all, and have everything controlled through our own Kconfig.
   Note: doing so will of course require that we have symbols available which might be used in upstream `prj.conf`

tejlmand (approved):
approved with reservations.

Would prefer to the solution being tested here: https://github.com/nrfconnect/sdk-zephyr/pull/725
but will let the decision on which solution to merge be at the discretion of the merger of this PR.

mbolivar-nordic (approved):
thanks for fixing the commit log!

PR #713, opened 2022-01-31
[nrf noup] action: clang: parallel execution

PerMac:
Bring back parallel execution to the clang action. In the upstream
they went to use a single dedicated powerfull agent to run this. We
need to keep this as we have to base on agents available in github's
cloud.

### DIFF
--- a/.github/workflows/clang.yaml
+++ b/.github/workflows/clang.yaml
@@ -21,12 +21,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


nvlsianpu (commented):
This downstream path is worth to be added. The Patch is not complex and gives a lot of timing improvement. Also command-line changes here look fine.

nvlsianpu (approved):

SebastianBoe (approved):

tejlmand (approved):

PR #716, opened 2022-02-01
[nrf noup] linker: only add Sx labels if SECURE_BOOT is enabled

kapi-no:
If CONFIG_SECURE_BOOT is not enabled then these symbols will
not be defined.

Ref: NCSDK-6211

### DIFF
--- a/include/zephyr/arch/arm/aarch32/cortex_m/scripts/linker.ld
+++ b/include/zephyr/arch/arm/aarch32/cortex_m/scripts/linker.ld
@@ -31,7 +31,7 @@
 
 #include <pm_config.h>
 
-#if CONFIG_NCS_IS_VARIANT_IMAGE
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:

tejlmand (changes requested):
Changes looks good, but please mark the commit  as a `fixup` commit, by having the following line in the commit message (For example line 3 in commit message)
```
fixup! [nrf noup] tree-wide: support NCS Partition Manager (PM) definitions
```

kapi-no:
Rebased my PR. @tejlmand, I changed the commit description according to your suggestion.

kapi-no:
Rebased after upmerge. The diff is the same

nvlsianpu (approved):

rlubos (approved):

PR #720, opened 2022-02-02
[nrf noup] soc: flash: Add tfm read service to nRF SoC Flash driver

sigvartmh:
This adds the functionality to read memory which are residing in the
secure part of flash from a non-secure application when building with
TF-M.

Currently we check if the address space is not inside the non-secure
applapplication address space. If the address space is outside we do a
request to TF-M so that the secure application can either return success
or fail if the address space is not allowed to be read.

Ref. NCSDK-13085


sigvartmh:
@carlescufi @SebastianBoe @joerchan:

So I'm just going to restart this discussion which is split over #662 and #689.

I do see the concern about putting this upstream but we don't have either PM_APP_ADDRESS or PM defined in upstream. So how do you see this patch fitting in upstream? 

> @sebas@ One noup commit relevant for SPM, which will be deprecated and then we can drop the commit. And one fromlist commit relevant for TF-M. That we would need to post a PR for.

> The patch adding tfm_read_service should be in a fromlist commit using DT[0].
[0] https://github.com/zephyrproject-rtos/zephyr/pull/41756/files

Is this purely a draft or can I just grab this and fromlist it?

> The patch adding PM support should be in a noup commit that removes the DT references.

After this is incorporated, is it better to try to map FLASH_AREA_OFFSET to something instead of removing the DT so that calling FLASH_AREA_OFFSET(image_0_nonsecure) would just resolve to what would be expected?

> The patch adding SPM support should be in it's own noup commit.

The patch does not contain any references to SPM it just moves around on the SPM defines but I can try to put them into their own commit so that they can easily be scrubbed once SPM is removed.

drivers/flash/soc_flash_nrf.c
```diff
@@ -153,12 +157,35 @@ static int flash_nrf_read(const struct device *dev, off_t addr,
...
+	// is not in the non-secure image we will not be able to read it
+	// directly and must instead invoke a secure service to read out
+	// the flash.
+#ifdef CONFIG_ARM_NONSECURE_FIRMWARE
+	// We assume that image_0_nonsecure is the only non-secure
+	// partition. This will need to be updated when it is also
+	// possible to execute from slot 1. Partitions that are assumed to
+	// be secure are: mcuboot, image_0, image_scratch, and storage.
+	if (!is_within_bounds(addr, len, FLASH_AREA_OFFSET(image_0_nonsecure),
+			      FLASH_AREA_SIZE(image_0_nonsecure))) {
```
sigvartmh, line 170: I think maybe we can try to map these in FLASH_LAYOUT with PM to what was used before.

SebastianBoe:
> I do see the concern about putting this upstream but we don't have either PM_APP_ADDRESS or PM defined in upstream. So how do you see this patch fitting in upstream?

The draft PR doesn't reference PM and can be sent upsteam.

> Is this purely a draft or can I just grab this and fromlist it?

I reviewed it and found no issues. But I see the draft PR says it is blocked by other commits. We need to check if the tfm_platform_mem_read service has been merged before posting this PR. Also this commit needs to be checked if has been merged upstream: https://github.com/zephyrproject-rtos/zephyr/pull/41756/commits/7a2d19e5d89c2312233905175b4fc256f7872910

> After this is incorporated, is it better to try to map FLASH_AREA_OFFSET to something instead of removing the DT so that calling FLASH_AREA_OFFSET(image_0_nonsecure) would just resolve to what would be expected?

If that's possible, then sure, that sounds great. But I'm not familiar enough with DT to say if that is safe/possible.

SebastianBoe (changes requested):
Need to avoid this noup commit by sending a PR upstream.

724f1bdecb0ddc2dc773c0ebfd1f176b896c4425

sigvartmh:
Yeah I'll remove 724f1bd just have it here for history atm.

SebastianBoe:
k, let me know when this is ready for review

drivers/flash/soc_flash_nrf.c
joerchan: @sigvartmh Could you include the tfm_read_ranges.h header and check against it instead? That should work for upstream I think.

sigvartmh: Seems we have the image_0_nonsecure already available downstream from https://github.com/nrfconnect/sdk-nrf/blob/59cd80ab8e3d7422fe2624104258cdf0bd2b26e9/include/flash_map_pm.h#L16

I only found `tfm_read_ranges.h` in the nrf repository.

sigvartmh:
Ping @SebastianBoe

sigvartmh:
Added reference to PR to the fromlist commit which is not upstream yet https://github.com/zephyrproject-rtos/zephyr/pull/42474

drivers/flash/soc_flash_nrf.c
sigvartmh: seems the defines are not always available (aka if you are not using mcuboot) in downstream. Added some more checks for it.

drivers/flash/soc_flash_nrf.c
```diff
@@ -153,12 +168,42 @@ static int flash_nrf_read(const struct device *dev, off_t addr,
...
 
-#if CONFIG_ARM_NONSECURE_FIRMWARE && CONFIG_SPM && USE_PARTITION_MANAGER \
-	&& CONFIG_SPM_SECURE_SERVICES
-	if (addr < PM_APP_ADDRESS) {
+	/* If the currently executing image is non-secure and the address
+	 * is not in the non-secure image we will not be able to read it
+	 * directly and must instead invoke a secure service to read out
+	 * the flash.
+	 */
+#ifdef CONFIG_ARM_NONSECURE_FIRMWARE
```
joerchan, line 176: This one should have been CONFIG_BUILD_WITH_TFM.

joerchan (approved):

sigvartmh:
@nvlsianpu I see that you have approved the upstream PR do you mind having a look at this one too?

drivers/flash/soc_flash_nrf.c
```diff
@@ -179,8 +179,13 @@ static int flash_nrf_read(const struct device *dev, off_t addr,
 	 * possible to execute from slot 1. Partitions that are assumed to
 	 * be secure are: mcuboot, image_0, image_scratch, and storage.
 	 */
+#if FLASH_AREA_LABEL_EXISTS(image_0_nonsecure)
 	if (!is_within_bounds(addr, len, FLASH_AREA_OFFSET(image_0_nonsecure),
 			      FLASH_AREA_SIZE(image_0_nonsecure))) {
+#else
+	if (!is_within_bounds(addr, len, PM_APP_ADDRESS, PM_APP_END_ADDRESS))
+#endif
+	{
```
SebastianBoe, line 188: Is it correct that the first if has a {, but the second does not?

sigvartmh: I was supposed to remove the one at line 184 but I guess having it in 186 would be cleaner?

drivers/flash/soc_flash_nrf.c
sigvartmh: That won't work with SPM

SebastianBoe (approved):

nvlsianpu (approved):

sigvartmh:
I've run the failing ci tests locally and they seem like false failures

joerchan (approved):

tejlmand (commented):
@sigvartmh please check if the `[nrf toup]` commits in this PR are candidates for future squash / fixup commits into other `[nrf noup]`.

I imagine https://github.com/nrfconnect/sdk-zephyr/pull/720/commits/030afa5040d9bca32adf3b33a6179a1f9591fc9f could be a squash candidate for squashing with c03da1480a3e0cfd2227928358dfe674dcbaac6c

mbolivar-nordic:

sigvartmh:
Functionality seems unwanted upstream https://github.com/zephyrproject-rtos/zephyr/pull/42474 I'll look into exposing this functionality for our use-case through the TF-M FWU Shim for MCUBoot.

PR #734, opened 2022-02-10
[nrf fromlist] boards: thingy53_nrf5340:  thingy53 bugs found on testing

zycz:
PR solves bugs related to machine learning application on thingy 53


emob-nordic (changes requested):
We need to add another commit that would enable regulator (`force_pwm_node`) in machine_learning application and add entry in `ug_thingy53.rst` that this regulator must be enabled if the application use any sensors connected to SPI3 line (confirm that) to avoid voltage drops and in the end false sensor readouts (please reword that and correct if im wrong).
The last commit i believe should not have any sauce tag.

emob-nordic:
An good place to mention about SPI enabling `force_pwm_mode` is in section `Thingy:53 application guide` you could add it as `.. note::` or `.. warning::`. Im adding @greg-fer  for doc updates.

shantha-14:
Thing53 will have its own release and it is too late for NCS 1.9.0

tejlmand (approved):

mbolivar-nordic:

MarekPieta:
This will go in together with the ongoing Zephyr upmerge. We can close it.

PR #740, opened 2022-02-17
[nrf fromlist] doc: application: tweak experimental features text

mbolivar-nordic:
Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/42876

We use a specific experimental Kconfig option when describing the
behavior of CONFIG_WARN_EXPERIMENTAL.

However, this option may not be experimental forever. This therefore
may go stale. Rather than try to keep generic documentation about
experimental options up to date with whatever the bluetooth subsystem
happens to consider experimental or not, use a placeholder CONFIG_FOO
instead of a real option.

### DIFF
--- a/doc/application/index.rst
+++ b/doc/application/index.rst
@@ -533,13 +533,13 @@ at CMake configure time if any experimental feature is enabled.
 
    CONFIG_WARN_EXPERIMENTAL=y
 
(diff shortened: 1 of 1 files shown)


kittydepa (approved):

mbolivar-nordic:
Rebased now that the breathe fix PR is in

PR #748, opened 2022-02-25
net: sockets: Fix build issue with posix

krish2718:
Cheery pick commit 697f415bbf ("net: sockets: Use struct timeval provided by libc")
from upstream Zephyr that fixes a build issue with using posix/time.h
along with net/socket.h in the same file.

### DIFF
--- a/include/net/socket_select.h
+++ b/include/net/socket_select.h
@@ -109,7 +109,6 @@ void ZSOCK_FD_SET(int fd, zsock_fd_set *set);
 #ifdef CONFIG_NET_SOCKETS_POSIX_NAMES
 
 #define fd_set zsock_fd_set
-#define timeval zsock_timeval
 #define FD_SETSIZE ZSOCK_FD_SETSIZE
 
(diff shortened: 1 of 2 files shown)


rlubos:
Hi, cherry-picked commits should be an almost exact copy of the upstream commit (including commit message). The only difference is the commit title, which should beging with `[nrf fromtree] ` prefix.

But I'd rather suggest waiting for the next upmerge, where this change should be integrated with our fork.

krish2718:
Thanks @rlubos I have a build failure for nRF5340 where it doesn't work with Zephyr because of https://github.com/zephyrproject-rtos/zephyr/issues/35512 and nRF doesn't build because of this fix missing in nRF. Anyways, we can wait for the upmerge, any ETA for that?

rlubos:
AFAIK the upmerge is going to start next week (this is not firm though, may be a subject of change).

krish2718:
Okay, closing this PR as its in the near future.

PR #749, opened 2022-02-25
Fix check_compliance.

tejlmand:
Check compliance is failing in sdk-zephyr.

This PR fixes check compliance for Kconfig by defining symbols that exists outside Zephyr.

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -58,7 +58,7 @@ jobs:
         git config --global user.name "Your Name"
         git remote -v
         # Sauce tag checks before rebasing
-        git log --oneline --first-parent origin/${BASE_REF}..HEAD | grep -E -v "\[nrf (mergeup|fromtree|fromlist|noup)\]" && { echo 'Sauce tag missing'; exit 1; }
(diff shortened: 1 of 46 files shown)


PerMac (approved):

frkv (approved):
LGTM

PerMac:
do we need a manifest update PR for this one?

tejlmand:
> do we need a manifest update PR in sdk-nrf for this one?

Created one here to run the builds for tfm and crypto:
https://github.com/nrfconnect/sdk-nrf/pull/7011

The changes should not have any impact, but it's always nice to run CI.
We can always decide not to merge the manifest PR, as there is not really a reason in this case.

tejlmand:
The changes in this PR triggers a Kconfig dependency loop when used in NCS.

Need to find a solution that is safe for all use cases.

carlescufi (changes requested):
Can we put all the Kconfig options that we inject here in a new file, `Kconfig.nrf` that we keep in a single patch?

mbolivar-nordic:
stale

PerMac:
Can we keep this open as a reminder, until Torsten is back? This fix is still needed.

jfischer-no:
> Can we keep this open as a reminder, until Torsten is back? This fix is still needed.

Still needed? Can it be closed?

PR #759, opened 2022-03-03
IPC service support

kapi-no:
In the _[nrf fromlist] bluetooth: hci: rpmsg: use ipc service library_, resolve the conflicts by discarding nrf53_sync_rtc changes which are not present yet in downstream.


tejlmand:
@kapi-no `[nrf fromlist]` commits should always have a reference to their upstream counterparts.

Also note that https://github.com/zephyrproject-rtos/zephyr/pull/43412 has been merged and can thus be changed into a `[nrf fromtree]` instead.

PR #760, opened 2022-03-03
Upmerge 22-03-03

gmarull:


pdunaj:
Hi @gmarull , my understanding is that `ed5e06420e123641e5471a576c907425ee51eba9` is the base commit for the upmerge. Can you confirm this is correct and won't change?

tejlmand (approved):
lgtm

PerMac:
@gmarull I see compliance is failing again
@carlescufi Scancode fails on files that were merged with red license check already in the upstream. Should it be ignored? https://github.com/zephyrproject-rtos/zephyr/pull/41729
https://github.com/zephyrproject-rtos/zephyr/pull/33948

subsys/ipc/ipc_service/backends/ipc_rpmsg_static_vrings.c
```diff
@@ -500,4 +502,26 @@ static int backend_init(const struct device *instance)
...
+	const struct backend_config_t *backend_config;
+	const struct backend_config_t backend_configs[] = {
+		DT_INST_FOREACH_STATUS_OKAY(BACKEND_CONFIG_DEFINE)
+	};
+
+	for (backend_config = backend_configs;
+	     backend_config < backend_configs + ARRAY_SIZE(backend_configs);
+	     backend_config++) {
+		if (backend_config->role == ROLE_HOST) {
+			memset((void *) backend_config->shm_addr, 0, backend_config->shm_size);
```
hubertmis, line 520: nit: should be enough to set 4 bytes:
```
memset((void *) backend_config->shm_addr, 0, VDEV_STATUS_SIZE);
```

PerMac:
6b578cf127 [nrf toup] ipc: backends: rpmsg: initialize shared memory to zero <- has an incorrect sauce tag and makes the compliance failing.

subsys/ipc/ipc_service/backends/ipc_rpmsg_static_vrings.c
gmarull: fixed up

PR #761, opened 2022-03-10
[nrf noup] net: mqtt: add native TLS support

MirkoCovizzi:
This commit adds an extra parameter in the configuration
structure to configure native TLS support at runtime.

Signed-off-by: Mirko Covizzi <mirko.covizzi@nordicsemi.no>

I found it's not necessary to cascade the change to every single library and sample since the MQTT context is always defined static, hence the `set_native_tls` parameter is `false` by default. If support for native TLS is needed, other commits can be added to address that.

### DIFF
--- a/doc/reference/networking/mqtt.rst
+++ b/doc/reference/networking/mqtt.rst
@@ -149,6 +149,7 @@ additional configuration information:
    tls_config->sec_tag_list = m_sec_tags;
    tls_config->sec_tag_count = ARRAY_SIZE(m_sec_tags);
    tls_config->hostname = MQTT_BROKER_HOSTNAME;
+   tls_config->set_native_tls = true;
 
 In this sample code, the ``m_sec_tags`` array holds a list of tags, referencing TLS
 credentials that the MQTT library should use for authentication. We do not specify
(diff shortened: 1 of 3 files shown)


rlubos (approved):
Not a fan of noup commits, but I guess at this point we don't have a better option.

Why is this change DNM?

MirkoCovizzi:
> Not a fan of noup commits, but I guess at this point we don't have a better option.
> 
> Why is this change DNM?

I have the habit of always putting DNM label when the PR is still under work in progress. I just forgot to remove it.

lemrey (approved):

eivindj-nordic (approved):

doc/reference/networking/mqtt.rst
```diff
@@ -161,6 +162,9 @@ Note, that TLS credentials referenced by the ``m_sec_tags`` array must be
 registered in the system first. For more information on how to do that, refer
 to :ref:`secure sockets documentation <secure_sockets_interface>`.
 
+Finally, we can optionally set ``set_native_tls`` to enable native TLS support
+instead of offloading TLS operations to the modem.
```
divipillai, line 166:
```suggestion
Finally, you can optionally set ``set_native_tls`` to enable native TLS support instead of offloading TLS operations to the modem.
```

divipillai (approved):

PR #762, opened 2022-03-11
Bluetooth: host: Accept duplicate device

ibaz-nordic:
When duplicate device is found, controller can either return a success or Invalid HCI Command Parameters (0x12).
Host now will not treat duplicate device as error (DRGN-15879).

Signed-off-by: Azizah Ibrahim <azizah.ibrahim@nordicsemi.no>

PR #763, opened 2022-03-11
[nrf noup] Bluetooth: host: Accept duplicate device

ibaz-nordic:
When duplicate device is found, controller can either return a success or Invalid HCI Command Parameters (0x12).
Host now will not treat duplicate device as error (DRGN-15879).

Signed-off-by: Azizah Ibrahim <azizah.ibrahim@nordicsemi.no>

### DIFF
--- a/subsys/bluetooth/host/id.c
+++ b/subsys/bluetooth/host/id.c
@@ -775,7 +775,10 @@ void bt_id_add(struct bt_keys *keys)
 
 	err = hci_id_add(keys->id, &keys->addr, keys->irk.val);
 	if (err) {
-		BT_ERR("Failed to add IRK to controller");
+		/* Duplicate device is OK */
+		if (err != BT_HCI_ERR_INVALID_PARAM) {
+			BT_ERR("Failed to add IRK to controller");
+		}
 		goto done;
 	}
 


subsys/bluetooth/host/id.c
```diff
@@ -775,7 +775,10 @@ void bt_id_add(struct bt_keys *keys)
 
 	err = hci_id_add(keys->id, &keys->addr, keys->irk.val);
 	if (err) {
-		BT_ERR("Failed to add IRK to controller");
+		/* Duplicate device is OK */
+		if (err != BT_HCI_ERR_INVALID_PARAM) {
```
MarekPieta, line 779: The fix does not work. The `hci_id_add` returns Zephyr error code (`-EIO` == `-5`) so the comparison would not catch it.
The actual HCI error code is not returned by the `hci_id_add`, so maybe we could handle the error in the `hci_id_add`?

Also is the `BT_HCI_ERR_INVALID_PARAM` returned only on duplicated IRK? If not then we may hide other errors from the user which is not good.

MarekPieta (changes requested):
See comment above

rugeGerritsen:
@ibaz-nordic , please see the comments in https://github.com/nrfconnect/sdk-zephyr/pull/762

subsys/bluetooth/host/id.c
ibaz-nordic: Cheers for the input.  I'll look at resolving the problem another way.

ibaz-nordic:
Will be replaced with another solution in upstream

PR #764, opened 2022-03-14
[nrf fromtree] modules: hal_nordic: Improve reservation of resources for BT_CTLR

anangl:
Instead of including from nrfx_glue.h a specific Zephyr Bluetooth
controller header file that defines PPI and GPIOTE resources to be
reserved for exclusive use by the controller, include a file with
only a fixed name and expect the chosen Bluetooth controller to
provide the location of this file in include paths. This way, when
a different Bluetooth controller implementation is used downstream,
a different file can be easily pointed to.


carlescufi (approved):

rugeGerritsen:
Lets wait with merging this one until the upstream PR is approved

rugeGerritsen (approved):

cvinayak (approved):

PR #765, opened 2022-03-14
[nrf fromlist] cmake: fix improper use of separate_arguments

hakonfam:
PR: https://github.com/zephyrproject-rtos/zephyr/pull/43761

The use of `separate_arguments()` in extra_flags.cmake and
unittest.cmake prevents the use of proper CMake lists.

A CMake list can be specified on command line as:
`-DMY_LIST="itemA;itemB;itemC"`

Users are expected to be able to provide extra compile flags using CMake
lists, like `-DEXTRA_CFLAGS="--flag1;--flag2"` but also using space
separated lists, like: `-DEXTRA_CFLAGS="--flag1 --flag2"`.

Instead of using `separate_arguments(...)` for this purposes, then it
is better to simply replace spaces with `;` to create a proper CMake
list, regardless of the input form used by the user.

This is same principle that is used for CONF_FILE, DTC_OVERLAY_FILE, and
other settings.


hakonfam:
Pulled in by earlier upmerge

PR #772, opened 2022-03-24
[nrf fromtree] drivers: uart_nrfx_uarte: Fix double Rx buffer releasing

KAGA164:
Fixes a case for the asynchronous UART API where
only single buffer was attached to the UART, after
filling this buffer release event was trigger first
time and after disabling UART the release event was
triggered again for the same buffer.

Signed-off-by: Kamil Gawor <Kamil.Gawor@nordicsemi.no>
(cherry picked from commit ba6d9139bbbda3211e0c1f65f4ccca0719f17fa5)


anangl (approved):

carlescufi (approved):

KAGA164:
I will try rebase

PR #782, opened 2022-04-11
OpenThread upmerge 20220407

edmont:
Selected OpenThread commits from upstream.

### DIFF
--- a/subsys/net/lib/openthread/platform/platform.c
+++ b/subsys/net/lib/openthread/platform/platform.c
@@ -21,8 +21,8 @@ void otSysInit(int argc, char *argv[])
 	ARG_UNUSED(argc);
 	ARG_UNUSED(argv);
 
-	platformAlarmInit();
 	platformRadioInit();
+	platformAlarmInit();
 }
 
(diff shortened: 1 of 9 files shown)


subsys/net/lib/openthread/platform/platform.c
```diff
@@ -21,8 +21,8 @@ void otSysInit(int argc, char *argv[])
 	ARG_UNUSED(argc);
 	ARG_UNUSED(argv);
 
-	platformAlarmInit();
 	platformRadioInit();
+	platformAlarmInit();
```
canisLupus1313, line 25: Why this order is changed? Don't we use timers in the driver?

edmont: I don't think we use `ot*` timing functions in the radio driver implementation. @lmaciejonczyk knows the details about the need of this change for aligning platform and radio timing.

rlubos (approved):

canisLupus1313 (approved):

lmaciejonczyk (approved):

PR #788, opened 2022-04-25
[nrf fromlist] ipc_service: static_vrings: Set WQ default type to PRIO_COOP

ArekBalysNordic:
Cherry-pick commit form 
upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/44981 
which fixes an issue with IPC's priority in the Matter Network.


carlocaione (approved):
LGTM but please not that the changes introduced by this commit are valid only and only if in the DT you do **not** have the `zephyr,priority` attribute. Otherwise the priority defined in the DT will be used instead of the fallback value.

dts/bindings/ipc/zephyr,ipc-openamp-static-vrings.yaml
```diff
@@ -42,4 +42,5 @@ properties:
       or for K_PRIO_PREEMPT(2)
         priority = <2 PRIO_PREEMPT>;
 
-      When this property is missing a default priority of 0 is assumed.
+      When this property is missing a default priority of <0 PRIO_COOP> is
```
pdunaj, line 45: `K_PRIO_COOP(0)` will set the highest coop priority. Isn't that rather radical?

pdunaj: ok. I guess it won't be obvious to select anyway. I would go for -1 so it would be matching system workqueue (to not starve the system in case of IPC data flood).
I guess if this is a real problem somebody will report it at some point.

emob-nordic (approved):

pdunaj (approved):

ArekBalysNordic:
@pdunaj Why do you need this commit right now? It is necessary for the NCS 2.0.0 release?

emob-nordic:
> @pdunaj Why do you need this commit right now? Is it necessary for the NCS 2.0.0 release?

Yes, it is.

ArekBalysNordic:
Closing -  New PR with [nrf fromtree] will be opened. This commit can be reopen: https://github.com/nrfconnect/sdk-zephyr/pull/812

PR #785, opened 2022-04-14
LwM2M plain CBOR content format support

VeijoPesonen:
Just to make clear this content format shouldn't be confused with LwM2M CBOR which is introduced in LwM2M v1.2.  Plain CBOR content format is part of the LwM2M version 1.1.

The content format can be pushed to upstream zephyr once the ZCBOR CBOR library has been upstreamed too.

PR #789, opened 2022-05-02
[nrf fromtree] net: lwm2m: CBOR and SenML CBOR content format support

VeijoPesonen:
Introduces two new content formats for LwM2M v1.1 - Raw CBOR and SenML CBOR. The changes have been cherry-picked from the upstream Zephyr repo. Replaces PR #785.


VeijoPesonen:
@jheiskan81 for your information.

VeijoPesonen:
Oh great, the sdk-zephyr PR style check is more strict than the upstream one. For the sdk-zephyr the style check seems to be run commit by commit and that causes issues.

rlubos:
> Oh great, the sdk-zephyr PR style check is more strict than the upstream one. For the sdk-zephyr the style check seems to be run commit by commit and that causes issues. The issue seen is fixed in a subsequent commit. @rlubos Any idea how this kind of issues have been solved earlier?

Actually, I'm surprised to hear that. @PerMac Is that really the case? I though the compliance check is executed in a same way both upstream and here.

PerMac:
We use the same compliance script as in the upstream.  In your case it is checkpatch script failing, and this one is supposed to run on every commit to check if they are fine. https://docs.zephyrproject.org/latest/contribute/guidelines.html?highlight=checkpatch#coding-style and https://docs.zephyrproject.org/latest/project/dev_env_and_tools.html?highlight=checkpatch#continuous-integration
I don't see it being stated explicitly that it run on each commit, but I think it can be deduced from the context. Also in the script used in both CIs checkpatch gets the full range of commits to check https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/ci/check_compliance.py#L206

Some of the checks in the CI don't run on every commit but just on top, e.g. twister checks or KConfig check being part of compliance, maybe hence the confusion?

VeijoPesonen:
> We use the same compliance script as in the upstream. In your case it is checkpatch script failing, and this one is supposed to run on every commit to check if they are fine.

What I would deduce from that is that something isn't quite right in upstream checks as those were passing with the same commits I have here.

* https://github.com/zephyrproject-rtos/zephyr/runs/6207975815?check_suite_focus=true

As a workaround I propose that I squash those two commits which touch the same parts of the code. Only question is that should I mention the both as the originating commits on the commit message or is mentioning only one sufficient? Either way both of those changes will get reverted in the next upmerge.

PerMac:
I was wrong then. As you pointed out, there must be some difference on how the compliance work in sdk-zephyr and upstream zephyr. I will try to clarify this with the upstream CI maintainers. I will ask if they can prepare a bit more detailed docs, where it will be stated which check should run on which scope (diff vs main, commit by commit, etc.) and especially the scope of checkpatch. 
As for the squashing strategy, @rlubos can you advise how to do this from the upmerger perspective? Maybe squashing all commits from this PR will be easier to handle later on?

VeijoPesonen:
@jheiskan81, @rlubos Based on discussing this offline I ended up editing the commit which was causing the compliance issue.

SeppoTakalo (approved):

PerMac:
@VeijoPesonen FYI I digged deeper and found that there is a difference in our compliance and the one in the upstream. 
https://github.com/nrfconnect/sdk-zephyr/blob/main/scripts/ci/check_compliance.py#L206
https://github.com/zephyrproject-rtos/zephyr/blame/cea72d7173f9ae274ec789f75340d452ad83002f/scripts/ci/check_compliance.py#L205
It points to https://github.com/zephyrproject-rtos/zephyr/commit/5a93798b2275f6dbf747c4e76baf854e902eb0dd
which solves our riddle. It seems we are not yet synced with this revert.

PerMac:
^^ FYI @carlescufi If our devs get hit by this difference again we might want to sync fromtree the compliance script.

VeijoPesonen:
@PerMac I'm seeking for your approval for the PR unless you know someone else who should do it ;)

PerMac:
Is there a follow up pr with manifest update in sdk-nrf?

jheiskan81:
https://github.com/nrfconnect/sdk-nrf/pull/7546

rlubos:
@VeijoPesonen I can't merge sdk-zephyr PR unless I see CI passing on the manifest update, please rebase your sdk-nrf PR

jheiskan81:
@rlubos OK sdk-zephyr and  sdk-nrf PR's are now rebased.

SeppoTakalo:
Failing test seems unrelated to this change. `nrf5340-audio...`

PR #790, opened 2022-05-05
tfm: Set TF-M as default for all NS boards

joerchan:
Enable TF-M by default for nordic nRF5340 and nrf9160 SoC boards.

Revert noup commit: "boards: disable BUILD_WITH_TFM on some non-secure builds"

NCSDK-14811
NCSDK-7558


joerchan:
DNM: This contains a squash commit that should be included as part of the current upmerge. This PR should be merged after the upmerge is finished.

lemrey:
Are there changes to the default amount of FLASH dedicated to the secure domain when using TFM, compared to `spm`?

joerchan:
> Are there changes to the default amount of FLASH dedicated to the secure domain when using TFM, compared to `spm`?

That depends on a lot on the configuration. Both SPM and TF-M have a minimal configuration that is 32 KB of flash.
TF-M in the normal configuration will take more flash usage by default, but it can be set depending on your use-case and what you enable of features.

carlescufi (approved):

boards/arm/circuitdojo_feather_nrf9160/Kconfig.defconfig
```diff
@@ -9,6 +9,21 @@ if BOARD_CIRCUITDOJO_FEATHER_NRF9160 || BOARD_CIRCUITDOJO_FEATHER_NRF9160_NS
 config BOARD
 	default "circuitdojo_feather_nrf9160"
 
+# By default, if we build for a Non-Secure version of the board,
+# enable building with TF-M as the Secure Execution Environment.
+config BUILD_WITH_TFM
+	default y if BOARD_CIRCUITDOJO_FEATHER_NRF9160_NS
```
anangl, line 15: Cannot this go upstream?

samples/tfm_integration/psa_crypto/prj.conf
```diff
@@ -11,6 +11,7 @@ CONFIG_LOG_DEFAULT_LEVEL=4
 #CONFIG_PSA_SHELL=y
 
 CONFIG_BUILD_WITH_TFM=y
+CONFIG_TFM_MINIMAL=n
```
anangl, line 14: This commit does not seem right. It makes sdk-zephyr dependent on sdk-nrf, where the `TFM_MINIMAL` option is defined.
Why is this actually needed? I could not find a place where the `TFM_MINIMAL` option is enabled by default for these samples, but probably I'm missing something.

joerchan: TFM_MINIMAL option is enabled by default globally.

boards/arm/circuitdojo_feather_nrf9160/Kconfig.defconfig
joerchan: It cannot go upstream, it does not compile. We only support this in downstream.

frkv (approved):
LGTM

carlescufi (changes requested):
Agree that referring to `TFM_MINIMAL` directly from the sdk-zephyr tree is probably not a good solution. @joerchan can we fix this on the sdk-nrf side?

joerchan:
> Agree that referring to TFM_MINIMAL directly from the sdk-zephyr tree is probably not a good solution. @joerchan can we fix this on the sdk-nrf side?

@carlescufi I don't know how

samples/tfm_integration/psa_crypto/prj.conf
carlescufi: @joerchan couldn't we disable this globally by default and the enable it in the sdk-nrf samples that require it?

joerchan: Changed TFM_MINIMAL to be a part of the choice symbol TFM_PROFILE_TYPE (as TFM_PROFILE_TYPE_MINIMAL) and having it as default enabled.
Then this can be solved by setting TFM_PROFILE_TYPE_NOT_SET, which is an upstream config.

PR #806, opened 2022-05-13
[nrf fromlist] mgmt/mcumgr/lib: Fix image list command crashing

de-nordic:
... when no device

It is possible, in case of two application images, to have no access
to one of devices of the secondary image.
When asserts are enabled, such situation causes crash even though the
image list command can handle it with no problem.
The commit removes the assert and adds additional swap type:
IMG_MGMT_SWAP_TYPE_UNKNOWN to indicate situations where it was
not possible to obtain swap type from boot_util.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/45636

Signed-off-by: Dominik Ermel <dominik.ermel@nordicsemi.no>
(cherry picked from commit e209b59c4c2aea3d3805326843b9984fc5ad2994)


anangl:
Superseded by #811.

PR #814, opened 2022-05-20
[nrf fromlist] nvs: implement NVS cache

Damian-Nordic:
The NVS data lookup time grows linearly with the number of
allocation table entries to walk through, meaning that if
some data pair in the NVS changes frequently, access to
other data pairs that change rarely can take a lot of time.

It is particularly visible when the NVS is used as the
settings backend since the backend needs to perform multiple
NVS reads to find a requested key.

Implement a simple cache that stores an address of the most
recent ATE for all NVS IDs that fall into the given cache
position. CRC8/16 is used as a hash function used to
distribute NVS IDs across the cache entries.

The cache entries are only invalidated when an NVS sector
is erased as part of the garbage collector task.

(cherry picked from commit 1b662b80e07d004e7d3524ea6aaf7ef0c2b9c612)
Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>


LuDuda (approved):

de-nordic (approved):

mbolivar-nordic (dismissed):
Since this is a fromlist commit, please add a link to the upstream PR in the commit description. This allows the upmergers to track the status upstream during the next upmerge and handle this patch appropriately. There is no need to use `git cherry-pick -x` (i.e. the "cherry picked from commit ....") here because the SHA in  your downstream fork is not what will get merged into upstream zephyr, so this is not useful in a fromlist patch.

carlescufi (approved):

PR #822, opened 2022-05-30
ci: Add Zephyr LWM2M test spec

fredaas:
Signed-off-by: Fredrik Ås <fredrik.aas@nordicsemi.no>

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -1,4 +1,13 @@
 # This is the Jenkins ci variant of the .github/labler.yaml
+"CI-iot-zephyr-lwm2m-test":
+  - "drivers/console/**/*"
+  - "drivers/flash/**/*"
+  - "subsys/dfu/boot/**/*"
+  - "subsys/net/ip/**/*"
+  - "subsys/net/lib/http/**/*"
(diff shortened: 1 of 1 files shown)


.github/test-spec.yml
```diff
@@ -1,4 +1,7 @@
 # This is the Jenkins ci variant of the .github/labler.yaml
+"CI-iot-zephyr-lwm2m-test":
+  - "**/*"
```
SeppoTakalo, line 3: I think this is too wide.

I would suggest for example:
```
drivers/console/**/*
drivers/flash/**/*
subsys/dfu/boot/**/*
subsys/net/ip/**/*
subsys/net/lib/http/**/*
subsys/net/lib/lwm2m//**/*
subsys/net/**/*
```

fredaas: Updated.

SeppoTakalo (approved):

shantha-14 (approved):

fredaas:
@jfischer-no Updated.

jfischer-no (approved):

nordicjm (approved):

PR #828, opened 2022-06-07
Reset device after load

fredaas:
This PR changes the default behavior when running gdb via `west debug` to reset the device after loading the debug client.

Signed-off-by: Fredrik Ås <fredrik.aas@nordicsemi.no>


SeppoTakalo:
Does this break the `west attach` on some cases?

PR #831, opened 2022-06-13
Implement PSA Protected Storage over Settings

superna9999:
This Implements the PSA Protected Storage API over Settings for the case when TF-M isn't available on the platform.


subsys/trusted_storage/Kconfig
```diff
@@ -0,0 +1,215 @@
...
+config TRUSTED_STORAGE_SETTINGS_AUTH_CRYPT_BACKEND
+	bool "Trusted Storage Authenticated Encryption Backend"
+	select TRUSTED_STORAGE_SETTINGS_HELPERS
+	help
+	  Build Trusted Storage Authenticated Encryption Backend.
+
+config TRUSTED_STORAGE_AUTH_CRYPT_BACKEND_PSA
+	bool "Trusted Storage Authenticated Encryption with PSA Crypto"
+	depends on MBEDTLS_BUILTIN_PSA_CRYPTO
+	depends on MBEDTLS_CIPHER
```
frkv, line 38: These should be converted to PSA_WANT_ALG_XXXX, but I think you need a PR from me in Zephyr before you can open up a pure PR upstream.

The PSA_WANT_ALG_XXXX is suported in our sdk-zephyr

subsys/trusted_storage/Kconfig
```diff
@@ -0,0 +1,215 @@
...
+	prompt "Protected Storage with Settings trust back-end"
+	depends on PROTECTED_STORAGE_SETTINGS
+	default PROTECTED_STORAGE_SETTINGS_TRUST_INTEGRITY
+	help
+	  Trust back-end to be used by the Protected Storage API.
+
+config PROTECTED_STORAGE_SETTINGS_TRUST_INTEGRITY
+	bool "Integrity"
+	depends on MBEDTLS_BUILTIN_PSA_CRYPTO
+	depends on MBEDTLS_MD
```
frkv, line 103: As commented before: Legacy => PSA_WANT_ALG_XXX

subsys/trusted_storage/Kconfig
```diff
@@ -0,0 +1,215 @@
...
+	prompt "Internal Trusted Storage with Settings trust back-end"
+	depends on INTERNAL_TRUSTED_STORAGE_SETTINGS
+	default INTERNAL_TRUSTED_STORAGE_SETTINGS_TRUST_CHACHAPOLY
+	help
+	  Trust back-end to be used by the Internal Trusted Storage API.
+
+config INTERNAL_TRUSTED_STORAGE_SETTINGS_TRUST_CHACHAPOLY
+	bool "Chachapoly Authenticated Encryption"
+	select TRUSTED_STORAGE_SETTINGS_AUTH_CRYPT_BACKEND
+	depends on MBEDTLS_CHACHAPOLY_AEAD_ENABLED
```
frkv, line 201: As commented before: Legacy => PSA_WANT_ALG_XXX

subsys/trusted_storage/src/auth_crypt_backend.c
```diff
@@ -0,0 +1,303 @@
...
+ * Actual implementation uses:
+ * - Hexadecimal UID imported as KEY
+ * - UID+Flags+Size as additional parameter
+ * - IV is an incrementing number at each encryption
+ * - TAG is left at the end of output data
+ */
+
+/* TODO Kconfig */
+#define AEAD_TAG_SIZE		16
+#define AEAD_IV_SIZE		12
```
frkv, line 35: Why isn't the IV size 16, here?

subsys/trusted_storage/src/auth_crypt_backend.c
```diff
@@ -0,0 +1,303 @@
...
+	psa_storage_uid_t uid;
+	psa_storage_create_flags_t flags;
+	size_t size;
+};
+
+static uint64_t g_iv_low;
+static uint64_t g_iv_high;
+
+/* Return an incrementing IV */
+static psa_status_t psa_ps_get_iv(uint8_t *iv, size_t iv_len)
```
frkv, line 56: I would appreciate a clearly marked API to handle initial vector and making a counter scheme internal to this. The fact that this is a static function and it has extra functionality beyond just getting the IV (i.e. implementing a counter scheme on some of the IV data) is 

I would also recommend hiding the storage of the actual IV beneath a clearly marked API in case the system supports running counters or externally sourced IV from hw peripherals

subsys/trusted_storage/src/auth_crypt_backend.c
```diff
@@ -0,0 +1,303 @@
...
+	psa_status_t status;
+
+	if (init_cb) {
+		status = init_cb();
+		if (status != PSA_SUCCESS) {
+			return -EIO;
+		}
+	}
+
+	if (get_random_cb) {
```
frkv, line 98: The unpacking of the random data is noisy here. If the marked API is get_iv when the data is needed, it will increase the cohesion and limit tight coupling.

subsys/trusted_storage/src/auth_crypt_backend.c
```diff
@@ -0,0 +1,303 @@
...
+	memset(data_buf, 0, TRUSTED_STORAGE_MAX_ASSET_SIZE);
+
+	return PSA_SUCCESS;
+}
+
+psa_status_t trusted_storage_auth_crypt_backend_set(psa_storage_uid_t uid,
+				const char *prefix, size_t data_length,
+				const void *p_data,
+				psa_storage_create_flags_t create_flags,
+				psa_status_t (*crypt_cb)(
```
frkv, line 203: I know the Zephyr coding standard is negative to typedefs, but I think that in this case the readability unfortunately suffers a lot. There are examples of function pointers for the type of callback you write here. Please change it to a typedef!

subsys/trusted_storage/src/auth_crypt_backend.h
```diff
@@ -0,0 +1,66 @@
...
+ * Returns 0 or a negative PSA error value if an error occurs.
+ */
+psa_status_t trusted_storage_auth_crypt_backend_get(psa_storage_uid_t uid,
+				const char *prefix, size_t data_size,
+				size_t data_offset, size_t data_length,
+				void *p_data, size_t *p_data_length,
+				psa_storage_create_flags_t create_flags,
+				size_t (*get_encrypted_size_cb)(
+					size_t data_size),
+				psa_status_t (*crypt_cb)(
```
frkv, line 32: Already commented: Typedef for the callback

subsys/trusted_storage/src/auth_crypt_backend.h
```diff
@@ -0,0 +1,66 @@
...
+
+/*
+ * Stores & authenticates the data within the trust implementation
+ * Returns 0 or a negative PSA error value if an error occurs.
+ */
+psa_status_t trusted_storage_auth_crypt_backend_set(psa_storage_uid_t uid,
+				const char *prefix, size_t data_length,
+				const void *p_data,
+				psa_storage_create_flags_t create_flags,
+				psa_status_t (*crypt_cb)(
```
frkv, line 49: Already commented: Typedef for the callback

subsys/trusted_storage/src/internal_trusted_storage/settings_trust_chachapoly.c
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	mbedtls_chachapoly_init(&ctx);
+
+	ret = mbedtls_chachapoly_setkey(&ctx, key);
+	if (ret < 0) {
+		goto cleanup;
+	}
+
+	ret = mbedtls_chachapoly_encrypt_and_tag(&ctx, input_len, iv_buf,
```
frkv, line 67: Specific reason this is not using PSA crypto APIs?

subsys/trusted_storage/src/auth_crypt_backend.c
superna9999: MbedTLS ChaChaPoly nonce is 12 bytes:
```
int mbedtls_chachapoly_starts( mbedtls_chachapoly_context *ctx,                 
                              const unsigned char nonce[12],                   
                               mbedtls_chachapoly_mode_t mode ); 
```

same for PSA AEAD with GCM:
```
#define PSA_AEAD_NONCE_LENGTH(key_type, alg) \                                  
    (PSA_BLOCK_CIPHER_BLOCK_LENGTH(key_type) == 16 ? \                          
          MBEDTLS_PSA_ALG_AEAD_EQUAL(alg, PSA_ALG_CCM) ? 13 : \                 
          MBEDTLS_PSA_ALG_AEAD_EQUAL(alg, PSA_ALG_GCM) ? 12 : \                 
          0 : \                                                                 
     (key_type) == PSA_KEY_TYPE_CHACHA20 && \                                   
          MBEDTLS_PSA_ALG_AEAD_EQUAL(alg, PSA_ALG_CHACHA20_POLY1305) ? 12 : \   
     0)       
```

subsys/trusted_storage/src/protected_storage/settings_backend.c
```diff
@@ -0,0 +1,173 @@
...
+}
+
+psa_status_t psa_ps_create_backend(psa_storage_uid_t uid,
+				   size_t capacity,
+				   psa_storage_create_flags_t create_flags)
+{
+	return PSA_ERROR_NOT_SUPPORTED;
+}
+
+psa_status_t psa_ps_set_extended_backend(psa_storage_uid_t uid,
```
Vge0rge, line 151: I don't see this in the TF-M repo. Is this solely intended to be used as a set function with an offset?

subsys/trusted_storage/src/auth_crypt_backend.c
frkv: Ah... Yes. There is likely a collision in terms here dependent on how ChaCha20 is described compared to e.g. AES. 

Probably a more accurate term could be "nonce" for the thing that is 12 bytes long

The ChaCha20 algorithm input is explained as this (in RFC7539)

   o  A 32-bit initial counter.  This can be set to any number, but will
      usually be zero or one.  It makes sense to use one if we use the
      zero block for something else, such as generating a one-time
      authenticator key as part of an AEAD algorithm.

   o  A 96-bit nonce.  In some protocols, this is known as the
      Initialization Vector.
  
From my understanding the counter and nonce is "concatenated" as part of the inititialization state of the cipher. But it isn't exactly described as an initialization vector (concatenating counter + nonce)

Maybe comment that the IV-size corresponds to a nonce as described in ChaCha20. It would be neccesary to update this to 16 bytes for AES operations

subsys/trusted_storage/src/protected_storage/settings_backend.c
superna9999: Hmm, I have a doubt now, I can't remember where I took the header

superna9999: Those functions returns `PSA_ERROR_NOT_SUPPORTED` in TF-M, this is why I did the same, but still added them in the backend in case we want to implement those

superna9999:
Added the typedefs and moved the nonce algo out, and fixed a few things.

The only remaining change is switching to `PSA_WANT_ALG_XXXX`

PR #835, opened 2022-06-21
[nrf fromtree] drivers: gpio: nrf: Fix pin no for getting GPIOTE channel

nordic-krch:
Incorrect GPIOTE channel was being freed because the pin number
being used is not the absolute pin but the pin within the port.

Origin commit: 4b3fe99b83e393777dedc4200c7209fcdf31a77f

Ref. NCSDK-15818.

Signed-off-by: Wael Barakat <waelsbarakat@gmail.com>


carlescufi (approved):

nordic-krch:
came with upmerge.

PR #838, opened 2022-06-24
Upmerge 20220622

tejlmand:
Zephyr Upmerge - 20220622


tejlmand:
@PerMac could you take a look at this commit:
https://github.com/nrfconnect/sdk-zephyr/commit/e226734ea8ca482284237a3ddc5a3092b73d1af7

I had to solve merge conflicts and would like you to verify the solution, especially around this:
https://github.com/nrfconnect/sdk-zephyr/blob/f61a410b6509e9e36f10a22fd4b64fad5f4286d3/.github/workflows/clang.yaml#L131

where you're `[nrf noup]` patch https://github.com/nrfconnect/sdk-zephyr/commit/e226734ea8ca482284237a3ddc5a3092b73d1af7 added `--subset`, but hat was when `csv` files were used.
This has changed to a `json` file, and i'm unsure if the same argument should be applied, but to the updated command execution with json input file.

PerMac:
> @PerMac could you take a look at this commit: [e226734](https://github.com/nrfconnect/sdk-zephyr/commit/e226734ea8ca482284237a3ddc5a3092b73d1af7)
> 
> I had to solve merge conflicts and would like you to verify the solution, especially around this:
> 
> https://github.com/nrfconnect/sdk-zephyr/blob/f61a410b6509e9e36f10a22fd4b64fad5f4286d3/.github/workflows/clang.yaml#L131
> 
> where your `[nrf noup]` patch [e226734](https://github.com/nrfconnect/sdk-zephyr/commit/e226734ea8ca482284237a3ddc5a3092b73d1af7) added `--subset`, but that was when `csv` files were used. This has changed to a `json` file, and i'm unsure if the same argument should be applied, but to the updated command execution with json input file.

changing .csv tp .json should do the trick. It also like this in the upstream now https://github.com/zephyrproject-rtos/zephyr/blob/main/.github/workflows/clang.yaml#L113
Our noup patch with subset is needed because we don't have an access to the runner (aws self-hosted) which the upstream is using, hence we have to use the github default (free) runners and to speed things up, run in parallel

tejlmand:
> changing .csv tp .json should do the trick. It also like this in the upstream now https://github.com/zephyrproject-rtos/zephyr/blob/main/.github/workflows/clang.yaml#L113
> Our noup patch with subset is needed because we don't have an access to the runner (aws self-hosted) which the upstream is using, hence we have to use the github default (free) runners and to speed things up, run in parallel

Did the changes. Thanks.

PerMac:
FYI During recent upmerges we were ignoring the Scancode check since the issues are already in the upstream code

edmont:
Hi @tejlmand, these two commits fix KRKNWK-14489, can they be cherry-picked? https://github.com/zephyrproject-rtos/zephyr/pull/47603/commits

edmont (approved):

jfischer-no (approved):

hermabe (approved):

de-nordic (approved):
mcumgr lib tested with smp_svr seems to work fine over BT and UART.
Tested on nrf52840dk.

eivindj-nordic (approved):

PR #844, opened 2022-07-04
Enable TF-M isolation level 2 by default

Vge0rge:
This defaults the TF-M regression tests to isolation level 2.
It is set in order to showcase a more secure configuration
by default. The default behavior of the sample is changed in
the prj.conf file. The yaml file which includes the test cases
enables testing for the three configurations: the library mode,
the IPC mode isolation level 1 and the IPC mode isolation level 2.

Ref: NCSDK-12585


joerchan (approved):

frkv (approved):
LGTM

frkv (approved):
LGTM (besides the commit to circumvent input size 0 which I'm assuming is going to be solved with new runtime)

PR #845, opened 2022-07-05
subsys: tls_credentials: add Trusted Credential storage backend

superna9999:
***This PR is designed to be moved to Zephyr RTOS when ready***

This PR adds a Trusted Storage backend to the tls_credentials API, using the PSA Protected Storage to store the credentials in a  trusted way.

Tested with:
```bash
west build -b nrf5340dk_nrf5340_cpuapp_ns tests/net/lib/tls_credentials -T net.tls_credentials.trusted_tfm
```


jfischer-no:
Can it be closed?

PR #850, opened 2022-07-14
workflow: clang: Add missing matrix.platform

PerMac:
squash! [nrf noup] ci: clang: parallel execution
To align with the upstream.


PerMac:
A hot fix for failing clang workflow. The result of this change will only be seen after it is merged (due to `on: pull_request_target` scope)

PerMac:
I tested it on my fork. Had to merged it first  and only then following PRs get the change applied

stephanosio (approved):

PR #853, opened 2022-07-22
manifest: PSA arch tests with increased timeout

Vge0rge:
This brings a change in the PSA arch tests where
we increase the timeout to 90 seconds for Nordic
devices. Since we run RSA key generation tests
this is necessary.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -235,7 +235,7 @@ manifest:
       groups:
         - tee
     - name: psa-arch-tests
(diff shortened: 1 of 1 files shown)


joerchan (approved):

oyvindronningstad (approved):

Vge0rge:
Closing since it is already merged by the upmerge.

PR #859, opened 2022-07-26
Import flash_area label removal changes

gmarull:
Import flash_area label removal changes to allow updating NCS. Also picked an unrelated commit so that patches can be cleanly applied.


gmarull:
closing for now, as these changes imply rework in PM as well.

PR #867, opened 2022-08-01
[nrf noup] scripts: add quarantine file

gopiotr:
Add scripts/quarantine.yaml file, which will be used in CI.

Signed-off-by: Piotr Golyzniak <piotr.golyzniak@nordicsemi.no>

### DIFF
--- a/scripts/quarantine.yaml
+++ b/scripts/quarantine.yaml
@@ -1,0 +1,186 @@
+# The configurations resulting as a product of scenarios and platforms
+# will be skipped if quarantine is used. More details here:
+# https://docs.zephyrproject.org/latest/guides/test/twister.html#quarantine
+
+- scenarios:
+    - drivers.modem.build
+  platforms:
+    - nrf52dk_nrf52832
+
+- scenarios:
+    - sample.tfm.secure_partition
+    - sample.tfm.secure_partition.ipc
+    - drivers.hwinfo.api
+  platforms:
+    - nrf9160dk_nrf9160_ns
+    - nrf5340dk_nrf5340_cpuapp_ns
+  comment: "https://projecttools.nordicsemi.no/jira/browse/NCSDK-7558"
+
+- scenarios:
+    - libraries.libc.newlib.thread_safety.userspace.stress
+    - libraries.libc.newlib_nano.thread_safety.userspace.stress
+  platforms:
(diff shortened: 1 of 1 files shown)


nordic-piks (approved):

jangalda-nsc (approved):

carlescufi (changes requested):
Should this really be in the zephyr tree? or we could consider it part of sdk-nrf?

PerMac:
@carlescufi We decided to consider sdk-zephyr as an independent project, i.e. we build it based on its own manifest. Therefore, sdk-nrf content is not available there. However, there shouldn't be any problem to checkout just a single quarantine file from sdk-nrf in the sdk-zephyr plan.
Keeping quarantine in sdk-zephyr will be in line with "keeping repo independent" but will require maintaining noup patch for this file. We can use squash! marking to keep only a single commit (affecting the history of quarantine).
Keeping quarantine in sdk-nrf will require downloading it in sdk-zephyr plan (seems trivial) but might be less transparent from the sdk-zephyr repo perspective. Also hot fixes for sdk-zephyr issues adding tests to quarantine will require a PR to sdk-nrf, so e.g. if some PR to sdk-zephyr introduces regression which we agree to temporarily have, we won't be able to modify the quarantine in the same PR to get it green. Also in reverse direction: it won't be possible to provide a fix for quarantined test and get it validated in the same PR by releasing the test from the quarantine. IMO this can be a bit tedious. Also during upmerge it might cause issues with being synced up: We see there is some regression introduced with the upmerge. We decide to quarantine failing tests until the proper solution is found. If quarantine file is in sdk-zephyr, it can be modified in within the upmerge PR and make the plan green. If the quarantine is in sdk-nrf we would then have to either update the quarantine in the main just to unlock the upmerge or play with the jenkins plan to point to some other download location.
Personally, I am in favored for the quarantine in sdk-zephyr, mainly due to the above inconveniences.

jangalda-nsc:
I agree with @PerMac. What's more quarantine.yaml is living documentation of sdk-zephyr.

nordic-babu (approved):

carlescufi (approved):

PerMac:
@gopiotr Please rebase

PerMac (approved):

PR #869, opened 2022-08-03
Upmerge 08 03 2022

nordicjm:


Kconfig.zephyr
```diff
@@ -843,6 +843,7 @@ endmenu
 
 config LEGACY_INCLUDE_PATH
 	bool "Allow for the legacy include paths (without the zephyr/ prefix)"
+	default y
```
gmarull, line 846: let's fix legacy includes instead (there's a script to automate the process in `scripts/utils`).

nordicjm: It will be an ongoing task but due to to needing lots of changing in various components e.g. docker images, this will have to be changed back to n during the next upmerge process

rlubos (approved):

anangl (approved):

tejlmand (changes requested):
Why are you reverting a `noup` commit AFTER the merge, just to re-apply it again ?

If for some reason a revert is needed, that should generally be done BEFORE the merge happens.
image

Commits in question:
049865ccdf6ab7a01f2e0f9d7239ab12e2b0624e
c95e9bc654ec4f663687a247d342df9ebd7c722f

tejlmand:
@de-nordic @nordicjm  are you looking into this error from CI ?
```
CMake Error at <path>/python3.8/site-packages/cmake/data/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
  <path>/python3.8/site-packages/cmake/data/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
 ...
```

nordicjm:
> Why are you reverting a `noup` commit AFTER the merge, just to re-apply it again ?

The reason for that revert is to bring in those fromtree commits, there was a conflict because TF-M is not enabled in the upstream boards file and one of the diffs searches for lines that are different in the source file, so the change was not needed for the upmerge itself but just for those fromtree patches which were requested by the board maintainer to be brought in. Should that still require reverting differently?
As for that error, both failing examples build fine here when building for native_posix_64 (no multilib on this machine) so it looks possibly like a toolchain issue

tejlmand:
> The reason for that revert is to bring in those fromtree commits, there was a conflict because TF-M is not enabled in the upstream boards file and one of the diffs searches for lines that are different in the source file, so the change was not needed for the upmerge itself but just for those fromtree patches which were requested by the board maintainer to be brought in. Should that still require reverting differently?

Not necessarily.
Fell free to keep as-is.

But still would like to know about: 5a5512f3f7
That commit is being reverted, but not re-applied.

tejlmand (changes requested):
Evil merges are not permitted. a1a049a934

Please do a proper revert of all conflicting commits first, so that we get a conflict free merge.

nordicjm:
> But still would like to know about: [5a5512f](https://github.com/nrfconnect/sdk-zephyr/commit/5a5512f3f76d00b62608dce9a9a56635dfd75e8a)
> That commit is being reverted, but not re-applied.

That's been reverted due to causing issues with CTF tests, a note has been added to the internal todo page to have the CTF tests fixed by the time of the next upmerge

tejlmand (approved):
lgtm

jfischer-no (approved):

emob-nordic (changes requested):
Bug in spotted in  `./samples/subsys/ipc/ipc_service/icmsg`, blocking for investigation.

emob-nordic (approved):
Discussed offline, accepting the bug in spsc_buf/ipc_icmsg as it was discovered in the upmerge and not caused by the upmerge.

PR #873, opened 2022-08-10
allow creation of all FAT type filesystems

owain-incus:
zephyr api layer between elm and user prevented creation of FAT32 and EXFAT; why?


nordicjm:
This will not be needed once https://github.com/nrfconnect/sdk-zephyr/pull/869 is merged

PR #874, opened 2022-08-10
[nrf fromtree] tfm: Allow enabling FPU hard ABI with TF-M

joerchan:
Allow enabling FPU with TF-M with the following limitations:
- Only IPC mode is supported by TF-M.
- Disallow FPU hard ABI when building the NS application, the TF-M build
system does not pass the flags correctly to all dependencies.

Exclude the non-secure TF-M application from being built by TF-M.
This also stops the building of the tfm_ns_api library that we were
linking against.
This library is defined by the tf-m regression tests.
Add the installed interface source files exported by the TF-M build
system compile them in the zephyr build system.


Vge0rge (approved):

arch/arm/core/aarch32/Kconfig
```diff
@@ -255,14 +255,16 @@ choice
```
SebastianBoe, line 255: "Add the installed interface source files exported by the TF-M build system compile them in the zephyr build system."

Language.

SebastianBoe: nvm. It's in a fromtree, so it's too late.

SebastianBoe (approved):
Great stuff!

PR #879, opened 2022-08-22
[nrf fromlist]: net: dhcp: Add support for restarting DHCP

krish2718:
If a L2 link has been established, then the DHCP is taking too long as
it has to go through its capped exponential backoff timers to trigger
discover (The DHCP starts immediately during init, this is itself wrong,
it should start on a link UP notification) that delays the DHCP for
few seconds to a minute.

And if we do stop and start DHCP then also it goes through the initial
delays (though configurable), which is also not ideal.

Add support for restarting DHCP without any delay, i.e., release and
send discover immediately.

This is also useful in case L2 switches to a different subnet, in this
case Zephyr doesn't restart DHCP automatically, this API can be used by
L2 apps/drivers to restart DHCP to get new subnet IP.


rlubos:
Can you link the upstream PR (I'm assuming there is one according the sauce tag).

rlubos:
Also, a bit cumbersome, but we need a manifest update PR in `sdk-nrf` for each downstream change to verify CI results.

krish2718:
https://github.com/zephyrproject-rtos/zephyr/pull/49152

krish2718:
https://github.com/nrfconnect/sdk-nrf/pull/8556

sachinthegreen (approved):

PR #876, opened 2022-08-16
[nrf fromlist]: Nrf wifi mgmt

krish2718:
This is based on pending https://github.com/nrfconnect/sdk-zephyr/pull/871 (will rebase once this is merged) and is a downstream PR of https://github.com/zephyrproject-rtos/zephyr/pull/49049

PR #880, opened 2022-08-22
[nrf noup] wifi_mgmt: Add support for non-offload TCP/IP

krish2718:
Wi-Fi offload has two parts control and data, when wifi_mgmt is used
for control the data part if assumed to be offloaded (WIFI/NET_OFFLOAD),
but some drivers do not support that assumption.

Extend the wifi_mgmt to take in a full Ethernet APIs so that wifi_mgmt
can be used with native TCP/IP stack.

### DIFF
--- a/drivers/wifi/esp_at/esp.c
+++ b/drivers/wifi/esp_at/esp.c
@@ -1088,7 +1088,7 @@ static void esp_iface_init(struct net_if *iface)
 }
 
 static const struct net_wifi_mgmt_offload esp_api = {
-	.iface_api.init = esp_iface_init,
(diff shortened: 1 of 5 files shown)


wbober:
https://github.com/nrfconnect/sdk-zephyr/pull/876#discussion_r951527218

krish2718:
> Please take a look here: [zephyrproject-rtos/zephyr#45565](https://github.com/zephyrproject-rtos/zephyr/pull/45565)

Thanks, just realised the duplicate effort, I have already updated the upstream PR and also left comments on the above one as well.

PR #884, opened 2022-08-22
[backport-v2.0][nrf fromtree] drivers: pinctrl_nrf: Fix nrf_pin_configure() implemen…

anangl:
…tation

This is a follow-up to commit fd0767557439d04c4a220a9857e901d8d71498db.

The above commit was supposed to introduce overriding of the S0S1
drive setting with S0D1 for TWI/TWIM peripherals, but since it did
not properly update the `nrf_pin_configure()` function (the `drive`
parameter was only added in the function signature, but then it was
not used...), the drive setting was in fact not overridden.
This commit corrects this embarrassing oversight.

(cherry picked from commit 32a87215d7aa97191097357f602237f7da4a3536)

### DIFF
--- a/drivers/pinctrl/pinctrl_nrf.c
+++ b/drivers/pinctrl/pinctrl_nrf.c
@@ -95,8 +95,8 @@ __unused static void nrf_pin_configure(pinctrl_soc_pin_t pin,
 		input = NRF_GPIO_PIN_INPUT_DISCONNECT;
 	}
 
-	nrf_gpio_cfg(NRF_GET_PIN(pin), dir, input, NRF_GET_PULL(pin),
-		     NRF_GET_DRIVE(pin), NRF_GPIO_PIN_NOSENSE);
+	nrf_gpio_cfg(NRF_GET_PIN(pin), dir, input, NRF_GET_PULL(pin), drive,
+		     NRF_GPIO_PIN_NOSENSE);
 }
 
 int pinctrl_configure_pins(const pinctrl_soc_pin_t *pins, uint8_t pin_cnt,


gmarull (approved):

tejlmand (approved):
lgtm

gopiotr:
Problem with failing CI arise due to the fact that we store quarantine file in separate CI repository. When Andrzej made PR to this particular branch `v3.0.99-ncs1-branch`, then tests in CI were taken also from this branch. Unfortunately quarantine file was not changed appropriately to this branch, but the newest one was used (taken from CI repo). At this moment there are no possibility to use older version of this quarantine to this particular PR, because it demands modification in entire `sdk-zephyr` CI workflow and it is useless to do only for this one PR. 

For the future we would like to avoid such situations by moving quarantine file directly into `sdk-zephyr` repository - https://github.com/nrfconnect/sdk-zephyr/pull/867. If it will be applied, then in PRs like this one, older quarantine could be taken together with older branch/revision.

At this moment until quarantine will be moved into `sdk-zephyr` I recommend to merge this PR on my own responsibility. Source of problem with failing tests `tests/subsys/logging/log_api/logging.log2_api...` comes from lack of `g++-multilib` package in used Docker image. Thus we have listed those tests in newest version of quarantine list. Moreover those tests pass locally when I don't use this docker image.

Taking it all into consideration, please @carlescufi, @mbolivar-nordic or @rlubos to merge this PR despite failing tests in CI, which do not concern changes performed by Andrzej.

gopiotr (approved):

PR #898, opened 2022-08-31
lib: os: spsc_pbuf: Fixing race and test failure on non-secure targets

nordic-krch:
Fixing race which was found when running `icmsg` ipc service sample (merged upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/49345).
https://github.com/zephyrproject-rtos/zephyr/pull/49345/commits/e632fdfdc3b15f0ab5d30154882a133f8b05f3dc
https://github.com/zephyrproject-rtos/zephyr/pull/49345/commits/014c7f9a4d2b3041dc29f34d2f0d513ba1a6aa68
https://github.com/zephyrproject-rtos/zephyr/pull/49345/commits/6a6a74101d3b4f4f067d9ab0ce949ee5f9bcfe6a
https://github.com/zephyrproject-rtos/zephyr/pull/49345/commits/f12b5a07f62cc8ce094e61bc1b598f16ca41fd39

Fixing failure of `zephyr/tests/lib/spsc_pbuf` seen on non-secure targets (e.g., nrf91_ns). Failure was cause by use of random number generator which cannot be called from an interrupt context (upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/49718).
https://github.com/zephyrproject-rtos/zephyr/pull/49718/commits/ca82e0b313e1378b5cc7613161b56f9fb7e94d96
https://github.com/zephyrproject-rtos/zephyr/pull/49718/commits/5b04bb7a403e64aa565eec8e8b4b6baa5cf0d96a

Ref. NCSDK-16467
Ref. NCSDK-16627
Ref. NCSDK-16308

### DIFF
--- a/Kconfig.zephyr
+++ b/Kconfig.zephyr
@@ -27,6 +27,8 @@ osource "$(KCONFIG_BINARY_DIR)/Kconfig.soc.defconfig"
 osource "soc/$(ARCH)/*/Kconfig.defconfig"
 # This loads the toolchain defconfigs
 osource "$(TOOLCHAIN_KCONFIG_DIR)/Kconfig.defconfig"
+# This loads the testsuite defconfig
+source "subsys/testsuite/Kconfig.defconfig"
 
 menu "Modules"
 
--- a/lib/os/spsc_pbuf.c
+++ b/lib/os/spsc_pbuf.c
@@ -12,6 +12,9 @@
 #include <zephyr/sys/byteorder.h>
 
 #define LEN_SZ sizeof(uint32_t)
+/* Amount of data that is left unused to distinguish between empty and full. */
+#define FREE_SPACE_DISTANCE sizeof(uint32_t)
+
 #define PADDING_MARK 0xFF
 
 #define GET_UTILIZATION(flags) \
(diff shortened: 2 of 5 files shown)


subsys/random/Kconfig
```diff
@@ -17,6 +17,7 @@ config TEST_RANDOM_GENERATOR
 
 choice RNG_GENERATOR_CHOICE
 	prompt "Random generator"
+	default XOSHIRO_RANDOM_GENERATOR if RNG_GENERATOR_DEFAULT_XOSHIRO && ENTROPY_HAS_DRIVER
```
tejlmand, line 20: > Kconfig does not allow to select an option which is a
> choice. Additional Kconfig option is added to allow that.

using a `default` for this probably won't work because a choice has a prompt, which means that if config in the choice is already active, then a selecting a symbol which then selects `RNG_GENERATOR_DEFAULT_XOSHIRO` which then changes the default choice will not have an effect because of the stuck symbol syndrome.

Ref: https://docs.zephyrproject.org/latest/build/kconfig/tips.html#stuck-symbols-in-menuconfig-and-guiconfig

tejlmand (changes requested):
Concerned this suffer the stuck symbol syndrome.

Need to perform some more testing before approving.

emob-nordic (approved):

subsys/random/Kconfig
nordic-krch: applied changes from upstream PR

nordic-krch: @tejlmand can you take another look?

nordic-krch: @tejlmand ping

tejlmand (changes requested):
https://github.com/zephyrproject-rtos/zephyr/pull/49718 has been merged.

Please update the `[nrf fromlist]` commits to reflect this.
(also the `[nrf fromlist]` are missing upstream PR reference in their commit messages, but updating them to `fromtree` also solves this 😉

tejlmand (approved):
lgtm

pdunaj (approved):

mbolivar-nordic (commented):
@nordic-krch looks like you are manually adding cherry pick information. I think it would be cleaner to just use `git cherry-pick -x <UPSTREAM_SHA>; git commit ---amend ; # add [nrf  fromtree]` or not add this information at all.

nordic-krch:
@mbolivar-nordic i typically use cherry-picking, though i didn't know the `-x` option. I don't recall if i added some manual changes here. Do I need to change something here?

ns-tolu (approved):

mbolivar-nordic:
> Do I need to change something here?

No, it was a request for the future.

PR #899, opened 2022-09-01
[nrf noup] wifi_stats: Use nRF Wi-Fi management

krish2718:
Until we merge with Zephyr's Wi-Fi management, use nRF Wi-Fi
management for Wi-Fi stats.

### DIFF
--- a/subsys/net/ip/Kconfig.stats
+++ b/subsys/net/ip/Kconfig.stats
@@ -127,7 +127,7 @@ config NET_STATISTICS_POWER_MANAGEMENT
 
 config NET_STATISTICS_WIFI
 	bool "Wi-Fi statistics"
-	depends on NET_L2_WIFI_MGMT
+	depends on NET_L2_WIFI_MGMT || NET_L2_NRF_WIFI_MGMT
 	default y
 	help
 	  Keep track of Wi-Fi related statistics. Note that this


subsys/net/ip/Kconfig.stats
```diff
@@ -127,7 +127,7 @@ config NET_STATISTICS_POWER_MANAGEMENT
 
 config NET_STATISTICS_WIFI
 	bool "Wi-Fi statistics"
-	depends on NET_L2_WIFI_MGMT
```
rlubos, line 130: how about making this `depends on NET_L2_WIFI_MGMT || NET_L2_RF_WIFI_MGMT` so statistics could work with either?

rlubos (approved):

sachinthegreen (approved):

subsys/net/ip/Kconfig.stats
```diff
@@ -127,7 +127,7 @@ config NET_STATISTICS_POWER_MANAGEMENT
...
+	depends on NET_L2_WIFI_MGMT || NET_L2_RF_WIFI_MGMT
```
rlubos, line 130: Actually, where is (will be) `NET_L2_RF_WIFI_MGMT` defined? CI check in NCS cannot find it (and neither can I).

krish2718: oops, It should be `NET_L2_RF_WIFI_MGMT`

PR #903, opened 2022-09-02
[nrf noup] Bluetooth: BT_L2CAP_ECRED is no longer experimental

hermabe:
BT_L2CAP_ECRED is not marked as experimental upstream and we qualify it
downstream.

To be squashed with 1e8c632358761d97cf2d659eb9fdc316b4558547 in the next
upmerge.

Signed-off-by: Herman Berget <herman.berget@nordicsemi.no>


Thalley (approved):

carlescufi (changes requested):
Why not do this upstream? This is a Host feature.

Thalley:
> Why not do this upstream? This is a Host feature.

It has already been removed upstream. The `select experimental` is only downstream.

carlescufi (approved):

PR #927, opened 2022-10-07
net: wifi: Add support for WPA supplicant

krish2718:
If WPA supplicant is enabled at build time, then use the WPA supplicant APIs, else use offload driver APIs.

### DIFF
--- a/subsys/net/l2/wifi/wifi_mgmt.c
+++ b/subsys/net/l2/wifi/wifi_mgmt.c
@@ -12,6 +12,9 @@ LOG_MODULE_REGISTER(net_wifi_mgmt, CONFIG_NET_L2_WIFI_MGMT_LOG_LEVEL);
 #include <zephyr/net/net_core.h>
 #include <zephyr/net/net_if.h>
 #include <zephyr/net/wifi_mgmt.h>
+#ifdef CONFIG_WPA_SUPP
+#include "supp_api.h"
+#endif
 
 static int wifi_connect(uint32_t mgmt_request, struct net_if *iface,
 			void *data, size_t len)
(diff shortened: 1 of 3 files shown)


subsys/net/l2/wifi/wifi_shell.c
```diff
@@ -68,12 +68,14 @@ static void handle_wifi_scan_result(struct net_mgmt_event_callback *cb)
 
 	if (scan_result == 1U) {
 		print(context.sh, SHELL_NORMAL,
-		      "\n%-4s | %-32s %-5s | %-4s | %-4s | %-5s    | %s\n",
-		      "Num", "SSID", "(len)", "Chan", "RSSI", "Sec", "MAC");
+		      "\n%-4s | %-32s %-5s | %-13s | %-4s | %-15s | %s\n",
```
rlubos, line 71: The changes in wifi_shell.c will come with an upmerge, let's not introduce them here. And for the future - you should rather cherry-pick upstream commits and tag them with `[nrf fromtree]` prefix instead of creating your own commits.

subsys/net/l2/wifi/wifi_mgmt.c
```diff
@@ -12,19 +12,25 @@ LOG_MODULE_REGISTER(net_wifi_mgmt, CONFIG_NET_L2_WIFI_MGMT_LOG_LEVEL);
...
 			void *data, size_t len)
 {
 	struct wifi_connect_req_params *params =
 		(struct wifi_connect_req_params *)data;
 	const struct device *dev = net_if_get_device(iface);
+#ifndef CONFIG_WPA_SUPP
 	struct net_wifi_mgmt_offload *off_api =
 		(struct net_wifi_mgmt_offload *) dev->api;
 
+
```
rlubos, line 29: Please keep the diff minimal, no unnecessary changes.

subsys/net/l2/wifi/wifi_shell.c
krish2718: Sure, done. I have dropped these changes as they are being taken care in upmerge.

sachinthegreen (approved):

rlubos (approved):

PR #940, opened 2022-10-21
[nrf noup] tf-m: Add implementation id Kconfig select to initial atte…

mswarowsky:
…station

The initial attestation service requires the Implementation id stored via the bl_storage library. As the implementation id is not needed by default and has no dependencies select is used.

As this relies on the bl_storage library of NSIB, which is a nrf only component noup commit is chosen

Signed-off-by: Markus Swarowsky <markus.swarowsky@nordicsemi.no>


joerchan (changes requested):

mswarowsky:
Moved to sdk-nrf so no changes here needed

PR #941, opened 2022-10-24
Flush stale response from previous command

hno-wires:
If the device responds after request timeout it's response will be sitting in the receive buffer when the next command is sent, which will process the earlier response as response to it's request.

Reset the semaphore when making a new request to synchronise request/response flow to the new request.


jfischer-no:
@hno-addiva Can you please submit your patch to upstream project, https://github.com/zephyrproject-rtos/zephyr, thanks.

PR #942, opened 2022-10-24
[nrf noup] tfm: Enable bootloaders in TF-M regression tests

joerchan:
Enable bootloaders in the TF-M regression test suite. Run NS and S test cases separately as there is not enough flash.

Signed-off-by: Joakim Andersson <joakim.andersson@nordicsemi.no>

### DIFF
--- a/samples/tfm_integration/tfm_regression_test/sample.yaml
+++ b/samples/tfm_integration/tfm_regression_test/sample.yaml
@@ -24,3 +24,12 @@ tests:
 
   sample.tfm.regression_ipc_lvl2:
     timeout: 200
+
+  # Regression tests are too large for bootloaders to be enabled.
+  # Run S and NS test suites separately.
+  sample.tfm.regression.bootloaders_s:
+    extra_args: CONFIG_BOOTLOADER_MCUBOOT=y CONFIG_SECURE_BOOT=y CONFIG_TFM_REGRESSION_NS=n
+    timeout: 200
+  sample.tfm.regression.bootloaders_ns:
+    extra_args: CONFIG_BOOTLOADER_MCUBOOT=y CONFIG_SECURE_BOOT=y CONFIG_TFM_REGRESSION_S=n
+    timeout: 200


samples/tfm_integration/tfm_regression_test/sample.yaml
```diff
@@ -24,3 +24,12 @@ tests:
...
     timeout: 200
+
+  # Regression tests are too large for bootloaders to be enabled.
+  # Run S and NS test suites separately.
+  sample.tfm.regression.bootloaders_s:
+    extra_args: CONFIG_BOOTLOADER_MCUBOOT=y CONFIG_SECURE_BOOT=y CONFIG_TFM_REGRESSION_NS=n
+    timeout: 200
+  sample.tfm.regression.bootloaders_ns:
+    extra_args: CONFIG_BOOTLOADER_MCUBOOT=y CONFIG_SECURE_BOOT=y CONFIG_TFM_REGRESSION_S=n
+    timeout: 200
```
SebastianBoe, line 35: @magnev FYI

SebastianBoe: @PerMac : Is it possible to have out-of-tree sample.yaml snippets ?

PerMac: @SebastianBoe Sorry, I don't know what you mean

PerMac:
If this configuration requires NCS to work (sdk-zephyr is not enough) than why not adding it to the nrf/samples? I guess code redundancy and possibility of getting out of sync with the base? Will `target_sources()` pointing towards original zephyr sample do the trick? I am not a fan of samples/tests pointing outside of  their directory since it makes debugging harder but I already have to live with it;) 
@carlescufi zephyr vs nrf manifest issue ^^

joerchan:
> If this configuration requires NCS to work (sdk-zephyr is not enough) than why not adding it to the nrf/samples? I guess code redundancy and possibility of getting out of sync with the base? Will `target_sources()` pointing towards original zephyr sample in CMakeLists.txt do the trick? I am not a fan of samples/tests pointing outside of their directory since it makes debugging harder but I already have to live with it;) @carlescufi zephyr vs nrf manifest issue ^^

@PerMac While that might work for this change there is a similar issue here: https://github.com/nrfconnect/sdk-zephyr/pull/93, it is the same problem, but not having this means that we won't be able to run this sample/test in sdk-zephyr manifest.

If we continue with your suggestion then it sounds like we should just completely disable this sample/test in sdk-zephyr and just rely on the sdk-nrf variant of it.

Note that having this in sdk-nrf is not a problem when it comes to the application sources of this since this is a special sample and all of the application source files actually reside in trusted-firmware-m repository.

joerchan:
@PerMac @carlescufi We will then copy these samples as tests in sdk-nrf and add our own configurations there.
https://github.com/nrfconnect/sdk-nrf/pull/9149

SebastianBoe:
This PR is deprecated by the approach in 9149

PR #946, opened 2022-10-26
[nrf fromtree] Bluetooth: Mesh: Added support for extended advertiser

Balaklaka:
Added improvements and fixes for extended advertiser to preform better for mesh. All fixes are taken from main branch.

Signed-off-by: Ingar Kulbrandstad [ingar.kulbrandstad@nordicsemi.no](mailto:ingar.kulbrandstad@nordicsemi.no)

### DIFF
--- a/subsys/bluetooth/mesh/Kconfig
+++ b/subsys/bluetooth/mesh/Kconfig
@@ -533,6 +533,30 @@ config BT_MESH_TX_SEG_MAX
 	  which leaves 56 bytes for application layer data using a
 	  4-byte MIC and 52 bytes using an 8-byte MIC.
 
+config BT_MESH_SEG_ACK_BASE_TIMEOUT
+	int "SegAck transmission base timeout"
+	default 150
+	range 150 400
+	help
+	  Defines a base timeout for the acknowledgment timer used to delay
+	  transmission of the Segmented Acknowledgment message.
+
+config BT_MESH_SEG_ACK_PER_HOP_TIMEOUT
+	int "SegAck transmission timeout per hop"
+	default 50
+	range 50 250
+	help
+	  Defines an additional per-hop timeout for the acknowledgment timer
+	  used to delay transmission of the Segmented Acknowledgment message.
+
+config BT_MESH_SEG_ACK_PER_SEGMENT_TIMEOUT
+	int "SegAck transmission timeout per segment not yet received"
+	default 0
+	range 0 100
+	help
+	  Defines an additional timeout for the acknowledgment timer for every
+	  segment not yet received.
+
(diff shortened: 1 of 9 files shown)


PavelVPV (approved):

subsys/bluetooth/mesh/Kconfig
```diff
@@ -620,6 +641,21 @@ config BT_MESH_RELAY_RETRANSMIT_INTERVAL
 	  messages, in milliseconds. Can be changed through runtime
 	  configuration.
 
+config BT_MESH_RELAY_BUF_COUNT
+	int "Number of advertising buffers for relayed messages"
+	default 32
```
omkar3141, line 646: I would suggest to keep this default low as we have found out. 
```suggestion
	default 7
```

subsys/bluetooth/mesh/Kconfig
```diff
@@ -620,6 +641,21 @@ config BT_MESH_RELAY_RETRANSMIT_INTERVAL
 	  messages, in milliseconds. Can be changed through runtime
 	  configuration.
 
+config BT_MESH_RELAY_BUF_COUNT
```
omkar3141, line 644: PR says add support for an extended advertiser. May be we should tune the NCS_SAMPLES_DEFAULT if BT_MESH is used to configure number of sets.

omkar3141 (changes requested):

subsys/bluetooth/mesh/Kconfig
omkar3141: Never mind.

omkar3141 (changes requested):

subsys/bluetooth/mesh/Kconfig
Balaklaka: Do we really have to change this value, I guess that it was set to 32 when first selected.
The users that need a lower value can change this in their configuration, and for future we can change it in main branch. I am worried about if this could have some side effect without us knowing it.

Balaklaka: Do we really have to change this value, I guess that it was set to 32 when first selected.
The users that need a lower value can change this in their configuration, and for future we can change it in main branch. I am worried about if this could have some side effect without us knowing it.

omkar3141: Ok. We should update this upstream main branch.

omkar3141 (approved):

nordicjm (changes requested):
All commits missing `cherry-picked from commit`

nordicjm (approved):

PR #949, opened 2022-10-26
[nrf noup] ci: clang: Use ubuntu-latest instead of zephyr_runner

PerMac:
squash! [nrf noup] ci: clang: parallel execution

zephyr_runner is upstream's proprietary agent inaccessible outside of upstream's CI. Use ubuntu-latest instead.

### DIFF
--- a/.github/workflows/clang.yaml
+++ b/.github/workflows/clang.yaml
@@ -11,7 +11,7 @@ jobs:
         with:
           access_token: ${{ github.token }}
   clang-build:
-    runs-on: zephyr_runner
+    runs-on: ubuntu-latest
     needs: clang-build-prep
     container:
       image: ghcr.io/zephyrproject-rtos/ci:v0.24.2


PerMac:
This patch unlocks sdk-zephyr CI which is hanging due to lack of access to zephyr_runner. Due to `on: pull_request_target` hanging of clang will affect this one as well and the PR has to be merge to see the effect.

PerMac:
The change was originally part of https://github.com/nrfconnect/sdk-zephyr/commit/cec61795f0946d0df409f923bca6a8659fe3ed9f which was reverted as no longer needed in the latest upmerge

stephanosio (approved):

anangl (approved):

PerMac:
@anangl Do you think we will need a manifest update PR for this one? I believe it won't affect nrf since this check only takes place on PRs to sdk-zephyr, and we can leave without immediate update of the sdk-nrf manifest

anangl:
I agree, there is no need to update the sdk-nrf manifest just for this change in sdk-zephyr.

PR #953, opened 2022-10-27
sample: tfm_regression_test: Sample.yaml fixes

joerchan:
Fix warning about Audit partition not being enabled in IPC model
Explicitly set Library model instead of disabling IPC model, this caused IPC model to still be enabled.


torsteingrindvik (approved):
Internal CI green- this PR fixes current broken CI.

frkv (approved):
LGTM

PR #956, opened 2022-11-01
mbedtls: Fix typos and improve prompt-less configuration 

frkv:
-Improves prompt-less configurations used to ensure specific features are turned off or on
 in PSA builds. 
-Fixes typo on asymmetric encryption/signing

ref: NCSDK-17840

### DIFF
--- a/modules/mbedtls/Kconfig.psa
+++ b/modules/mbedtls/Kconfig.psa
@@ -21,18 +21,13 @@ config PSA_WANT_ALG_HMAC_DRBG
 
 endmenu # RNG support
 
-menu "PSA Key support"
-
 config PSA_HAS_KEY_SUPPORT
 	bool
 	default y
-	depends on PSA_WANT_KEY_TYPE_DERIVE 		|| \
-		   PSA_WANT_KEY_TYPE_HMAC 		|| \
+	depends on PSA_WANT_KEY_TYPE_DERIVE		|| \
+		   PSA_WANT_KEY_TYPE_HMAC		|| \
 		   PSA_WANT_KEY_TYPE_AES 		|| \
-		   PSA_WANT_KEY_TYPE_ARIA		|| \
-		   PSA_WANT_KEY_TYPE_CAMELLIA		|| \
 		   PSA_WANT_KEY_TYPE_CHACHA20		|| \
-		   PSA_WANT_KEY_TYPE_DES		|| \
 		   PSA_WANT_KEY_TYPE_ECC_KEY_PAIR	|| \
 		   PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY	|| \
 		   PSA_WANT_KEY_TYPE_RSA_KEY_PAIR	|| \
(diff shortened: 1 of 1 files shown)


modules/mbedtls/Kconfig.psa
```diff
@@ -165,7 +165,7 @@ config PSA_HAS_HASH_SUPPORT
 config PSA_WANT_ALG_SHA_1
 	bool
 	prompt "PSA SHA1 support" if !PSA_PROMPTLESS
-	default y if !PSA_DEFAULT_OFF
```
SebastianBoe, line 186: Is the correct Jira issue referenced?

I didn't find any more details there.

frkv: Wrong jira in commit-message. NCSDK-17333 => NCSDK-17**8**33. Fixed!

modules/mbedtls/Kconfig.psa
```diff
@@ -179,11 +159,13 @@ config PSA_HAS_HASH_SUPPORT
...
 		   PSA_WANT_ALG_RIPEMD160	|| \
 		   PSA_WANT_ALG_MD5
+	help
+	  Prompt-less configuration that states that hash is supported.
 
 config PSA_WANT_ALG_SHA_1
 	bool
 	prompt "PSA SHA1 support" if !PSA_PROMPTLESS
-	default y if !PSA_DEFAULT_OFF
+	default y if !PSA_DEFAULT_OFF && !BUILD_WITH_TFM
```
SebastianBoe, line 168: "The default for PSA_WANT_ALG_SHA_1 is adjusted to
 not enable this when BUILD_WITH_TFM is enabled. Note that the
 defconfig in nrf (kconfig.psa.defconfig) for TF-M is already set
 to not enable SHA-1 by default, but this isn't enough to turn the
 feature off."

Why is this not enough? I would have thought it would be enough.

joerchan: Because the crypto-backends don't look at these configurations. They just call `hash_compute`.
So we need an explicit "NOT_SUPPORTED" return before calling those functions.

```
#if !defined(PSA_WANT_ALG_SHA_1)
    if (alg == PSA_ALG_SHA_1) {
        return PSA_ERROR_NOT_SUPPORTED;
    }
#endif
...
#if defined(PSA_CRYPTO_DRIVER_HAS_HASH_SUPPORT_CC3XX)
    status = cc3xx_hash_compute(alg, input, input_length, hash, hash_size,
            hash_length);
...
#endif /* PSA_CRYPTO_DRIVER_HAS_HASH_SUPPORT_CC3XX */
#if defined(PSA_CRYPTO_DRIVER_HAS_HASH_SUPPORT_OBERON)
    status = oberon_hash_compute(alg, input, input_length, hash, hash_size,
            hash_length);
...
#endif /* PSA_CRYPTO_DRIVER_HAS_HASH_SUPPORT_OBERON */
```

SebastianBoe: Should we delete Kconfig.psa.defconfig then?

Since it is not executed early enough to have any effect?

Or is there some configuration where it does have effect?

Vge0rge (approved):

PR #957, opened 2022-11-03
Upmerge 2022.11.03

desowin:
Merge upstream up to commit cd16a8388f71a6cce0cea871f75f6d4ac8f56da9


desowin:
@krish2718 Could you please check whether I properly fixed the conflicts in `subsys/net/ip/net_private.h` when reapplying your changes?

@PerMac Could you please check whether I properly fixed the conflicts in `.github/workflows/clang.yaml` when reapplying your changes?

subsys/net/ip/net_private.h
```diff
@@ -28,7 +28,7 @@ union net_mgmt_events {
 #if defined(CONFIG_NET_DHCPV4)
 	struct net_if_dhcpv4 dhcpv4;
 #endif /* CONFIG_NET_DHCPV4 */
-#if defined(CONFIG_NET_L2_WIFI_MGMT)
+#if defined(CONFIG_NET_L2_WIFI_MGMT) || defined(CONFIG_NET_L2_NRF_WIFI_MGMT)
```
krish2718, line 31: This is not needed, `CONFIG_NET_L2_NRF_WIFI_MGMT` was removed.

desowin: Ok. I'll drop the `[nrf noup] net: ip: Fix nRF Wi-Fi management events` then as this is the only line it changed.

nordicjm:
This commit seems to be reverted then not re-applied:
```
commit 98bc3a7a6b3917ce5a058e7a7271d3036edffb88
Author: Krishna T <krishna.t@nordicsemi.no>
Date:   Wed Sep 28 17:31:32 2022 +0530

    [nrf noup] net: ip: Fix nRF Wi-Fi management events
    
    When using nRF Wi-Fi management it uses the net management events with
    information, but the size of the events is not enough to hold the Wi-Fi
    events as the proper size is defined only for Zephyr's Wi-Fi management.
    
    Use the proper size for both nRF and Zephyr's Wi-Fi management.
    
    Note: If CONFIG_NET_DHCPV4 is enabled, then this works, as size is now large
    enough for Wi-Fi management events.
    
```

desowin:
> This commit seems to be reverted then not re-applied

@nordicjm It was reapplied with conflict fixed but then I removed it due to it being unnecessary. See https://github.com/nrfconnect/sdk-zephyr/pull/957#discussion_r1012550358 How to deal with such cases?

nordicjm:
If it's not needed then it's fine to drop.

anangl:
Commits [[nrf noup] ci: clang: Use ubuntu-latest instead of zephyr_runner](https://github.com/nrfconnect/sdk-zephyr/pull/957/commits/64bdbc8bdd83fc8a3768b93c4c633c09acd23cb8) and [[nrf noup] ci: clang: Remove incorrect zephyr path](https://github.com/nrfconnect/sdk-zephyr/pull/957/commits/5ffa9ba782773742782783af757b6854fa0a24fb) should be squashed with [[nrf noup] ci: clang: parallel execution](https://github.com/nrfconnect/sdk-zephyr/pull/957/commits/9d00c7f0de76a979107742118b3b892fd22fb786).

anangl (approved):

PerMac:
after squashing the clang.yaml looks correct. I will run some tests on my fork of sdk-zephyr to verify we won't get any surprises after the upmerge since I am not sure if this change from the upstream won't cause us problems
```
-      - name: Cleanup
+      - name: Clone cached Zephyr repository
+        continue-on-error: true
         run: |
-          # hotfix, until we have a better way to deal with existing data
-          rm -rf zephyr zephyr-testing
+          git clone /github/cache/zephyrproject/zephyr .
+          git remote set-url origin ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}
```

anangl:
Regarding [[nrf noup] ci: clang: parallel execution](https://github.com/nrfconnect/sdk-zephyr/pull/957/commits/a48257efcb5ed29589ae5d01113a4c1c4eaeeb48):
```
(cherry picked from commit ae483a6bad7ad9c46b243fdaa149d2f20e8617d6
then squashed a3c7869f348491ecd25fae687a8d4e4a5928302f in bringing back
clang-build-prep to resolve conflict and then squashed in
c8787bd14c99f2dcbdd26a9786e01ad6c8311c63)
```
Looking at https://github.com/zephyrproject-rtos/zephyr/commit/e00a3cb600b40f62c29c6263d011057826e712da, it seems that you should not bring back the clang-build-prep job just to resolve the conflict when applying a3c7869f348491ecd25fae687a8d4e4a5928302f.

desowin:
> Looking at [zephyrproject-rtos/zephyr@e00a3cb](https://github.com/zephyrproject-rtos/zephyr/commit/e00a3cb600b40f62c29c6263d011057826e712da), it seems that you should not bring back the clang-build-prep job just to resolve the conflict when applying [a3c7869](https://github.com/nrfconnect/sdk-zephyr/commit/a3c7869f348491ecd25fae687a8d4e4a5928302f).

I have updated the commit removing the clang-build-prep, thanks.

desowin:
> after squashing the clang.yaml looks correct. I will run some tests on my fork of sdk-zephyr to verify we won't get any surprises after the upmerge since I am not sure if this change from the upstream won't cause us problems

@PerMac I guess all I have to do is to wait for your feedback on this.

PerMac:
The state of this workflow doesn't affect the PR, since the code from main is used for the PR checks. I will give it a deeper look so we won't have to deal with it after the upmerge.

jfischer-no (approved):

PerMac:
I verified the current clang.yaml workflow on my fork and it works. "Clone cached Zephyr repository" staged indeed failed but since it has `continue-on-error: true` the workflow continued and passed. I opt for leaving it as it is, without adding extra noup patch removing this section (less conflicts in future). One has to expand the logs to see that there was some issue, therefore I believe it shouldn't cause confusion. -> https://github.com/PerMac/sdk-zephyr/actions/runs/3393962978/jobs/5641922330

PavelVPV:
> 2 builds are failing due to flash overflows https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-zephyr/job/PR-957/3/#showFailuresLink:
> 
> zephyr/samples/subsys/mgmt/mcumgr/smp_svr/sample.mcumgr.smp_svr.bt on nrf52dk_nrf52832 -> has strickt platform_allow. https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml#L10 Since we cannot use nrf52840dk due to lack of cryptocell in sdk-zephyr (we try nrf52833 instead) the sample is no longer built on any platform in the context of sdk-zephyr.
> 
> zephyr/tests/bluetooth/mesh_shell/bluetooth.mesh.mesh_shell on qemu_cortex_m3 https://github.com/zephyrproject-rtos/zephyr/blob/main/tests/bluetooth/mesh_shell/testcase.yaml#L3 -> the same. Strict allow list and due to the overflow we will have no platform coverage left in sdk-zephyr
> 
> @carlescufi should we add the failing builds to the quarantine and live without having them covered in sdk-zephyr? Other option might be trying to optimize the configurations or fixing platform_allow lists not to artificially block the building so we can build on some other platform. In the upstream overflows are just skipped (no errors unless a platform is on the integration_platforms list). CC @cvinayak @de-nordic

Hi @PerMac, regarding `mesh_shell`, I can create a "working" prj.conf file for `qemu_cortex_m3` for `mesh_shell`. Will it be OK if I do it in upstream and then add the commit to this PR (this may take some time to get PR merged) or it will be easier for you if I create a fix straight into this PR?

PerMac:
@PavelVPV I believe it has to go through the upstream. I guess after you create the upstream PR we can add the commit as [nrf fromtree] to this upmerge if it won't get merge fast enough.

de-nordic:
> 
> zephyr/samples/subsys/mgmt/mcumgr/smp_svr/sample.mcumgr.smp_svr.bt on nrf52dk_nrf52832 -> has strickt platform_allow. https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml#L10 Since we cannot use nrf52840dk due to lack of cryptocell in sdk-zephyr (we try nrf52833 instead) the sample is no longer built on any platform in the context of sdk-zephyr.
> 
 
This has been discussed offline with @PerMac, and I have tested possibility of adding nrf52dk_nrf52832 to the list but the sample did not fit there either.
I have been able to finally make the nrf52dk_nrf52832  and the nrf52dk_nrf52832  builds possible, by disabling logging in mcumgr and file systems, and removing some of non essential commands of smp_svr. The problem is that may actions only allowed the sample to build with narrow amount of flash left free (usage was more than 99% in both cases).
This means that although the sample build, it will not work in reality as there is not enough space in image slot to perform swap operation.
Without changes to flash allocation, for example by dropping support for file system within the sample and removing the partition, we just prove that the sample is able to build for platforms, but it will not really work; it will boot and some smp_svr commands will work, but uploading image will not.

PavelVPV:
> @PavelVPV I believe it has to go through the upstream. I guess after you create the upstream PR we can add the commit as [nrf fromtree] to this upmerge if it won't get merge fast enough.

The fix for mesh_shell is merged here: https://github.com/zephyrproject-rtos/zephyr/pull/51968

desowin:
> This means that although the sample build, it will not work in reality as there is not enough space in image slot to perform swap operation.
> Without changes to flash allocation, for example by dropping support for file system within the sample and removing the partition, we just prove that the sample is able to build for platforms, but it will not really work; it will boot and some smp_svr commands will work, but uploading image will not.

@carlescufi How this should be handled?

nordicjm:
I have got nrf52dk_nrf52832 building with bluetooth at 85% flash usage (186KB of 220KB) with the following changes:
```
CONFIG_CLOCK_CONTROL_NRF_SHELL=n
CONFIG_DEVICE_SHELL=n
CONFIG_DEVMEM_SHELL=n
CONFIG_FLASH_SHELL=n
CONFIG_BT_USER_PHY_UPDATE=n
CONFIG_BT_PHY_UPDATE=n
CONFIG_USE_SEGGER_RTT=n
CONFIG_BT_CTLR_LE_PING=n
CONFIG_BT_CTLR_PHY=n
CONFIG_LOG_MODE_MINIMAL=y
```

de-nordic:
> CONFIG_CLOCK_CONTROL_NRF_SHELL

I think we should redefine what for the smp_svr sample is for. I think that all of the above settings should be set by default to prj.conf, except for `CONFIG_LOG_MODE_MINIMAL=y`.
The base smp_svr should have image upload and some basic OS group things with all the other stuff disabled, as the main purpose is to show image upload capabilities.
We can have other sample.yml entries to allow testing other MCUmgr capabilities.
The problem we may face is such feature isolation may prevent us seeing some some collisions between features, and problems with estimating values for resource configs, like heap or stack.

SeppoTakalo (approved):

manuargue:
> > @manuargue can you please provide a fix?
> 
> I sent one: [zephyrproject-rtos/zephyr#52116](https://github.com/zephyrproject-rtos/zephyr/pull/52116)
> 
> @desowin please take this patch as fromtree or fromlist

sorry for seeing late the notifications, and thanks for the fix @mbolivar-nordic

desowin:
Fixed the checkpatch issues, the remaining ones cannot be fixed because they are either outside our control (commits from upstream) or is due to Revert-before-merge of an old commit.

nordicjm (approved):

gmarull (approved):

jfischer-no (approved):

PavelVPV (approved):

rlubos (approved):

de-nordic (approved):
MCUmgr tested on smp_svr sample, on nrf52840dk_nrf52840 seems to work fine.

PerMac (approved):

cvinayak (approved):

PR #961, opened 2022-11-09
[nrf_fromtree] nrf53: use icmsg as the default ipc transport

hubertmis:
This PR is intended to test the patch which will be pushed to upstream Zephyr. It is not intended to be merged in sdk-zephyr directly.

This PR cherry picks icmsg with multi endpoint capability and enables it as a default IPC transport for nRF53.


Thalley:
I'm not familiar with icmsg :) 

@hubertmis Is this an alternative to rpmsg? Or rather an alternative to openamp? 

I assume that this won't break anything, as we can still just select the current rpmsg/openamp solution in Zephyr, right? 

Out of curiosity, what is the difference between this, and what we have now in Zephyr?

hubertmis:
> @hubertmis Is this an alternative to rpmsg? Or rather an alternative to openamp?

OpenAMP is an implementation of RPMsg specification. Icmsg is other IPC protocol, not compatible with RPMsg (so also not compatible with OpenAMP). The main benefit of icmsg is much leaner memory layout. The main drawback is lack of zero-copy model, so for passing big amount of data additional buffer allocation scheme should be used (like Zephyr's network buffers).

Both sides of the communication must use the same IPC protocol, so if App core uses icmsg for HCI channel, then Net must use icmsg as well for HCI. Please note that it is also possible that HCI uses OpenAMP (rpmsg-complicant), while nrf_rpc uses icmsg. But such configuration would increase memory consumption to keep two IPC implementation libraries.

If you use ipc_service API from Zephyr, switching from OpenAMP to icmsg should be more-less transparent.

jfischer-no:
Outdated, please close when it is no longer needed.

hubertmis:
Stale PR

PR #965, opened 2022-11-11
[nrf fromtree] net: lwm2m: fix hanging deregistration

tokangas:
Fixed a deadlock in deregistration by moving mutex unlocking.


rlubos:
@tokangas Wasn't this commit already included in the upmerge? It's present on the `main` branch.

PR #971, opened 2022-11-14
[nrf noup] boards: thingy53_nrf5340: Add common partition map

MarekPieta:
Change introduces common static Partition Manager configuration.

Jira: NCSDK-18033


tejlmand (approved):
lgtm

tejlmand:
`[nrf noup]` exception granted for the following reasons:
- it's adding NCS specific files which doesn't exists upstream, hence no risk of merge conflicts
- it's extending board functionality which cannot be done oot, for example in sdk-nrf repo.

anangl (approved):
LGTM

PR #981, opened 2022-11-18
test-spec: matter: update files in CI-matter-test

doublemis1:
Signed-off-by: Michał Szablowski <michal.szablowski@nordicsemi.no>

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -84,7 +84,19 @@
   - "**/*"
 
 "CI-matter-test":
-  - "**/*"
+  - "include/dfu/**/*"
+  - "include/mgmt/mcumgr/**/*"
+  - "soc/arm/nordic_nrf/**/*"
+  - "subsys/dfu/**/*"
+  - "subsys/settings/**/*"
+  - "subsys/net/**/*"
+  - "subsys/mgmt/mcumgr/**/*"
+  - "drivers/net/**/*"
+  - "samples/hci_rpmsg/**/*"
+  - any:
+    - "subsys/bluetooth/**/*"
+    - "!subsys/bluetooth/mesh/**/*"
+    - "!subsys/bluetooth/audio/**/*"
 
 "CI-find-my-test":
   - "**/*"


LuDuda (approved):

.github/test-spec.yml
```diff
@@ -84,7 +84,20 @@
...
-  - "**/*"
+  - "include/dfu/**/*"
+  - "include/mgmt/mcumgr/**/*"
+  - "soc/arm/nordic_nrf/**/*"
+  - "subsys/dfu/**/*"
+  - "subsys/settings/**/*"
+  - "subsys/net/**/*"
+  - "subsys/mgmt/mcumgr/**/*"
+  - "drivers/net/**/*"
+  - "drivers/wifi/**/*"
```
Damian-Nordic, line 95: nRF700X drivers are in sdk-nrf, not zephyr, so this is probably not needed.

.github/test-spec.yml
```diff
@@ -84,7 +84,20 @@
...
+  - "samples/hci_rpmsg/**/*"
```
Damian-Nordic, line 96: Are these samples entries needed? Do our tests depend on mcumgr samples?

.github/test-spec.yml
```diff
@@ -84,7 +84,20 @@
...
+  - "samples/subsys/mgmt/mcumgr/smp_svr/**/*"
+  - any:
+    - "subsys/bluetooth/**/*"
```
Damian-Nordic, line 99: probably only `bluetooth/host` and `bluetooth/controller` are needed? At least `bluetooth/audio` shouldn't trigger our tests.

Damian-Nordic (approved):

.github/test-spec.yml
doublemis1: Do you mean hci_rpmsg?
We use that in wifi sample.

Regarding next record mcumgr, I thought we were basing our implementation of SMP support on this example

Damian-Nordic: We use SMP server, but not the sample, so this shouldn't affect us, I think.

doublemis1: ok, removed

thst-nordic (approved):

PR #993, opened 2022-11-25
[nrf noup] Update doc of tfm/psa_crypto sample

mswarowsky:
The sample is requesting an initial attestation token, which requires running the provisioning sample before, so the device has the needed key material programmed.
This updates the documentation to make users aware of this and as the provisioning sample is in sdk-nrf, a noup is used.

Signed-off-by: Markus Swarowsky <markus.swarowsky@nordicsemi.no>

### DIFF
--- a/samples/tfm_integration/psa_crypto/README.rst
+++ b/samples/tfm_integration/psa_crypto/README.rst
@@ -197,6 +197,12 @@ We need to reset the board manually after flashing the image to run this code.
 On nRF5340 and nRF9160:
 =======================
 
+.. note::
+   To be able to create an initial attestation token (IAT), the device has to be provisioned
+   by generating a Hardware Unique key (HUK) and an identity key.
+   An example on how to do this can be found in the TF-M provisioning sample of the nRF Connect SDK.
+
+
 Build Zephyr with a non-secure configuration
 (``-DBOARD=nrf5340dk_nrf5340_cpuapp_ns`` or ``-DBOARD=nrf9160dk_nrf9160_ns``).
 


joerchan:
@mswarowsky We have these test-cases in sdk-nrf/tests/tfm/ can we document it there instead?
We don't intend to use these samples anymore, only our copies.

samples/tfm_integration/psa_crypto/README.rst
```diff
@@ -197,6 +197,13 @@ We need to reset the board manually after flashing the image to run this code.
 On nRF5340 and nRF9160:
 =======================
 
+.. note::
+   To be able to create a initial attestation token (IAT) the device has to be provisioned,
+   which is done by the :ref:`provisioning_image` sample.
```
mia-ko, line 202: See if writing the reference like this will fix the build:
:ref:`nrf:provisioning_image`

mswarowsky: Sadly that didn't work

samples/tfm_integration/psa_crypto/README.rst
```diff
@@ -197,6 +197,12 @@ We need to reset the board manually after flashing the image to run this code.
 On nRF5340 and nRF9160:
 =======================
 
+.. note::
+   To be able to create a initial attestation token (IAT) the device has to be provisioned,
+   by generating a Hardware Unique key (HUK) and an identity key.
+   An example how to do this, can be found in the TF-M provisioning sample of NCS.
```
joerchan, line 203: Create a reference?

mswarowsky: Not possible as this is in sdk-zephyr :(

mbolivar-nordic: We use intersphinx to allow links between docsets. E.g. from sdk-nrf, you can say

```
:ref:`zephyr:some_zephyr_ref`
```

I believe it may be possible from sdk-zephyr to say:

```
:ref:`nrf:some_nrf_ref`
```

mswarowsky: Then the doc build in the CI here fails, but can try it again

mbolivar-nordic: Never mind, then.

joerchan (approved):

samples/tfm_integration/psa_crypto/README.rst
mia-ko, line 203:
```suggestion
   To be able to create an initial attestation token (IAT), the device has to be provisioned
   by generating a Hardware Unique key (HUK) and an identity key.
   An example on how to do this can be found in the TF-M provisioning sample of the nRF Connect SDK.
```

mia-ko (changes requested):

mia-ko (approved):

PR #994, opened 2022-11-28
[nrf noup] boards: thingy53_nrf5340: Enable MCUboot by default

MarekPieta:
Change enables MCUboot bootloader by default to allow programming samples and applications without external programmer (using MCUboot serial recovery).

Jira: NCSDK-18263

### DIFF
--- a/boards/arm/thingy53_nrf5340/Kconfig.defconfig
+++ b/boards/arm/thingy53_nrf5340/Kconfig.defconfig
@@ -8,6 +8,12 @@ if BOARD_THINGY53_NRF5340_CPUAPP || BOARD_THINGY53_NRF5340_CPUAPP_NS
 config BOARD
 	default "thingy53_nrf5340_cpuapp"
 
+config BOOTLOADER_MCUBOOT
+	default y if !MCUBOOT
+
+config BOARD_ENABLE_CPUNET
+	default y if !MCUBOOT
+
 # By default, if we build for a Non-Secure version of the board,
 # enable building with TF-M as the Secure Execution Environment.
 config BUILD_WITH_TFM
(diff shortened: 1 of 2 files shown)


nordicjm:
@MarekPieta Needs rebase

Damian-Nordic (approved):

boards/arm/thingy53_nrf5340/Kconfig.defconfig
```diff
@@ -8,6 +8,9 @@ if BOARD_THINGY53_NRF5340_CPUAPP || BOARD_THINGY53_NRF5340_CPUAPP_NS
 config BOARD
 	default "thingy53_nrf5340_cpuapp"
 
+config BOOTLOADER_MCUBOOT
+	default y if !MCUBOOT
```
anangl, line 12: Wouldn't just `default y` be sufficient when the `BOOTLOADER_MCUBOOT` option itself: https://github.com/nrfconnect/sdk-zephyr/blob/c6184a9ff7382e5c699c35b11c9b4e096a7f247d/Kconfig.zephyr#L749?

tejlmand:
why is this a `nrf noup` ?
Can't this PR go upstream.

boards/arm/thingy53_nrf5340/Kconfig.defconfig
tejlmand: no, cause dependencies are not inherited.

So this needs either the current `if !MCUBOOT` or a `depends on !MCUBOOT`

MarekPieta: Removing this dependency results in adding MCUboot child image in a never ending-loop (MCUboot child image seems to add MCUboot child image recursively).

MarekPieta:
Hi @tejlmand,
nordic,pm-ext-flash is directly related to the Partition Manager which is specific to the NCS

de-nordic:
> Hi @tejlmand, nordic,pm-ext-flash is directly related to the Partition Manager which is specific to the NCS. That's why I think we should keep it as `[nrf noup]`

The binding description is only available from sdk-nrf.

de-nordic (approved):

carlescufi (approved):

tejlmand (approved):

kapi-no (approved):

PR #996, opened 2022-11-30
[nrf fromtree] Bluetooth: Audio: Fix MCC disconnect issue

rick1082:
When MCC disconnected, it would memset the entire instance. This causes issues, because the subscription parameters would also be memset, causing the `notify` callback to be set to NULL. If this was done as part of a disconnect, the GATT layer would attempt to call the notify callback as part of cleaning up the ACL, but if the callback was set to NULL (from the memset) this caused a fatal error.

Signed-off-by: Emil Gydesen <emil.gydesen@nordicsemi.no>
(cherry picked from commit e80c70fd051da70afc4bec8203e82c0265d1ca99)

### DIFF
--- a/subsys/bluetooth/audio/mcc.c
+++ b/subsys/bluetooth/audio/mcc.c
@@ -74,27 +74,6 @@ struct mcs_instance_t {
 #endif /* CONFIG_BT_MCC_OTS */
 	uint16_t content_control_id_handle;
 
-	struct bt_gatt_subscribe_params player_name_sub_params;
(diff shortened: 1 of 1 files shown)


gWacey (approved):

nordicjm (changes requested):
No cherry-picked line

pirun (approved):

nordicjm (approved):

asbjornsabo (approved):
I have already reviewed, and approved, this PR upstream.
It is also needed in NCS.

PR #1004, opened 2022-12-02
Revert "[nrf noup] boards: thingy53_nrf5340: Add common partition map"

MarekPieta:
This reverts commit 9a8680372fdb6e09f3d6537c8c6751dd5c50bf86.


kapi-no (approved):

tejlmand (changes requested):
I don't believe we should revert, it's just more work for ourselves, if anything, this should be merged instead #994

MarekPieta:
The PR could Thingy;53 known issue from NCS 2.2.0.  Still I understand it was rejected and we need to live with the known issue.

MarekPieta:
This would not go into NCS 2.2.0. There is no point to introduce it later on. Closed.

PR #1005, opened 2022-12-04
Revert "[nrf noup] kconfig: Disallow FPU for SPM"

joerchan:
This reverts commit ee89a68db75bf0d52dbfc9ee0ad9b4e4dfc18500.

Signed-off-by: Joakim Andersson <joakim.andersson@nordicsemi.no>

### DIFF
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -785,7 +785,6 @@ menu "Floating Point Options"
 config FPU
 	bool "Floating point unit (FPU)"
 	depends on CPU_HAS_FPU
-	depends on !IS_SPM
 	help
 	  This option enables the hardware Floating Point Unit (FPU), in order to
 	  support using the floating point registers and instructions.


hakonfam (approved):
Are you not supposed to use some special commit message tag to make sure that the tooling used for rebases automatically removes some patch commit? 

Like here https://github.com/nrfconnect/sdk-zephyr/pull/919 ? Or is that only when a patch removes parts of another commit?

joerchan:
> Are you not supposed to use some special commit message tag to make sure that the tooling used for rebases automatically removes some patch commit?
> 
> Like here #919 ? Or is that only when a patch removes parts of another commit?

I think this is correct, your example is a fixup, and not a revert. Due to the rebase of sdk-zephyr we don't have any reverts left to compare.
@carlescufi @mbolivar-nordic Is this correct?

carlescufi:
> > Are you not supposed to use some special commit message tag to make sure that the tooling used for rebases automatically removes some patch commit?
> > Like here #919 ? Or is that only when a patch removes parts of another commit?
> 
> I think this is correct, your example is a fixup, and not a revert. Due to the rebase of sdk-zephyr we don't have any reverts left to compare. @carlescufi @mbolivar-nordic Is this correct?

I am quite sure you are right, @joerchan. But I will ask @mbolivar-nordic to take another look just in case.

mbolivar-nordic:
@joerchan is right -- just use `git revert`. The oss-history tooling looks for reverts when it's figuring out what commits to leave in NCS history after a rebase, so this is the only supported way to do this, and it's why upmerges always include a stack of reverts for downstream commits that conflict, followed by reapplications of conflict-free versions.

I could have sworn we had this documented somewhere, but I couldn't find it, so I wrote some new internal documentation to point people to if this comes up again and will point you all to it.

rlubos (approved):

PR #1006, opened 2022-12-05
[nrf fromlist] drivers: usb_dc_nrfx: Do not uninit on detach

desowin:
It is possible for nrfx usbd driver to send NRFX_USBBD_EP_ABORTED during endpoint disable. The event is passed to event handler registered with nrfx_usbd_init(). The nrfx_usbd_uninit() removes the registered event handler, replacing it with NULL. If any event is sent after uninit, the NULL pointer is executed and device crashes.

Do not uninit nrfx usbd driver on detach so it is possible for the usb_disable() to disable all the endpoints.

Fixes: 460ca86527dd ("drivers: usb_dc_nrfx: Always allow endpoint disable")

Upstream PR: zephyrproject-rtos/zephyr#52752

(cherry picked from commit fa9a0e31223b3138d87056eabb32b0edb4a18310)
Signed-off-by: Tomasz Moń <tomasz.mon@nordicsemi.no>

### DIFF
--- a/drivers/usb/device/usb_dc_nrfx.c
+++ b/drivers/usb/device/usb_dc_nrfx.c
@@ -1264,7 +1264,6 @@ static void usbd_work_handler(struct k_work *item)
 int usb_dc_attach(void)
 {
 	struct nrf_usbd_ctx *ctx = get_usbd_ctx();
-	nrfx_err_t err;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


desowin:
This is not really needed in NCS because the commit it fixes has not yet been upmerged.

PR #1007, opened 2022-12-05
Added nrfjprog support to Adafruit feather nrf52840 board

studiofuga:
The Adafruit Feather nrf52840 has a jlink port that can be used for flashing and debugging.
This patch adds the support for it.


nordicjm:
This patch should be applied directly to upstream zephyr https://github.com/zephyrproject-rtos/zephyr/ since that is where the board is. Then these changes will be picked up and reflected here.

nordicjm:
Closing this PR, needs to be submitted upstream.

PR #1019, opened 2023-01-02
Export the last ICMP sequence number

krish2718:
This is a temporary hack needed by WFA DUT to return the sequence number to the WTS. Need to replaced with proper ICMP public API.

Signed-off-by: Susan Rapaka <susan.rapaka@nordicsemi.no>
Signed-off-by: Ravi dondaputi <radi.dondaputi@nordicsemi.no>


krish2718:
This is no longer needed and implemented as part of wfa_dut sample.

PR #1021, opened 2023-01-04
[nrf noup] boards: thingy53_nrf5340: Set MCUBoot to use all of RAM

sigvartmh:
Change enables MCUBoot to use all available RAM on thingy53 as it require more RAM than what is reserved as secure. This will make thingy53 a little bit less secure than our other boards. Since this is a prototype board I think it should be okay as we require the enabled MCUBoot features for it to be programmable without a debugger.

Jira: NCSDK-17250

### DIFF
--- a/boards/arm/thingy53_nrf5340/Kconfig.defconfig
+++ b/boards/arm/thingy53_nrf5340/Kconfig.defconfig
@@ -14,6 +14,9 @@ config BOOTLOADER_MCUBOOT
 config BOARD_ENABLE_CPUNET
 	default y if !MCUBOOT
 
+config MCUBOOT_USE_ALL_AVAILABLE_RAM
+	default y if BOOTLOADER_MCUBOOT
+
 # By default, if we build for a Non-Secure version of the board,
 # enable building with TF-M as the Secure Execution Environment.
 config BUILD_WITH_TFM


sigvartmh:
Related nRF PR that requires this change https://github.com/nrfconnect/sdk-nrf/pull/9780

boards/arm/thingy53_nrf5340/Kconfig.defconfig
```diff
@@ -14,6 +14,9 @@ config BOOTLOADER_MCUBOOT
 config BOARD_ENABLE_CPUNET
 	default y if !MCUBOOT
 
+config MCUBOOT_USE_ALL_AVAILABLE_RAM
+	default y if BOOTLOADER_MCUBOOT
+
```
hakonfam, line 19: We should move this to `sdk-nrf` instead. Just define it in some of our kconfig and then make sure to have `depends on` the board you want to enable it for.

hakonfam (changes requested):

boards/arm/thingy53_nrf5340/Kconfig.defconfig
MarekPieta: Some of the features (like defining common static PM configuration, enabling cpunet and MCUboot bootloader) are already added as `[nrf noup]`. If change affect only a specific board in the NCS, I think it may be more reasonable to put it here (it would be way easier for the users to localize source of the change).

MarekPieta: btw. @sigvartmh, could MCUboot clean up the used non-secure RAM before booting application? That may allow us to prevent leaking data to non-secure code, right ?

sigvartmh: Sure I'll add one. I was just had a moment of laziness. I thought about then I thought it doesn't make sense that MCUBoot should use more than the available secure RAM unless you are doing something very custom and then your already a little bit on your own.

I'll add the function however this will affect the boot time.

nordicjm: I can't view the linked issue, can you explain what enabling `MCUBOOT_USE_ALL_AVAILABLE_RAM` does?

sigvartmh: `MCUBOOT_USE_ALL_AVAILABLE_RAM ` will make MCUBoot use all the RAM available on the chip. Maybe I should add this wording.

So the way it works after the PR is that MCUboot will have just the `secure_ram` available while before this PR mcuboot would have the start address of the `nonsecure_ram` this could potentially leak data from secure non-aware to non-secure. Since MCUBoot does not clean up residual data in RAM.

nordicjm: How does this work with PCD for updating the network core? Not all RAM can be used

sigvartmh:
Moved the Kconfig default out of Zephyr and into NCS

sigvartmh:
https://github.com/nrfconnect/sdk-nrf/pull/9780/files#diff-20a0ebbac4434a507dd7e71b80070ffa6d2d99756fbda1fe2ccb08a78007e86cR118

PR #1024, opened 2023-01-12
[nrf fromtree] net: wifi: Add support for regulatory domain configura…

krish2718:
…tion

Wi-Fi bands are regulated by a governing body depending on operating country, add support for the user to provide a country of operation as a hint to the Wi-Fi chipset.

Ideally if the chipset supports this is all handled internally, in that case "get" is useful but for testing and other usecases add a "set" as well, similar to "iw reg set" or "country_code=" configuration in hostapd/wpa_supplicant in Linux world.

This add a new offload API operation "reg_domain" that can be used to either get or set the regulatory information.

The validation is left to the underlying chipset, shell only does basic validation, (XY/00).

This is just a regulatory hint to the chipset, there could be other regulatory hints e.g., beacon that can override this configuration, so, an additional option to force this setting despite other hints is also given for testing purposes.

FYI, the standard database used is [1].

[1] - https://git.kernel.org/pub/scm/linux/kernel/git/sforshee/wireless-regdb.git/tree/db.txt

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/53541

### DIFF
--- a/include/zephyr/net/wifi.h
+++ b/include/zephyr/net/wifi.h
@@ -12,6 +12,8 @@
 #ifndef ZEPHYR_INCLUDE_NET_WIFI_H_
 #define ZEPHYR_INCLUDE_NET_WIFI_H_
 
+#define WIFI_COUNTRY_CODE_LEN 2
+
 /* Not having support for legacy types is deliberate to enforce
  * higher security.
  */
--- a/include/zephyr/net/wifi_mgmt.h
+++ b/include/zephyr/net/wifi_mgmt.h
@@ -40,6 +40,7 @@ enum net_request_wifi_cmd {
 	NET_REQUEST_WIFI_CMD_PS_MODE,
 	NET_REQUEST_WIFI_CMD_TWT,
 	NET_REQUEST_WIFI_CMD_PS_CONFIG,
+	NET_REQUEST_WIFI_CMD_REG_DOMAIN,
 };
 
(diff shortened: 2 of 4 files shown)


rlubos (approved):

sachinthegreen (approved):

krish2718:
After multiple re-runs and debugging, the problem lies with `gitlint` in the compliance. Due to https://github.com/jorisroovers/gitlint/blob/main/CHANGELOG.md#v090-2017-12-03 using a single commit along with `--commits` won't work, not sure how this got missed as this was `0.9.0` but we use `0.18.0`?

So, this needs a fix.

carlescufi:
> After multiple re-runs and debugging, the problem lies with `gitlint` in the compliance. Due to https://github.com/jorisroovers/gitlint/blob/main/CHANGELOG.md#v090-2017-12-03 using a single commit along with `--commits` won't work, not sure how this got missed as this was `0.9.0` but we use `0.18.0`? @carlescufi
> 
> So, this needs a fix.

It's because we just started using this and all PRs so far had more than one commit I believe

carlescufi (changes requested):
Please change the gitlint commit message so it amends:
`[nrf noup] ci: NCS-specific CI tweaks`

krish2718:
Depends on #1027

carlescufi (approved):

PR #1025, opened 2023-01-13
thingy53: enable MCUmgr Bluetooth support

nordicjm:
Ports the CDC console support for thingy53, and MCUmgr handler changes and adds a no-up patch to enable MCUmgr support with img_mgmt if MCUboot is enabled.


de-nordic (approved):

nordicjm:
@MarekPieta give this a try, the Kconfig is `CONFIG_BOARD_ENABLE_MCUMGR_BT_OTA_DFU`

nordicjm (commented):
https://github.com/zephyrproject-rtos/zephyr/pull/54008 needs to be cherry-picked as a fromlist

kapi-no:
@shantha-14, is it okay for us to move forward with this change? The main purpose of this PR is to improve UX for the Thingy:53 target in NCS.

shantha-14:
> @shantha-14, is it okay for us to move forward with this change?

We need to twister tests to be made green, we will try to get this in for release

boards/arm/thingy53_nrf5340/Kconfig
```diff
@@ -59,6 +59,24 @@ config DOMAIN_CPUNET_BOARD
 	  another board. For example hci_rpmsg on the nRF5340_cpunet for
 	  Bluetooth applications.
 
+config BOARD_ENABLE_MCUMGR_BT_OTA_DFU
+	bool "MCUmgr and OTA DFU"
+	select MCUMGR
+	select MCUMGR_SMP_BT
+	select MCUMGR_CMD_IMG_MGMT
```
MarekPieta, line 66: Can we also enable this part automatically?

```
# Enable custom SMP request to erase settings partition.                                             
CONFIG_MCUMGR_GRP_ZEPHYR_BASIC=y                                                                     
CONFIG_MCUMGR_GRP_BASIC_CMD_STORAGE_ERASE=y 
```

MarekPieta: System workq stack may need to be updated to prevent overflows in default configurations:
```
# Increase system workqueue stack size to prevent stack overflows while handling                     
# SMP requests.                                                                                      
CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=4096
```

MarekPieta: Consider also updating:

* Bluetooth buffers
* enabling SMP reassembly feature
* updating peripheral preferred connection parameters (+enable CONFIG_MCUMGR_SMP_BT_CONN_PARAM_CONTROL)

This could ensure quicker DFU. @mkapala-nordic works on it already so please align with him to avoid duplicating work.

boards/arm/thingy53_nrf5340/Kconfig
```diff
@@ -59,6 +59,24 @@ config DOMAIN_CPUNET_BOARD
...
+	select MCUMGR_CMD_OS_MGMT
+	select NORDIC_QSPI_NOR
```
MarekPieta, line 68: Can we also by default enable: `CONFIG_IMG_ERASE_PROGRESSIVELY`?
Consider also aligning QSPI flash configuration to defaults that were used by Thigny:53:
```
CONFIG_NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE=4096                                                   
CONFIG_NORDIC_QSPI_NOR_STACK_WRITE_BUFFER_SIZE=16 
```

MarekPieta (commented):
A few additional alignments for Thingy:53 seem needed here

boards/arm/thingy53_nrf5340/Kconfig
nordicjm: Fixed

boards/arm/thingy53_nrf5340/Kconfig
nordicjm: Fixed except workqueue stack size, I do not want to force something which might be set to a different size by applications

mkapala-nordic: Shouldn't BT buffers also be increased for CPUNET?

boards/arm/thingy53_nrf5340/Kconfig
```diff
@@ -59,6 +59,52 @@ config DOMAIN_CPUNET_BOARD
...
+	default 498 if BOARD_ENABLE_MCUMGR_BT_OTA_DFU
+
+config BT_BUF_ACL_RX_SIZE
+	default 502 if BOARD_ENABLE_MCUMGR_BT_OTA_DFU
+
+config BT_BUF_ACL_TX_SIZE
+	default 502 if BOARD_ENABLE_MCUMGR_BT_OTA_DFU
+
+config BT_CTLR_DATA_LENGTH_MAX
+	default 251 if BOARD_ENABLE_MCUMGR_BT_OTA_DFU
```
kapi-no, line 94: We need to match this configuration in the Network Core image

boards/arm/thingy53_nrf5340/Kconfig
```diff
@@ -59,6 +59,52 @@ config DOMAIN_CPUNET_BOARD
...
+config BT_L2CAP_TX_MTU
+	default 498 if BOARD_ENABLE_MCUMGR_BT_OTA_DFU
+
+config BT_BUF_ACL_RX_SIZE
+	default 502 if BOARD_ENABLE_MCUMGR_BT_OTA_DFU
+
+config BT_BUF_ACL_TX_SIZE
+	default 502 if BOARD_ENABLE_MCUMGR_BT_OTA_DFU
+
+config BT_CTLR_DATA_LENGTH_MAX
```
kapi-no, line 93: Is it okay to configure the Bluetooth Controller configuration in the Application image? I think this could produce a warning during the build time as the Bluetooth Controller configuration is not available in the Application image of the nRF5340

boards/arm/thingy53_nrf5340/Kconfig
nordicjm: You can't know if this option has been selected from the network core image so have no way to know

nordicjm: You can't know if this option has been selected from the network core image so have no way to know

boards/arm/thingy53_nrf5340/Kconfig
kapi-no: See the child image configuration in the SMP Server sample:

https://github.com/nrfconnect/sdk-zephyr/blob/main/samples/subsys/mgmt/mcumgr/smp_svr/child_image/hci_rpmsg.conf

nordicjm: As above:

> You can't know if this option has been selected from the network core image so have no way to know

So either (for the network core) these are always left at their defaults or set to their maximums whenever bluetooth is enabled for the network core - which do we want to go with?

boards/arm/thingy53_nrf5340/Kconfig
kapi-no: I don't think that this particular option is available in the application image as the Bluetooth Controller configuration should be disabled there. It is not present in the autoconf.h file from the SMP server sample (built for nRF5340 DK):

```cat build//zephyr/include/generated/autoconf.h | grep 'BT_CTLR'```

It may be the case that this configuration option does not have any impact on the building stage, but it seems unnecessary

File: boards/arm/thingy53_nrf5340/Kconfig (hunk 1 above)
Comment (kapi-no) [reply boards/arm/thingy53_nrf5340/Kconfig:66]: @nordicjm. we had a discussion today on how to solve it as @mkapala-nordic already signalled this issue today morning. We were considering changing BT buffer defaults whenever the ```BT``` configuration is enabled in the Network Core.

@mkapala-nordic, do I recall correctly? Did you have time to create a commit that shows this approach?

File: boards/arm/thingy53_nrf5340/Kconfig (hunk 4 above)
Comment (nordicjm) [reply boards/arm/thingy53_nrf5340/Kconfig]: Moved

File: boards/arm/thingy53_nrf5340/Kconfig (hunk 3 above)
Comment (nordicjm) [reply boards/arm/thingy53_nrf5340/Kconfig]: Fixed to maximum by default

File: boards/arm/thingy53_nrf5340/Kconfig [hunk 5]
```diff
@@ -67,4 +114,27 @@ config DOMAIN_CPUAPP_BOARD
...
+config BT_BUF_EVT_RX_SIZE
+	default 255
+
+config BT_BUF_EVT_DISCARDABLE_SIZE
+	default 255
+
+config BT_BUF_CMD_TX_SIZE
+	default 255
+
+config BT_CTLR_DATA_LENGTH_MAX
```
kapi-no, line 135: It seems that the HCI RPMsg child image does not build for Thingy:53 because of RAM overflows.

```region `RAM' overflowed by 12120 bytes```

Unfortunately, the HCI RPMsg sample in the upstream sets the number of buffers to 16 and this configuration cannot be changed using defaults in this file. We may consider adding the hci_rpmsg.conf to each Bluetooth sample to make it work.

Comment (nordicjm) [reply boards/arm/thingy53_nrf5340/Kconfig:136]: Fixed to allow 3 connections by default, builds with:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      184820 B       222 KB     81.30%
             RAM:       47016 B        64 KB     71.74%
           SRAM1:          0 GB        64 KB      0.00%
        IDT_LIST:          0 GB         2 KB      0.00%
```
Could possibly extend this to 4 connections?

Comment (kapi-no) [reply boards/arm/thingy53_nrf5340/Kconfig:136]: Thanks, that's a better solution. We will see during our tests if 3 connections are enough. Being too close to 100% of RAM usage may trigger more maintenance work in the trigger as the Zephyr Bluetooth stack keeps growing and could potentially trigger RAM overflow after some time.

**Comment by pirun** [comment]
Not sure if it related to this PR or mobile tool.

I've test with platform nRF21540dk, nRF5340dk, LE Audio dk.
I try to test [ SMP_SVR sample](https://github.com/nrfconnect/sdk-zephyr/tree/main/samples/subsys/mgmt/mcumgr/smp_svr) with mobile nRF Connect Device Manager.
[IOS ](https://github.com/NordicSemiconductor/IOS-nRF-Connect-Device-Manager) works fine.
[Android](https://github.com/NordicSemiconductor/Android-nRF-Connect-Device-Manager) app will crash once Android connected to SMP_SVR.

**Review by mkapala-nordic — APPROVED** [review]

File: samples/subsys/mgmt/mcumgr/smp_svr/src/bluetooth.c [hunk 6]
```diff
@@ -58,9 +58,24 @@ BT_CONN_CB_DEFINE(conn_callbacks) = {
...
+}
+
 void start_smp_bluetooth_adverts(void)
 {
+	int rc;
+
 	k_work_init(&advertise_work, advertise);
+	rc = bt_enable(bt_ready);
 
 	k_work_submit(&advertise_work);
```
MarekPieta, line 77: Should we actually start advertising here? Shouldn't we wait for Bluetooth readiness (so that submit the work only from `bt_ready` callback?

MarekPieta (approved):

samples/subsys/mgmt/mcumgr/smp_svr/src/bluetooth.c
kapi-no: Good catch! Seems like a rebase error since the upstream code is correct:

https://github.com/zephyrproject-rtos/zephyr/commit/4516e7161c8e3ca93239e189741f8c60ce678fb3#

See the force push diff here: https://github.com/nrfconnect/sdk-zephyr/compare/0388cfd8f0612f8eb46308215c58756e58c3b5e8..04b9f936ebdfec8ed2c72c7fc05985463192b99a

kapi-no:
Rebased the PR on the main branch. The file diff hasn't changed

kapi-no:
Rebased to fix conflicts

shantha-14 (approved):

PR #1029, opened 2023-01-17
[nrf fromlist] net: ppp: Fix PPP connection with host

rlubos:
A follow up to commit 1d7a077e11 - apparently the PPP interface should not be brought entirely down internally, as this can break further communication with the host. Because of that, reintroduce functionality that used to be covered by net_if_carrier_down() (which basically skips the L2 enable(false) call), but limited to PPP scope only.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/53828

Signed-off-by: Robert Lubos <robert.lubos@nordicsemi.no>


trantanen (approved):
Verified with PR https://github.com/nrfconnect/sdk-nrf/pull/9939 that PPP works as before

de-nordic (approved):

PR #1052, opened 2023-02-07
lib: tftp_client: Add data handler in TFTP client

junqingzou:
Currently the client relies on buffer from application to save received data, which limits the size of remote file to the size of allocated buffer.

Add an alternative data handler to send the received data (max 512 bytes each) to application directly, with the merit of 
.No limitation on the size of remote file
.No memory copy in tftp_client library
.No buffer allocation in application required

Signed-off-by: Jun Qing Zou <jun.qing.zou@nordicsemi.no>

### DIFF
--- a/include/zephyr/net/tftp.h
+++ b/include/zephyr/net/tftp.h
@@ -19,9 +19,22 @@
 extern "C" {
 #endif
 
+/**
+ * @typedef tftp_data_handler_t
+ *
+ * Handler to handle data received from TFTP server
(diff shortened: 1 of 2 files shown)


rlubos:
@junqingzou Did you open an upstream PR?

junqingzou:
Submitted to Zephyr upstream with TFTP PUT support, too
https://github.com/zephyrproject-rtos/zephyr/pull/54643

PR #1054, opened 2023-02-08
[nrf noup] Bluetooth: Mesh: zero randomization for friend's adv

alxelax:
Friend's replies on LPN's polls do not assume randomization in advertiser. Zero randomization will help to optimize time when LPN keeps receiving window open and save power.

Signed-off-by: Aleksandr Khromykh <aleksandr.khromykh@nordicsemi.no>

### DIFF
--- a/subsys/bluetooth/mesh/Kconfig
+++ b/subsys/bluetooth/mesh/Kconfig
@@ -812,7 +812,7 @@ config BT_MESH_LPN_INIT_POLL_TIMEOUT
 config BT_MESH_LPN_SCAN_LATENCY
 	int "Latency for enabling scanning"
 	range 0 50
-	default 15
+	default 2
 	help
 	  Latency in milliseconds that it takes to enable scanning. This
 	  is in practice how much time in advance before the Receive Window
--- a/subsys/bluetooth/mesh/adv_ext.c
+++ b/subsys/bluetooth/mesh/adv_ext.c
@@ -13,6 +13,9 @@
 #include <zephyr/bluetooth/bluetooth.h>
 #include <zephyr/bluetooth/hci.h>
 #include <zephyr/bluetooth/mesh.h>
+#if defined(CONFIG_BT_LL_SOFTDEVICE)
+#include <sdc_hci_vs.h>
+#endif
 
 #include "common/bt_str.h"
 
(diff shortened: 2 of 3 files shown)


subsys/bluetooth/mesh/Kconfig
```diff
@@ -812,7 +812,7 @@ config BT_MESH_LPN_INIT_POLL_TIMEOUT
 config BT_MESH_LPN_SCAN_LATENCY
 	int "Latency for enabling scanning"
 	range 0 50
-	default 15
+	default 4
```
PavelVPV, line 815: Probably this should be 2. Could you try to test it with 2?

subsys/bluetooth/mesh/adv_ext.c
```diff
@@ -173,6 +192,13 @@ static int adv_start(struct bt_mesh_ext_adv *adv,
 
 	adv->timestamp = k_uptime_get();
 
+	if (adv->tag & BT_MESH_FRIEND_ADV) {
```
PavelVPV, line 195: It is probably not needed to set zero randomness here as it is constant configuration IIRC. Could you check without this code?

subsys/bluetooth/mesh/Kconfig
alxelax: It works. I'll set 2.

subsys/bluetooth/mesh/adv_ext.c
alxelax: Yep, you're right

Andrewpini:
I see that there is some include bug in one of the provisioner tests in twister. Do you have a fix for this issue?

subsys/bluetooth/mesh/adv_ext.c
```diff
@@ -456,6 +479,13 @@ int bt_mesh_adv_enable(void)
 		if (err) {
 			return err;
 		}
+
+		if (IS_ENABLED(CONFIG_BT_LL_SOFTDEVICE) && adv->tag & BT_MESH_FRIEND_ADV) {
```
PavelVPV, line 483: I think you need to wrap body of `set_adv_randomness` into ifdef, not the whole function, to use `IS_ENABLED`.

alxelax:
> I see that there is some include bug in one of the provisioner tests in twister. Do you have a fix for this issue?

@Andrewpini, fixed

PavelVPV (changes requested):
I guess you need to pool new version of commit from upstream PR

PavelVPV (approved):

Andrewpini (approved):

tmon-nordic:
The fromlist seems to be the only option because there is no matching upstream bug, so it won't be merged before 3.3.0 is released. However, what is the real reason behind the noup? Why is it not applicable to upstream?

alxelax:
> The fromlist seems to be the only option because there is no matching upstream bug, so it won't be merged before 3.3.0 is released. However, what is the real reason behind the noup? Why is it not applicable to upstream?

Hi @tmon-nordic, to skip using random delay during advertisement in Bluetooth controller the custom HCI command is used. It exists, unfortunately, only in SDC (softdevice). There is no such command in both Zephyr's host\controller. As a BLE Mesh maintainer, I am not happy with this either but I do not see another way to use this feature.

PR #1065, opened 2023-02-16
Bluetooth: Host: Add PAWR - Advertiser

pirun:
Add bt_le_per_adv_param_v2 to set pawr timing.
Add callback for subevent data request.


Thalley:
@pirun Is there any plans to add this to upstream Zephyr as well? For the host point of view, it can be added like this, even if the Zephyr LL does not support it yet

hermabe:
> @pirun Is there any plans to add this to upstream Zephyr as well? For the host point of view, it can be added like this, even if the Zephyr LL does not support it yet

Agree, all host changes should ideally be contributed upstream. One issue is how to test it, as the Zephyr LL does not support PAwR.

Thalley:
> Agree, all host changes should ideally be contributed upstream. One issue is how to test it, as the Zephyr LL does not support PAwR.

Indeed. On the nRF5340 it is easy to use Zephyr upstream on the host/appcore, and then e.g. SoftDevice LL (if that supports this feature) on the netcore. That's how I've been testing features not available for the Zephyr LL in the past.

PR #1069, opened 2023-02-18
[nrf fromlist] net: l2: ethernet: Fix double free

krish2718:
In the case of no ARP entry, the incoming packet is added to the ARP's pending queue, while ARP is being resolved. Here a reference is taken by the ARP layer to the packet to avoid it being freed, but the Ethernet immediately puts down the reference and send the ARP packet to the driver.

If the ARP request fails for some reason, L2 returns failure to net_if which then puts down the reference and the packet will be freed as the reference count is now zero.

But the packet is still in the ARP's pending queue and after timeout ARP will put down the reference causing double free bus fault (double free message is only seen if the CONFIG_NET_PKT_LOG_LEVEL_DBG is enabled, so, a bit hard to debug.

Fix this but not putting down the reference in Ethernet that was taken by ARP and let ARP timeout handling free the pending packet.

(cherry picked from commit 6934eaaae8777b2caebcde49bc1f83ae53073274)

### DIFF
--- a/subsys/net/l2/ethernet/arp.c
+++ b/subsys/net/l2/ethernet/arp.c
@@ -744,6 +744,19 @@ void net_arp_clear_cache(struct net_if *iface)
 	k_mutex_unlock(&arp_mutex);
 }
 
+int net_arp_clear_pending(struct net_if *iface, struct in_addr *dst)
+{
+	struct arp_entry *entry = arp_entry_find_pending(iface, dst);
+
+	if (!entry) {
+		return -ENOENT;
+	}
+
+	arp_entry_cleanup(entry, true);
+
+	return 0;
+}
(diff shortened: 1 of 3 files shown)


tmon-nordic (changes requested):
[nrf fromlist] tag must be accompanied with "Upstream PR: <link to upstream PR>", not cherry picked from hash (which will change once the upstream PR is merged).

tmon-nordic (approved):

rlubos (changes requested):
I don't think this is a valid change, see https://github.com/zephyrproject-rtos/zephyr/pull/54949#pullrequestreview-1305438500

rlubos (approved):

PR #1074, opened 2023-02-21
[backport-v1.9][nrf fromtree] soc: nrf53: Add workaround for anomaly 160

anangl:
Cherry-pick a set of upstream commits to get the workaround for nRF53 anomaly 160 in NCS 1.9.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
-@Library("CI_LIB") _
+@Library("CI_LIB@v1.9-branch") _
 
 def pipeline = new ncs.sdk_zephyr.Main()
 
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -54,6 +54,14 @@ config CODE_DATA_RELOCATION_SRAM
 	  config is used to create an MPU entry for the SRAM space used for code
 	  relocation.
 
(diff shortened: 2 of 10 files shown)


PerMac:
I see twister is failing due to 
`FileNotFoundError: [Errno 2] No such file or directory: 'zephyr/scripts/quarantine.yaml'`
It is because the CI call link to quarantine file to use but it is not in the given version. I think this can be fixed in 2 places:
- add a quarantine file in this PR at (zephyr/)scripts/quarantine.yaml with the content:
```
- scenarios:
   - None
  platforms:
   - all

```
This will fulfill the CI command by passing a "mocked" quarantine

The other solution would be to modify the CI side and modify/create a CI versioning for 1.9 and modify the twister call there to not ask for quarantine.

I think the first option is easier and maybe even better, because if some issues appear latter on, which wouldn't be fixed in 1.9 than those can be added to quarantine

thst-nordic:
it is using the latest (2.2.99) version of the CI configuration,  we probably should switch this to v1.9

mbolivar-nordic (approved):

cvinayak (approved):

kruithofa (approved):

PR #1075, opened 2023-02-21
[nrf fromtree] soc: nrf53: Add workaround for anomaly 160

anangl:
Cherry-pick a set of upstream commits to get the workaround for nRF53 anomaly 160 in NCS.

### DIFF
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -45,6 +45,14 @@ config CODE_DATA_RELOCATION_SRAM
 	  config is used to create an MPU entry for the SRAM space used for code
 	  relocation.
 
+config ARM_ON_ENTER_CPU_IDLE_HOOK
+	bool
+	help
+	  Enables a hook (z_arm_on_enter_cpu_idle()) that is called when
+	  the CPU is made idle (by k_cpu_idle() or k_cpu_atomic_idle()).
+	  If needed, this hook can be used to prevent the CPU from actually
+	  entering sleep by skipping the WFE/WFI instruction.
+
 rsource "core/aarch32/Kconfig"
 rsource "core/aarch32/Kconfig.vfp"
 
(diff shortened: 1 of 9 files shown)


PerMac:
@gmarull @thst-nordic Doc building is failing on this PR with the below error.  Have you seen it before/know the fix?

WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
error: externally-managed-environment

gmarull:
> @gmarull @thst-nordic Doc building is failing on this PR with the below error. Have you seen it before/know the fix?
> 
> WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag. error: externally-managed-environment

We should actually not be building PDF on PRs, https://github.com/zephyrproject-rtos/zephyr/commit/e11d82ba31db95e5de74eff02c5626bcd235b854 can be imported

gmarull:
> @gmarull I see. However, I fail to see if this will address the issue seen in this PR. It looks like something on the environment configuration side. I checked and it looks like all pdf docs building wrokflows, from last 12h failed. Even the "scheduled" ones (no pr). https://github.com/nrfconnect/sdk-zephyr/actions/workflows/doc-build.yml I will open a separate gh issue for this workflow, since it is not only issue of this PR.

we should actually disable pdf build in our fork, in NCS we don't use PDF at all.

PerMac:
I see e.g. this https://github.com/python/cpython/issues/102134 It seems this might be related to same recent change in ubuntu?

cvinayak (approved):

carlocaione (approved):

PR #1078, opened 2023-02-22
Add BL5340PA development kit

ahedin25:
The BL5340PA contains a nRF21540 FEM.  It requires the Nordic MSPL FEM driver and the Laird FEM driver.


jfischer-no:
I do not think it is acceptable for this repository, should be part of upstream Zephyr or you own downstream project.

PR #1087, opened 2023-03-02
Remove dependency on MBEDTLS_BUILTIN for MBEDTLS_DEBUG

vili-nordic:


joerchan:
@vili-nordic As Robert mentioned, the (cherry-picked from <commit-sha>) message is missing, and there is a tbd in the message also.

SebastianBoe (approved):

PR #1096, opened 2023-03-08
Zephyr upmerge 06032023

jfischer-no:
Zephyr upmerge based on upstream commit https://github.com/zephyrproject-rtos/zephyr/commit/6d9adf2e8af476819fce802d326a02226d98ed0c


jfischer-no:
> I see the following which have been reverted and not reapplied:

> e498ffb1e3 Revert "[nrf noup] samples: psa_crypto: update documentation"

sample is removed upstream by commit c63fb217601fb35903536db1c258571b8cd81b9b ("samples: tfm_integration: Remove psa_crypto")

> bc6b19d20b Revert "[nrf noup] boards: arm: thingy53: Disable USB CDC added by MCUBoot"

Thanks, fixed

nordicjm (approved):

subsys/mgmt/mcumgr/Kconfig
```diff
@@ -7,7 +7,7 @@ menuconfig MCUMGR
 	depends on NET_BUF
 	depends on ZCBOR
 	depends on CRC
-	select BOOT_IMAGE_ACCESS_HOOKS if (SOC_NRF5340_CPUAPP_QKAA && MCUMGR_CMD_IMG_MGMT)
+	select BOOT_IMAGE_ACCESS_HOOKS if (SOC_NRF5340_CPUAPP_QKAA && MCUMGR_GRP_IMG)
```
nordicjm: Can accept this change for upmerge as it replicates the existing code with new Kconfig name but this actually wrong, an nrf53 user only wants boot image hooks _if_ they are supporting network core firmware updates, if they are not, this should be disabled, as or right now it can't be disabled at all which is undesirable and wrong. @de-nordic

de-nordic: I can see problem here, although I do not think there is not Kconfig magic to fix that. Only fix I can see here is to directly select the BOOT_IMAGE_ACCESS_HOOKS for configurations (samples, applications, test) that have net core as second image.

nordicjm: A quick fix would be `IMPLY` as it allows people to disable it, probably better would be to tie this to the value of updateable image count and if PCD is enabled in the bootloader

PR #1098, opened 2023-03-08
[nrf noup] Bluetooth: Mesh: Fix adv randomness bug

Andrewpini:
Fixes issue where randomness can be removed for advertising sets that have to handle other adv types than the BT_MESH_FRIEND_ADV tag type.

### DIFF
--- a/subsys/bluetooth/mesh/adv_ext.c
+++ b/subsys/bluetooth/mesh/adv_ext.c
@@ -482,7 +482,9 @@ int bt_mesh_adv_enable(void)
 			return err;
 		}
 
-		if (IS_ENABLED(CONFIG_BT_LL_SOFTDEVICE) && adv->tag & BT_MESH_FRIEND_ADV) {
+		if (IS_ENABLED(CONFIG_BT_LL_SOFTDEVICE) &&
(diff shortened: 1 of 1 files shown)


PavelVPV (approved):

subsys/bluetooth/mesh/adv_ext.c
```diff
@@ -482,7 +482,7 @@ int bt_mesh_adv_enable(void)
 			return err;
 		}
 
-		if (IS_ENABLED(CONFIG_BT_LL_SOFTDEVICE) && adv->tag & BT_MESH_FRIEND_ADV) {
+		if (IS_ENABLED(CONFIG_BT_LL_SOFTDEVICE) && adv->tag == BT_MESH_FRIEND_ADV) {
```
alxelax, line 485: Will it be better to use `IS_ENABLED(CONFIG_BT_MESH_ADV_EXT_FRIEND_SEPARATE)` instead to let compiler explicitly know it is possible to cut out?

Andrewpini: Sure 👍

alxelax (approved):

PR #1099, opened 2023-03-09
modules: canopennode: Updated CANopenNode driver to v2.0

alibenpeng:
This commit updates the existing zephyr driver for canopennode from v1.3 to v2.0 and adds support for all new upstream driver features. The motivation for this was that I needed the ASCII gateway feature, which is not available in v1.3.
I ran checkpatch on my own code, but the stuff in the objdict-directory of the example is autogenerated and won't pass. But I hardly even touched that, so it should be fine.


jfischer-no (changes requested):
What is the reason to do it for NCS?

alibenpeng:
> What is the reason to do it for NCS?

Actually I wasn't sure where to submit this. Should this go directly into upstream Zephyr?

alibenpeng:
> > > What is the reason to do it for NCS?
> > 
> > 
> > Actually I wasn't sure where to submit this. Should this go directly into upstream Zephyr?
> 
> Yes, please.

Alright, I'll close this then.

PR #1101, opened 2023-03-10
[nrf fromtree] net: dns: Check existing DNS servers before reconfigure

stig-bjorlykke:
In dns_resolve_reconfigure() check if the DNS servers already exist before cancel all ongoing queries. This will solve an issue with getaddrinfo() returning DNS_EAI_CANCELED when receiving a retransmitted DHCP offer and when receiving a IPv6 Router Advertisement.

### DIFF
--- a/subsys/net/lib/dns/resolve.c
+++ b/subsys/net/lib/dns/resolve.c
@@ -1386,6 +1386,59 @@ int dns_resolve_close(struct dns_resolve_context *ctx)
 	return ret;
 }
 
+static bool dns_server_exists(struct dns_resolve_context *ctx,
+			      const struct sockaddr *addr)
+{
+	for (int i = 0; i < SERVER_COUNT; i++) {
+		if (IS_ENABLED(CONFIG_NET_IPV4) && (addr->sa_family == AF_INET) &&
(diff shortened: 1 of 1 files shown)


subsys/net/lib/dns/resolve.c
```diff
@@ -1386,6 +1386,63 @@ int dns_resolve_close(struct dns_resolve_context *ctx)
 	return ret;
 }
 
+static bool dns_server_exists(struct dns_resolve_context *ctx,
+			      const struct sockaddr *addr)
+{
+	for (int i = 0; i < SERVER_COUNT; i++) {
+#if defined(CONFIG_NET_IPV4)
+		if ((addr->sa_family == AF_INET) &&
```
rlubos, line 1394: I'd rather use C-style check, i. e. `if (IS_ENABLED(CONFIG_NET_IPV4) && (addr->sa_family == AF_INET) &&`.

rlubos (commented):
LGTM, just one suggestion.

subsys/net/lib/dns/resolve.c
stig-bjorlykke: Fixed.

jfischer-no:
Please rebase. Btw, next upmerge will start very soon.

PR #1105, opened 2023-03-24
nrf53: Halt app CPU before flashing netcore

swojo:
Ongoing IPC communication between the app and net cores can cause flashing to fail on the nrf53 . This is because the network core is flashed first and the app core is not halted during this time. Attempted IPC communication can then lead to assertions/resets during the flashing process.  

This PR proposes to halt the app core before halting and flashing the network core for the nrf53.

### DIFF
--- a/scripts/west_commands/runners/nrfjprog.py
+++ b/scripts/west_commands/runners/nrfjprog.py
@@ -343,6 +343,8 @@ class NrfJprogBinaryRunner(ZephyrBinaryRunner):
             net_hex.write_hex_file(net_hex_file)
             app_hex.write_hex_file(app_hex_file)
 
+            # halt app CPU before flashing network core
+            program_commands.append(['nrfjprog', '--halt', '--coprocessor', 'CP_APPLICATION'])
             add_program_cmd(net_hex_file, 'CP_NETWORK', [])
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
Hello and thanks for the patch. If we are going to take this patch, we would want it merged into the upstream zephyr (https://github.com/zephyrproject-rtos/zephyr) repository first, so it can be brought into the nRF Connect SDK's sdk-zephyr as part of our regular upstream synchronization procedure.

I am asking around internally to see if this is the best way to solve this problem and will get back to you as soon as I have a response. Once we know the answer, I'd like to ask that you send the patch upstream, where it can be merged.

Thanks again!

swojo:
> Hello and thanks for the patch. If we are going to take this patch, we would want it merged into the upstream zephyr (https://github.com/zephyrproject-rtos/zephyr) repository first, so it can be brought into the nRF Connect SDK's sdk-zephyr as part of our regular upstream synchronization procedure.
> 
> I am asking around internally to see if this is the best way to solve this problem and will get back to you as soon as I have a response. Once we know the answer, I'd like to ask that you send the patch upstream, where it can be merged.
> 
> Thanks again!

Hi,
Any update on whether you agree with this solution?

PR #1116, opened 2023-04-05
settings: allows multiple settings storage backends simultaneously

faisalSaleemSetec:


faisalSaleemSetec:
> How does this correspond to upstream PR [zephyrproject-rtos/zephyr#53140](https://github.com/zephyrproject-rtos/zephyr/pull/53140)?

oops, my bad.. i was suppose to push it locally..

PR #1131, opened 2023-04-18
[nrf noup] Bluetooth: Mesh: Update default adv sets for mesh

PavelVPV:
Change default mesh configuration when running with SoftDevice Controller to get better performance.

Why it is changed this way:
- This configuration is only tested on SDC and may not give the same performance on ZLL, thus can't be set in upstream;
- `default <N> if BT_LL_SOFTDEVICE` can't be used as it won't work for nRF53 board where the controller is not present in the app core.

### DIFF
--- a/subsys/bluetooth/Kconfig.adv
+++ b/subsys/bluetooth/Kconfig.adv
@@ -30,6 +30,7 @@ config BT_EXT_ADV_LEGACY_SUPPORT
 config BT_EXT_ADV_MAX_ADV_SET
 	int "Maximum number of simultaneous advertising sets"
 	range 1 64
+	default 7 if BT_MESH
 	default 1
 	help
 	  Maximum number of simultaneous Bluetooth advertising sets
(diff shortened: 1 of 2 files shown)


Andrewpini (approved):

jfischer-no (changes requested):
I do not think that performance tuning should be part of the [nrf noup] patches.

Balaklaka (approved):

PavelVPV:
> I do not think that performance tuning should be part of the [nrf noup] patches.

The explanation is given in the commit message. We want this to be a `default` configuration for mesh in NCS. Another option is to create an overlay .conf file for all mesh samples in sdk-nrf, but this won't work for any non-sample project. So the only option I see currently is to do it this way. Please provide better suggestion if you have one.

PavelVPV:
Putting `DNM` until fix nightly CI.

PavelVPV:
Closing as found a way how to achieve this in sdk-nrf repo.

PR #1140, opened 2023-04-29
net: wifi_mgmt: Support to provide raw scan data

krish2718:
Support to provide raw beacon and probe response data to registered application.

### DIFF
--- a/include/zephyr/net/wifi_mgmt.h
+++ b/include/zephyr/net/wifi_mgmt.h
@@ -113,6 +113,7 @@ enum net_event_wifi_cmd {
 	NET_EVENT_WIFI_CMD_IFACE_STATUS,
 	NET_EVENT_WIFI_CMD_TWT,
 	NET_EVENT_WIFI_CMD_TWT_SLEEP_STATE,
+	NET_EVENT_WIFI_CMD_RAW_SCAN_RESULT,
 };
 
 #define NET_EVENT_WIFI_SCAN_RESULT				\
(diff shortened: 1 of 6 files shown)


include/zephyr/net/wifi_mgmt.h
```diff
@@ -274,11 +278,23 @@ enum wifi_twt_sleep_state {
 	WIFI_TWT_STATE_AWAKE = 1,
 };
 
+#ifdef CONFIG_WIFI_FEAT_RAW_SCAN_RESULTS
+struct wifi_raw_scan_result {
+	int8_t rssi;
+	int frame_length;
+	unsigned short frequency;
+	uint8_t data[0];
```
krish2718, line 286: This can also be fixed: `CONFIG_WIFI_RAW_SCAN_RESULT_LENGTH`

include/zephyr/net/wifi_mgmt.h
```diff
@@ -328,6 +344,11 @@ void wifi_mgmt_raise_iface_status_event(struct net_if *iface,
 void wifi_mgmt_raise_twt_event(struct net_if *iface,
 		struct wifi_twt_params *twt_params);
 void wifi_mgmt_raise_twt_sleep_state(struct net_if *iface, int twt_sleep_state);
+#ifdef CONFIG_WIFI_FEAT_RAW_SCAN_RESULTS
+#define MAX_RAW_DATA_PER_EVENT_LIMIT 42
```
krish2718, line 348: this can be removed

include/zephyr/net/wifi_mgmt.h
ajayparida: This is as per suggestion given by @udaynordic. 
Currently in the event, the memory will be allocated as per the frame length and passed.
instead of passing CONFIG_WIFI_RAW_SCAN_RESULT_LENGTH size data.

ajayparida:
@sachinthegreen @krish2718 @rado17 @udaynordic

rado17 (approved):

PR #1149, opened 2023-05-04
[nrf toup] Fix app dependency on offsets.h generated header

joerchan:
Fix issue of rare race-condititon for source file included in 'app' target for the generated header file offsets.h

zephyr/include/zephyr/linker/linker-defs.h:26:10:
| fatal error: offsets.h: No such file or directory
|     26 | #include <offsets.h>
|        |          ^~~~~~~~~~~
|  compilation terminated.

Fix this by adding the generated headers offsets.h and version.h to the zephyr_interface target instead and remove the zephyr_generated_interface target that app has no dependency on.


CMakeLists.txt
```diff
@@ -1344,7 +1342,6 @@ configure_linker_script(
   ${APP_SMEM_ALIGNED_DEP}
   ${KOBJECT_LINKER_DEP}
   ${CODE_RELOCATION_DEP}
-  zephyr_generated_headers
```
MarekPieta, line 1347: So that the former `zephyr_generated_headers` would be taken into account here in another way ?

MarekPieta (approved):
I do not feel like a Zephyr build system expert, but to me it seems fine

CMakeLists.txt
joerchan: Thanks. No it should be zephyr_interface here I think.

joerchan: Actually, it already is, see line below.

So the change is that zephyr_interface now also includes offsets.h and version.h.
So zephyr_interface now includes everything that was previously in zephyr_generated_headers.

tejlmand (changes requested):
> Fix this by adding the generated headers offsets.h and version.h to the zephyr_interface target instead and remove the zephyr_generated_interface target that app has no dependency on.

this is actually not fixing anything, afaict this just move dependency from one target-level dependency (zephyr_generated_headers) to another target-level dependency (zephyr_interface). Doing so won't solve anything as I believe what is missing in this case is a file-level dependency.

Secondly in this case, the linker doesn't depends on everything from the zephyr_interface, so adding such broad dependency should be done with care.

For more info, see also the CMake docs, where such an example is given:
https://cmake.org/cmake/help/latest/command/add_custom_command.html#example-generating-files-for-multiple-targets

joerchan:
Closing this, a workaround is applied upstream and we can wait for the upstream PR to resolve itself and come down in the regular upmerge.

PR #1152, opened 2023-05-05
[nrf noup] cmake: west: Use unix-style path for west on windows

nordicjm:
Fixes an issue whereby the back-slashes in the WEST_PYTHON variable were wrongly treated as escape sequence codes.

### DIFF
--- a/cmake/modules/west.cmake
+++ b/cmake/modules/west.cmake
@@ -106,3 +106,9 @@ else()
     set(WEST WEST-NOTFOUND CACHE INTERNAL "West")
   endif()
 endif()
+
+if(DEFINED WEST_PYTHON AND WIN32)
+  # Replace windows back-slashes in path with forward-slashes to prevent issues
+  # with the back-slashes being wrongly interpreted as escape sequence codes.
+  file(TO_CMAKE_PATH "${WEST_PYTHON}" WEST_PYTHON)
+endif()


sarah-read:
This works for me on Windows, thanks for the fix.

cmake/modules/west.cmake
```diff
@@ -106,3 +106,9 @@ else()
     set(WEST WEST-NOTFOUND CACHE INTERNAL "West")
   endif()
 endif()
+
+if(DEFINED WEST_PYTHON AND WIN32)
+  # Replace windows back-slashes in path with forward-slashes to prevent issues
+  # with the back-slashes being wrongly interpreted as escape sequence codes.
+  file(TO_CMAKE_PATH "${WEST_PYTHON}" WEST_PYTHON)
```
tejlmand, line 113: `WEST_PYTHON` path is already adjusted here:
https://github.com/nrfconnect/sdk-zephyr/blob/c2859f736bfc959b91cca472ba32563afccea0bc/cmake/modules/west.cmake#L15-L18

so we should move this adjustment upwards and let CMake take care of handling the OS, that is if on Linux and path already contains `/`, then CMake will keep those.

And while doing so, feel free to adjust `get_filename_component(... REAL_PATH)` to the never `file(REAL_PATH ...)`.
That can be done because Zephyr now requires CMake >=3.20.

tejlmand (changes requested):
Why is this a `[nrf noup]` ?

nordicjm:
Superseded by https://github.com/zephyrproject-rtos/zephyr/pull/57886

PR #1158, opened 2023-05-08
[nrf fromtree] net: update tfm dependency for psa

pjasinski990:
TFM is not the only provider for PSA API, which means the PSA config shouldn't depend only on it.

Signed-off-by: Piotr Jasiński <piotr.jasinski@nordicsemi.no>
(cherry picked from commit e010dcee5d7c582dbfd1e197f53c777ac640f362)


rlubos:
@pjasinski990 Please open a manifest update in sdk-nrf.

rlubos (approved):

PR #1171, opened 2023-05-11
Bluetooth: Controller: Upstream Zephyr Bluetooth Controller fixes for NCS 2.4.0

cvinayak:


cvinayak:
not a priority to have latest open source controller changes in NCS 2.4.0

PR #1176, opened 2023-05-15
mgmt: mcumgr: Supress false error log output

pirun:
img_mgmt_state_read try to open mcuboot_primary_1 (ram_flash partition) without checking whether it exists or not in application.
We don't enable CONFIG_FLASH_SIMULATOR in application in order to save ram usage.

This commit introduces error log output.
nrfconnect/sdk-zephyr@df0fa6d

Generally it would failed when [opening slot2 partition](https://github.com/nrfconnect/sdk-zephyr/blob/main/subsys/mgmt/mcumgr/grp/img_mgmt/src/zephyr_img_mgmt.c#L712) which makes CI failed when test DFU.
This error message could be founded on thingy53 and nrf5340 when https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/working_with_nrf/nrf53/nrf5340.html#simultaneous-multi-image-dfu feature is used.
2023_05_15_mcumgr_img_grp_failed_to_open_slot2

We have two options.
1. Skipped opening slot 2 image when CONFIG_FLASH_SIMULATOR is not enabled.
2. Change `img_mgmt_erased_val` error to warning message.

### DIFF
--- a/subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
+++ b/subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
@@ -255,6 +255,11 @@ img_mgmt_state_read(struct smp_streamer *ctxt)
 	     zcbor_list_start_encode(zse, 2 * CONFIG_MCUMGR_GRP_IMG_UPDATABLE_IMAGE_NUMBER);
 
 	for (i = 0; ok && i < 2 * CONFIG_MCUMGR_GRP_IMG_UPDATABLE_IMAGE_NUMBER; i++) {
+		/* Slot 2 mcuboot_primary_1 only exists when ram_flash is enabled*/
+		if (i == 2 && !IS_ENABLED(CONFIG_FLASH_SIMULATOR)) {
+			continue;
+		}
+
 		int rc = img_mgmt_read_info(i, &ver, hash, &flags);
 		if (rc != 0) {
 			continue;
--- a/subsys/mgmt/mcumgr/grp/img_mgmt/src/zephyr_img_mgmt.c
+++ b/subsys/mgmt/mcumgr/grp/img_mgmt/src/zephyr_img_mgmt.c
@@ -709,7 +709,7 @@ img_mgmt_erased_val(int slot, uint8_t *erased_val)
 
 	rc = flash_area_open(area_id, &fa);
 	if (rc != 0) {
-		LOG_ERR("Failed to open flash area ID %u: %d", area_id, rc);
+		LOG_WRN("Failed to open flash area ID %u: %d", area_id, rc);
 		return MGMT_ERR_EUNKNOWN;
 	}
 


subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
```diff
@@ -255,6 +255,11 @@ img_mgmt_state_read(struct smp_streamer *ctxt)
 	     zcbor_list_start_encode(zse, 2 * CONFIG_MCUMGR_GRP_IMG_UPDATABLE_IMAGE_NUMBER);
 
 	for (i = 0; ok && i < 2 * CONFIG_MCUMGR_GRP_IMG_UPDATABLE_IMAGE_NUMBER; i++) {
+		/* Slot 2 mcuboot_primary_1 only exists when ram_flash is enabled*/
+		if (i == 2 && !IS_ENABLED(CONFIG_FLASH_SIMULATOR)) {
+			continue;
```
nordicjm, line 260: This is not acceptable, slot 2 does not mean it is an nrf5340, it could be used on any number of devices and is being considered for the nrf9160 with MCUmgr client support

nordicjm (changes requested):

subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
koffes: I think this PR can go upstream. The easiest fix is to do the ERR->WRN, but I am not familiar enough in this territory to suggest something better off the top of my head.

nordicjm: It is an error if someone tries to open a flash area that doesn't exist, so replacing it with a warning doesn't make sense

pirun: Maybe img_mgmt_state_read should not read an unsexisted partition.
Is there any way to prevent this?

nordicjm: Updatable image numbers should not be set to 2 for the application if it can't use both updatable images, that'd be a bug in the application configuration.

pirun: That's what Nordic designed for nRF5340.
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/working_with_nrf/nrf53/nrf5340.html#simultaneous-multi-image-dfu

> Note
The application core does not have direct access to the network core flash memory. The update image is passed indirectly using RAM. Because of this, the mcuboot_primary_1 must be stored in ram_flash region. To enable providing such region on the device, see [CONFIG_FLASH_SIMULATOR](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/kconfig/index.html#CONFIG_FLASH_SIMULATOR).

nordicjm: It might be used by nrf5340 for network core update but that does not mean that is the only use for it, I am the upstream mcumgr maintainer for zephyr.

de-nordic (changes requested):
> Generally it would failed when [opening slot2 partition](https://github.com/nrfconnect/sdk-zephyr/blob/main/subsys/mgmt/mcumgr/grp/img_mgmt/src/zephyr_img_mgmt.c#L712) which makes CI failed when test DFU.

Then fix CI. 
Subsystem may fail with errors and app may recover anyway. That is what happens here.
Generally changing  behavior for all errors that may appear here, only with use of some Kconfig specific to single driver instance that may be cause of an error, is not acceptable.
If you want to suppress the message you need to do that to operations on specific slot only.

koffes (commented):
I am asked to review this PR, but is it in a reviewable state now @pirun ?

nordicjm:
> I am asked to review this PR, but is it in a reviewable state now @pirun ?

This PR is not acceptable, my NACK will remain

pirun:
This PR should be closed since it's a workaround not a fix.
There is a ticket https://nordicsemi.atlassian.net/browse/NCSDK-21790 to fix the issue.

PR #1178, opened 2023-05-15
Add ACK for disconnect

krish2718:
Pulled a dependent commit from #1140  will rebase once its merged.

### DIFF
--- a/include/zephyr/net/wifi_mgmt.h
+++ b/include/zephyr/net/wifi_mgmt.h
@@ -114,6 +114,7 @@ enum net_event_wifi_cmd {
 	NET_EVENT_WIFI_CMD_TWT,
 	NET_EVENT_WIFI_CMD_TWT_SLEEP_STATE,
 	NET_EVENT_WIFI_CMD_RAW_SCAN_RESULT,
+	NET_EVENT_WIFI_CMD_DISCONNECT_COMPLETE,
 };
 
 #define NET_EVENT_WIFI_SCAN_RESULT				\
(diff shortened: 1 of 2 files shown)


sachinthegreen (approved):

include/zephyr/net/wifi_mgmt.h
```diff
@@ -139,6 +140,9 @@ enum net_event_wifi_cmd {
 
 #define NET_EVENT_WIFI_RAW_SCAN_RESULT                          \
 	(_NET_WIFI_EVENT | NET_EVENT_WIFI_CMD_RAW_SCAN_RESULT)
+
+#define NET_EVENT_WIFI_DISCONNECT_COMPLETE			\
```
markaj-nordic, line 144: Where is this one used? Is it that with this change we will still receive the NET_EVENT_WIFI_DISCONNECT_RESULT in the application with the difference being now we block the disconnect command until it really succeeds?

krish2718: No, this PR only adds the event, this is sent https://github.com/nrfconnect/sdk-hostap/pull/94/files#diff-ff43beace1c9eb2129211e73d03f9385752af39391aa954185b48b9891e1c9b4R305

markaj-nordic: Turned out the `NET_EVENT_WIFI_CMD_DISCONNECT_COMPLETE` was really not used :) Thanks for the change

rlubos (approved):

subsys/net/l2/wifi/wifi_mgmt.c
```diff
@@ -381,3 +381,15 @@ void wifi_mgmt_raise_raw_scan_result_event(struct net_if *iface,
...
 #endif /* CONFIG_WIFI_MGMT_RAW_SCAN_RESULTS */
+
+void wifi_mgmt_raise_disconnect_complete_event(struct net_if *iface,
+					       int status)
+{
+	struct wifi_status cnx_status = {
+		.status = status,
+	};
+
+	net_mgmt_event_notify_with_info(NET_EVENT_WIFI_CMD_DISCONNECT_COMPLETE,
```
kkasperczyk-no, line 392: Sorry for post merge review, but I verified it doesn't work. I can be fixed by replacing `NET_EVENT_WIFI_CMD_DISCONNECT_COMPLETE` with the `NET_EVENT_WIFI_DISCONNECT_COMPLETE`.

PR #1180, opened 2023-05-16
[nrf noup] modules: mbedtls: Update PSA naming and fix dependencies

joerchan:
This PR addresses a mix of issues.

Using the same naming as the official PSA documentation for PSA algorithms.

Fixes missing missing dependency auto-enabling of PSA key types.

Fix missing dependency for cipher module with PSA stream cipher.

Various other small fixes, see commit messages.

### DIFF
--- a/modules/mbedtls/Kconfig
+++ b/modules/mbedtls/Kconfig
@@ -5,6 +5,7 @@
 
 config ZEPHYR_MBEDTLS_MODULE
 	bool
+
 config MBEDTLS_PROMPTLESS
 	bool
 	help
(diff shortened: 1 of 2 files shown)


vili-nordic (approved):

SebastianBoe (changes requested):
Why noup?

Seems like something we would either want to upstream or live with.

joerchan:
> Why noup?
> 
> Seems like something we would either want to upstream or live with.

IT's a fixup on another noup commit.

SebastianBoe (approved):

vili-nordic (approved):

SebastianBoe (approved):

PR #1199, opened 2023-06-05
ieee802154 nrf5 callout timeouts

ankuns:
Workaround for nRF5340 network core crashes. 

Added Kconfig IEEE802154_NRF5_CALLOUT_TIMEOUT_MS.

Waiting for callbacks ending tx and cca operations are limited by Kconfig-urable `CONFIG_IIEEE802154_NRF5_CALLOUT_TIMEOUT_MS`
Waiting for end of delayed transmission (nrf5_tx_at) is limited to `time_ahead + CONFIG_IEEE802154_NRF5_CALLBACK_TIMEOUT_MS` 
Waiting for end of energy detection is limited to `duration + CONFIG_IEEE802154_NRF5_CALLBACK_TIMEOUT_MS` 

When frames cease to be received, the radio driver is polled with period `IEEE802154_NRF5_RX_THREAD_HEARTBEAT_PERIOD_MS` (60s by default for serialized driver) to check if network core crashed.

Also:
Cherry-picked commit: https://github.com/zephyrproject-rtos/zephyr/commit/174b7f47757845e531c33583dde372671cc5b27c which ensures that `nrf_802154_serialization_error` never returns.

Related ticket: KRKNWK-17057

### DIFF
--- a/drivers/ieee802154/Kconfig.nrf5
+++ b/drivers/ieee802154/Kconfig.nrf5
@@ -78,6 +78,17 @@ config IEEE802154_NRF5_DELAY_TRX_ACC
 	  Accuracy of the clock used for scheduling radio delayed operations (delayed transmission
 	  or delayed reception), in ppm.
 
+config IEEE802154_NRF5_CALLOUT_TIMEOUT_MS
+	int "Timeout in milliseconds of waiting for a callout that finishes a requested operation"
+	default 10000
+	help
+	  This parameter limits the maximum time that can pass between
+	  a request to the nRF 802.15.4 Radio Driver to the corresponding asynchronous callout
+	  notifying the result of the operation. Reaching this timeout is a fatal error
+	  of the nRF 802.15.4 Radio Driver thus rather big values are advised.
+	  Set this value to 0 to have no limit (equivalent to K_FOREVER) which turns off the
+	  timing-out feature.
+
 config IEEE802154_NRF5_LOG_RX_FAILURES
 	bool "Frame reception failures logging"
 	help
--- a/drivers/ieee802154/Kconfig.nrf5
+++ b/drivers/ieee802154/Kconfig.nrf5
@@ -89,6 +89,17 @@ config IEEE802154_NRF5_CALLOUT_TIMEOUT_MS
 	  Set this value to 0 to have no limit (equivalent to K_FOREVER) which turns off the
 	  timing-out feature.
 
+config IEEE802154_NRF5_RX_THREAD_HEARTBEAT_PERIOD_MS
+	int "Period in milliseconds of polling nRF 802.15.4 Radio Driver to detect potential malfunctions"
+	default 60000 if NRF_802154_SER_HOST
+	default 0
+	help
+	  The nRF 802.15.4 Radio Driver is polled with period set by this parameter in order to
+	  detect any potential malfunctions of the driver. This option is useful when
+	  the radio driver implementation is provided by separate core through serialization.
+	  To keep energy usage low the value should be set to big values.
+	  Value equal to 0 turns off the polling completely.
+
 config IEEE802154_NRF5_LOG_RX_FAILURES
 	bool "Frame reception failures logging"
 	help
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -122,94 +122,102 @@ static void nrf5_get_eui64(uint8_t *mac)
 	memcpy(mac + index, &factoryAddress, sizeof(factoryAddress) - index);
 }
 
+static void nrf5_rx_frame_process(struct nrf5_802154_data *nrf5_radio,
+				struct nrf5_802154_rx_frame *rx_frame)
+{
+	struct net_pkt *pkt = NULL;
+	uint8_t pkt_len;
+	uint8_t *psdu;
+
+	__ASSERT_NO_MSG(rx_frame->psdu);
+
+	/* rx_mpdu contains length, psdu, fcs|lqi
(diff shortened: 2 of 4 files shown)


drivers/ieee802154/Kconfig.nrf5
```diff
@@ -78,6 +78,28 @@ config IEEE802154_NRF5_DELAY_TRX_ACC
 	  Accuracy of the clock used for scheduling radio delayed operations (delayed transmission
 	  or delayed reception), in ppm.
 
+config IEEE802154_NRF5_CALLBACK_TIMEOUT_MS
```
jciupis, line 81: I'd opt for `CALLOUT` instead of `CALLBACK`.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -620,7 +688,7 @@ static int nrf5_tx(const struct device *dev,
 		nrf_802154_channel_get(), nrf_802154_tx_power_get());
 
 	/* Wait for the callback from the radio driver. */
-	k_sem_take(&nrf5_radio->tx_wait, K_FOREVER);
+	nrf5_callback_sem_take(&nrf5_radio->tx_wait);
```
jciupis, line 691: How would this code behave if a transmission with delay larger than `IEEE802154_NRF5_CALLBACK_TIMEOUT_MS` was scheduled? AFAIU the current implementation would throw a fatal error, which is not the intended behavior.

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -286,6 +346,14 @@ static int nrf5_energy_scan_start(const struct device *dev,
 		if (nrf_802154_energy_detection(duration * 1000) == false) {
 			nrf5_data.energy_scan_done = NULL;
 			err = -EPERM;
+		} else {
```
jciupis, line 349: There are several places in this PR where this pattern:
```c
    } else {
#if (CONFIG_IEEE802154_NRF5_CALLBACK_TIMEOUT_MS != 0)
        // foo
#endif
    }
```
is used. It's tempting to replace it with:
```c
    } else if (CONFIG_IEEE802154_NRF5_CALLBACK_TIMEOUT_MS) {
        // foo
    }
```
It looks cleaner and might make maintenance easier because compilation errors will not slip through even if the feature is disabled by default. Please check if such a change is feasible (might not be if the code refers to variables that must be `#ifdef'ed`).

drivers/ieee802154/ieee802154_nrf5.c
ankuns: You are right, however `IEEE802154_NRF5_CALLBACK_TIMEOUT_MS` is expected to be large. OT requests delayed transmits milliseconds before the transmisios because otherwise whole transmitter is considered busy.

Taking into account `tx_at` value from `nrf5_tx_at` (simillar to ed duration) would be a solution to the problem, however the defensive nature of the timeout would be somehow weakened.
A user can always set `IEEE802154_NRF5_CALLBACK_TIMEOUT_MS` to even higher value.

Do you still think we should take `tx_at` into account (or please propose other in-code-solution)?
Otherwise we could add some warning in the Kconfig help or leave it as it is now.

jciupis: > the defensive nature of the timeout would be somehow weakened.

I think then it would be adequate, not weakened. When an immediate operation such as standalone CCA or transmission are requested, their execution is expected to start immediately. The user specifies with Kconfig that they're willing to wait _x_ milliseconds for such operation to finish. In my opinion the same wait time of _x_ milliseconds should be applied to an operation that starts in an arbitrary moment in time other than "now".

I agree that currently OpenThread only uses small delays when requesting delayed transmission and the default timeout is expected to work correctly in that case. Still I think there's no need to put a limitation on the user if there's no need to do that. In my opinion it would be best to take `tx_at` into account.

drivers/ieee802154/Kconfig.nrf5
ankuns: Fixed

drivers/ieee802154/ieee802154_nrf5.c
ankuns: I think I fixed this, please have a look.

drivers/ieee802154/ieee802154_nrf5.h
```diff
@@ -76,6 +76,11 @@ struct nrf5_802154_data {
 	 */
 	energy_scan_done_cb_t energy_scan_done;
 
+#if (CONFIG_IEEE802154_NRF5_CALLBACK_TIMEOUT_MS != 0)
```
jciupis, line 79: One remaining occurrence to be renamed

drivers/ieee802154/ieee802154_nrf5.c
jciupis: Thank you.

drivers/ieee802154/ieee802154_nrf5.c
ankuns: I think I fixed as you pointed, please have a look

Code related to `ed_timer` must remain, the `ed_timer` is in the struct. If I left `ed_timer` unconditionally, for sake of having all paths compiled, there would be a waste of ram in case ed_timer is not used.

jciupis: That makes sense, thanks.

jciupis (approved):

rlubos (approved):

ahasztag:
What about handling delayed operations?

ahasztag (approved):

doublemis1 (approved):

PR #1200, opened 2023-06-06
[nrf fromlist] net: wifi_mgmt: Reject TWT setup till IP address is configured

ajayparida:
If a user tries to enable TWT too early in the connection, then we might enter TWT sleep even before DHCP is completed, this can result in packet loss as when we wakeup we cannot receive traffic and completing DHCP itself can take multiple intervals. Though static ip address can be assigned too. Reject TWT till Wi-Fi interface has
a valid IP address.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/58728


sachinthegreen (approved):

ajayparida:
Closed as it's part of https://github.com/nrfconnect/sdk-zephyr/pull/1220

PR #1216, opened 2023-06-22
net: wifi: Fix power save timeout data type

krish2718:
This should be an unsigned integer. Also, add a comment to explain this feature.

Upstream-Pr: https://github.com/zephyrproject-rtos/zephyr/pull/59638

### DIFF
--- a/subsys/net/l2/wifi/wifi_shell.c
+++ b/subsys/net/l2/wifi/wifi_shell.c
@@ -593,8 +593,13 @@ static int cmd_wifi_ps(const struct shell *sh, size_t argc, char *argv[])
 		shell_fprintf(sh, SHELL_NORMAL, "PS wake up mode: %s\n",
 				config.ps_params.wakeup_mode ? "Listen interval" : "DTIM");
 
-		shell_fprintf(sh, SHELL_NORMAL, "PS timeout: %d ms\n",
-				config.ps_params.timeout_ms);
+		if (config.ps_params.timeout_ms) {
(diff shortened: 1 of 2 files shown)


VivekUppunda (approved):

rado17 (approved):

subsys/net/l2/wifi/wifi_shell.c
```diff
@@ -586,8 +586,13 @@ static int cmd_wifi_ps(const struct shell *sh, size_t argc, char *argv[])
 		shell_fprintf(sh, SHELL_NORMAL, "PS wake up mode: %s\n",
 				config.ps_params.wakeup_mode ? "Listen interval" : "DTIM");
 
-		shell_fprintf(sh, SHELL_NORMAL, "PS timeout: %d ms\n",
-				config.ps_params.timeout_ms);
+		if (!config.ps_params.timeout_ms) {
+			shell_fprintf(sh, SHELL_NORMAL, "PS timeout: %d ms\n",
+					config.ps_params.timeout_ms);
+		} else {
+			shell_fprintf(sh, SHELL_NORMAL, "PS timeout: disabled\n");
```
ajayparida, line 593: Please check, if "timeout_ms" is valid also it will print as disabled.

ajayparida (changes requested):

subsys/net/l2/wifi/wifi_shell.c
krish2718: Thanks, I have fixed it in upstream, waiting for it to merge and then cherry-pick.

ajayparida (approved):

PR #1222, opened 2023-06-28
[DNM]net: NET filter

MarGasiorek:
Simple implementation of the network filter based on the hooks.


MarGasiorek:
Incorrect PR, it will be replaced by new one.

PR #1223, opened 2023-06-29
Find My: fixes for minor release

kapi-no:
If disconnection has been triggered in between the security update and the call to `smp_pairing_complete` we need to abort the pairing.

The disconnection may have been triggered by `bt_unpair`, in that case the keys will have been erased and it will lead to an assertion to continue as if nothing happened.

To resolve this issue, at the beginning of `smp_pairing_complete` the `status` is set to `BT_SMP_ERR_UNSPECIFIED` if there is no connection.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/59807

### DIFF
--- a/subsys/bluetooth/host/smp.c
+++ b/subsys/bluetooth/host/smp.c
@@ -1632,7 +1632,15 @@ static void smp_pairing_complete(struct bt_smp *smp, uint8_t status)
 {
 	struct bt_conn *conn = smp->chan.chan.conn;
 
-	LOG_DBG("status 0x%x", status);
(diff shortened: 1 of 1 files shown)


theob-pro (approved):

mkapala-nordic (approved):
As it was merged in upstream already, will it be changed to fromtree?

kapi-no:
Changed the [nrf fromlist] to [nrf fromtree]. No diff has changed

rlubos (approved):

PR #1225, opened 2023-06-30
net: NET filter

MarGasiorek:
The simple implementation of the net filer based on hooks.


subsys/net/ip/connection.c
```diff
@@ -576,6 +579,13 @@ enum net_verdict net_conn_input(struct net_pkt *pkt,
 	uint8_t pkt_family = net_pkt_family(pkt);
 	uint16_t src_port = 0U, dst_port = 0U;
 
+	#if defined(CONFIG_NET_FILTER)
```
rlubos, line 582: Preprocessor directives shouldn't be indented.

rlubos: Or, alternatively to ifdef-slicing the code, you could provide a dummy (empty) implementations for those functions in the header file. It's a common practice we use in net subsystem to keep the code a bit cleaner.

include/zephyr/net/net_filter.h
```diff
@@ -0,0 +1,126 @@
...
+ * @brief Network filter hook.
+ *
+ * @param pf protocol family.
+ * @param hooknum hook ID number.
+ * @param pkt network packet.
+ * @return verdict about the packet.
+ */
+enum net_verdict nf_hook(uint8_t pf, unsigned int hooknum, struct net_pkt *pkt);
+
+static inline enum net_verdict nf_prerouting_hook(uint8_t pf, struct net_pkt *pkt)
```
rlubos, line 104: All public functions should be documented.

include/zephyr/net/net_filter.h
```diff
@@ -0,0 +1,126 @@
...
+ * @defgroup net_filter Network Filter API
+ * @ingroup networking
+ * @{
+ */
+
+/** @brief hook callback type. */
+typedef enum net_verdict (nf_hook_fn_t)(struct net_pkt *pkt);
+
+/** @brief Hook IDs for IP. */
+typedef enum {
```
rlubos, line 36: We don't typedef enums in Zephyr according to Coding Guidelines.

include/zephyr/net/net_filter.h
```diff
@@ -0,0 +1,126 @@
...
+	NF_IP_PRE_ROUTING = 0,
+	NF_IP_LOCAL_IN,
+	NF_IP_FORWARD,
+	NF_IP_LOCAL_OUT,
+	NF_IP_POST_ROUTING,
+	NF_IP_NUMHOOKS
+} nf_ip_hooks;
```
rlubos, line 43: Maybe `ip_hook_type` would be a bit more descriptive?

include/zephyr/net/net_filter.h
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * Several callback functions can be registered with the same hook.
+ * The priority value determines the order of callbacks function and
+ * their calling order. The callback function for which priority has
+ * been set to lower value (e.g. -100) will be called before callback
+ * function with higher priority value (e.g. 0).
+ */
+struct nf_hook_cfg {
+	nf_hook_fn_t *hook_fn;  /**< callback function */
+	unsigned int hooknum;   /**< hook ID number */
```
rlubos, line 55: There's a dedicated enum for the value defined just above. Also, I'd prefer `hook type` instead of `hook ID number`.

include/zephyr/net/net_filter.h
```diff
@@ -0,0 +1,126 @@
...
+ */
+
+/** @brief hook callback type. */
+typedef enum net_verdict (nf_hook_fn_t)(struct net_pkt *pkt);
+
+/** @brief Hook IDs for IP. */
+typedef enum {
+	NF_IP_PRE_ROUTING = 0,
+	NF_IP_LOCAL_IN,
+	NF_IP_FORWARD,
```
rlubos, line 39: There's already IPv6 routing module, so I assume this hook could be used there?

subsys/net/pkt_filter/CMakeLists.txt
```diff
@@ -7,3 +7,9 @@ zephyr_library_sources(base.c)
 zephyr_library_sources_ifdef(CONFIG_NET_L2_ETHERNET ethernet.c)
 
 endif()
+
+if(CONFIG_NET_FILTER)
+zephyr_library()
```
rlubos, line 12: Please double check - I'm not actually sure if two libraries can be defined in a single file (especially if you don't provide the library name).

subsys/net/pkt_filter/Kconfig
```diff
@@ -19,4 +19,19 @@ module-help = Enables packet filter output debug messages
 source "subsys/net/Kconfig.template.log_config.net"
 endif # NET_PKT_FILTER
 
+config NET_FILTER
```
rlubos, line 22: I'm actually a bit uncertain, whether NET_FILTER isn't too broad of a name for the module. Maybe NET_IP_FILTER? I'm open for discussion though.

subsys/net/pkt_filter/net_filter.c
```diff
@@ -0,0 +1,217 @@
...
+	}
+	k_spin_unlock(&entry->lock, key);
+}
+
+static void nf_lock_and_register_hook(struct nf_hook_entries *entry,
+				      struct nf_hook_entry *hook)
+{
+	k_spinlock_key_t key = k_spin_lock(&entry->lock);
+
+	if (sys_slist_is_empty(&entry->hook_root)) {
```
rlubos, line 87: Is this extra case really needed? I believe we could `sys_slist_prepend()` even on an empty list.

subsys/net/pkt_filter/net_filter.c
```diff
@@ -0,0 +1,217 @@
...
+	k_spin_unlock(&entry->lock, key);
+}
+
+enum net_verdict nf_hook(uint8_t pf, unsigned int hooknum, struct net_pkt *pkt)
+{
+	__ASSERT_NO_MSG(pkt != NULL);
+	struct nf_hook_entries *entry_head = nf_hook_entry_head(pf, hooknum);
+	enum net_verdict result = NET_CONTINUE;
+
+	if (entry_head && !sys_slist_is_empty(&entry_head->hook_root)) {
```
rlubos, line 118: IMO it'd be cleaner just to return early in this case, just like below.

subsys/net/pkt_filter/net_filter.c
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	struct nf_hook_entries *entry_head = nf_hook_entry_head(cfg->pf, cfg->hooknum);
+
+	if (!entry_head) {
+		NET_WARN("No entry");
+		return -ENOENT;
+	}
+
+	struct nf_hook_entry *new_hook = k_malloc(sizeof(struct nf_hook_entry));
```
rlubos, line 153: Maybe instead of getting heap involved, we could just make a requirement that `struct nf_hook_cfg` should be placed in the static memory? Or if really dynamic allocation is needed, it could be done by the app itself.

rlubos (commented):
First, brief look over the module.

rlubos:
Some documentation would be nice (like the one from confluence).

include/zephyr/net/net_filter.h
MarGasiorek: OK, I change it to
```suggestion
enum nf_ip_hooks{
```

include/zephyr/net/net_filter.h
MarGasiorek: This is a hook type categories for IP hooks, I'm not sure if `ip_hook_type` it's a good name for this enum.
If it is a case I can change it to list of #define-s

include/zephyr/net/net_filter.h
MarGasiorek: OK, I changed description to `hook type category` however I want to avoid assign specific type for this value. When this module evaluate in future then can be different hooks categories for e.g. [brige](https://en.wikipedia.org/wiki/Netfilter#/media/File:Netfilter-packet-flow.svg): `brouting`

include/zephyr/net/net_filter.h
MarGasiorek: Done.

subsys/net/ip/connection.c
MarGasiorek: Done

subsys/net/pkt_filter/net_filter.c
MarGasiorek: Right, this additional case is not needed.

subsys/net/pkt_filter/net_filter.c
MarGasiorek: done

subsys/net/pkt_filter/net_filter.c
MarGasiorek: Yes, I agree. I used linux approach, but now I see that it not make a sense.

include/zephyr/net/net_filter.h
MarGasiorek: OK, we can discuss offline where I can add this hook (or maybe others too).

subsys/net/pkt_filter/CMakeLists.txt
MarGasiorek: I move it to another directory with separate cmake file and Kconfig.

subsys/net/pkt_filter/Kconfig
MarGasiorek: In linux it's jut netfilter, NET_IP_FILTER sounds good when we look to current implementation (now I'm focusing on the IP filtering) but it can be used also for other purposes.
What do you think about NET_HOOKS_FILTER?

rlubos: `NET_HOOKS_FILTER` sounds a bit awkward to me, I'd prefer to stick to `NET_FILTER` if we don't want to use IP. Although it may seem a bit ambiguous, given there's `NET_PKT_FILTER` already.

MarGasiorek:
The commit is replaced by https://github.com/nrfconnect/sdk-zephyr/pull/1226

PR #1226, opened 2023-07-04
[nrf fromtree] net: pkt_filter: Introduce additional hooks for pkt_filter

MarGasiorek:
The additional hooks provide infrastructure to construct rules on another network stack levels. Main benefit of this approach is packets are pre-parsed and e.g. IP filter is easier to implement. These hooks are equivalent of prerouting and local_in in linux's netfilter.


rlubos (commented):
Wouldn't we need IP send rules as well?

rlubos:
Also, I'm missing the concept of the filter priority from the original concept, will it be added?

MarGasiorek:
I answer in order:
1. Yes, and do you think we need one more hook instead of this https://github.com/nrfconnect/sdk-zephyr/blob/743f99947369088ecba6820a720ede8054893866/subsys/net/ip/net_core.c#L436. If answer is yes, then please help me choose correct place 
2. No, in this approach (small steps) I believe user will prepare all rules in one place and then use functions for append or prepend. In future we can consider add priority, what is your opinion?

One more thing, I want to add additional argument to test function https://github.com/nrfconnect/sdk-zephyr/blob/743f99947369088ecba6820a720ede8054893866/include/zephyr/net/net_pkt_filter.h#L38C1-L38C74. This argument will be a pointer to void:
t`ypedef bool (npf_test_fn_t)(struct npf_test *test, struct net_pkt *pkt, **void* data**);` and allow to pass additional data from hooks, e.g. ip header. For backward compatibility NULL is also acceptable and data validity must be done directly in test function. Do you think it will be acceptable for zephyr community?

rlubos:
> 1. Yes, and do you think we need one more hook instead of this https://github.com/nrfconnect/sdk-zephyr/blob/743f99947369088ecba6820a720ede8054893866/subsys/net/ip/net_core.c#L436
>    
>    If answer is yes, then please help me choose correct place to 'throw in hook'.

I think you need to elaborate, I don't think I follow. I was asking about send hooks and here we have recv one.

> No, in this approach (small steps) I believe user will prepare all rules in one place and then use functions for append or prepend. In future we can consider add priority, what is your opinion?

If it's not needed (for now) I see no point adding it at the moment.

> This argument will be a pointer to void:
typedef bool (npf_test_fn_t)(struct npf_test *test, struct net_pkt *pkt, void* data);
and allow to pass additional data from hooks, e.g. ip header. For backward compatibility NULL is also acceptable and data validity must be done directly in test function. Do you think it will be acceptable for zephyr community?

I don't see a problem with that, but it should be well documented on what kind of data to expect at what level. Although, is that really needed - at least in terms of IP header? Couldn't the header be read directly from the net_pkt?

subsys/net/ip/ipv4.c
```diff
@@ -330,6 +330,11 @@ enum net_verdict net_ipv4_input(struct net_pkt *pkt)
 
 	net_pkt_set_family(pkt, PF_INET);
 
+	if (!net_pkt_filter_ipv4_recv_ok(pkt)) {
```
rlubos, line 333: Is the location of this hook correct? For IPv6 we call the hook before verifying the destination IP address, here we call it after.

MarGasiorek:
> > 1. Yes, and do you think we need one more hook instead of this https://github.com/nrfconnect/sdk-zephyr/blob/743f99947369088ecba6820a720ede8054893866/subsys/net/ip/net_core.c#L436
> >    
> >    If answer is yes, then please help me choose correct place to 'throw in hook'.
> 
> I think you need to elaborate, I don't think I follow. I was asking about send hooks and here we have recv one.
> 
Sorry, my mistake, I thought about this hook: https://github.com/nrfconnect/sdk-zephyr/blob/743f99947369088ecba6820a720ede8054893866/subsys/net/ip/net_if.c#L341
> > No, in this approach (small steps) I believe user will prepare all rules in one place and then use functions for append or prepend. In future we can consider add priority, what is your opinion?
> 
> If it's not needed (for now) I see no point adding it at the moment.
> 
> > This argument will be a pointer to void:
> > typedef bool (npf_test_fn_t)(struct npf_test *test, struct net_pkt _pkt, void_ data);
> > and allow to pass additional data from hooks, e.g. ip header. For backward compatibility NULL is also acceptable and data validity must be done directly in test function. Do you think it will be acceptable for zephyr community?
> 
> I don't see a problem with that, but it should be well documented on what kind of data to expect at what level. Although, is that really needed - at least in terms of IP header? Couldn't the header be read directly from the net_pkt?
Now I'm not sure what is enough, but I want to know if additional argument can be a issue or not to get correct development direction

subsys/net/ip/ipv4.c
MarGasiorek: Moved to more appropriate place

subsys/net/pkt_filter/Kconfig
```diff
@@ -12,6 +12,28 @@ config NET_PKT_FILTER
...
+config NET_PKT_FILTER_IPV6_HOOK
+	bool "Additional network packet filtering hook inside IPv6 stack"
+	depends on NET_IPV6
+	help
+	  This additional hook provides infrastructure to construct custom
+	  rules on the IP packet.
+
+config NET_PKT_FILTER_LOCAL_IN_HOOK
+	bool "Additional network packet filtering hook for connection input"
+	depends on CONFIG_NET_IP
```
rlubos, line 32:
```suggestion
	depends on NET_IP
```

subsys/net/pkt_filter/base.c
```diff
@@ -25,6 +25,50 @@ struct npf_rule_list npf_recv_rules = {
...
+#endif /* CONFIG_NET_PKT_FILTER_IPV6_HOOK */
+
+/*
+ * Helper function
+ */
+static struct npf_rule_list *get_ip_rules(uint8_t pf)
+{
+	switch (pf) {
+	case PF_INET:
+	#ifdef CONFIG_NET_PKT_FILTER_IPV4_HOOK
```
rlubos, line 56: #ifdefs shouldn't be indented

tests/net/npf/src/main.c
```diff
@@ -72,6 +75,34 @@ static struct net_pkt *build_test_pkt(int type, int size, struct net_if *iface)
...
+
+	size = (family == AF_INET) ? sizeof(struct net_ipv4_hdr) :
+		(family == AF_INET6) ? sizeof(struct net_ipv6_hdr) : 0U;
+
+	pkt = net_pkt_rx_alloc_with_buffer(iface, size, family, 0, K_NO_WAIT);
+	zassert_not_null(pkt, "");
+
+	if (family == AF_INET) {
+		ret = net_ipv4_create(pkt, (struct in_addr *)src,
+						(struct in_addr *)dst);
```
rlubos, line 93: Some weird indentation here - generally we try to align with the opening bracket when we split the argument list.

tests/net/npf/src/main.c
```diff
@@ -307,4 +342,156 @@ ZTEST(net_pkt_filter_test_suite, test_npf_address_mask)
...
+	zassert_true(net_pkt_filter_ip_recv_ok(pkt_v6), "");
+
+	zassert_true(npf_remove_all_ipv6_recv_rules(), "");
+	net_pkt_unref(pkt_v6);
+	net_pkt_unref(pkt_v4);
+}
+
+ZTEST(net_pkt_filter_test_suite, test_npf_ip_filter)
+{
+	test_npf_ipv4_address_filtering();
```
rlubos, line 493: IMO we should just split IPv4/IPv6 testing into two separate test cases.

rlubos (commented):
Looks reasonable. For upstream PR, I'd recommend to move tests to a separate commit.

subsys/net/pkt_filter/Kconfig
MarGasiorek: Nice catch, thanks

tests/net/npf/src/main.c
MarGasiorek: Now should be better.

subsys/net/pkt_filter/base.c
MarGasiorek: Fixed

tests/net/npf/src/main.c
MarGasiorek: I fully agree, however file author did it another way:
https://github.com/nrfconnect/sdk-zephyr/blob/01e2e4c9ce8294d74032a2d33e0dcb1b74802568/tests/net/npf/src/main.c#L304-L308
Should I change my TC and this above as well?

rlubos: I don't think there's a need to modify existing tests, just update the new ones. Unless you really want to.

MarGasiorek: I think that common approach allow us keep clear code.

MarGasiorek: OK, I see that I cannot change this TC
https://github.com/nrfconnect/sdk-zephyr/blob/01e2e4c9ce8294d74032a2d33e0dcb1b74802568/tests/net/npf/src/main.c#L287
So I update only my part.

MarGasiorek:
> Looks reasonable. For upstream PR, I'd recommend to move tests to a separate commit.

Separate commit in separate PRs, or separate commits in the same PR?

rlubos:
> > Looks reasonable. For upstream PR, I'd recommend to move tests to a separate commit.
> 
> Separate commit in separate PRs, or separate commits in the same PR?

Same PR, 2 commits.

MarGasiorek:
> > > Looks reasonable. For upstream PR, I'd recommend to move tests to a separate commit.
> > 
> > 
> > Separate commit in separate PRs, or separate commits in the same PR?
> 
> Same PR, 2 commits.

Done.

rlubos (approved):

konradderda (commented):
LGTM, thanks! :+1:

include/zephyr/net/net_pkt_filter.h
```diff
@@ -67,6 +67,12 @@ struct npf_rule_list {
 extern struct npf_rule_list npf_send_rules;
 /** @brief rule list applied to incoming packets */
 extern struct npf_rule_list npf_recv_rules;
+/** @brief rule list applied for local incoming packets */
+extern struct npf_rule_list npf_local_in_recv_rules;
```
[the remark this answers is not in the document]
MarGasiorek: I used name convection similar to linux, we can add in future similar function i different place e.g. for bridge: `npf_local_brige_recv_rules()`.
But I agree, currently 'in' part not make a sense, anyway it is already in zephyr.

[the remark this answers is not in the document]
konradderda: I agree as we discussed it f2f some time ago. I tried to remove this draft comment from the review but GitHub did not cooperate/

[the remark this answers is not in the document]
MarGasiorek: no worries, it will be a comment for 'future generations'.

PR #1228, opened 2023-07-05
[nrf fromlist] soc: arm: nrf53: workaround pop lr after wfi crash

ankuns:
On nRF5340 net core it was observed that when `wfi` instruction was followed by `pop {r0, lr}` in the `arch_cpu_idle` function, the value of `lr` sometimes got read as 0 from memory despite having correct value stored in the memory.

This commit inserts additional `nop` instruction after waking up to delay access to the memory.


carlescufi (changes requested):
Please use a hook or introduce one if not present. We cannot have this downstream in the core arch code, especially now that it is being refactored in Zephyr.

carlescufi:
> Please use a hook or introduce one if not present. We cannot have this downstream in the core arch code, especially now that it is being refactored in Zephyr.

@carlocaione perhaps could suggest how to have a hook like this.

LuDuda (approved):

carlocaione:
Without using a real function hook that would impact too much on the idle thread I would suggest to inject a macro carrying the few assembly instructions, like we do for example on RISC-V here: https://github.com/zephyrproject-rtos/zephyr/blob/8560fb319249e3fad3122f8541df4c4beef6e5e2/arch/riscv/core/isr.S#L146-L148 so that the NOPs would be introduced by the preprocessor at compile time.

Somewher soc-specific:
```c
#define WAIT_ON_WFI \
    nop \
    nop \
    nop \ 
    nop
```

In `cpu_idle.S` (upstream-able with a good justification):
```
...
#if defined(CONFIG_WAIT_ON_WFI)
    WAIT_ON_WFI
#endif
...
```

ankuns:
@carlescufi I agree with you, I am preparing upstreamable patch, but it takes time, can we go with this for 2.4.1 ?

ankuns:
Content of this PR will be likely superseded by cherry-pick of commit in https://github.com/nrfconnect/sdk-zephyr/pull/1231 which is hopefully an upstreamable solution. It gives exactly the same assembly code.

carlocaione (approved):

rlubos (approved):

rlubos:
For some reason twister status on GH was not updated, but the latest run passed:
https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-zephyr/job/PR-1228/3/

AryanNordic:
ooh, this patch is beautiful. Just came across this from somewhere else.

PR #1231, opened 2023-07-06
soc: arm: nrf53: workaround pop lr after wfi crash

ankuns:
On nRF5340 net core it was observed that when `wfi` instruction was followed by `pop {r0, lr}` in the `arch_cpu_idle` function, the value of `lr` sometimes got read as 0 from memory despite having correct value stored in the memory.

This commit inserts additional `nop` instruction after waking up to delay access to the memory.

The cpu_ilde.S generic code can inline SoC-specific macro when `CONFIG_ARM_ON_EXIT_CPU_IDLE` is enabled.
The `CONFIG_ARM_ON_EXIT_CPU_IDLE` is enabled by default for nRF5340 net core only.

For nRF5340 the SoC specific code expands to 4 nops as proposed in PR for NCS 2.4.1 https://github.com/nrfconnect/sdk-zephyr/pull/1228 commit a0413b0262334c2d6a0162fe49bd1d688f611924


ankuns:
- [ ] Target it to upstream
- [ ] update commit sauce tag with [nrf fromtree] or [nrf fromlist]

arch/arm/Kconfig
```diff
@@ -53,6 +53,12 @@ config ARM_ON_ENTER_CPU_IDLE_HOOK
 	  If needed, this hook can be used to prevent the CPU from actually
 	  entering sleep by skipping the WFE/WFI instruction.
 
+config ARM_ON_EXIT_CPU_IDLE_HOOK
+	bool
+	help
+	  Enables a pssibility to inject SoC-specific code just after WFI/WFE
```
rlubos, line 59:
```suggestion
	  Enables a possibility to inject SoC-specific code just after WFI/WFE
```

ankuns:
@carlescufi @carlocaione @rlubos , hopefully this is "upstreamable" version of https://github.com/nrfconnect/sdk-zephyr/pull/1228 (commit a0413b0262334c2d6a0162fe49bd1d688f611924) could you have a look and decide if it is good enough to go upstream? If yes I'll open PR to upstream, update this one with [nrf fromlist] + link to the PR, and replace https://github.com/nrfconnect/sdk-zephyr/pull/1228 with cherry-pick of this.

arch/arm/Kconfig
```diff
@@ -53,6 +53,12 @@ config ARM_ON_ENTER_CPU_IDLE_HOOK
 	  If needed, this hook can be used to prevent the CPU from actually
 	  entering sleep by skipping the WFE/WFI instruction.
 
+config ARM_ON_EXIT_CPU_IDLE_HOOK
```
carlocaione, line 56: The problem is that this is mimicking `CONFIG_ARM_ON_ENTER_CPU_IDLE_HOOK` that is indeed a real hook, this is not, this is an injected macro basically.

So, we have two options here:
1. We keep mirroring what is done for `CONFIG_ARM_ON_ENTER_CPU_IDLE_HOOK` that is we add a real `z_arm_on_exit_cpu_idle` function that is called with all the the needed pre/post-amble of `push / pop` and we keep this name.
2. We use the current solution but we change symbol name and we justify it why this is not an hook but an instructions injection by macro (to me, that makes much more sense because you are saving at least a couple of instructions but on the downside you must write your routine in assembly, while with the hook you can write it in C)

arch/arm/Kconfig
```diff
@@ -53,6 +53,12 @@ config ARM_ON_ENTER_CPU_IDLE_HOOK
...
+	bool
+	help
+	  Enables a pssibility to inject SoC-specific code just after WFI/WFE
+	  instructions of cpu idle implementation.
```
carlocaione, line 60: If we are going with the macro injection the description must be way more informative, see https://github.com/zephyrproject-rtos/zephyr/blob/630555d9f6ead7530b3d1e5a5fba96104607550c/arch/riscv/Kconfig#L63-L79 for reference

arch/arm/core/aarch32/cpu_idle.S
```diff
@@ -64,6 +68,11 @@ SECTION_FUNC(TEXT, z_arm_cpu_idle_init)
 	dsb
 	\wait_instruction
 
+#if defined(CONFIG_ARM_ON_EXIT_CPU_IDLE_HOOK)
+	/* Inline the macro provided by SoC-specific code */
+	arm_on_exit_cpu_idle_hook
```
carlocaione, line 73: If we are going with the macro this should be all caps. If we are using an hook small caps are fine.

carlocaione (changes requested):

arch/arm/Kconfig
ankuns: I will pick 2nd option. Calling hook written in C is not an option because it might push sth on the stack, memory access fust after wfi is this thigh we need to avoid.

carlocaione: yes, this is exactly the kind of justification that is needed upstream

arch/arm/Kconfig
ankuns: fixed

arch/arm/Kconfig
ankuns: fixed, picked 2nd option

arch/arm/Kconfig
ankuns: added detailed description

arch/arm/core/aarch32/cpu_idle.S
ankuns: renamed to `SOC_ON_EXIT_CPU_IDLE` (no hook word, all caps)

ankuns:
@carlocaione I think I addressed what you pointed, 2nd round of review please.

soc/arm/nordic_nrf/nrf53/soc_cpu_idle.h
```diff
@@ -0,0 +1,21 @@
...
+
+
+#if defined(_ASMLANGUAGE)
+
+.macro SOC_ON_EXIT_CPU_IDLE
+	nop
+	nop
+	nop
+	nop
+.endm
```
carlocaione, line 19: This should be a `define` not a `.macro`.

arch/arm/Kconfig
```diff
@@ -53,6 +53,21 @@ config ARM_ON_ENTER_CPU_IDLE_HOOK
...
+	  instructions of the cpu idle implementation.
+
+	  Enabling this option requires that the SoC provides a soc_cpu_idle.h
+	  header file which defines SOC_ON_EXIT_CPU_IDLE macro guarded by
+	  _ASMLANGUAGE. The SOC_ON_EXIT_CPU_IDLE macro is expanded just after
+	  WFI/WFE instructions before any memory access is performed. The purpose
+	  of the SOC_ON_EXIT_CPU_IDLE is to perform an action that mitigate issues
+	  observed on some SoCs caused by a memory access following WFI/WFE
+	  instructions. For this reason the macro is used instead of a hook
+	  function.
```
carlocaione, line 69:
```suggestion
	  Enabling this option requires that the SoC provides a soc_cpu_idle.h
	  header file which defines SOC_ON_EXIT_CPU_IDLE macro guarded by
	  _ASMLANGUAGE.

	  The SOC_ON_EXIT_CPU_IDLE macro is expanded just after
	  WFI/WFE instructions before any memory access is performed. The purpose
	  of the SOC_ON_EXIT_CPU_IDLE is to perform an action that mitigate issues
	  observed on some SoCs caused by a memory access following WFI/WFE
	  instructions.
```

carlocaione (changes requested):

soc/arm/nordic_nrf/nrf53/soc_cpu_idle.h
carlocaione: right, because you need:
```c
#define SOC_ON_EXIT_CPU_IDLE \
    nop; \
    nop; \
    nop; \
    nop;
```

ankuns: fixed

carlocaione (approved):

rlubos (approved):

PR #1234, opened 2023-07-07
[nrf noup] include: net: add socket option level IPPROTO_ALL

MarkusLassila:
Add socket option level IPPROTO_ALL to socket_ncs.h.

PR that depends on this: https://github.com/nrfconnect/sdk-nrf/pull/11715

### DIFF
--- a/include/zephyr/net/socket_ncs.h
+++ b/include/zephyr/net/socket_ncs.h
@@ -25,7 +25,7 @@ extern "C" {
 #define SOCK_NATIVE 0x80000000
 #define SOCK_NATIVE_TLS 0x40000000
 
-/* NCS specific TLS options */
+/* NCS specific TLS level socket options */
 
 /** Socket option to set DTLS handshake timeout, specifically for nRF sockets.
  *  The option accepts an integer, indicating the total handshake timeout,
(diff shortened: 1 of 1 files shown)


include/zephyr/net/socket_ncs.h
```diff
@@ -121,6 +112,26 @@ extern "C" {
...
+
+/** IPv4 and IPv6 protocol level (pseudo-val) for nRF sockets. */
+#define IPPROTO_ALL 512
+/** sockopt: disable all replies to unexpected traffics */
+#define SO_SILENCE_ALL 30
+
+/* NCS specific IPPROTO_IP level socket options */
+
+/** sockopt: disable IPv4 ICMP replies */
+#define SO_IP_ECHO_REPLY 31
```
lemrey, line 126: Non-zero enables ICMP echo replies on IPv4. Perhaps this should be "enable IPv4 ICMP replies"? (same goes for the IPv6 socket option)

MarkusLassila: Changed: disable -> enable

MarkusLassila:
> 

Thank you, I tried to adapt to the comment style, but missed this. Is there somewhere where the comment guideline would be documented?

rlubos (approved):

lemrey (approved):
Looks good. It's documented [here ](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/releases_and_maturity/developing/code_base.html#oss-repositories-downstream-project-history)

PR #1236, opened 2023-07-10
[nrf fromtree] libc/picolibc: Remove -T /dev/null linker arguments

nordicjm:
Now that the gcc compiler driver uses the -T flag instead of -Wl,-T, we can remove the hack here that kept the picolibc specs file from inserting the picolibc linker script.

Signed-off-by: Keith Packard <keithp@keithp.com>
(cherry picked from commit 4e7930b099771ecc5ad5c05b95a158cc0ceee0d2)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -469,8 +469,9 @@ if(CONFIG_USERSPACE)
 endif()
 
 get_property(TOPT GLOBAL PROPERTY TOPT)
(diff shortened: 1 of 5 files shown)


rlubos (approved):

rlubos:
CI passed: https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-zephyr/view/change-requests/job/PR-1236/1/

PR #1244, opened 2023-07-14
modules: mbedtls: Make PSA key types user configurations

joerchan:
Make the PSA key type configurations be user configurations instead
of implictly enabled based on other configurations.
While AES is the most likely to be used in this case the user can
choose different block ciphers for block cipher modes, and others
stream ciphers also exists.
This also allows applications to enable key generation without any
algorithms for ECC and RSA keys.
Also add the full key types configurations, even though not all may be
supported.

Add help text for for unsafe PSA algorithms. These should not be used
for security.

Use same text naming as PSA spec for RIPEMD-160 and SHA-1.

### DIFF
--- a/modules/mbedtls/Kconfig.psa
+++ b/modules/mbedtls/Kconfig.psa
@@ -179,7 +179,10 @@ config PSA_HAS_HASH_SUPPORT
 
 config PSA_WANT_ALG_SHA_1
 	bool
-	prompt "PSA SHA-1 support" if !PSA_PROMPTLESS
+	prompt "PSA SHA-1 support (weak)" if !PSA_PROMPTLESS
+	help
+	  Warning: The SHA-1 hash is weak and deprecated and is only recommended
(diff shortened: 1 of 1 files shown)


modules/mbedtls/Kconfig.psa
```diff
@@ -179,7 +179,10 @@ config PSA_HAS_HASH_SUPPORT
 
 config PSA_WANT_ALG_SHA_1
 	bool
-	prompt "PSA SHA1 support" if !PSA_PROMPTLESS
+	prompt "PSA SHA-1 support (unsafe)" if !PSA_PROMPTLESS
```
SebastianBoe, line 182:
```suggestion
	prompt "PSA SHA-1 support (weak)" if !PSA_PROMPTLESS
```

SebastianBoe (approved):

mswarowsky (approved):

PR #1252, opened 2023-07-24
Psa ps

mswarowsky:


subsys/trusted_storage/include/psa/error.h
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2019-2021, Arm Limited. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ */
```
mswarowsky, line 6: If somebody knows how we handle the copyright for copied files, I'm happy to get input.

subsys/trusted_storage/src/auth_crypt_nonce.c
```diff
@@ -0,0 +1,90 @@
...
+
+	/* On overflow */
+	if (g_nonce_low == 0) {
+		++g_nonce_high;
+	}
+
+	/* Copy low first */
+	memcpy(nonce, &g_nonce_low, MIN(nonce_len, sizeof(g_nonce_low)));
+
+	/* Then high is there's free space */
```
vili-nordic, line 54: broken english

subsys/trusted_storage/src/auth_crypt_backend.c
```diff
@@ -0,0 +1,225 @@
...
+
+cleanup_objects:
+	/* Remove all object if an error occurs */
+	trusted_storage_remove_object(uid, prefix, TRUSTED_STORAGE_FILENAME_SUFFIX_NONCE);
+	trusted_storage_remove_object(uid, prefix, TRUSTED_STORAGE_FILENAME_SUFFIX_DATA);
+
+cleanup:
+	memset(&additional_data, 0, sizeof(additional_data));
+	memset(nonce, 0, sizeof(nonce));
+	memset(aead_buf, 0, AEAD_MAX_BUF_SIZE);
```
vili-nordic, line 198: safe memset?

subsys/trusted_storage/src/auth_crypt_backend.h
```diff
@@ -0,0 +1,61 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef __TRUSTED_STORAGE_AUTH_CRYPT_BACKEND_H_
+#define __TRUSTED_STORAGE_AUTH_CRYPT_BACKEND_H_
+
+#include <psa/error.h>
+#include <psa/storage_common.h>
+
+/* Initialize the Auth/Crypt crypto backend */
```
vili-nordic, line 13: this is a typedef that does not initialize anything?

subsys/trusted_storage/src/auth_crypt_nonce.h
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) 2022 Nordic Semiconductor ASA
```
vili-nordic, line 2: wrong year?

subsys/trusted_storage/src/settings_helpers.c
```diff
@@ -0,0 +1,318 @@
...
+#include <stdint.h>
+#include <stddef.h>
+#include <string.h>
+#include <errno.h>
+
+#include "settings_helpers.h"
+
+/* Helper to fill filename with a suffix */
+#define TRUSTED_STORAGE_FILENAME_FILL(filename, prefix, uid, suffix)                               \
+	snprintf((filename), TRUSTED_STORAGE_FILENAME_MAX_LENGTH + 1,                              \
```
vili-nordic, line 22: snprintf return value not checked (not obvious to me if it's safe to skip)

subsys/trusted_storage/src/auth_crypt_backend.c
```diff
@@ -0,0 +1,225 @@
...
+#define TRUSTED_STORAGE_FILENAME_SUFFIX_NONCE ".nonce"
+
+/*
+ * AEAD based Authenticated Encrypted trust implementation
+ *
+ * Actual implementation uses:
+ * - Hexadecimal UID imported as KEY
+ * - UID+Flags+Size as additional parameter
+ * - NONCE is an incrementing number at each encryption
+ * - TAG is left at the end of output data
```
vili-nordic, line 31: tag is the before output data other places?

subsys/trusted_storage/src/auth_crypt_backend.c
mswarowsky: Do we have a safe memset in zehphyr?

subsys/trusted_storage/src/auth_crypt_backend.c
mswarowsky: In the mbedTLS implementation, it looks like the tag comes before the encrypted data,
But do you think we can remove that here?

subsys/trusted_storage/src/auth_crypt_nonce.h
mswarowsky: Updated that on all files now

subsys/trusted_storage/src/auth_crypt_backend.c
```diff
@@ -0,0 +1,225 @@
...
+ */
+
+#define AEAD_TAG_SIZE   16
+#define AEAD_NONCE_SIZE 12
+
+/* Max storage size for the encrypted or decrypted output */
+#define AEAD_MAX_BUF_SIZE ROUND_UP(TRUSTED_STORAGE_MAX_ASSET_SIZE + AEAD_TAG_SIZE, AEAD_TAG_SIZE)
+
+/* 128bits AEAD key pattern: uid low, uid */
+#define AEAD_KEY_PATTERN "%08x%08x"
```
vili-nordic, line 41: why not use PRIx64 instead of making two 32-bit ints?

mswarowsky: switched to hashing to this got removed

subsys/trusted_storage/src/settings_helpers.c
mswarowsky: Added it at places where this is used

mswarowsky: Added the extra check

subsys/trusted_storage/src/aead/aead_ctr_nonce.c
```diff
@@ -0,0 +1,87 @@
...
+
+#include "aead_nonce.h"
+
+/*
+ * NONCE is an incrementing 128bits number
+ */
+
+#define NONCE_MAX_LENGTH 16
+
+static uint64_t g_nonce_low;
```
frkv, line 26: This can be a structure, so you can do one single call to generate random (using sizeof)

samples/subsys/trusted_storage/src/main.c
```diff
@@ -13,6 +13,42 @@
...
+
+#include "settings/settings_file.h"
+#include <zephyr/device.h>
+#include <zephyr/fs/fs.h>
+#include <zephyr/fs/littlefs.h>
+
+#define STORAGE_PARTITION	storage_partition
+#define STORAGE_PARTITION_ID	FIXED_PARTITION_ID(STORAGE_PARTITION)
+
+/* NFFS work area strcut */
```
vili-nordic, line 28:
```suggestion
/* NFFS work area struct */
```

mswarowsky: It stands for Newtron Flash File System and seems to be an FS that Zephyr had before so this is most likely just copy-pasted from somewhere.... So changing to LittelFS

subsys/trusted_storage/src/aead/aead_key_huk.c
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	if (!hw_unique_key_are_any_written()) {
+		result = hw_unique_key_write_random();
+		if (result != HW_UNIQUE_KEY_SUCCESS) {
+			return 0;
+		}
+	}
+
+	return 0;
```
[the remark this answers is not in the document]
mswarowsky: Due to the this was called in the system init the error would haven been ignored anyway. 

But I forgot to remove that function, as the HUK should be written by the application, as it requires a reboot on the 52840, which die application might wanna schedule.

subsys/trusted_storage/src/aead/trusted_backend_aead.c
```diff
@@ -0,0 +1,326 @@
...
+				       TRUSTED_STORAGE_AEAD_BACKEND_FILENAME_SUFFIX_SIZE);
+	if (status != PSA_SUCCESS) {
+		return status;
+	}
+
+	return storage_remove_object(uid, prefix,
+				     TRUSTED_STORAGE_AEAD_BACKEND_FILENAME_SUFFIX_FLAGS);
+}
+
+uint32_t trusted_get_support_backend(void)
```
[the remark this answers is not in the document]
mswarowsky: `trusted_get_support` is what it should be, apparently vs code*s refactoring feature fooled me here....

subsys/trusted_storage/src/aead/aead_ctr_nonce.c
mswarowsky: Good idea, changed it

subsys/trusted_storage/src/storage_backend_settings.c
```diff
@@ -0,0 +1,162 @@
...
+#define TRUSTED_STORAGE_SETTINGS_BACKEND_FILENAME_UID_LENGTH (sizeof(uint64_t) * 2)
+
+/* Storage pattern: prefix, uid low, uid high, suffix */
+#define TRUSTED_STORAGE_SETTINGS_BACKEND_FILENAME_PATTERN "%s%08x%08x%s"
+
+/* Max filename length aligned on Settings File backend max length */
+#define TRUSTED_STORAGE_SETTINGS_BACKEND_FILENAME_MAX_LENGTH 32
+
+/* Helper to fill filename with a suffix */
+#define TRUSTED_STORAGE_SETTINGS_BACKEND_FILENAME_FILL(filename, filename_size, prefix, uid,       \
```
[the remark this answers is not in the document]
mswarowsky: Yeah also thought of it, done it now :)

subsys/trusted_storage/src/storage_backend_settings.c
```diff
@@ -0,0 +1,162 @@
...
+	int ret;
+
+	if (object_size == 0 || object_data == NULL || prefix == NULL || suffix == NULL) {
+		return PSA_ERROR_INVALID_ARGUMENT;
+	}
+
+	ret = TRUSTED_STORAGE_SETTINGS_BACKEND_FILENAME_FILL(
+		path, TRUSTED_STORAGE_SETTINGS_BACKEND_FILENAME_MAX_LENGTH, prefix, uid, suffix);
+	if (ret < 0 || ret >= TRUSTED_STORAGE_SETTINGS_BACKEND_FILENAME_MAX_LENGTH) {
+		return error_to_psa_error(ret);
```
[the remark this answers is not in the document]
mswarowsky: Na that was not correct but now wit han own function for creating the filename, this is obsolete

subsys/trusted_storage/Kconfig
```diff
@@ -0,0 +1,195 @@
+# Trusted storage subsystem configuration options
```
joerchan, line 1: We've put both trusted storage and protected storage in this Kconfig, so I think it is better named as "secure storage", which is the term used to describe both of them elsewhere as well.

subsys/trusted_storage/include/psa/error.h
joerchan: I suggest we put the PSA headers under the `ext/` folder.
This one is allowed to have different licenses, and it more accurately represents how we got these headers.

subsys/trusted_storage/src/aead/aead_crypt_mbedtls_chachapoly.c
```diff
@@ -0,0 +1,135 @@
...
+	return 16 + data_size;
+}
+
+psa_status_t trusted_storage_aead_encrypt(const void *key_buf, size_t key_len,
+					  const void *nonce_buf, size_t nonce_len,
+					  const void *add_buf, size_t add_len,
+					  const void *input_buf, size_t input_len, void *output_buf,
+					  size_t output_size, size_t *output_len)
+{
+	mbedtls_chachapoly_context ctx;
```
joerchan, line 37: Why do we need an mbedtls backend for aead?

subsys/trusted_storage/Kconfig
```diff
@@ -0,0 +1,195 @@
...
+
+# Copyright (c) 2023 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+menuconfig TRUSTED_STORAGE
```
joerchan, line 6: I suggested we namespace this under PSA.

I can't see that we are adding a generic interface using PSA ITS as the backend, so I don't think there is any reason to avoid using PSA here.

subsys/trusted_storage/src/aead/aead_crypt_mbedtls_chachapoly.c
mswarowsky: The idea is to try to upstream that and I got told therefore we should have an mbedtls option, as there might be the risk that they don't what to use psa api. As ITS will be called as part of other PSA calls.
@Vge0rge correct me if I'm wrong, I remember disscussing that with you.

subsys/trusted_storage/Kconfig
mswarowsky: So having it called secure storage subsystem?
Not sure if I understood you correctly.

mswarowsky: Disscued offline

mswarowsky:
Closing this as we will take the changes as part of sdk-nrf here https://github.com/nrfconnect/sdk-nrf/pull/11968

subsys/trusted_storage/include/psa/error.h
carlescufi: Although it is not strictly forbidden to include 3rd-party headers and source code anywhere in the tree, I agree with @joerchan that having them in `ext/psa` or whatever other subfolder you choose promotes better structure and a clean division of ownership of the headers.

PR #1260, opened 2023-08-01
mgmt/mcumgr/grp/img: Cherry-pick MCUboot direct XIP support

MarekPieta:
Cherry-pick MCUboot direct XIP support.

### DIFF
--- a/subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
+++ b/subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
@@ -9,6 +9,7 @@
 #include <zephyr/toolchain.h>
 #include <string.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/dfu/mcuboot.h>
 
 #include <zcbor_common.h>
 #include <zcbor_decode.h>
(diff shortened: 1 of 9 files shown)


subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
```diff
@@ -518,6 +596,53 @@ int img_mgmt_set_next_boot_slot(int slot, bool confirm)
...
+			rc = IMG_MGMT_RET_RC_UNKNOWN;
+		}
+	}
+	flash_area_close(fa);
+
+#if defined(CONFIG_MCUMGR_GRP_IMG_STATUS_HOOKS)
+	if (slot == active_slot && confirm) {
+		uint16_t ret_group;
+		/* Confirm event is only sent for active slot */
+		(void)mgmt_callback_notify(MGMT_EVT_OP_IMG_MGMT_DFU_CONFIRMED, NULL, 0, &ret_rc,
```
MarekPieta, line 638: `ret_rc` seems not declared here. This must be aligned to prevent build failures

MarekPieta: Aligned locally

MarekPieta:
> Note that MCUmgr MCUboot info may significantly change.

What do you mean @de-nordic? Do you think we should desist from cherry-picking it at this point?
@philips77 please confirm if you want to test with the MCUmgr MCUboot info in the current version.

de-nordic:
> > Note that MCUmgr MCUboot info may significantly change.
> 
> What do you mean @de-nordic? Do you think we should desist from cherry-picking it at this point? @philips77 please confirm if you want to test with the MCUmgr MCUboot info in the current version.

@nordicjm  Do you agree with the proposed command specification as done in https://github.com/zephyrproject-rtos/zephyr/pull/57797/files?

nordicjm:
> > > Note that MCUmgr MCUboot info may significantly change.
> > 
> > 
> > What do you mean @de-nordic? Do you think we should desist from cherry-picking it at this point? @philips77 please confirm if you want to test with the MCUmgr MCUboot info in the current version.
> 
> @nordicjm Do you agree with the proposed command specification as done in https://github.com/zephyrproject-rtos/zephyr/pull/57797/files?

Yes, at present it has the code in the file to do it but in future I suggest we replace it and just use it to directly query the `blinfo` group of settings once someone stops moaning over https://github.com/zephyrproject-rtos/zephyr/pull/59025

mkapala-nordic (approved):

philips77:
> please confirm if you want to test with the MCUmgr MCUboot info in the current version.

It depends on what is inside the info and how is it will be released. If you plan to release these features separately, so d-xip goes in 2.5, but info later, than I need to test d-xip only and together with info later. Or if they can be enabled separately, so one does not depend on the other.

kapi-no:
This PR should be closed as the support for Direct-XIP in this PR  is outdated. The support will come as nrf fromtree commits from a different PR set.

PR #1263, opened 2023-08-01
[nrf noup] settings: nvs: use dedicated lookup cache hash function

Damian-Nordic:
Introduce `NVS_LOOKUP_CACHE_FOR_SETTINGS` Kconfig option that enables a dedicated hash function for the NVS lookup cache that takes advantage of the NVS ID allocation scheme used by the NVS settings backend. As such, this option should only be used if an application uses NVS via the settings layer.

The commit is a `noup` patch because the original change seems not generic enough to be accepted by the Zephyr community: https://github.com/zephyrproject-rtos/zephyr/pull/61619.

### DIFF
--- a/subsys/fs/nvs/Kconfig
+++ b/subsys/fs/nvs/Kconfig
@@ -27,6 +27,15 @@ config NVS_LOOKUP_CACHE_SIZE
 	  Number of entries in Non-volatile Storage lookup cache.
 	  It is recommended that it be a power of 2.
 
+config NVS_LOOKUP_CACHE_FOR_SETTINGS
+	bool "Non-volatile Storage lookup cache optimized for settings"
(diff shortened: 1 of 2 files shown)


SylwesterKonczyk (approved):

LuDuda (approved):

LuDuda:
@Damian-Nordic what about to remove DNM?

Damian-Nordic:
> @Damian-Nordic what about to remove DNM?

I'm OK with that but maybe we should upstream this first?

Damian-Nordic:
FYI, removed DNM but only the second commit is exlusive for this PR. The first one is already upstreamed and a part of the current upmerge.

mariuszpos (approved):

LuDuda (approved):
LGTM :+1:

PR #1265, opened 2023-08-04
kernel: cherry-pick custom boot banner and set to nRF connect SDK

nordicjm:
[nrf fromtree] kernel: Add support to override banner
    
    This is useful for Zephyr customers to put their custom banners.

    [nrf noup] kernel: Set boot banner to NCS
    
    Updates the boot banner to show that it is booting nRF Connect
    SDK instead of Zephyr OS.

This looks like the following:
```
*** Booting nRF Connect SDK v3.3.99-ncs1-2860-gb75c68264f49 ***
Hello World! nrf52840dk_nrf52840
```

Which shows the sdk-zephyr version, not the sdk-nrf version, however, with the manifest update on https://github.com/nrfconnect/sdk-nrf/pull/11956 then it shows correctly:
```
*** Booting nRF Connect SDK v2.4.0-853-g22d42f40a900 ***
Hello World! nrf52840dk_nrf52840
```
If for some reason, someone does not have git, will show:
```
*** Booting nRF Connect SDK 2.4.99 ***
Hello World! nrf52840dk_nrf52840
```


krish2718 (approved):

kernel/Kconfig
```diff
@@ -331,7 +331,7 @@ config BOOT_BANNER
 config BOOT_BANNER_STRING
 	string "Boot banner string"
 	depends on BOOT_BANNER
-	default "Booting Zephyr OS build"
+	default "Booting nRF Connect SDK"
```
tejlmand, line 334: does this need to be a `noup`, can't we place the default overwrite in sdf-nrf here: https://github.com/nrfconnect/sdk-nrf/blob/main/Kconfig.nrf ?

tejlmand (changes requested):
Do we really need a `[nrf noup]` patch for this ?

nordicjm (commented):
@tejlmand this just does not work at all, BUILD_VERSION is always set even though it's not a custom build version, the original way worked

tejlmand (approved):
> @tejlmand this just does not work at all, BUILD_VERSION is always set even though it's not a custom build version, the original way worked

yes, I see with the changes to `banner.c` in ba9359d605162a492cb2f2bd8533b6a51a32aa67 we need a `noup` patch.

But when I made this comment: https://github.com/nrfconnect/sdk-zephyr/pull/1265#discussion_r1284370248, there was only a Kconfig change and no changes to `banner.c`, ref old commit: cfaa7524c0f32538ac89ad0d0bdae272f2708b96.

Approved.

tejlmand (changes requested):
temp block for extra testing

tejlmand (approved):
approved, but technically the 2nd commit should be a `[nrf fromtree]` now because upstream PR has been merged.

https://github.com/zephyrproject-rtos/zephyr/pull/61635

PR #1274, opened 2023-08-11
[nrf noup] boards: thingy53_nrf5340: Fix available RAM

krish2718:
fixup! [nrf noup] boards: thingy53_nrf5340: Add common partition map

Because of the missing shared RAM partition, the partition manager wrongly calcualted the usable RAM for applications (non-shared) showing as 504KB instead of 440KB.

### DIFF
--- a/boards/arm/thingy53_nrf5340/pm_static_thingy53_nrf5340_cpuapp.yml
+++ b/boards/arm/thingy53_nrf5340/pm_static_thingy53_nrf5340_cpuapp.yml
@@ -53,3 +53,7 @@ pcd_sram:
   address: 0x20000000
   size: 0x2000
   region: sram_primary
+shared_sram:
+  address: 0x20070000
+  size: 0x10000
+  region: sram_primary


VivekUppunda (approved):

sigvartmh (approved):

sachinthegreen (approved):

krish2718:
This fix is incorrect, this shared RAM entry isn't needed in PM as that is automatically incuded if CPUNET is being used. The problem is CPUNET isn't used by Wi-Fi, so, when PM is used the Shared RAM is counted as available RAM.

Two queries:

1. If CPUNET isn't needed, then can we use the shared RAM for CPUAPP? I guess yes but needs DTS changes.
2. If PM is in use, how to convey to PM that CPUNET isn't used and either exclude/include Shared RAM (Depends on #1).

krish2718:
@hakonfam @sigvartmh Any comments on https://github.com/nrfconnect/sdk-zephyr/pull/1274#issuecomment-1683641715

krish2718:
Abandoning this, went with the fix https://github.com/nrfconnect/sdk-nrf/pull/12018/commits/233ae3f376b286f6c016fdff9267e7fc5fa900cd

PR #1275, opened 2023-08-11
[nrf noup]: ci: Only run lwm2m tests when needed

rugeGerritsen:
There is no need to run those for all PRs.
For example: When there are only changes to Bluetooth, we don't want this CI to run.

Adding these filters may speedup the time to merge for PRs to sdk-nrf

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -21,7 +21,7 @@
   - "subsys/testsuite/ztest/**/*"
 
 "CI-lwm2m-test":
-  - "**/*"
+# Not necessary to run tests on changes to this repo.
 
 "CI-boot-dfu-test":
   - "subsys/mgmt/mcumgr/**/*"


.github/test-spec.yml
```diff
@@ -21,7 +21,13 @@
   - "subsys/testsuite/ztest/**/*"
 
 "CI-lwm2m-test":
```
PFnord, line 23: As we have checked, in the lwm2m carrierlib team, it seems we might not need any test run on zephyr-sdk modifications. Would it be possible to leave this as blank? or remove the label entirely?

thst-nordic: yes just remove line 24, and it won't be run

rugeGerritsen:
Last push was a rebase

.github/test-spec.yml
rugeGerritsen: That worked!

guwa (approved):

PFnord (approved):

PR #1279, opened 2023-08-15
Fix parse syscalls

sprhawk:
I encountered two problems:

1. the `toolchain/common.h` is included during parse_syscalls.py ( during I enabled CONFIG_APPLICATION_DEFINED_SYSCALL)
2. I'm using ccls, and it will generate .ccls-cache under applicaton root, it will be parsed by `parse_syscalls.py`


de-nordic:
Rebase needed to the new main.

PR #1286, opened 2023-08-23
[nrf noup] ci: Only run crypto and tf-m tests when needed

magnev:
Extended the github CI filter for crypto and TF-M tests to reduce the CI load and possibly speed up PRs.

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -31,7 +31,13 @@
   - "samples/subsys/mgmt/mcumgr/smp_svr/**/*"
 
 "CI-tfm-test":
-  - "**/*"
+  - "boards/arm/nrf5340dk_nrf5340/**/*"
+  - "boards/arm/nrf9160dk_nrf9160/**/*"
+  - "drivers/entropy/*"
+  - "dts/arm/nordic/nrf5340*"
+  - "dts/arm/nordic/nrf9160*"
+  - "modules/trusted-firmware-m/**/*"
+  - "samples/tfm_integration/**/*"
 
 "CI-ble-test":
   - "**/*"
@@ -58,7 +64,16 @@
   - "**/*"
 
 "CI-crypto-test":
-  - "**/*"
+  - "boards/arm/nrf52840dk_nrf52840/**/*"
+  - "boards/arm/nrf5340dk_nrf5340/**/*"
+  - "boards/arm/nrf9160dk_nrf9160/**/*"
+  - "drivers/entropy/*"
+  - "drivers/serial/**/*"
+  - "dts/arm/nordic/nrf52840*"
+  - "dts/arm/nordic/nrf5340*"
+  - "dts/arm/nordic/nrf9160*"
+  - "include/drivers/serial/**/*"
+  - "modules/mbedtls/**/*"
 
 "CI-fem-test":
   - "**/*"


.github/test-spec.yml
```diff
@@ -58,7 +68,22 @@
...
+  - "drivers/timer/**/*"
+  - "dts/arm/nordic/nrf52840*"
+  - "dts/arm/nordic/nrf5340*"
+  - "dts/arm/nordic/nrf9160*"
+  - "include/drivers/net/**/*"
+  - "include/drivers/serial/**/*"
+  - "include/net/net*"
+  - "include/net/socket*"
+  - "subsys/net/**/*"
+  - "subsys/random/**/*"
```
SebastianBoe, line 86:
```suggestion
  - "subsys/random/**/*"
  - "modules/mbedtls/**/*"
```

SebastianBoe (approved):

.github/test-spec.yml
```diff
@@ -31,7 +31,17 @@
   - "samples/subsys/mgmt/mcumgr/smp_svr/**/*"
 
 "CI-tfm-test":
-  - "**/*"
+  - "boards/arm/nrf5340dk_nrf5340/**/*"
+  - "boards/arm/nrf9160dk_nrf9160/**/*"
+  - "drivers/entropy/**/*"
+  - "drivers/serial/**/*"
+  - "drivers/timer/**/*"
```
joerchan, line 38: I don't understand why serial and timer is included.
If the logic is that TF-M has a serial and timer drivers, then it is TF-M specific drivers and not the zephyr ones, so that should not be a trigger.

.github/test-spec.yml
```diff
@@ -31,7 +31,17 @@
   - "samples/subsys/mgmt/mcumgr/smp_svr/**/*"
 
 "CI-tfm-test":
-  - "**/*"
+  - "boards/arm/nrf5340dk_nrf5340/**/*"
+  - "boards/arm/nrf9160dk_nrf9160/**/*"
+  - "drivers/entropy/**/*"
```
joerchan, line 36: Because of PSA entropy driver?

.github/test-spec.yml
```diff
@@ -58,7 +68,22 @@
   - "**/*"
 
 "CI-crypto-test":
-  - "**/*"
+  - "boards/arm/nrf52840dk_nrf52840/**/*"
+  - "boards/arm/nrf5340dk_nrf5340/**/*"
+  - "boards/arm/nrf9160dk_nrf9160/**/*"
+  - "drivers/entropy/**/*"
+  - "drivers/net/**/*"
```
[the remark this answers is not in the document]
magnev: I can agree with that, so I will remove net

.github/test-spec.yml
```diff
@@ -31,7 +31,17 @@
...
+  - "boards/arm/nrf9160dk_nrf9160/**/*"
+  - "drivers/entropy/**/*"
+  - "drivers/serial/**/*"
+  - "drivers/timer/**/*"
+  - "dts/arm/nordic/nrf5340*"
+  - "dts/arm/nordic/nrf9160*"
+  - "include/drivers/serial/**/*"
+  - "modules/trusted-firmware-m/**/*"
+  - "samples/tfm_integration/**/*"
+  - "subsys/random/**/*"
```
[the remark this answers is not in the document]
magnev: I just saw that it was compiled into the crypto tests. Not sure where it is used, so I will remove it

.github/test-spec.yml
magnev: Yes, I can reduce the filter to only check that file

.github/test-spec.yml
magnev: It might not be needed, I just saw that crypto tests compile source code in those directories so I assumed that tf-m tests do the same. I have perhaps been a bit too strict on the filter values, but I want to catch issues as early as possible. Especially  with serial since we use it extensively in our test system

joerchan (approved):

PR #1293, opened 2023-08-25
[nrf fromlist] tfm: Harden and Enforce TF-M Initial Attestion Key

joerchan:
Enforce that the initial attestation partition has the required
initial attestation key provisioned.

If the initial attestation key (IAK) is not present during boot of
TF-M the system will panic during initialization.

Introduce Kconfig option in zephyr build system that reflects the TF-M
cmake config variable with the same default value for dummy provisioning
and have it satisfy the IAK present requirement.
This configuration is not suitable for production, and by having this
in zephyr configuration we can have this as part of the hardened
configuration check.


mswarowsky (approved):

SebastianBoe (approved):

de-nordic:
I think this needs rebase or to be closed.

PR #1296, opened 2023-09-05
[nrf fromtree] bluetooth: host: convert conn address on identity reso…

kapi-no:
…lution

Fix an issue in the CCC configuration when the Peripheral device maintains two or more distinct connections on different identities with the same peer. The issue occurs when the local device performs the pairing and the bonding procedure on a connection associated with one of the Bluetooth identities. During the identity resolution, the peer address field in the CCC descriptor is converted from the RPA-type address to the Identity Address. However, the destination address on the remaining connection objects associated with other Bluetooth identities is not converted. Due to this, their CCC configuration is reset and GATT indications and notification fail to be sent even if the Central device subscribed to them.

Added necessary code to iterate over all connection objects during the identity resolution phase and aligned their destination address from the RPA-type to the Identity Address.

Signed-off-by: Kamil Piszczek <Kamil.Piszczek@nordicsemi.no>
(cherry picked from commit 42c904526b45bbc0fc87171567b88790783f0fbe)

### DIFF
--- a/subsys/bluetooth/host/smp.c
+++ b/subsys/bluetooth/host/smp.c
@@ -3813,6 +3813,20 @@ static uint8_t smp_id_add_replace(struct bt_smp *smp, struct bt_keys *new_bond)
 	return 0;
 }
 
+struct addr_match {
+	const bt_addr_le_t *rpa;
+	const bt_addr_le_t *id_addr;
+};
+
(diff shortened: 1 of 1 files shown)


hermabe (approved):

MarekPieta (approved):

kapi-no:
Will be merged as part of the upmerge process

PR #1308, opened 2023-09-18
nrf53: Add RTC pretick

ankuns:
This PR adds pretick feature to nRF5340 to workaround YOPAN-165.

Jira: KRKNWK-17580

The commit https://github.com/nordic-krch/fw-nrfconnect-zephyr/commit/a0016385c41c9a6528f8ea98b5ec3cd30fd9495c has been cherry-picked (with just conflicts resolved) to not infringe authorship of the commit.

Additionally the pretick is extended to allow usage of pretick feature with non-zero Kconfig option `NRF_RTC_TIMER_USER_CHAN_COUNT`. Just one CC is required for pretick feature and user channels can also be used.

The RTC0 events (generated by MPSL) are handled by sniffing RTC0 instead of relying on MPSL-provided special RTC0.CC[3]

Commits are [nrf fromtree]-tagged cherry-picks from already merged https://github.com/zephyrproject-rtos/zephyr/pull/63148


ankuns:
FYI: @nordic-krch @ahasztag @martintv

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -137,12 +143,177 @@ bool z_arm_on_enter_cpu_idle(void)
...
+		uint32_t rtc1_pretick_cc_val = (rtc1_counter + rtc1_ticks_to_next_event - 1U) & NRF_RTC_COUNTER_MAX;
+		static bool rtc1_pretick_cc_set_on_time = false;
+
+		if (rtc1_pretick_cc_val != nrf_rtc_cc_get(NRF_RTC1, RTC1_PRETICK_CC_CHAN)) {
+			/* The CC for pretick needs to be updated. */
+			nrf_rtc_cc_set(NRF_RTC1, RTC1_PRETICK_CC_CHAN, rtc1_pretick_cc_val);
+
+			/* Let's check if we updated CC on time so that the CC can take effect. */
+			rtc1_counter = nrf_rtc_counter_get(NRF_RTC1);
+			if (((rtc1_pretick_cc_val - rtc1_counter) & NRF_RTC_COUNTER_MAX) < 2) {
```
nordic-krch, line 161: based on this @anangl comment: https://github.com/zephyrproject-rtos/zephyr/blob/0933a4b2949889ca924b10aff05b1093dabe7ba0/drivers/timer/nrf_rtc_timer.c#L242
it probably should be `< 3`. @anangl can you confirm?

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -137,12 +143,177 @@ bool z_arm_on_enter_cpu_idle(void)
...
+				/* The COUNTER value is close enough to the expected pretick CC,
+				 * that the pretick event generation is not guaranteed.
+				 */
+				rtc1_pretick_cc_set_on_time = false;
+			} else {
+				/* The written rtc1_pretick_cc is guaranteed to to trigger compare event. */
```
nordic-krch, line 167: There is also a case when next evt was `now+1` and pretick then was set to `now` (the now that was read at the beginning of the function). Second `now` read could progress by 1 tick so `rtc1_pretick_cc_val - rtc1_counter` will give 0xffffff so comment won't be true but it will mean that event just expired so there will be ISR and there is no need for pretick. Am I right or messed it up? If yes then maybe add something like `or user event just expired`.

soc/arm/nordic_nrf/nrf53/soc.c
ankuns: fixed to `3`

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -137,12 +218,200 @@ bool z_arm_on_enter_cpu_idle(void)
...
+
+		barrier_dmem_fence_full();
+		bool rtc_scheduled = soc_rtc_ticks_to_next_event_get(NRF_RTC1, RTC1_PRETICK_SELECTED_CC_MASK, rtc_counter, &rtc_ticks_to_next_event);
+
+		uint32_t rtc0_ticks_to_next_event = 0U;
+		uint32_t rtc0_counter = nrf_rtc_counter_get(NRF_RTC0);
+		barrier_dmem_fence_full();
+		if (soc_rtc_ticks_to_next_event_get(NRF_RTC0, RTC0_PRETICK_SELECTED_CC_MASK, rtc0_counter, &rtc0_ticks_to_next_event)) {
+			/* An event is scheduled on RTC0. Need */
+			uint32_t diff = rtc_counter_diff_get(NRF_RTC0, NRR_RTC1);
```
ahasztag, line 235:
```suggestion
			uint32_t diff = rtc_counter_diff_get(NRF_RTC0, NRF_RTC1);
```

typo

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -137,12 +218,199 @@ bool z_arm_on_enter_cpu_idle(void)
...
+	if (ok_to_sleep) {
+		uint32_t rtc_ticks_to_next_event = 0U;
+
+		uint32_t rtc_counter = nrf_rtc_counter_get(NRF_RTC1);
+
+		barrier_dmem_fence_full();
+		bool rtc_scheduled = soc_rtc_ticks_to_next_event_get(NRF_RTC1, RTC1_PRETICK_SELECTED_CC_MASK, rtc_counter, &rtc_ticks_to_next_event);
+
+		uint32_t rtc0_ticks_to_next_event = 0U;
+		uint32_t rtc0_counter = nrf_rtc_counter_get(NRF_RTC0);
```
ahasztag, line 231: I think that instead of reading rtc0 directly a better idea would be first calling `rtc_counter_diff_get` and then base everything on the acquired value of the RTC1 counter (so the RTC0 counter value would be simply RTC1 counter value + `diff` and the RTC0 is not read directly outside of `rtc_counter_diff_get`). There are lots of operation between reading the RTC1 and RTC0 counter values and there is no protection against RTC1 ticking in the mean time. This way the value RTC0 can already be ahead of RTC1 and CC events from RTC0 could be treated as being "earlier" although in reality they are later then the RTC1 events.

So my suggestion would be:
```
uint32_t diff = rtc_counter_diff_get(NRF_RTC0, NRR_RTC1);
uint32_t rtc_counter = nrf_rtc_counter_get(NRF_RTC1);
uint32_t rtc0_counter = rtc_counter + diff;
```

Reading RTC1 counter after diff also makes sure it has the newest possible value.

ahasztag (changes requested):

soc/arm/nordic_nrf/nrf53/soc.c
ahasztag: I do not see a reason why you can't.

You want to know the difference to the nearest event of RTC0 and RTC1 from a _common_ point in time. Currently the reference time for RTC0 and RTC1 is different, and it shouldn't.

ahasztag: (Of course earlier I meant `uint32_t rtc0_counter = (rtc_counter + diff) & NRF_RTC_COUNTER_MAX;`, I forgot about `& NRF_RTC_COUNTER_MAX`)

soc/arm/nordic_nrf/nrf53/soc.c
ankuns: @nordic-krch I believe I'm handling the case you mentioned correctly, please have a look.

soc/arm/nordic_nrf/nrf53/soc.c
ankuns: Fixed, please have a look.

soc/arm/nordic_nrf/nrf53/Kconfig.soc
```diff
@@ -47,6 +49,25 @@ config SOC_NRF53_ANOMALY_160_WORKAROUND
 	depends on SYS_CLOCK_EXISTS
 	select ARM_ON_ENTER_CPU_IDLE_HOOK
 
+config SOC_NRF53_RTC_PRETICK
+	bool
+	depends on !NRF_802154_RADIO_DRIVER
```
ankuns, line 54: Once https://github.com/nrfconnect/sdk-nrf/pull/12316 moves one CC from from nrf_rtc_timer.c to RTC0, get rid of 
```
depends on !NRF_802154_RADIO_DRIVER
```

soc/arm/nordic_nrf/nrf53/Kconfig.soc
```diff
@@ -10,11 +10,13 @@ config SOC_NRF5340_CPUAPP
 	select CPU_HAS_FPU
 	select ARMV8_M_DSP
 	select HAS_POWEROFF
+	imply SOC_NRF53_RTC_PRETICK
 
 config SOC_NRF5340_CPUNET
 	bool
 	select ARM_ON_EXIT_CPU_IDLE
 	imply SOC_NRF53_ANOMALY_160_WORKAROUND_NEEDED
+	imply SOC_NRF53_RTC_PRETICK
```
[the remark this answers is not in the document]
ankuns: See commit named `nrf53: disable WDT_NRFX if pretick is use on nrf5340 cpunet`

soc/arm/nordic_nrf/nrf53/soc.c
nordic-krch: yes, imo this case is handled correctly. I was only referring to the statement in the comment.

ankuns: comment updated

soc/arm/nordic_nrf/nrf53/Kconfig.soc
```diff
@@ -47,6 +49,25 @@ config SOC_NRF53_ANOMALY_160_WORKAROUND
 	depends on SYS_CLOCK_EXISTS
 	select ARM_ON_ENTER_CPU_IDLE_HOOK
 
+config SOC_NRF53_RTC_PRETICK
+	bool
```
nordic-krch, line 53: Imo it should have prompt. Otherwise you cannot do
`SOC_NRF53_RTC_PRETICK=n` in project config. Many may want to disable it given that in room temperature issue is so extremely rare.

ankuns: done

soc/arm/nordic_nrf/nrf53/Kconfig.soc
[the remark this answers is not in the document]
ankuns: Commit mentioned above shown issues with zome twister tests. Chenged approach and replaced it with commit named `
nrf53: disallow simultaneous pretick and WDT_NRFX`

soc/arm/nordic_nrf/nrf53/Kconfig.soc
ankuns: done

nordic-krch (approved):

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -123,6 +134,76 @@ static bool nrf53_anomaly_160_check(void)
...
+			else {
+				/* CC that fires no earlier than already found. */
+			}
+		}
+	}
+
+	return result;
+}
+
+static void rtc_counter_synchronized_get(NRF_RTC_Type * rtc_a, NRF_RTC_Type * rtc_b, uint32_t * counter_a, uint32_t * counter_b)
```
ahasztag, line 196: Didn't we want this function to be availble to other modules?

ankuns: That was my idea first, however it is required for pretick only, and I have no idea where to put it to allow it to be reused.

ahasztag (approved):

ahasztag (approved):

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -123,25 +134,359 @@ static bool nrf53_anomaly_160_check(void)
...
+	/* The value stored is irrelevant. If any exception took place after
+	 * cpu_idle_prepare_monitor_begin, the the local monitor is cleared and
+	 * the store fails returning 1.
+	 * See Arm v8-M Architecture Reference Manual:
+	 *   Chapter B9.2 The local monitors
+	 *   Chapter B9.4 Exclusive access instructions and the monitors
+	 * See Arm Cortex-M33 Processor Technical Reference Manual
+	 *   Chapter 3.5 Exclusive monitor
+	 */
+	return __STREXB(0U, &cpu_idle_prepare_monitor_dummy);
```
martintv, line 231: I tested this PR, and this works. 
but isn't this function now relying on that __STREXB instruction uses R0 to store the result of the strex instruction?... since functions return values in R0 this works as longs as strexb instruction does that... but assembler might use other regsisters than R0, might it not?

seems safer to me to explicitly compare cpu_idle_prepare_monitor_dummy to something afterwards.

but I might be missing something?

ankuns: We do not rely on specific register r0 here, but on return value of `__STREXB`. I'm aware of AAPCS and I know return values are generally in r0. But `__STREXB` is `__STATIC_FORCEINLINE`, the function is inlined, the compiler inlined also the static function `cpu_idle_prepare_monitor_end`. Compiler picked whichever registers it wished. 

If there was no inlining then the compiler would need to pick `r0` as a result of strexb, or generate a code that ensures that. 

Please have a look on disassembly below. Compiler picked, `r2` as a result of `strexb` instruction (the success/failure of store) , and then compared the `r2` to 0 (`cbnz r2, ... ` instruction) treating `r2` as the result of `cpu_idle_prepare_monitor_end`.

I don't think there is a need to check the value written, just the fact that store succeeded matters and that result is returned through whichever register compiler selected.
```
01023d24 <z_arm_on_enter_cpu_idle>:
{
 1023d24:	b5f8      	push	{r3, r4, r5, r6, r7, lr}
 */
__STATIC_FORCEINLINE uint32_t __STREXB(uint8_t value, volatile uint8_t *addr)
{
   uint32_t result;

   __ASM volatile ("strexb %0, %2, %1" : "=&r" (result), "=Q" (*addr) : "r" ((uint32_t)value) );
 1023d26:	491e      	ldr	r1, [pc, #120]	; (1023da0 <z_arm_on_enter_cpu_idle+0x7c>)
 1023d28:	2300      	movs	r3, #0
 1023d2a:	e8c1 3f42 	strexb	r2, r3, [r1]
	if (cpu_idle_prepare_monitor_end() == 0) {
 1023d2e:	bb52      	cbnz	r2, 1023d86 <z_arm_on_enter_cpu_idle+0x62>
		ok_to_sleep = cpu_idle_prepare_allows_sleep;
 1023d30:	4b1c      	ldr	r3, [pc, #112]	; (1023da4 <z_arm_on_enter_cpu_idle+0x80>)
 1023d32:	781d      	ldrb	r5, [r3, #0]
```

martintv (approved):

anangl (approved):

ankuns:
Rebased due to upmerge pr went in https://github.com/nrfconnect/sdk-zephyr/pull/1294

PR #1313, opened 2023-09-21
[nrf noup] ci: scripts: add quarantine file

PerMac:
fixup! [nrf noup] ci: scripts: add quarantine file

### DIFF
--- a/scripts/quarantine.yaml
+++ b/scripts/quarantine.yaml
@@ -79,3 +79,10 @@
   platforms:
     - nrf52dk_nrf52832
   comment: "Ram overflows, also in the upstream"
+
+- scenarios:
+    - sample.psa_crypto
+  platforms:
+    - nrf5340dk_nrf5340_cpuapp_ns
+    - nrf9160dk_nrf9160_ns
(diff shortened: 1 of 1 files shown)


mswarowsky (approved):

de-nordic (approved):

de-nordic:
Shouldn't this be paired with sdk-nrf PR for manifest update?

PerMac:
hmm... This change should have no effect at all on sdk-nrf CI, just affecting sdk-zephyr's CI

de-nordic:
> hmm... This change should have no effect at all on sdk-nrf CI, just affecting sdk-zephyr's CI. But we will need nrf updated, yes. I will create on

I am asking because the sdk-nrf PRs are also a method to inform that sdk-zephyr is moving forward in sdk-nrf west.
In this case this is not a problem, but if you have some changes that may conflict and you push PR basing on the current sdk-nrf you may end up with conflicts, because sdk-zephyr has extras on top, and then it is worth seeing what is bringing the change in and whether it is safe to rebase your sdk-zephyr PR, or there are extras that you need to wait for.

PR #1334, opened 2023-10-04
 net: dhcpv6: Add Zephyr DHCPv6 client 

rlubos:


jukkar:
Don't we need also the fix d68db1ecfa512bc416bfa906d74f441cba2ab8fb
`net: dhcpv6: Fix params check`
from upstream?

rlubos:
> Don't we need also the fix d68db1ecfa512bc416bfa906d74f441cba2ab8fb `net: dhcpv6: Fix params check` from upstream?

True, thanks for reminding! Just pushed the additional commit.

jukkar (approved):

jtguggedal (approved):

PR #1335, opened 2023-10-04
[nrf fromtree] Bluetooth: att: don't re-use the ATT buffer for confir…

jori-nordic:
…mations

If the peer is a zephyr host, there is no problem, as the Zephyr host limits sending parallel REQs and INDs.

But the spec allows sending those in parallel, and it may end up that the re-used REQ buffer hasn't been destroyed when an indication comes.

Only re-use the buffer when enqueuing ATT responses.

This means that we may run out of buffers if the peer sends too many indications and our application also sends a lot of commands/notifications.

The rationale for this is that having to handle a lot of requests is a more plausible scenario (e.g. being discovered by multiple peers) than handling lots of parallel indications.

Signed-off-by: Jonathan Rico <jonathan.rico@nordicsemi.no>
(cherry picked from commit 7093538c425b1bd3dd4bce3dfd3708a8281bd033)

### DIFF
--- a/subsys/bluetooth/host/att.c
+++ b/subsys/bluetooth/host/att.c
@@ -670,10 +670,12 @@ static struct net_buf *bt_att_chan_create_pdu(struct bt_att_chan *chan, uint8_t 
 
 	switch (att_op_get_type(op)) {
 	case ATT_RESPONSE:
-	case ATT_CONFIRMATION:
-		/* Use a timeout only when responding/confirming */
+		/* Use a timeout only when responding */
 		timeout = BT_ATT_TIMEOUT;
 		re_use = true;
+		break;
+	case ATT_CONFIRMATION:
+		timeout = BT_ATT_TIMEOUT;
 		break;
 	default:
 		timeout = K_FOREVER;
@@ -701,7 +703,7 @@ static struct net_buf *bt_att_chan_create_pdu(struct bt_att_chan *chan, uint8_t 
 			 * This is better than an assert as an assert would
 			 * allow a peer to DoS us.
 			 */
-			LOG_ERR("already processing a transaction on chan %p", chan);
+			LOG_ERR("already processing a REQ/RSP on chan %p", chan);
 
 			return NULL;
 		}


hermabe (approved):

alwa-nordic (approved):

de-nordic:
@jori-nordic Some sdk-nrf update maybe?

jori-nordic:
> Some sdk-nrf update maybe

I thought that was a solved problem, that it triggered the right CI automatically, isn't that the case?

de-nordic:
sdk-zephyr is part of sdk-nrf setup, and this will not get merged unless pointed by the sdk-nrf west manifest.
Currently it has been only tested in context of sdk-zephyr itself so it will not get merged because its effects on sdk-nrf is unknown.

jori-nordic:
Here it is then: https://github.com/nrfconnect/sdk-nrf/pull/12554

IMO, the CI should most definitely trigger sdk-nrf _before_ having to create the manifest PR. It's unnecessary friction/noise for PR authors, reviewers and release engineers.

de-nordic:
> Here it is then: [nrfconnect/sdk-nrf#12554](https://github.com/nrfconnect/sdk-nrf/pull/12554)
> 
> IMO, the CI should most definitely trigger sdk-nrf _before_ having to create the manifest PR. It's unnecessary friction/noise for PR authors, reviewers and release engineers.

How would CI figure your environment in case when you depend on changes in more than one repo, for example hal and sdk-zephyr?

jori-nordic:
Could be solved with a label e.g. `test-sdk-nrf` or something like that. Then you manually add it. Going further, the nordic bot could even open the PR.

Point is, there are a bunch of backports like this one that only change sdk-zephyr, with no other deps, and the PR process could be streamlined.

I've also seen times, usually at release, that there are multiple redundant manifest updates. 
The bot could also detect this, so we'd have a sort of merge-queue instead of having to do busy-work keeping up with which manifest PR is merged or going to be merged and rebasing, and retriggering CI wasting nodes etc..

PR #1336, opened 2023-10-04
Don't set a minimum value for rx_timeout_slab if HW timers are used

pres2300:
For a description of this issue, see [ticket 62828](https://github.com/zephyrproject-rtos/zephyr/issues/62828#event-10477372536).

I'm only developing on an nrf52840, but did my best to make it compatible with other use cases.  It'd be great if someone could confirm this change to ensure there are no unintended side-effects.


drivers/serial/uart_nrfx_uarte.c
```diff
@@ -850,9 +857,14 @@ static int uarte_nrfx_rx_enable(const struct device *dev, uint8_t *buf,
 	 * way but timeouts would always occur later than expected,  most likely
 	 * after ~3 ticks.
 	 */
+#ifdef UARTE_HW_ASYNC
```
nordic-krch, line 860: Do you see any gain by doing that? See comment above:
```
Note that lower value would work in a similar way but timeouts would always occur later than expected,  most likely after ~3 ticks.
```
If  timeout is less than 3 RTC ticks algorithm  for setting RTC timeout (in nrf_rtc_timer.c) will fail on first try and be repeated which will lead to 3 tick timeout or more so it's better to immediately start with 3 ticks.

pres2300: With my changes I no longer receive spurious `UART_RX_RDY` events.  An explanation about why this happens is given in [ticket 62828](https://github.com/zephyrproject-rtos/zephyr/issues/62828#event-10477372536)

In summary when `rx_timeout` is set to a low value it is possible to trigger the `UART_RX_RDY` event incorrectly due to how `rx_timeout_slab` gets set.  I'm not sure if the consideration needs to be made when setting `rx_timeout_slab` when HW timers are used.

anangl: > If timeout is less than 3 RTC ticks algorithm for setting RTC timeout (in nrf_rtc_timer.c) will fail on first try and be repeated which will lead to 3 tick timeout or more so it's better to immediately start with 3 ticks.

@nordic-krch This can happen, but not necessarily always. When kernel realizes that the tick at which the timeout is to be scheduled has already passed, it does not set up a system timer timeout for it, but it calls the handler immediately (thus, the comment you pointed out is not entirely accurate). We should let the kernel cope with the timeouts needed by the UART driver, not try to do a part of its job here. BTW, this rounding up code is anyway incorrect, as it uses `CONFIG_SYS_CLOCK_TICKS_PER_SEC` instead of `CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC`, and these two values may not be the same.

drivers/serial/uart_nrfx_uarte.c
```diff
@@ -850,9 +857,14 @@ static int uarte_nrfx_rx_enable(const struct device *dev, uint8_t *buf,
...
+	data->async->rx_timeout_slab = timeout / RX_TIMEOUT_DIV;
```
anangl, line 861: Only this one line should stay here. Everything else in the surrounding `#ifdef` and the preceding comment should be removed.

drivers/serial/uart_nrfx_uarte.c
```diff
@@ -65,6 +65,13 @@ LOG_MODULE_REGISTER(uart_nrfx_uarte, CONFIG_UART_LOG_LEVEL);
 #define UARTE_ANY_ASYNC 1
 #endif
 
+#if (defined(CONFIG_UART_0_NRF_HW_ASYNC) && defined(CONFIG_UART_0_NRF_HW_ASYNC_TIMER)) || \
+	(defined(CONFIG_UART_1_NRF_HW_ASYNC) && defined(CONFIG_UART_1_NRF_HW_ASYNC_TIMER)) || \
+	(defined(CONFIG_UART_2_NRF_HW_ASYNC) && defined(CONFIG_UART_2_NRF_HW_ASYNC_TIMER)) || \
+	(defined(CONFIG_UART_3_NRF_HW_ASYNC) && defined(CONFIG_UART_3_NRF_HW_ASYNC_TIMER))
+#define UARTE_HW_ASYNC 1
+#endif
+
```
anangl, line 74: This should not be added.

anangl (changes requested):
Such PR should be created upstream, i.e. in https://github.com/zephyrproject-rtos/zephyr/.

pres2300:
> Such PR should be created upstream, i.e. in https://github.com/zephyrproject-rtos/zephyr/.
> 
> @pres2300 Please let me know if you want to do it yourself or if I should do it (as the assignee of [zephyrproject-rtos/zephyr#62828](https://github.com/zephyrproject-rtos/zephyr/issues/62828)).

I can make the changes you suggest to this PR and then create a new PR upstream.  I'll comment back here when I get around to doing that (hopefully sometime this week).

pres2300:
@anangl I have opened a PR upstream: https://github.com/zephyrproject-rtos/zephyr/pull/64094

pres2300:
I'm not quite sure what happened to this PR with all the commits, but this has been resolved upstream with https://github.com/zephyrproject-rtos/zephyr/commit/52c55177ba71af328c6069ea56c007e6eb5cdf1d

Should this PR be closed or remain open until synced with upstream?

anangl:
> Should this PR be closed or remain open until synced with upstream?

There is no need to keep this PR open. Please close it.

pres2300:
Closing as this fix has been merged upstream: https://github.com/zephyrproject-rtos/zephyr/pull/64094

PR #1346, opened 2023-10-10
[nrf fromlist] nrf53: fix RTC pretick for RTC rescheduling by other i…

ankuns:
…nterrupts

It might happen that while some interrupt handler other than for RTC0 or RTC1 (e.g. for RADIO) is executed, the scheduled pretick CC triggers. This starts pretick pulses due to the loop through IPC. The change in pretick schedule did not stop the pretick pulses going through IPC loop, what caused heavy increase in power consumption.

This commit fixes this behavior.
Added also clarifications for Kconfig option `SOC_NRF53_RTC_PRETICK`.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/63597


martintv (approved):

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -224,6 +225,15 @@ static bool cpu_idle_prepare_monitor_end(void)
 	return __STREXB(0U, &cpu_idle_prepare_monitor_dummy);
 }
 
+static void rtc_pretick_finish_previous(void)
+{
+	NRF_IPC->PUBLISH_RECEIVE[CONFIG_SOC_NRF53_RTC_PRETICK_IPC_CH_TO_NET] &=
+			~IPC_PUBLISH_RECEIVE_EN_Msk;
+
+	nrf_rtc_event_clear(NRF_RTC1, NRF_RTC_CHANNEL_EVENT_ADDR(RTC1_PRETICK_CC_CHAN));
```
jciupis, line 233: What do you think about adding 
```c
NRF_WDT->TASKS_STOP = 1;
```
in this function as well?

ankuns: Stopping WDT happens from idle task when we are ok to sleep and the pretick CC did not fire yet.

ivaniushkov (approved):
tested manually, the power consumption for periodic advertiser is ok now

soc/arm/nordic_nrf/nrf53/soc.c
jciupis: Stopping it here could lower the power consumption and wouldn't change the solution functionally. Let's proceed with what we have though to minimize the risk that some additional race or bug slips through when we stop the WDT here.

jciupis (approved):

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -133,6 +133,7 @@ static bool nrf53_anomaly_160_check(void)
 
 	return true;
 }
+#endif /* CONFIG_SOC_NRF53_ANOMALY_160_WORKAROUND */
```
anangl, line 136: Just a detail.
This causes that `z_arm_on_enter_cpu_idle` is defined always, even if neither of those two workarounds uses it. That hook might be potentially needed for some other purpose, so I think it would be good to wrap its definition in
```
#if defined(CONFIG_SOC_NRF53_ANOMALY_160_WORKAROUND) || \
	(defined(CONFIG_SOC_NRF53_RTC_PRETICK) && defined(CONFIG_SOC_NRF5340_CPUNET))
```

soc/arm/nordic_nrf/nrf53/soc.c
```diff
@@ -361,6 +373,7 @@ bool z_arm_on_enter_cpu_idle(void)
 		}
 #endif
 	}
+#endif
```
anangl, line 376:
```suggestion
#endif /* CONFIG_SOC_NRF53_ANOMALY_160_WORKAROUND */
```

anangl (approved):

soc/arm/nordic_nrf/nrf53/soc.c
ankuns: done

soc/arm/nordic_nrf/nrf53/soc.c
ankuns: done

PR #1348, opened 2023-10-11
[nrf noup] modules: mbedtls: Disable mbedtls_hardware_poll with CC3xx…

rlubos:
… enabled

squash! [nrf noup] modules: mbedtls: Disable configurations in Kconfig.tls-generic

Precompiled CC3xx libraries silently provide mbedtls_hardware_poll function (used internally by the library on initialization), which creates ambiguity with Zephyr-provided implementation. In result, the CC3xx entropy source is broken when CONFIG_MBEDTLS_ZEPHYR_ENTROPY is enabled. Workaround this, by disabling the Zephyr implementation if CC3xx is enabled, and let mbed TLS use the implementation from the CC library instead.


rlubos:
@anangl GH suggested you as a reviewer :)

katgiadla (approved):

anangl (approved):

PR #1351, opened 2023-10-11
[nrf noup] ci: Allow non-github upstream PRs

krish2718:
fixup! [nrf noup] ci: NCS-specific CI tweaks

OSS projects still use mailing-lists, so, using a URL for patchwork or a mailing should be allowed instead of failing.

The regex is made too wide intentionally to allow all sorts, even the http URLs.

### DIFF
--- a/scripts/gitlint/ncs.py
+++ b/scripts/gitlint/ncs.py
@@ -46,7 +46,7 @@ class NCSSauceTags(CommitRule):
                     commit.message.title):
                 return [RuleViolation(self.id, 'Invalid mergeup commit title')]
         elif tag == 'fromlist':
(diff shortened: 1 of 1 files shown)


krish2718:
I don't think this needs a manifest update, but let me know if I need to raise one.

sachinthegreen (approved):

carlescufi:
> I don't think this needs a manifest update, but let me know if I need to raise one.

It does not.

carlescufi (changes requested):
This is Zephyr, there will always be a PR in a `fromlist` patch.

carlescufi:
Closing

PR #1354, opened 2023-10-16
[nrf noup] Migrate sample child image configuration to sysbuild

nordicjm:


nordicjm:
@alwa-nordic @asbjornsabo @hermabe @Thalley please review

hermabe (commented):
I dont know exactly how sysbuild works, but the config for hci_rpmsg matches the existing one for the child_image so fine by me.

Some questions:
- Are you planning on removing the child_image configs and machinery? Now there seems to be two ways to build this sample.
- Why only this sample and not the others?
- Why in sdk-zephyr instead of upstream?

nordicjm:
>     * Are you planning on removing the child_image configs and machinery? Now there seems to be two ways to build this sample.

They will both co-exist for a period of time, then child/parent image support will be removed
 
>     * Why only this sample and not the others?

These are the only child_image folders in sdk-zephyr
 
>     * Why in sdk-zephyr instead of upstream?

Child/parent is an ncs thing, the configurations copied here do not exist nor would they be used in zephyr. If they *should* be present in upstream zephyr, then that would be the fault of whomever added them to sdk-zephyr instead of submitting them to zephyr

hermabe (approved):

PR #1370, opened 2023-10-31
sysbuild: Add alternative hex file and sysbuild signing script support

nordicjm:
[nrf noup] cmake: flash: Add support for alternative hex file
    
    Adds support for an alternative hex file, this is needed for
    sysbuild partition manager to flash the combined hex file

    [nrf fromlist] cmake: Add support for sysbuild-set signing script
    
    Allows a sysbuild project to specify a signing script file to
    use instead of the default zephyr one


cmake/flash/CMakeLists.txt
```diff
@@ -36,8 +36,17 @@ function(runners_yaml_append_config)
...
+
+  if(SYSBUILD)
+    zephyr_get(alt_file VAR ALT_SIGNED_HEX_FILE SYSBUILD)
+  endif()
+
   if(CONFIG_BUILD_OUTPUT_HEX)
-    get_runners_prop(hex_file hex "${KERNEL_HEX_NAME}")
+    if(SYSBUILD AND DEFINED alt_file)
+      set(hex "${alt_file}")
+    else()
```
tejlmand, line 47: not directly related to those changes themselves, but I was wondering if this would be a use-case where `sysbuild` should be able to support flashing of specific binaries / hex files.

That way, sysbuild could mark the `app` image as build only, and instead specify its own alternate image as a custom hex file to flash.

nordicjm: Could be a possible expansion, though there area things like https://github.com/zephyrproject-rtos/zephyr/pull/64551 which throw a spanner in the works

hakehuang: > not directly related to those changes themselves, but I was wondering if this would be a use-case where `sysbuild` should be able to support flashing of specific binaries / hex files.

it is not support well for `west flash`, as it heavyily depends on  CMakeCache.txt and runners.yaml, which is generated by west build

cmake/flash/CMakeLists.txt
```diff
@@ -36,8 +36,17 @@ function(runners_yaml_append_config)
...
   runners_yaml_append("  # Build outputs:")
   runners_yaml_append("  elf_file: ${elf}")
+
+  if(SYSBUILD)
+    zephyr_get(alt_file VAR ALT_SIGNED_HEX_FILE SYSBUILD)
+  endif()
+
   if(CONFIG_BUILD_OUTPUT_HEX)
-    get_runners_prop(hex_file hex "${KERNEL_HEX_NAME}")
+    if(SYSBUILD AND DEFINED alt_file)
```
tejlmand, line 45: you already tested for `if(SYSBUILD)` above, so if `SYSBUILD` is false then `alt_file` will be undefined.

So would be enough to just do:
```suggestion
    if(DEFINED alt_file)
```

tejlmand (approved):
minor nit, but otherwise looks good.

57300 (approved):

PR #1371, opened 2023-10-31
wifi_shell: Remove __wifi_args_to_params function

D-Triveni:
Remove the __wifi_args_to_params function from wifi_shell.c, as it is now part of a wifi_credentials library.


krish2718:
This is still used by `wifi_mgmt` and `wifi_shell`, and these are upstream modules and `wifi_credentials` is NCS only.

PR #1372, opened 2023-11-01
Zperf multicast improvements

krish2718:
Multicast receive for Zperf.

### DIFF
--- a/subsys/net/lib/zperf/zperf_shell.c
+++ b/subsys/net/lib/zperf/zperf_shell.c
@@ -178,6 +178,39 @@ static int parse_ipv4_addr(const struct shell *sh, char *host, char *port,
 	return 0;
 }
 
+static int zperf_bind_host(const struct shell *sh,
+			   size_t argc, char *argv[],
+			   struct zperf_download_params *param)
+{
+	int ret;
+
+	/* Parse options */
+	if (argc >= 2) {
+		param->port = strtoul(argv[1], NULL, 10);
+	} else {
+		param->port = DEF_PORT;
+	}
+
+	if (argc >= 3) {
+		char *addr_str = argv[2];
+		struct sockaddr addr;
+
+		memset(&addr, 0, sizeof(addr));
+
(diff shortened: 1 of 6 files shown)


subsys/net/lib/zperf/zperf_shell.c
```diff
@@ -191,7 +191,7 @@ static int cmd_setip(const struct shell *sh, size_t argc, char *argv[])
 		if (zperf_get_ipv6_addr(argv[start + 1], argv[start + 2],
 					&shell_ipv6) < 0) {
 			shell_fprintf(sh, SHELL_WARNING,
-				      "Unable to set IP\n");
+				      "Unable to set %s address (%s disabled)\n", "IPv6", "IPv4");
```
rado17, line 227: Why pass const string as argument instead having it as part of the fmt itself?

subsys/net/lib/zperf/zperf_shell.c
```diff
@@ -1110,7 +1110,7 @@ void zperf_shell_init(void)
 {
 	int ret;
 
-	if (IS_ENABLED(CONFIG_NET_IPV6) && MY_IP6ADDR) {
+	if (IS_ENABLED(MY_IP4ADDR_SET) && MY_IP6ADDR) {
```
rado17, line 1150: Should this be IS_ENABLED(MY_IP6ADDR_SET)

subsys/net/lib/zperf/zperf_shell.c
krish2718: To save memory, the format strings are RO and can be re-used if they are same saving memory.

subsys/net/lib/zperf/zperf_shell.c
krish2718: oops, need to fix upstream as well, good catch.

krish2718: updated with the fix.

rado17 (approved):

sachinthegreen (approved):

PR #1379, opened 2023-11-06
[nrf noup] ci: Add CI-run-zephyr-twister test spec

jangalda-nsc:
CI-run-zephyr-twister is used to determine if twister tests from sdk-zephyr should be run on PR to sdk-nrf repository.

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -1,4 +1,19 @@
 # This is the Jenkins ci variant of the .github/labler.yaml
+
+"CI-run-zephyr-twister":
+  - any:
+      - "!.github/**/*"
+      - "!doc/**/*"
+      - "!CODEOWNERS"
+      - "!LICENSE"
+      - "!**/*.rst"
+      - "!VERSION"
+      - "!submanifests/**/*"
+      - "!MAINTAINERS.yml"
+      - "!version.h.in"
+      - "!Jenkinsfile"
+      - "!**/*.md"
+
 "CI-iot-zephyr-lwm2m-test":
   - "drivers/console/**/*"
   - "drivers/flash/**/*"


.github/test-spec.yml
```diff
@@ -1,4 +1,17 @@
...
+      - "!.github/**/*"
+      - "!.vscode/**/*"
+      - "!doc/**/*"
+      - "!test-manifests/**/*"
+      - "!CODEOWNERS"
+      - "!LICENSE"
+      - "!**/*.rst"
+      - "!VERSION"
+      - "!ncs_version.h.in"
+
```
thst-nordic, line 14: add more? 

!submanifests/**/*
!scripts/**/*
!snippets/**/*
!scripts/**/*
!CODEOWNERS
!MAINTAINERS.yml
!version.h.in
!Jenkinsfile
"!**/*.md"

.github/test-spec.yml
```diff
@@ -1,4 +1,17 @@
...
+  - any:
+      - "!.github/**/*"
+      - "!.vscode/**/*"
+      - "!doc/**/*"
+      - "!test-manifests/**/*"
+      - "!CODEOWNERS"
+      - "!LICENSE"
+      - "!**/*.rst"
+      - "!VERSION"
+      - "!ncs_version.h.in"
```
thst-nordic, line 13: remove `ncs_`

DatGizmo (approved):

.github/test-spec.yml
PerMac: IMO we can leave the scripts to be traced (not excluded) for now. Later on we can play with a smarter scope selection. E.g. since recently, twister vastly improved on providing unit tests for itself and those are triggered as a github workflow,. However, I think many of those haven't came to sdk-zephyr yet. Also, the test_plan script which we will use in sdk-nrf has means of descoping certain patterns from being added to the scope. If we expand it somehow to reliably work with changes in zephyr we could leave the decision to be made by that script

thst-nordic (approved):

.github/test-spec.yml
PerMac: why snippets are excluded?

PR #1388, opened 2023-11-08
[nrf fromtree] boards: arm: add nrf9131ek_nrf9131

maxd-nordic:
This patch adds the nRF9131-EK board.

(cherry-picked from 386b65800896262779336e83552c8a44ba96f7a2)


maxd-nordic:
this is slightly different to the upstream version due to a change in tfm

carlescufi (approved):

anangl (changes requested):
> this is slightly different to the upstream version due to a change in tfm

As a fromtree commit, it cannot be slightly different. The diff must be the same as in upstream.

maxd-nordic:
> > this is slightly different to the upstream version due to a change in tfm
> 
> As a fromtree commit, it cannot be slightly different. The diff must be the same as in upstream.

How do I fix this? Our version of Zephyr is not up to date enough to support the exact upstream version.

anangl:
> How do I fix this? Our version of Zephyr is not up to date enough to support the exact upstream version.

Cherry-pick more commits from upstream. If that is not feasible, add a temporary `[nrf noup]` commit with the extra changes you need.

anangl (approved):

PR #1403, opened 2023-11-29
[nrf noup] modules: openthread: Add PSA implementation for PBDKF2 genkey

joerchan:
Add implementation of opentherad pbkdf2 generate key using PSA functions.

Using assert instead of return code until openthread function signature is changed.

noup: zephyr needs to be using mbedtls 3.5.0 to support PBKDF2.


joerchan:
Closing this, can be moved to upstream zephyr once mbedtls 3.5.0 is integrated, required for PBKDF2 support from mbedtls.

PR #1409, opened 2023-12-01
[nrf fromtree] soc: arm: Remove CPU_HAS_NRF_IDAU's dependencies

mswarowsky:
CPU_HAS_NRF_IDAU is depending on SOC_SERIES_NRF91X and SOC_NRF5340_CPUAPP.

This makes it more difficult to have an out-of-tree nrf SOC.

It is also an unnecessary dependency. There is no prompt for CPU_HAS_NRF_IDAU, so it will not show up in menuconfig and it won't be possible to enable it from a users Kconfig fragment.

The only way to enable it is to select it, and those that select this option can themselves make sure that they only do so when appropriate.

Also, move NRF_SPU options out to SOC Kconfig files to also make out-of-tree SoCs possible. With the added benefit of not polluting the common soc/arm/Kconfig.

(cherry picked from commit 6348a973c38eeed5e18e6821dcb38a1c2f78ac76)

### DIFF
--- a/soc/arm/nordic_nrf/nrf53/Kconfig.soc
+++ b/soc/arm/nordic_nrf/nrf53/Kconfig.soc
@@ -98,6 +98,18 @@ config SOC_DCDC_NRF53X_HV
 	help
 	  Enable nRF53 series System on Chip High Voltage DC/DC converter.
 
+config NRF_SPU_FLASH_REGION_SIZE
+	hex
+	default 0x4000
+	help
+	  FLASH region size for the NRF_SPU peripheral
+
+config NRF_SPU_RAM_REGION_SIZE
+	hex
+	default 0x2000
+	help
+	  RAM region size for the NRF_SPU peripheral
+
 config SOC_NRF_GPIO_FORWARDER_FOR_NRF5340
 	bool
 	depends on NRF_SOC_SECURE_SUPPORTED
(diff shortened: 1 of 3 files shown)


vili-nordic (approved):

anangl (changes requested):
Please use for these two new commits titles that differ at least a bit from each other (for example, add "for nRF53" and "for nRF91", respectively). In the current form, it is confusing.

mswarowsky:
> Please use for these two new commits titles that differ at least a bit from each other (for example, add "for nRF53" and "for nRF91", respectively). In the current form, it is confusing.

Oh yeah, added nRF 53 and 91 in the commit message

soc/arm/nordic_nrf/nrf53/Kconfig.soc
```diff
@@ -98,6 +98,32 @@ config SOC_DCDC_NRF53X_HV
 	help
 	  Enable nRF53 series System on Chip High Voltage DC/DC converter.
 
+config NRF_SPU_FLASH_REGION_SIZE
+	hex
+	default 0x4000
+	help
+	  FLASH region size for the NRF_SPU peripheral
+
+config NRF_SPU_FLASH_START_ADDR_ALIGNMENT
```
anangl, line 107: Why not `NRF_SPU_FLASH_REGION_ALIGNMENT`, actually?

anangl (approved):

soc/arm/nordic_nrf/nrf53/Kconfig.soc
mswarowsky: good question changed it

anangl (approved):

PR #1411, opened 2023-12-04
modules: hal_nordic: nRF 802.15.4 customizable asserts

ankuns:
Recent nRF 802.15.4 Radio Driver provides assert abstraction layer. The assert abstraction layer is implemented in Zephyr in following ways depending on the `NRF_802154_ASSERT_CHOICE` Kconfig choice.

`NRF_802154_ASSERT_ZEPHYR_MINIMAL` (default) gives ability to still perform run-time checking of the nRF 802.15.4 Radio Driver operation with minimum memory overhead and configurable behavior on fault detection regardless of the `CONFIG_ASSERT` Kconfig option value.

`NRF_802154_ASSERT_ZEPHYR` gives ability to use asserts provided and configurable by Zephyr including the ability to turn off the run-time checking of the nRF 802.15.4 Radio Driver operation.


modules/hal_nordic/nrf_802154/nrf_802154_assert_handler.c
```diff
@@ -0,0 +1,26 @@
...
+
+#if defined(CONFIG_NRF_802154_ASSERT_ZEPHYR_MINIMAL)
+
+__weak void nrf_802154_assert_handler(void)
+{
+#ifdef CONFIG_USERSPACE
+	/* User threads aren't allowed to induce kernel panics; generate
+	 * an oops instead.
+	 */
+	if (k_is_user_context()) {
```
hubertmis, line 18: Why `k_panic()` is used here at all? Isn't `k_oops()` a better choice?

ankuns: Based on https://github.com/zephyrproject-rtos/zephyr/blob/main/lib/os/assert.c#L26

hubertmis (approved):

modules/hal_nordic/Kconfig
```diff
@@ -205,13 +205,39 @@ endif
...
+choice NRF_802154_ASSERT_CHOICE
+	prompt "nRF 802.15.4 assert implementation"
+	default NRF_802154_ASSERT_ZEPHYR_MINIMAL
+
+config NRF_802154_ASSERT_ZEPHYR_MINIMAL
+	bool "nRF 802.15.4 minimal assertions"
+	help
+	  This option provides minimal run-time checking of the nRF 802.15.4 Radio Driver operation
+	  even if kernel-wide CONFIG_ASSERT is disabled. In case of abnormal condition the function
+	  `nrf_802154_assert_handler()` is called. No file and line debug information are not provided
```
jciupis, line 225:
```suggestion
	  `nrf_802154_assert_handler()` is called. File and line debug information are not provided
```

jciupis (approved):

modules/hal_nordic/Kconfig
ankuns: fixed

anangl (approved):

PR #1415, opened 2023-12-06
[nrf fromlist] Fix tests and samples for TF-M with partition-manager

joerchan:
Make sure that the flash partition is valid.
With TF-M enabled the storage_partition can be used as memory that is
known to be configured as non-secure flash region.
The slot1_ns_partition partition is only define when TF-M is built with bootloader.


nordicjm (approved):

joerchan:
DNM: Can maybe replace this with the changes from: https://github.com/zephyrproject-rtos/zephyr/pull/66279

Instead of enabling a bootloader for the slot1_ns_partition  to be valid, we can enable settings for the storage_partition to be valid.

anangl (approved):

PR #1429, opened 2023-12-22
Wi-Fi shell improvs

krish2718:


subsys/net/l2/wifi/wifi_shell.c
```diff
@@ -1655,7 +1655,8 @@ SHELL_STATIC_SUBCMD_SET_CREATE(wifi_commands,
 	SHELL_CMD_ARG(disconnect, NULL, "Disconnect from the Wi-Fi AP",
 		  cmd_wifi_disconnect,
 		  1, 0),
-	SHELL_CMD_ARG(ps, NULL, "Configure Wi-F PS on/off, no arguments will dump config",
+	SHELL_CMD_ARG(ps, NULL, "Configure Wi-F power save state\n"
```
sachinthegreen: Configure or display Wi-Fi power save state\n

sachinthegreen (changes requested):

krish2718:
FYI, for a single line change in help, had to pull in many of the scan re-work fromtree commits, as upmerge is still pending.

sachinthegreen (approved):

include/zephyr/net/wifi_mgmt.h
```diff
@@ -41,6 +41,12 @@ extern "C" {
 #define WIFI_MGMT_SCAN_SSID_FILT_MAX 0
 #endif /* CONFIG_WIFI_MGMT_SCAN_SSID_FILT_MAX */
 
+#ifdef CONFIG_WIFI_MGMT_SCAN_CHAN_MAX_MANUAL
+#define WIFI_MGMT_SCAN_CHAN_MAX_MANUAL CONFIG_WIFI_MGMT_SCAN_CHAN_MAX_MANUAL
+#else
+#define WIFI_MGMT_SCAN_CHAN_MAX_MANUAL 1
+#endif /* CONFIG_WIFI_MGMT_SCAN_CHAN_MAX_MANUAL */
```
trantanen, line 48: In which case we hit the else branch? Can there be some other `wifi_mgmt` implementation than the one in `subsys/net/l2/wifi`?

krish2718: This is a workaround for a bug in few cellular/location samples, the Wi-Fi headers are included unconditionally (`nrf/include/net/wifi_location_common.h`) and if the Wi-Fi isn't enabled, this poses a problem, as Kconfig isn't included. This was first instance of a Kconfig being used in the header files.

trantanen: Thanks for the info. I know some/many people prefer to use as little conditions as possible which causes these issues. I'll keep this in mind to see if there would be better solution for this.

rado17 (approved):

PR #1430, opened 2023-12-29
wifi: Implement AP events

krish2718:


rado17 (approved):

include/zephyr/net/wifi_mgmt.h
```diff
@@ -519,6 +562,18 @@ struct wifi_raw_scan_result {
...
 #endif /* CONFIG_WIFI_MGMT_RAW_SCAN_RESULTS */
 
+/** AP mode - connected STA details */
+struct wifi_ap_sta_info {
+	/** Link mode, see enum wifi_link_mode */
+	enum wifi_link_mode link_mode;
+	/** MAC address */
+	uint8_t mac[WIFI_MAC_ADDR_LEN];
+	/** MAC address length */
+	uint8_t mac_length;
```
sachinthegreen, line 572: Is this needed (as in would there be a case where it would be different than WIFI_MAC_ADDR_LEN) ?

sachinthegreen (changes requested):

include/zephyr/net/wifi_mgmt.h
krish2718: I don't think so, but its the norm in Zephyr L2 to not hardcode MAC lengths, if you look at this file we use same logic for scan results as well. We can clean that up in a separate PR.

sachinthegreen (approved):

PR #1441, opened 2024-01-11
[nrf noup] drivers: ieee802154: temporary fix for ACK header IE config

maciejbaczmanski:
Skip configuring ACK header IE if `EnableCsl()`
has been called with arguments:

- `aCslPeriod` == 0
- `aShortAddr` == IEEE802154_NO_SHORT_ADDRESS_ASSIGNED
- `aExtAddr` == NULL

This happens during RCP startup in `Radio::Init()` implementation.
It will be fixed on OT and Zephyr level, but this commit is needed
to prevent coprocessor samples from asserting on boot.

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -886,6 +886,19 @@ static int nrf5_configure(const struct device *dev,
 		uint8_t short_addr_le[SHORT_ADDRESS_SIZE];
 		uint8_t element_id;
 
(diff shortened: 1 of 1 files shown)


jciupis (changes requested):
Please rework as discussed f2f

jciupis (approved):
Dirty but acceptable, let's merge it and start working towards a proper fix upstream

canisLupus1313 (approved):

carlescufi (approved):

PR #1455, opened 2024-01-18
[nrf fromlist] drivers: ieee802154: fix nRF5 Rx error handling

jciupis:
The current implementation implicitly assumes that if the device is configured to have the capability of acting as a CSL endpoint then in case a delayed reception with matching ID finishes with a timeout no action is needed. This assumption is correct when RxOnWhenIdle mode is disabled because the transition to sleep is done automatically by the driver below. However, it's wrong when RxOnWhenIdle is enabled. This commit fixes that case by adding a call to event handler that notifies the higher layer about the event and allows it to transition to RxOff if needed.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/67774

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -763,6 +763,7 @@ static int nrf5_init(const struct device *dev)
 
 	nrf5_get_capabilities_at_boot();
 
+	nrf5_radio->rx_on_when_idle = true;
 	nrf5_radio_cfg->irq_config_func(dev);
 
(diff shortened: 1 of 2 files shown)


ankuns:
Commit https://github.com/zephyrproject-rtos/zephyr/pull/67774/commits/b85cb9aacaf9cf95b76e04c6e404c705430c6d97 upushed to upstream PR, please update here.

jciupis:
@ankuns done.

ankuns (approved):

dawidprzybylo (approved):

PR #1474, opened 2024-01-29
[nrf fromtree] Bluetooth: Mesh: enable access responses randomization

HaavardRei:
Enable by default the access layer responses random delays. Commit also adapts all mesh models, samples and
babblesim tests to use random delay functionality correctly.

(cherry picked from commit 4e7d64b1b45247de1e1d71a4402f5637b0fd0afe)

### DIFF
--- a/subsys/bluetooth/mesh/Kconfig
+++ b/subsys/bluetooth/mesh/Kconfig
@@ -639,6 +639,7 @@ config BT_MESH_LABEL_NO_RECOVER
 
 menuconfig BT_MESH_ACCESS_DELAYABLE_MSG
 	bool "Access layer tx delayable message"
+	default y
 	help
 	  Enable following of the message transmitting recommendations, the Access layer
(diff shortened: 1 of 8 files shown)


alxelax (approved):

PavelVPV (changes requested):
Since you are pulling mesh changes from upstream, could you please also cherry-pick this commit https://github.com/zephyrproject-rtos/zephyr/commit/27b1f4eb7f9 so we don't miss it later when downstream other mesh stack changes? It is a stack change and goes between the last downstreamed mesh stack commit https://github.com/nrfconnect/sdk-zephyr/commit/9aa5cc1391f and the one you cherry-picked.

I also a bit concerned now because we skipped couple of babblesim tests, but this commit changes bsim tests. We may get conflict at some point if we keep skipping bsim tests.

PavelVPV (approved):

PR #1477, opened 2024-02-05
Pull nRF54L15 and nRF54H20 from upstream

gmarull:


dts/arm/nordic/nrf54l15_cpuapp_peripherals.dtsi
```diff
@@ -0,0 +1,412 @@
...
+};
+
+i2c20: i2c@c6000 {
+	compatible = "nordic,nrf-twim";
+	#address-cells = <1>;
+	#size-cells = <0>;
+	reg = <0xc6000 0x1000>;
+	clock-frequency = <I2C_BITRATE_STANDARD>;
+	interrupts = <198 NRF_DEFAULT_IRQ_PRIORITY>;
+	status = "disabled";
```
rakons, line 109: Missing: easydma-maxcnt-bits

rakons: The same for others twim compatible device declarations.

rakons (changes requested):

dts/arm/nordic/nrf54l15_cpuapp_peripherals.dtsi
masz-nordic: Why is this value in devicetree? It should be taken from MDK.
`TWIMxx_EASYDMA_MAXCNT_MAX` is the define to be used.

masz-nordic: MDK exposes per-instance defines. Looking at SHIM code, I think it would be enough to change this line:
https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/i2c/i2c_nrfx_twim.c#L422-L423
to `.max_transfer_size = BIT_MASK(TWIM##idx##_EASYDMA_MAXCNT_MAX)`.
Or am I missing something here?

rakons: There was a long discussion about this topic in the past. The decision taken was that we will move more and more data into DT. Personally - I believe it is better than defining it internally in the driver and taking it out by some macro magic. It gives you full view of the system on DTS level and supports your decision on the specific timer or other device taken to do the job. We have to live with the fact that device management in Zephyr is DTS centric.

gmarull: It is well known that using peripheral indices concatenation is a problem in the Zephyr model because standard instantiation just relies on DT instances. We're the only vendor left with such a custom scheme, but one day we'll need to migrate. Using DT means this tiny bit won't need any fix in the future.

masz-nordic: @rakons As far as I know, this decision was reversed until we have a solution that automatically generates devicetree from IPS. The main issue here is maintaining the DTS in sync with IPS, as this example shows.

gmarull: Added https://github.com/nrfconnect/sdk-zephyr/pull/1477/commits/97019d8f9ebb2062b7afa30d430ca54bcd9465ef

masz-nordic:
>I am not sure if this is part of this chips migration - but it seems that we are still missing here Flash driver implementation.

It's in separate PR here: https://github.com/zephyrproject-rtos/zephyr/pull/68309

gmarull:
@rakons ping

rakons (approved):

gmarull:
> The commits are signed with your nordicsemi.no email but the committer has teslabs.com email. Please either change the committer or key used to sign the commits.

thanks, fixed

gmarull:
> > The commits are signed with your nordicsemi.no email but the committer has teslabs.com email. Please either change the committer or key used to sign the commits.
> 
> This should've been caught by compliance, strange!

no, we don't check GPG signatures, just signed-off lines.

tmon-nordic:
> > The commits are signed with your nordicsemi.no email but the committer has teslabs.com email. Please either change the committer or key used to sign the commits.
> 
> This should've been caught by compliance, strange!

Compliance only checks `author`, `committer` and `Signed-off-by` lines and completely ignores `gpgsig` field.

tmon-nordic (approved):
`west build -d build-multicore-hello -b nrf54h20pdk_nrf54h20_cpuapp samples/multicore/hello_world/` built and run successfully on nRF54H20-PDK.

gmarull:
> THe last two fromlists can be replaced with fromtrees because the PRs are now merged upstream.

done

PR #1487, opened 2024-02-09
Fix print of ssid

kapbh:


jukkar (approved):

rlubos (approved):

SeppoTakalo (approved):

jfischer-no (approved):

subsys/net/l2/wifi/wifi_shell.c
```diff
@@ -116,26 +106,29 @@ static void handle_wifi_scan_result(struct net_mgmt_event_callback *cb)
 	const struct wifi_scan_result *entry =
 		(const struct wifi_scan_result *)cb->info;
 	uint8_t mac_string_buf[sizeof("xx:xx:xx:xx:xx:xx")];
+	const struct shell *sh = context.sh;
```
trantanen, line 109: `context.sh` can be `NULL` if this shell implementation is included into an application and wifi scan is triggered outside of the shell commands in this file. E.g., you trigger wifi scan through the code and not using these shell commands, this result handler would still be called causing an assert in shell because of the NULL shell pointer.

Before this PR, `printk` was used when shell pointer was NULL.

You should probably request shell uart from the shell backend.

PR #1503, opened 2024-02-16
cherry pick modem subsystem

tomi-font:


SeppoTakalo (approved):

drivers/gnss/Kconfig.generic
```diff
@@ -0,0 +1,26 @@
+# Copyright 2023 Google LLC
```
MarkusLassila, line 1: Are these GNSS changes somehow intertwined with the modem driver changes? I was not expecting to see these in here.

tomi-font: Yeah they are to some extent as I encountered some merge conflicts without taking in those GNSS changes.

MarkusLassila: I think that the modem driver stuff is likely OK, as we are the only user for many of those, however, I am a bit worried about these GNSS changes. Ping @trantanen.

MarkusLassila (approved):
OK from SLM perspective. We need the modem driver changes.

drivers/gnss/Kconfig.generic
MarkusLassila: Sorry, wrong Tommi. Ping @tokangas.

tokangas: The Zephyr GNSS API is already present in sdk-zephyr, but we don't use it currently. So, these commits should be fine.

rlubos (approved):

PR #1514, opened 2024-02-20
[nrf fromlist] dts: Add uicr nfct_pins_as_gpios prop on 5340_ns

Vge0rge:
Add a uicr node for the NRF5340 non-secure target. The UICR is mapped as secure only so the non-secure app doesn't have access to it but we still want to configure the nfct_pins_as_gpios property the same way for both secure and non-secure targets.

The uicr option in the non-secure target doesn't have the register property since it is not usable.

Ref: NCSDK-24437


mswarowsky (approved):

SebastianBoe (approved):

nordicjm (commented):
Revert needed

PR #1520, opened 2024-02-22
[nrf fromlist] net: ensure the shell sh is valid before call shell_pr…

kapbh:
…intf

It is possible that the `sh` was not set before use. This change adds a NULL check for `sh` in the following macros: PR, PR_SHELL, PR_ERROR, PR_INFO, and PR_WARNING.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/68812


rlubos (approved):
Please open a manifest PR.

krish2718 (approved):

krish2718:
> I have commented on the upstream PR as follows to try to make this work better at least for some backends: [zephyrproject-rtos/zephyr#68812 (comment)](https://github.com/zephyrproject-rtos/zephyr/pull/68812#issuecomment-1959329541)

We were waiting for some resolution for that PR before we pull it downstream, but couple of teams have already reported this crash, hence decided to do a `fromlist` to fix NCS and then during `upmerge` we can handle any changes with `fromtree`.

rlubos:
Is the crash that you're referring to, the same that was fixed upstream with https://github.com/zephyrproject-rtos/zephyr/commit/6b1b8157cd769b818293fca69cbc6108f1544296?

krish2718:
> Is the crash that you're referring to, the same that was fixed upstream with [zephyrproject-rtos/zephyr@6b1b815](https://github.com/zephyrproject-rtos/zephyr/commit/6b1b8157cd769b818293fca69cbc6108f1544296)?

Yes, I did see that, that will resolve the crash, but unless a shell command is issued and sets the context, all events would be dropped, so, we need this PR to fix the prints so that events are not dropped.

trantanen:
> > I have commented on the upstream PR as follows to try to make this work better at least for some backends: [zephyrproject-rtos/zephyr#68812 (comment)](https://github.com/zephyrproject-rtos/zephyr/pull/68812#issuecomment-1959329541)
> 
> We were waiting for some resolution for that PR before we pull it downstream, but couple of teams have already reported this crash, hence decided to do a `fromlist` to fix NCS and then during `upmerge` we can handle any changes with `fromtree`.

The problem is that without my suggestion, WiFi scan results are not shown in NCS 2.6.0 like they are in NCS 2.5.0 because there is no upmerge anymore before the release.
I agree we need some quick fix to unlock some nightly tests (our team's nightly is also red on a couple of tests because of this bug). So from my perspective, feel free to merge this PR as it is as quickly as possible. However, for NCS 2.6.0 I'd like to see a solution that doesn't have regression in terms of how much information is printed in our basic scenarios with UART/RTT shell backends.

trantanen:
> > However, for NCS 2.6.0 I'd like to see a solution that doesn't have regression in terms of how much information is printed in our basic scenarios with UART/RTT shell backends.
> 
> Agreed.

Now that the upmerge PR is merged, is there a plan to incorporate that into NCS 2.6.0?

PR #1527, opened 2024-02-23
[nrf fromlist] Bluetooth: Controller: Allow 0 ISO TX/RX buffer count

yuxCai:
Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/69382

There is no need to allocate TX/RX buffers when the device is an receiver/transmitter only.


cvinayak (changes requested):
Refer to reviews upstream

Thalley (changes requested):
See https://github.com/zephyrproject-rtos/zephyr/pull/69382#pullrequestreview-1898235364

PR #1528, opened 2024-02-23
[nrf fromlist] drivers: ieee802154: gracefully handle invalid Ack tim…

jciupis:
…estamp

The nRF IEEE 802.15.4 driver might report a received Ack frame with invalid timestamp, if the timestamp could not have been taken. The upper layers are not prepared to handle such a case as they expect that for a received frame, the timestamp is always present and valid.

This commit detects this situation and handles it gracefully by reporting the transmission as failed as if no Ack was received.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/69388


LuDuda (approved):
LGTM 👍

Damian-Nordic (approved):

PR #1535, opened 2024-02-26
Bring LTO support to NCS

kapi-no:
Cherry-pick attempt to bring the LTO feature from zephyr upstream to NCS

Base PRs:
https://github.com/zephyrproject-rtos/zephyr/pull/66392
https://github.com/zephyrproject-rtos/zephyr/pull/68530
+ Some additional fromtree commits added on top of the `zephyr/scripts/build/gen_isr_tables.py` script to avoid conflicts


rakons (approved):

LuDuda:
Testing it right now. But if one more rebase / CI run is needed anyway, i would consider to include above commits :) Else we may need to do it in a subsequent PR.

LuDuda (approved):
LGTM :+1:

Let's merge this. And follow up with Radek's addition as it allows also to build Matter samples.

PR #1536, opened 2024-02-26
[nrf fromtree] Bluetooth: Mesh: Enable CDP1 by default

PavelVPV:
CDP1 is mandatory to qualify a mesh device for mesh protocol v1.1. The dependency on BT_MESH_MODEL_EXTENSIONS is removed because if a node contains only base models (for example Generic OnOff Server), this API will consume extra RAM bytes which won't be used.

(cherry-picked from commit e30b7a84bc421dc16e87c47deab1921186604830)

### DIFF
--- a/subsys/bluetooth/mesh/Kconfig
+++ b/subsys/bluetooth/mesh/Kconfig
@@ -712,7 +712,7 @@ config BT_MESH_MODEL_EXTENSIONS
 
 config BT_MESH_COMP_PAGE_1
 	bool "Support for Composition Data Page 1"
-	depends on BT_MESH_MODEL_EXTENSIONS
+	default y
 	help
 	  Enable support for Composition Data Page 1.
 
@@ -720,6 +720,7 @@ config BT_MESH_MODEL_EXTENSION_LIST_SIZE
 	int "Model extensions list size"
 	depends on BT_MESH_COMP_PAGE_1
 	range 0 255
+	default 0 if !BT_MESH_MODEL_EXTENSIONS
 	default 10
 	help
 	  This option specifies how many models relations can be saved.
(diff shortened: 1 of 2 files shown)


alxelax (approved):

subsys/bluetooth/mesh/Kconfig
```diff
@@ -712,14 +712,15 @@ config BT_MESH_MODEL_EXTENSIONS
...
 config BT_MESH_COMP_PAGE_1
 	bool "Support for Composition Data Page 1"
-	depends on BT_MESH_MODEL_EXTENSIONS
+	default y
 	help
 	  Enable support for Composition Data Page 1.
 
 config BT_MESH_MODEL_EXTENSION_LIST_SIZE
 	int "Model extensions list size"
 	depends on BT_MESH_COMP_PAGE_1
```
Balaklaka, line 721: Should this not rather be dependent on BT_MESH_MODEL_EXTENSIONS, then the default does not need to be 0.

jfischer-no (approved):

subsys/bluetooth/mesh/Kconfig
PavelVPV: No, there are 2 API: `bt_mesh_model_extend` and `bt_mesh_model_correspond`. The first one depends on `CONFIG_BT_MESH_MODEL_EXTENSIONS`, the second one does not. A user can still have the list size > 0 and use only `bt_mesh_model_correspond` API keeping `CONFIG_BT_MESH_MODEL_EXTENSIONS` disabled. `CONFIG_BT_MESH_MODEL_EXTENSIONS` allocates extra RAM in model structs for storing model extension tree, which is not needed when only corresponding relationship is used between models.

Balaklaka (approved):

PR #366, opened 2020-10-08
[nrf toup] shell: increase stack size for OpenThread

edmont:
Default shell stack size is not always enough when used with OpenThread. Increasing it to the found adequate value.

Signed-off-by: Eduardo Montoya <eduardo.montoya@nordicsemi.no>

PR #1560, opened 2024-03-05
[nrf fromlist] kconfig: shell: increase shell stack size for OpenThread

maciejbaczmanski:
Shell stack size is too low for OpenThread without joiner functionality, causing overflow. This commit increases it.

upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/69783

Signed-off-by: Maciej Baczmanski <maciej.baczmanski@nordicsemi.no>
(cherry picked from commit 4fb493e49fef245e62df21620382e027a7200907)

### DIFF
--- a/subsys/shell/Kconfig
+++ b/subsys/shell/Kconfig
@@ -40,9 +40,8 @@ endif
 
 config SHELL_STACK_SIZE
 	int "Shell thread stack size"
-	default 3168 if OPENTHREAD_SHELL && OPENTHREAD_JOINER
+	default 3168 if OPENTHREAD_SHELL
 	default 3072 if 64BIT
-	default 2616 if OPENTHREAD_SHELL
 	default 2048 if MULTITHREADING
 	default 0 if !MULTITHREADING
 	help


canisLupus1313 (approved):

edmont (approved):

jfischer-no (changes requested):
Does it fix stack size for NCS or native Zephyr?

maciejbaczmanski:
> Does it fix stack size for NCS or native Zephyr?

For NCS. Currently shell stack size is too small for some variants causing overflow on openthread samples

LuDuda (approved):

maciejbaczmanski:
> > > Does it fix stack size for NCS or native Zephyr?
> > 
> > 
> > For NCS. Currently shell stack size is too small for some variants causing overflow on openthread samples
> 
> Then it is better to resolve it in NCS not in Zephyr, maybe like other application and samples in NCS do it. How many releases will the stack be sufficient and we see another fromlist patch?

This is a bugfix. As it can be seen in changed file, value of stack size is not changed. Incorrect assigning of value for opentrhead with no joiner functionality has been removed. PR has been approved upstream. Resolving it in NCS would only cause doubling of configs with the same value.

jfischer-no:
Your commit message: "Shell stack size is too low for OpenThread without joiner
functionality, causing overflow. This commit increases it."

>As it can be seen in changed file, value of stack size is not changed. Incorrect assigning of value for opentrhead with no joiner functionality has been removed. 

Is it just me who sees two different statements?

> Resolving it in NCS would only cause doubling of configs with the same value.

But you answer to my question "Does it fix stack size for NCS or native Zephyr?" is "For NCS.". According to you answer this value does not applies to native Zephyr RTOS but to NCS only, therefore is should be fixed in NCS.

Also changed here: https://github.com/nrfconnect/sdk-zephyr/pull/366/commits/4b94355440861c1973a6f976cc46760d4bfc77db

carlescufi:
> > > > Does it fix stack size for NCS or native Zephyr?
> > > 
> > > 
> > > For NCS. Currently shell stack size is too small for some variants causing overflow on openthread samples
> > 
> > 
> > Then it is better to resolve it in NCS not in Zephyr, maybe like other application and samples in NCS do it. How many releases will the stack be sufficient and we see another fromlist patch?
> 
> This is a bugfix. As it can be seen in changed file, value of stack size is not changed. Incorrect assigning of value for opentrhead with no joiner functionality has been removed. PR has been approved upstream. Resolving it in NCS would only cause doubling of configs with the same value.

I discussed with @maciejbaczmanski and we agreed that this does apply upstream. The commit message will be amended.

SeppoTakalo (approved):

jfischer-no:
> I discussed with @maciejbaczmanski and we agreed that this does apply upstream. The commit message will be amended.

okay. I will merged it asap depending on rebase process.

PR #1563, opened 2024-03-06
[nrf fromtree] manifest: remove .git suffixes from repo-paths

gmarull:
Some entries had the .git suffix in the repo-path property. While it does not affect e.g. clones, it makes the project URL (built using base url and repo-path) unusable to build other URLs (e.g. to point to a commit, release tag, etc.).

Signed-off-by: Gerard Marull-Paretas <gerard@teslabs.com>
(cherry picked from commit 16a60630830d19c0be6c3118c9dd3f33d03d0ac7)

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -40,77 +40,77 @@ manifest:
         - babblesim
     - name: babblesim_base
       remote: babblesim
-      repo-path: base.git
+      repo-path: base
       path: tools/bsim/components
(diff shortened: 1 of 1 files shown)


aescolar (approved):
Ok, but is the signoff a mistake?

gmarull:
> Ok, but is the signoff a mistake?

nope, I'm a contractor

PR #1576, opened 2024-03-14
[nrf fromlist] scripts: runners: nrf: Check device family

maxd-nordic:
This patch makes the nrfjprog runner check the device family before flashing. Otherwise, it is very easy to break some devices that have multiple different chips.

(cherry picked from commit e0aa35d74a5aaa6b2fa9f684385c8e8fcda2ad30)
Signed-off-by: Maximilian Deubel <maximilian.deubel@nordicsemi.no>


maxd-nordic:
We should have this in place before the new Thingy is shipped since it's mostly relevant for that device.

PR #1582, opened 2024-03-19
[nrf noup] DNM: drivers: misc: add temporary SPU driver

gmarull:
Add a temporary hack to handle SPU related stuff until this gets migrated properly upstream.

**NOTE**: Please, point west.yml to this PR if you need SPU, this should unblock some BLE-related tasks while the driver gets ported upstream.


pdunaj:
@KAGA164 , please have a look - I expect it was you or @maje-emb that worked on SPU configuration in next

gmarull:
Alternative merged upstream, will be brought by the upmerge

PR #1571, opened 2024-03-08
nRF54L15: Add initial TF-M support

Vge0rge:

PR #1583, opened 2024-03-19
Misc. TF-M+54L support

SebastianBoe:


SebastianBoe:
I intend to make the toup's into fromlist when I have a green CI and review.

drivers/flash/soc_flash_nrf_rram.c
```diff
@@ -103,11 +107,20 @@ static void commit_changes(size_t len)
...
 
-static void rram_write(off_t addr, const void *data, size_t len)
+static int rram_write(off_t addr, const void *data, size_t len)
 {
 	nrf_rramc_config_t config = {.mode_write = true, .write_buff_size = WRITE_BUFFER_SIZE};
+	int err = 0;
 
+#ifdef CONFIG_BUILD_WITH_TFM
+	err = tfm_platform_nrf_rramc_config_set(config.mode_write, config.write_buff_size);
+	if(err) {
```
de-nordic, line 117: Do you want the #ifdef CONFIG_BUILD_WITH_TFM in code to make clear the intention of the call tfm_platform_nrf_rramc_config_set here or would it be possible to provide static inline wrapper for the function that would evaluate to nothing when !CONFIG_BUILD_WITH_TFM ?

SebastianBoe: I feel it would be misleading to call a function called

tfm_platform_nrf_rramc_config_set

that in reality does nothing.

But I am open for refactoring.

Do you know who in upstream Zephyr would review this patch?

I would like to write it in a way that can be upstreamed without feedback.

de-nordic: Me for RRAM, probably Andrzej too.

SebastianBoe: OK, I'll ask you two for a review after I get a green CI :)

Then hopefully I won't have to do too much paperwork updatign the fromlist many times.

mswarowsky (commented):
fine for me, I guess the toups are getting fromlist once #1571 is in

SebastianBoe:
reorganizing

PR #1588, opened 2024-03-25
Upmerge 25-03-2024

gmarull:


boards/nordic/nrf54l15pdk/nrf54l15pdk_nrf54l15_cpuapp.dts
```diff
@@ -173,3 +174,29 @@
...
+&spi00 {
+	status = "okay";
+	cs-gpios = <&gpio2 5 GPIO_ACTIVE_LOW>;
+	pinctrl-0 = <&spi0_default>;
+	pinctrl-1 = <&spi0_sleep>;
+	pinctrl-names = "default", "sleep";
+
+	mx25r64: mx25r6435f@0 {
+		compatible = "jedec,spi-nor";
+		status = "okay";
```
tejlmand: why is this `status = "okay";` when corresponding commit here https://github.com/nrfconnect/sdk-zephyr/commit/113f05cbc93fb7775ab5512fd56e718d908a03cf is: `status = "disabled";` ?
@gmarull

gmarull: Because https://github.com/nrfconnect/sdk-zephyr/commit/113f05cbc93fb7775ab5512fd56e718d908a03cf is yet another evil merge. fromtree/list that need to be reverted (because of conflicts), are reapplied from the upstream PR (where eg tuned HWMv2 changes are not present). Upstream has status "okay".

nvlsianpu: I have updated upstream PR (so it has status = "disabled" now).

gmarull: It looks like it can't be applied cleanly, because it depends on [aa1690f8c6a3b8cfcf548103830acac41228bb2d](https://github.com/zephyrproject-rtos/zephyr/commit/aa1690) (so it was likely tuned further to skip conflicts).

shanthanordic (approved):
approved with agreed scope and mitigation with failures

carlescufi:
Verified that the failing checks are not related to downstream changes.

PerMac:
> Verified that the failing checks are not related to downstream changes.
> Had to disable the clang check due to new incompatibilities @PerMac.

I believe we already had it disabled for a while and I think it's a second time when it auto-enabled itself during upmerge .

PR #1589, opened 2024-03-25
[nrf fromtree] Samples: Bluetooth: Fix PAwR sample failed to set subevent data

ryanjh:
Be respectful of PAwR subevents while scheduling scan activities. The radio will be swtiched from scan to PAwR when it is closed to the next subevent interval.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/70678

### DIFF
--- a/samples/bluetooth/periodic_adv_rsp/src/main.c
+++ b/samples/bluetooth/periodic_adv_rsp/src/main.c
@@ -295,7 +295,8 @@ int main(void)
 	}
 
 	while (num_synced < MAX_SYNCS) {
-		err = bt_le_scan_start(BT_LE_SCAN_PASSIVE, device_found);
+		/* Enable continuous scanning */
+		err = bt_le_scan_start(BT_LE_SCAN_PASSIVE_CONTINUOUS, device_found);
 		if (err) {
 			printk("Scanning failed to start (err %d)\n", err);
 			return 0;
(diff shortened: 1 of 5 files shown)


samples/bluetooth/periodic_adv_rsp/src/main.c
```diff
@@ -295,7 +295,14 @@ int main(void)
 	}
 
 	while (num_synced < MAX_SYNCS) {
-		err = bt_le_scan_start(BT_LE_SCAN_PASSIVE, device_found);
+		struct bt_le_scan_param scan_param = {
+			.type = BT_LE_SCAN_TYPE_PASSIVE,
+			.options = BT_LE_SCAN_OPT_FILTER_DUPLICATE,
+			.interval = per_adv_params.interval_min,
+			.window = per_adv_params.interval_min,
```
Thalley, line 302: Scan interval and window are units of 0.625 ms, whereas the periodic advertising interval are in units of 1.25ms. 

Is this intentional? If so, please add a comment to explain why we are setting these to the half the value of the PA interval.

ryanjh: thanks for pointing it out. Those values should be double.

ryanjh: reused the macros for contiuous scanner. https://github.com/zephyrproject-rtos/zephyr/pull/70044

rugeGerritsen (approved):

hermabe (approved):

PR #1591, opened 2024-03-26
Add dwell time range check, Add channel validity check

kapbh:


subsys/net/l2/wifi/wifi_shell.c
```diff
@@ -764,7 +769,7 @@ static int wifi_scan_args_to_params(const struct shell *sh,
 static int cmd_wifi_scan(const struct shell *sh, size_t argc, char *argv[])
 {
 	struct net_if *iface = net_if_get_first_wifi();
-	struct wifi_scan_params params = { 0 };
+	struct wifi_scan_params params = { .dwell_time_active = 50, .dwell_time_passive = 130, };
```
sachinthegreen, line 772: Can we please create defines for `50` and `130` in maybe `https://github.com/nrfconnect/sdk-zephyr/blob/main/include/zephyr/net/wifi_mgmt.h` instead of using magic numbers ?

sachinthegreen (changes requested):

jukkar (approved):

sachinthegreen (approved):

PR #1595, opened 2024-03-28
[nrf noup] dfu/boot/mcuboot fix img. confirmation in case of PARTITION_MANAGER

nvlsianpu:
Active partition ID need to be extracted basing on PARTITION_MANAGER products.

ref.:NCSDK-26693

This is alternative patch to https://github.com/nrfconnect/sdk-nrf/pull/14633
similar solution spotted there: https://github.com/nrfconnect/sdk-nrf/blob/main/applications/nrf_desktop/src/modules/dfu.c#L81

### DIFF
--- a/subsys/dfu/boot/mcuboot.c
+++ b/subsys/dfu/boot/mcuboot.c
@@ -34,8 +34,34 @@
 #define BOOT_HEADER_MAGIC_V1 0x96f3b83d
 #define BOOT_HEADER_SIZE_V1 32
 
+#if USE_PARTITION_MANAGER
+#include <pm_config.h>
+
+#if CONFIG_BUILD_WITH_TFM
+	#define PM_ADDRESS_OFFSET (PM_MCUBOOT_PAD_SIZE + PM_TFM_SIZE)
+#else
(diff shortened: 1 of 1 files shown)


de-nordic (approved):
Cumbersome but simple and completely addressing the issue.

nordicjm (approved):

subsys/dfu/boot/mcuboot.c
```diff
@@ -34,8 +34,26 @@
 #define BOOT_HEADER_MAGIC_V1 0x96f3b83d
 #define BOOT_HEADER_SIZE_V1 32
 
+#if USE_PARTITION_MANAGER
+
```
MarekPieta, line 38: I think we may also need this one: `#include <pm_config.h>`

MarekPieta (approved):

sigvartmh:
I think this is probably the better way going about it. I'm just a bit uncertain how we reach for the active slot but I'll investigate

subsys/dfu/boot/mcuboot.c
nvlsianpu: but yes, will apply your request

de-nordic (approved):

PR #1601, opened 2024-04-04
[nrf fromtree] net: lib: http: Correct http_client edge-case behavior

glarsennordic:
Correct various small edge-case behaviors that have been accidentally introduced in the http_client.

- http_client_req no longer incorrectly returns -ETIMEDOUT on NULL HTTP resonse. -ETIMEDOUT is now only returned when the underlying TLS socket times out.
- http_client_req now returns -ECONRESET upon incomplete (but non-NULL) HTTP response. The request callback is no longer called in this case (as with any other error state).
- http_wait_data has been refactored slightly to increase clarity.

Signed-off-by: Georges Oates_Larsen <georges.larsen@nordicsemi.no>

(cherry picked from commit https://github.com/zephyrproject-rtos/zephyr/pull/70911/commits/30c7d0920c9b03dd15f9ac1913738c290c003275)

### DIFF
--- a/subsys/net/lib/http/http_client.c
+++ b/subsys/net/lib/http/http_client.c
@@ -392,7 +392,11 @@ static void http_client_init_parser(struct http_parser *parser,
 	settings->on_url = on_url;
 }
 
-static void http_data_final_null_resp(struct http_request *req)
+/* Report a NULL HTTP response to the caller.
+ * A NULL response is when the HTTP server intentionally closes the TLS socket (using FINACK)
+ * without sending any HTTP payload.
+ */
+static void http_report_null(struct http_request *req)
 {
 	if (req->internal.response.cb) {
 		NET_DBG("Calling callback for Final Data"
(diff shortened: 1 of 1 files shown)


glarsennordic:
@gmarull Is it possible to get this included in the ongoing upmerge? This is a bugfix for IRIS-8638

plskeggs (approved):

subsys/net/lib/http/http_client.c
```diff
@@ -438,24 +464,28 @@ static int http_wait_data(int sock, struct http_request *req, int32_t timeout)
...
 		} else if (ret < 0) {
+			ret = -errno;
 			goto error;
 		}
 		if (fds[0].revents & (ZSOCK_POLLERR | ZSOCK_POLLNVAL)) {
+			ret = -errno;
 			goto error;
 		} else if (fds[0].revents & ZSOCK_POLLHUP) {
 			/* Connection closed */
 			LOG_DBG("Connection closed");
```
jayteemo, line 478: can this and the one below be moved to the `closed:` label?

jayteemo (approved):

subsys/net/lib/http/http_client.c
glarsennordic: Hypothetically, yes, and I started to do that, but I ultimately thought the code was a little easier to read this way.

I don't care a ton either way, but if you also don't care too much I will leave as-is to speed up review/merge

rlubos:
We'd need a manifest PR for this (to verify CI).

tony-le-24 (approved):

glarsennordic:
Ah! Right. Thank you. I'll make that happen

jukkar (approved):

tony-le-24:
confirmed that it works on my CI, built and tested locally, the issue was gone.

glarsennordic:
@rlubos looks like CI is green on the manifest PR after latest rebase

PR #1607, opened 2024-04-09
[nrf noup] entropy: Add fake entropy nRF PRNG driver needed to run Radio Test and BLE on nRF54h20

KAGA164:
This adds temporary entropy driver simulation for
nRF54h20 device since final entropy source is not
available yet. This driver is needed to unblock further development on nRF54H20.

### DIFF
--- a/drivers/entropy/CMakeLists.txt
+++ b/drivers/entropy/CMakeLists.txt
@@ -34,6 +34,13 @@ zephyr_library_sources_ifdef(CONFIG_ENTROPY_BT_HCI             entropy_bt_hci.c)
 zephyr_library_sources_ifdef(CONFIG_ENTROPY_GECKO_SE           entropy_gecko_se.c)
 zephyr_library_sources_ifdef(CONFIG_ENTROPY_PSA_CRYPTO_RNG     entropy_psa_crypto.c)
 
+if (CONFIG_FAKE_ENTROPY_NRF_PRNG)
+  zephyr_library_sources(fake_entropy_nrf_prng.c)
+
+  message(WARNING "\nA nRF PRNG is used, which does not produce real random bits."
+  "This is not secure and should therefore never be used in a product.")
+endif()
+
 if (CONFIG_BUILD_WITH_TFM)
   target_include_directories(${ZEPHYR_CURRENT_LIBRARY} PRIVATE
     $<TARGET_PROPERTY:tfm,TFM_BINARY_DIR>/api_ns/interface/include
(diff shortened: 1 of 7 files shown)


drivers/entropy/CMakeLists.txt
```diff
@@ -34,6 +34,13 @@ zephyr_library_sources_ifdef(CONFIG_ENTROPY_BT_HCI             entropy_bt_hci.c)
 zephyr_library_sources_ifdef(CONFIG_ENTROPY_GECKO_SE           entropy_gecko_se.c)
 zephyr_library_sources_ifdef(CONFIG_ENTROPY_PSA_CRYPTO_RNG     entropy_psa_crypto.c)
 
+if (CONFIG_FAKE_ENTROPY_NRF_PRNG)
+  zephyr_library_sources(fake_entropy_nrf_prng.c)
+
+  message(WARNING "\nA nRF PRNG is used, which does not produce real random bits."
```
seko-nordic, line 40: not sure if it's a good idea to expose this publicly, even with a warning. Can we put it to some private test repo instead?

maje-emb (approved):

SebastianBoe (changes requested):
This could cause customers to end up with fake entropy in their nrf53/91, etc. products.

This risk needs to mitigated somehow.

pdunaj:
> This could cause customers to end up with fake entropy in their nrf53/91, etc. products.
> 
> This risk needs to mitigated somehow.

This will be mid release change that must be gone by 2.7. Nobody should pick it up for production.
Additionally this can be mitigated by adding depend on nrf54x. @KAGA164, can you add that depend on just in case.

drivers/entropy/CMakeLists.txt
pdunaj: let's move this discussion outside of PR

SebastianBoe:
"Nobody should pick it up for production."
There is nothing stopping a customer from using the NCS master branch for production.

"Additionally this can be mitigated by adding depend on nrf54x. @KAGA164, can you add that depend on just in case."

Yes, something trivial like this is what I'm looking for.

KAGA164:
> "Nobody should pick it up for production." There is nothing stopping a customer from using the NCS master branch for production.
> 
> "Additionally this can be mitigated by adding depend on nrf54x. @KAGA164, can you add that depend on just in case."
> 
> Yes, something trivial like this is what I'm looking for.

@SebastianBoe @pdunaj I added additional dependency for nRF54Hx series. Anyway customer would need to modify or overlay DTS configuration to use this implementation so it hard to use this driver accidentally

SebastianBoe (approved):

PR #1615, opened 2024-04-10
fixup! [nrf noup] tree-wide: support NCS Partition Manager (PM) defin…

de-nordic:
…itions

Allow SB_CONFIG_PARTITION_MANAGER to control inclusion of parition_manager.cmake.

### DIFF
--- a/cmake/modules/kernel.cmake
+++ b/cmake/modules/kernel.cmake
@@ -245,5 +245,7 @@ if("${CMAKE_EXTRA_GENERATOR}" STREQUAL "Eclipse CDT4")
 endif()
 
 if(ZEPHYR_NRF_MODULE_DIR)
-  include(${ZEPHYR_NRF_MODULE_DIR}/cmake/partition_manager.cmake)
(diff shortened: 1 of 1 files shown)


de-nordic:
Yeah, breaks everything that does not use sysbuild

cmake/modules/kernel.cmake
```diff
@@ -244,6 +244,6 @@ if("${CMAKE_EXTRA_GENERATOR}" STREQUAL "Eclipse CDT4")
   eclipse_cdt4_generator_amendment(1)
 endif()
 
-if(ZEPHYR_NRF_MODULE_DIR)
+if(ZEPHYR_NRF_MODULE_DIR AND (CONFIG_PARTITION_MANAGER OR SB_CONFIG_PARTITION_MANAGER))
   include(${ZEPHYR_NRF_MODULE_DIR}/cmake/partition_manager.cmake)
```
nordicjm, line 248: Not the right path for sysbuild

de-nordic:
Not the right aproach.

PR #1617, opened 2024-04-11
[nrf fromlist] boards: nrf54l15pdk_nrf54l15_cpuapp: yaml for PDK 0.3.0

barsok:
add twister yaml to support PDK 0.3.0 in tests

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/71375

### DIFF
--- a/boards/arm/nrf54l15pdk_nrf54l15/nrf54l15pdk_nrf54l15_cpuapp_0_3_0.yaml
+++ b/boards/arm/nrf54l15pdk_nrf54l15/nrf54l15pdk_nrf54l15_cpuapp_0_3_0.yaml
@@ -1,0 +1,20 @@
+# Copyright (c) 2024 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+identifier: nrf54l15pdk_nrf54l15_cpuapp@0.3.0
+name: nRF54l15-PDK-nRF54l15-Application
+type: mcu
+arch: arm
+toolchain:
+  - gnuarmemb
+  - xtools
+  - zephyr
+ram: 256
+flash: 1536
+supported:
+  - counter
+  - gpio
+  - i2c
+  - spi
+  - watchdog
+  - i2s


boards/arm/nrf54l15pdk_nrf54l15/nrf54l15pdk_nrf54l15_cpuapp_0_3_0.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2024 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+identifier: nrf54l15pdk_nrf54l15_cpuapp@0.3.0
```
MarekPieta, line 4: Since we need to use another approach for `identifier` (comparing to upstream Zephyr), maybe could explicitly inform about it in the commit message or make this commit `[nrf noup]`?

MarekPieta (approved):

kapi-no:
@barsok, could you open manifest bump PR in sdk-nrf or leverage your old one? Your PR can't be merged here unless the sdk-nrf PR counterpart passes CI.

 We would like very much to have this change in NCS as soon as possible.

FYI, your Zephyr upstream PR is merged.

boards/arm/nrf54l15pdk_nrf54l15/nrf54l15pdk_nrf54l15_cpuapp_0_3_0.yaml
barsok: The new format is in the equivalent PR upstream. I hope that this PR can be replaced with the one from the upstream, hence [nrf fromlist]. I hope it's ok

anangl (approved):

boards/arm/nrf54l15pdk_nrf54l15/nrf54l15pdk_nrf54l15_cpuapp_0_3_0.yaml
ArekBalysNordic: I'm curious and worried, why do we want to add @0.3.0 instead of using `nrf54l15pdk_nrf54l15_cpuapp` for v0.3.0 and set `nrf54l15pdk_nrf54l15_cpuapp@0.2.x`? I think the 0.3.0 version is officially dedicated for future usage and 0.2.x will be deprecated soon. For example, in Matter samples, it is very demanding to maintain both versions because this is a completely different build target. After all, it is additionally at least 4 files more per sample, and we have 5 samples on nrf54L...

barsok: Our test setups are configured for using PDKs in revision 0.2 and it is decided that this customer sampling release will be based on NCS 2.6 and PDK 0.2.  PDK 0.3.0 is supported as well by adding "@0.3.0" in west commands. NCS 2.7 will use PDK 0.7.0 by default (https://nordicsemi.atlassian.net/wiki/spaces/NCS/pages/657981900/NCS+2.7.0+HW+DK+Versions)

ArekBalysNordic: Are we going to remove those configurations (@0.3.0) after customer sampling at all? So it is needed only for the release, right?

barsok: 0.2 most likely will be removed, 0.3 will stay as most sampled customers have these boards. I am not sure what will be the default, possibly an even newer board (need to follow up)

ArekBalysNordic: And this is the only way to deal with it? Can we, for example, create a Kconfig for that instead of creating a different build type? And why does this target not have PWM support? It is required by our two samples.

gmarull:
commit diff only due to hwmv2, so ok

PR #1619, opened 2024-04-11
[nrf noup] tests: Allow UARTE new tests to run with HW model v…

nordic-bami:
…1 board names

NRFX-5524: Verify coverage for UARTE - extend testing


nordic-piks (approved):

gmarull (changes requested):
not sure why we have hwmv2 names in here, please, revert the original commit.

nordic-segl (approved):

gmarull (changes requested):
please, revert original commit, then cherry-pick by adjusting the commits from upstream instead.

PR #1621, opened 2024-04-11
Revert "[nrf fromlist] tests: Extend tests coverage for UARTE driver"

nordic-bami:
Needs to be reverted and re-applied from upstream with alginments
Signed-off-by: Bartosz Miller <bartosz.miller@nordicsemi.no>


nordic-segl (approved):

rlubos (changes requested):
Please use a standard revert syntax (git defautl message), otherwise this commit will mess up our tools. Here's an example of proper revert:
https://github.com/nrfconnect/sdk-zephyr/commit/5fc568680ed76e96c8c8674704e4e1a4617bcb49

nordic-piks (approved):

PR #1625, opened 2024-04-11
[nrf fromlist] drivers: nrf: rram: Support TF-M

SebastianBoe:
fromlist: https://github.com/zephyrproject-rtos/zephyr/pull/71825

Non-secure images cannot reference NRF_RRAMC_NS because NRF_RRAMC_NS
does not exist.

TF-M will configure RRAMC according to these Kconfig's before booting
the non-secure image so we ifdef out this code.

Also, rewrite the implementation of commit_changes to also work when
the commit task is not available.

### DIFF
--- a/drivers/flash/soc_flash_nrf_rram.c
+++ b/drivers/flash/soc_flash_nrf_rram.c
@@ -17,6 +17,26 @@
 
 #include <zephyr/../../drivers/flash/soc_flash_nrf.h>
 
+/* Note that it is supported to compile this driver for both secure
+ * and non-secure images, but non-secure images cannot call
+ * nrf_rramc_config_set because NRF_RRAMC_NS does not exist.
+ *
+ * Instead, when TF-M boots, it will configure RRAMC with this static
+ * configuration:
+ *
+ * nrf_rramc_config_t config = {
+ *   .mode_write = true,
+ *   .write_buff_size = WRITE_BUFFER_SIZE
+ * };
+ *
+ * nrf_rramc_ready_next_timeout_t params = {
+ *   .value = CONFIG_NRF_RRAM_READYNEXT_TIMEOUT_VALUE,
+ *   .enable = true,
+ * };
+ *
+ * For more details see NCSDK-26982.
+ */
+
 LOG_MODULE_REGISTER(flash_nrf_rram, CONFIG_FLASH_LOG_LEVEL);
 
 #define RRAM DT_INST(0, soc_nv_flash)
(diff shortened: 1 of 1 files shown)


drivers/flash/soc_flash_nrf_rram.c
```diff
@@ -58,6 +58,30 @@ BUILD_ASSERT((PAGE_SIZE % (WRITE_LINE_SIZE) == 0),
...
+ *   .write_buff_size = WRITE_BUFFER_SIZE
+ * };
+ *
+ * nrf_rramc_ready_next_timeout_t params = {
+ *   .value = CONFIG_NRF_RRAM_READYNEXT_TIMEOUT_VALUE,
+ *   .enable = true,
+ * };
+ *
+ * For more details see NCSDK-26982.
+ */
```
de-nordic, line 79: No conditional comments, looks strange, like leftover. Just do comment above the `#if !defined()`, should be enough

drivers/flash/soc_flash_nrf_rram.c
```diff
@@ -85,29 +109,54 @@ static inline bool is_within_bounds(off_t addr, size_t len, off_t boundary_start
 }
 
 #if WRITE_BUFFER_ENABLE
-static void commit_changes(size_t len)
+static void commit_changes(off_t addr, size_t len)
 {
+#ifdef NRF_RRAMC
```
de-nordic, line 114: Where is this one coming from?

drivers/flash/soc_flash_nrf_rram.c
SebastianBoe: OK

vili-nordic (approved):

drivers/flash/soc_flash_nrf_rram.c
```diff
@@ -85,29 +105,53 @@ static inline bool is_within_bounds(off_t addr, size_t len, off_t boundary_start
...
 	}
+#endif
 
 	if ((len % (WRITE_BUFFER_MAX_SIZE)) == 0) {
 		/* Our last operation was buffer size-aligned, so we're done. */
 		return;
 	}
 
+#if !defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
 	nrf_rramc_task_trigger(NRF_RRAMC, NRF_RRAMC_TASK_COMMIT_WRITEBUF);
```
nvlsianpu, line 123: ARG_UNUSED(addr);

nvlsianpu (approved):

de-nordic (approved):

drivers/flash/soc_flash_nrf_rram.c
SebastianBoe: will fix :)

jaz1-nordic (approved):

mswarowsky (approved):

PR #1645, opened 2024-04-16
[nrf fromlist] wifi: shell: Support WPA auto personal security mode

rado17:
Applications need to run a scan to identify the security mode before attempting the connection which adds to the time taken for connection to be established.
To avoid the initial scan, support auto security mode which will enable STA to choose between WPA, WPA2 and WPA3, based on the network configuration.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/71044


include/zephyr/net/wifi.h
```diff
@@ -49,6 +49,8 @@ enum wifi_security_type {
 	WIFI_SECURITY_TYPE_WEP,
 	/** WPA-PSK security. */
 	WIFI_SECURITY_TYPE_WPA_PSK,
+	/** WPA/WPA2/WPA3 PSK security. */
+	WIFI_SECURITY_TYPE_WPA_AUTO,
```
krish2718, line 53: WPA_AUTO_PERSONAL might limit to personal security.

rado17: Done

krish2718 (approved):

sachinthegreen (approved):

PR #1655, opened 2024-04-19
[nrf noup] samples: psa_crypto: Remove support for Nordic boards

mswarowsky:
We have our own psa crypto samples to show how to used PSA crypto with NCS.

This sample still uses CONFIG_MBEDTLS_BUILTIN which is not supported anymore, therefore removing the support for it in NCS.

Ref: NCSDK-17944


vili-nordic (approved):

SebastianBoe (approved):

mswarowsky:
No Manifest PR as it just disables a sample and it not so important, that it an be pulled in by another manifest PR

PR #1663, opened 2024-04-24
[nrf fromtree] boards: arm: mark counter as supported for nRF54L15

nordic-krch:
Mark counter as supported for nRF54L15 to allow running counter_basic_api by twister.

Signed-off-by: Magdalena Pastula <magdalena.pastula@nordicsemi.no>
(cherry picked from commit 76f990730f629b011b7c1c34d706fa87bd0585d3)


anangl (changes requested):
The commit is different from the upstream one (a different file is modified), so it should be taken as `[nrf fromlist]`.

nordicjm (approved):

PR #1665, opened 2024-04-24
[nrf fromtree] net: arp: Add support for gratuitous ARP transmission

rlubos:
Add support for gratuitous ARP transmission by Zephyr network stack. This allows to prematurely fill the peer ARP table, so there's no need to send an explicit request when peer needs to send an actual packet.

The gratuitous ARP is send when the network interface is brought up (joins the network) or a new IP address is added. The gratuitous ARP request is also sent periodically, with a configurable interval time. The gratuitous ARP should also be sent whenever MAC address of the interface is changed, but as Zephyr only allows to do this when interface is down, this is already covered by the first case (interface brought up).

### DIFF
--- a/include/zephyr/net/ethernet.h
+++ b/include/zephyr/net/ethernet.h
@@ -971,6 +971,24 @@ static inline bool net_eth_get_vlan_status(struct net_if *iface)
 }
 #endif
 
+/**
+ * @brief Check if the given interface is a VLAN interface.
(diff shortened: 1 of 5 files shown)


rlubos:
Extra cherry-picks to get `net_eth_is_vlan_interface()` cleanly.

jukkar (approved):

PR #1666, opened 2024-04-24
Introduce pm config

adamkondraciuk:


soc/arm/nordic_nrf/nrf54h/poweroff.c
```diff
@@ -0,0 +1,179 @@
...
+
+	/* Re-enable domain retention. */
+	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, true);
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, true);
+
+	z_pm_soc_resume();
+}
+
+/* Function called during local domain suspend to RAM. */
+int z_pm_sys_suspend(void)
```
hubertmis, line 96: I would rename this and similar functions to clearly indicate they are about `suspend_to_ram` not about suspend to idle or suspend to disk.

arch/arm/core/cortex_m/pm_s2ram.c
hubertmis, line 1: This looks like a generic Zepyr code. I would consider upstreaming. @gmarull , FYI

hubertmis (changes requested):

dts/bindings/mmu_mpu/arm,armv6m-mpu.yaml
```diff
@@ -10,3 +10,8 @@ include: base.yaml
 properties:
   reg:
     required: true
+
+  arm,num-mpu-regions:
```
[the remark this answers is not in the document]
gmarull: what we should avoid is to disalign with upstream, all this code should be developed upstream-first.

soc/arm/nordic_nrf/nrf54h/poweroff.c
```diff
@@ -0,0 +1,179 @@
...
+#define RAMBLOCK_POWER_ID	    0
+#define RAMBLOCK_CONTROL_OFF	    0
+#elif defined(NRF_RADIOCORE)
+#define RAMBLOCK_CONTROL_BIT_ICACHE 2
+#define RAMBLOCK_CONTROL_BIT_DCACHE 3
+#define RAMBLOCK_POWER_ID	    0
+#define RAMBLOCK_CONTROL_OFF	    0
+#else
+#error "Unsupported domain."
+#endif
```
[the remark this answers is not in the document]
hubertmis: One idea: how about wrapping this with `#ifndef RAMBLOCK_CONTROL_BIT_ICACHE` and pass the values through a build system

soc/arm/nordic_nrf/nrf54h/soc.c
```diff
@@ -40,7 +40,9 @@ static void power_domain_init(void)
 	 */
 
 	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, true);
+#if defined(CONFIG_HW_REVISION_SOC1)
 	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, true);
+#endif
 
 	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, true);
 	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, true);
```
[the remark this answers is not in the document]
hubertmis: For caches write-back in `z_arm_on_enter_cpu_idle_prepare()` to be bullet-proof, the function must be called after https://github.com/nrfconnect/sdk-zephyr/blob/99187c89524710cba572477a2cc95ab8d309b563/arch/arm/core/cortex_m/cpu_idle.S#L113

soc/nordic/common/poweroff.c
```diff
@@ -6,6 +6,7 @@
 #include <zephyr/sys/poweroff.h>
 #include <zephyr/toolchain.h>
 
+#if !defined(CONFIG_SOC_SERIES_NRF54HX)
```
hubertmis, line 9: This `#if` shows that this file is not "common" for Nordic SoCs, so it should be moved to another directory than soc/nordic/common.

soc/nordic/nrf54h/poweroff.c
```diff
@@ -0,0 +1,182 @@
...
+#error "Unsupported domain."
+#endif
+
+static void suspend_common(void)
+{
+	if (IS_ENABLED(CONFIG_DCACHE)) {
+		/* Flush, disable and power down DCACHE */
+		sys_cache_data_flush_all();
+		sys_cache_data_disable();
+		nrf_memconf_ramblock_control_enable_set(NRF_MEMCONF, RAMBLOCK_POWER_ID,
```
hubertmis, line 36: A little updated procedure is described in this PR: https://github.com/nrfconnect/doc-internal/pull/93

soc/nordic/nrf54h/poweroff.c
```diff
@@ -0,0 +1,182 @@
...
+		sys_cache_instr_disable();
+		nrf_memconf_ramblock_control_enable_set(NRF_MEMCONF, RAMBLOCK_POWER_ID,
+							RAMBLOCK_CONTROL_BIT_ICACHE, false);
+	}
+
+	/* Disable retention */
+	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
+
+	k_cpu_idle();
```
hubertmis, line 51: `k_cpu_idle()` should not be called from s2ram functions. `k_cpu_idle()` should contain some boilerplate around caches to safely suspend the CPU.

soc/nordic/nrf54h/poweroff.c
```diff
@@ -0,0 +1,182 @@
...
+
+void z_sys_poweroff(void)
+{
+	nrf_resetinfo_resetreas_local_set(NRF_RESETINFO, 0);
+	nrf_resetinfo_restore_valid_set(NRF_RESETINFO, false);
+
+	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, false);
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, false);
+
+	nrf_lrcconf_task_trigger(NRF_LRCCONF010, NRF_LRCCONF_TASK_SYSTEMOFFREADY);
```
hubertmis, line 62: This task must be triggered after `suspend_common()` because of the timing (there is a timeout in System Controller).
After `suspend_common()` this function must call `WFI`. The code after `WFI` is reachable in some scenarios, so `CODE_UNREACHABLE;` below is invalid.

soc/nordic/nrf54h/poweroff.c
```diff
@@ -0,0 +1,182 @@
...
+		/* Power up and re-enable DCACHE */
+		nrf_memconf_ramblock_control_enable_set(NRF_MEMCONF, RAMBLOCK_POWER_ID,
+							RAMBLOCK_CONTROL_BIT_DCACHE, true);
+		sys_cache_data_enable();
+	}
+
+	/* Re-enable domain retention. */
+	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, true);
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, true);
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, true);
```
hubertmis, line 93: The domains shall not be forced on here. I guess this is a workaround for the incomplete `k_cpu_idle()` implementation for nRF54H. I would at least add a TODO comment here, because when `k_cpu_idle()` will be implemented, we'll encounter surprises like unexpected forcing power domains ON.

soc/nordic/nrf54h/poweroff.c
```diff
@@ -0,0 +1,182 @@
...
+{
+	/* empty */
+}
+
+bool pm_s2ram_mark_check_and_clear(void)
+{
+	bool unretained_wake;
+	bool restore_valid;
+
+	unretained_wake = nrf_resetinfo_resetreas_local_get(NRF_RESETINFO) &
```
hubertmis, line 129: This code shall check if any other reset reasons are set in the RESETINFO. If they are, the function shall return `false`.

soc/nordic/nrf54h/soc.c
```diff
@@ -46,8 +46,9 @@ static void power_domain_init(void)
 	 *  WFI the power domain will be correctly retained.
 	 */
 
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, true);
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, true);
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN,
+			!IS_ENABLED(CONFIG_HALTIUM_CPURAD));
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
```
hubertmis, line 51: This code shall be moved around `k_cpu_idle()` implementation.

hubertmis (approved):
I marked multiple remarks that require attention. However, they can be fixed in the future PRs. For now, this PR should allow moving forward in the task force.

soc/nordic/nrf54h/poweroff.c
```diff
@@ -0,0 +1,182 @@
...
+{
+	if (!z_pm_do_suspend()) {
+		return;
+	}
+
+	/*
+	 * Save the CPU context (including the return address),set the SRAM
+	 * marker and power off the system.
+	 */
+	(void)pm_s2ram_suspend(z_pm_sys_suspend);
```
hubertmis, line 156: It seems this function should be called with IRQs disabled (wrapped with `__disable_irq()` - `__enable_irq()`).

If IRQs are enabled, we risk a race condition: after the NVIC or MPU state is saved in RAM, ZLI is triggered, resulting in a change in NVIC or MPU. When ZLI ends, this function resumes execution storing CPU registers, configuring LRCCONF, etc. When the CPU wakes up and restores NVIC and MPU, it restores incorrect values, saved before ZLI was triggered.

If IRQs are disabled, ZLI wouldn't be handled until we exit this function, `WFI` would result in `NOP`, so the system state would be coherent.

PR #1667, opened 2024-04-24
net: lib: coap_client: Improve cancel function

plskeggs:
Improve coap_client_cancel_requests(). Ensure it can be called from a callback. Clear pending reqs. Report error to waiting callbacks. Clear active flag.
    
This is useful when the network becomes unavailable or prior to disconnecting in order to save power.
    
(cherry picked from commit f04f8b242285dcd56b9e458de8ea233a14237b9b)

### DIFF
--- a/include/zephyr/net/coap_client.h
+++ b/include/zephyr/net/coap_client.h
@@ -138,6 +138,17 @@ int coap_client_req(struct coap_client *client, int sock, const struct sockaddr 
 		    struct coap_client_request *req, struct coap_transmission_parameters *params);
 
 /**
+ * @brief Cancel all current requests.
+ *
+ * This is intended for canceling long-running requests (e.g. GETs with the OBSERVE option set)
+ * which has gone stale for some reason.
+ *
+ * @param client Client instance.
+ */
+void coap_client_cancel_requests(struct coap_client *client);
+
+
+/**
  * @}
  */
 
--- a/include/zephyr/net/coap_client.h
+++ b/include/zephyr/net/coap_client.h
@@ -84,6 +84,7 @@ struct coap_client_internal_request {
 	uint32_t last_id;
 	uint8_t request_tkl;
 	bool request_ongoing;
+	atomic_t in_callback;
 	struct coap_block_context recv_blk_ctx;
 	struct coap_block_context send_blk_ctx;
 	struct coap_pending pending;
@@ -140,13 +141,13 @@ int coap_client_req(struct coap_client *client, int sock, const struct sockaddr 
 /**
  * @brief Cancel all current requests.
  *
- * This is intended for canceling long-running requests (e.g. GETs with the OBSERVE option set)
- * which has gone stale for some reason.
+ * This is intended for canceling long-running requests (e.g. GETs with the OBSERVE option set,
+ * or a block transfer) which have gone stale for some reason. It is also intended for responding
+ * to network connectivity issues.
  *
  * @param client Client instance.
  */
 void coap_client_cancel_requests(struct coap_client *client);
-
 
 /**
  * @}
--- a/subsys/net/lib/coap/coap_client.c
+++ b/subsys/net/lib/coap/coap_client.c
@@ -811,8 +811,17 @@ static int handle_response(struct coap_client *client, const struct coap_packet 
 	}
 fail:
 	client->response_ready = false;
-	internal_req->request_ongoing = false;
+	if (ret != 0 || !coap_request_is_observe(&internal_req->request)) {
+		internal_req->request_ongoing = false;
+	}
 	return ret;
+}
+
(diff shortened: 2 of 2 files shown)


plskeggs:
@juhaylinen @SeppoTakalo -- I would like your early review of this. If you agree with the approach then I will create an upstream PR and add the appropriate sauce tag here.

include/zephyr/net/coap_client.h
```diff
@@ -137,6 +137,16 @@ int coap_client_init(struct coap_client *client, const char *info);
 int coap_client_req(struct coap_client *client, int sock, const struct sockaddr *addr,
 		    struct coap_client_request *req, struct coap_transmission_parameters *params);
 
+/**
+ * @brief Cancel CoAP Requests
+ *
+ * When network connectivity is lost, terminate all pending requests.
```
SeppoTakalo, line 143: Cancellation should be possible even without any network errors.

For example CoAP block-wise transfer, we might see during the transfer that the FOTA image is going to be rejected. So no need to download rest of the blocks.

subsys/net/lib/coap/coap_client.c
```diff
@@ -385,6 +387,25 @@ static void report_callback_error(struct coap_client_internal_request *internal_
...
+			coap_pending_clear(&client->requests[i].pending);
+			k_mutex_unlock(&client->send_mutex);
+			report_callback_error(&client->requests[i],
+					      COAP_RESPONSE_CODE_GATEWAY_TIMEOUT);
+			client->requests[i].request_ongoing = false;
+		}
+	}
+	atomic_clear(&coap_client_recv_active);
+	/* Wait until after zsock_poll() times out. */
+	k_sleep(K_MSEC(COAP_PERIODIC_TIMEOUT));
```
SeppoTakalo, line 405: Why does it have to sleep?
I don't really understand.

include/zephyr/net/coap_client.h
```diff
@@ -137,6 +137,16 @@ int coap_client_init(struct coap_client *client, const char *info);
...
+ *
+ * @param client Client instance.
+ * @return zero when operation started successfully or negative error code otherwise.
+ */
+int coap_client_cancel_req(struct coap_client *client);
```
SeppoTakalo, line 148: Do you have to call this from within the callback, or outside, or does it matter?
Is it safe to call from callback?

SeppoTakalo (commented):
I think the API is OK.

Clarify:
* Can the cancellation be called within the callback (like when we receive blockwise Block N=0).
* Does the cancellation work for blockwise?

subsys/net/lib/coap/coap_client.c
plskeggs: Well, it can take up to COAP_PERIODIC_TIMEOUT milliseconds before zsock_poll() times out. Once it does, handle_poll() will be able to return to its caller and clean up. It "feels right" to allow that to happen before returning so the system is idle again.

include/zephyr/net/coap_client.h
plskeggs: Good question. I will devise a test for this.

plskeggs: Good question. I am making a small change so it should be ok to call it either way.

glarsennordic (commented):
I have some comments, but I left them on the upstream zephyr PR

glarsennordic (approved):

subsys/net/lib/coap/coap_client.c
```diff
@@ -27,6 +27,7 @@ static atomic_t coap_client_recv_active;
 static int send_request(int sock, const void *buf, size_t len, int flags,
 			const struct sockaddr *dest_addr, socklen_t addrlen)
 {
+	LOG_HEXDUMP_DBG(buf, len, "Send CoAP Request:");
```
SeppoTakalo, line 30: I think this is a bit too much debug information?
Is this intended to be left here or was it just a left over from development? We can debug CoAP messages properly in Wireshark if needed.

SeppoTakalo (approved):

subsys/net/lib/coap/coap_client.c
plskeggs: Yes, it may be too much at times. It is inconvenient now to change this as my upstream PR merged already.

Ideally we had another log level -- in some systems I have seen "insane" in addition to "debug".

In this case, I probably should have gated it with a macro.

plskeggs:
@carlescufi I updated this PR to be `fromtree` instead of `fromlist` now that the upstream PR has merged.

juhaylinen (approved):

plskeggs:
No longer necessary; the upstream PR merged and was included in the recent upmerge.

PR #1676, opened 2024-04-30
[nrf fromlist] ipc: ipc_service: icbmsg backend: workaround endpoint binding deadlock

kapi-no:
This change works around the issue with the semaphore timeout during
the Bluetooth HCI driver initialization when the bt_enable function
is called in the context of the System Workqueue thread. This issue
only affects platform that use the IPC service and its ICBMsg backend
(e.g. the nRF54H20 DK target).

The bt_enable function, when called in the System Workqueue context,
results in a deadlock, as the waiting semaphore of the Bluetooth HCI
driver times out:

bt_hci_driver: Endpoint binding failed with -11

During the Bluetooth HCI driver open operation in the context of the
bt_enable function, the driver code waits using the semaphore for the
endpoint binding process of the IPC service module to finalize. The
issue occurs when the waiting occurs in the System Workqueue context.
The ICBMsg backend from the IPC service schedules a system work during
the endpoint registration, in which it finalizes the binding operation,
also in the System Workqueue context. As the Bluetooth HCI driver
with its wait operation keeps the System Workqueue context busy, the
endpoint binding cannot be completed by the ICBMsg backend before the
HCI driver semaphore timeout.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/72377

This change is a temporary workaround solution to streamline the development of applications for the nRF54H20 DK target that use the App Event Manager module and operate in the System Workqueue context (e.g. nRF Desktop application). The proper solution is planned for the Bluetooth initialization process that will replace the current workaround.

Ticket for a more long-term solution:

https://nordicsemi.atlassian.net/browse/NCSDK-27318

### DIFF
--- a/subsys/ipc/ipc_service/backends/ipc_icbmsg.c
+++ b/subsys/ipc/ipc_service/backends/ipc_icbmsg.c
@@ -110,6 +110,12 @@ LOG_MODULE_REGISTER(ipc_icbmsg,
 /** Registered endpoints count mask in flags. */
 #define FLAG_EPT_COUNT_MASK 0xFFFF
 
+/** Workqueue stack size for bounding processing (this configuration is not optimized). */
+#define EP_BOUND_WORK_Q_STACK_SIZE (512U)
+
+/** Workqueue priority for bounding processing. */
+#define EP_BOUND_WORK_Q_PRIORITY (CONFIG_SYSTEM_WORKQUEUE_PRIORITY)
+
 enum msg_type {
 	MSG_DATA = 0,		/* Data message. */
 	MSG_RELEASE_DATA,	/* Release data buffer message. */
@@ -193,6 +199,9 @@ struct control_message {
 };
 
 BUILD_ASSERT(NUM_EPT <= EPT_ADDR_INVALID, "Too many endpoints");
+
+/* Work queue for bounding processing. */
+static struct k_work_q ep_bound_work_q;
 
 /**
  * Calculate pointer to block from its index and channel configuration (RX or TX).
@@ -672,7 +681,7 @@ static int send_bound_message(struct backend_data *dev_data, struct ept_data *ep
  */
 static void schedule_ept_bound_process(struct backend_data *dev_data)
 {
-	k_work_submit(&dev_data->ep_bound_work);
+	k_work_submit_to_queue(&ep_bound_work_q, &dev_data->ep_bound_work);
 }
 
 /**
@@ -1117,6 +1126,17 @@ static int backend_init(const struct device *instance)
 {
 	const struct icbmsg_config *conf = instance->config;
 	struct backend_data *dev_data = instance->data;
+	static K_THREAD_STACK_DEFINE(ep_bound_work_q_stack, EP_BOUND_WORK_Q_STACK_SIZE);
+	static bool is_work_q_started;
+
+	if (!is_work_q_started) {
+		k_work_queue_init(&ep_bound_work_q);
+		k_work_queue_start(&ep_bound_work_q, ep_bound_work_q_stack,
+				   K_THREAD_STACK_SIZEOF(ep_bound_work_q_stack),
+				   EP_BOUND_WORK_Q_PRIORITY, NULL);
+
+		is_work_q_started = true;
+	}
 
 	dev_data->conf = conf;
 	dev_data->is_initiator = (conf->rx.blocks_ptr < conf->tx.blocks_ptr);


subsys/ipc/ipc_service/backends/ipc_icbmsg.c
```diff
@@ -1117,10 +1121,15 @@ static int backend_init(const struct device *instance)
...
 	struct backend_data *dev_data = instance->data;
+	static K_THREAD_STACK_DEFINE(ep_bound_work_q_stack, EP_BOUND_WORK_Q_STACK_SIZE);
 
 	dev_data->conf = conf;
 	dev_data->is_initiator = (conf->rx.blocks_ptr < conf->tx.blocks_ptr);
 	k_mutex_init(&dev_data->mutex);
+	k_work_queue_init(&dev_data->ep_bound_work_q);
+	k_work_queue_start(&dev_data->ep_bound_work_q, ep_bound_work_q_stack,
+			   K_THREAD_STACK_SIZEOF(ep_bound_work_q_stack),
+			   CONFIG_SYSTEM_WORKQUEUE_PRIORITY, NULL);
```
MarekPieta, line 1132: Can we add a `#define` for the priority to avoid directly referring to system workq priority here?

subsys/ipc/ipc_service/backends/ipc_icbmsg.c
```diff
@@ -1117,10 +1121,15 @@ static int backend_init(const struct device *instance)
 {
 	const struct icbmsg_config *conf = instance->config;
 	struct backend_data *dev_data = instance->data;
+	static K_THREAD_STACK_DEFINE(ep_bound_work_q_stack, EP_BOUND_WORK_Q_STACK_SIZE);
 
 	dev_data->conf = conf;
 	dev_data->is_initiator = (conf->rx.blocks_ptr < conf->tx.blocks_ptr);
 	k_mutex_init(&dev_data->mutex);
+	k_work_queue_init(&dev_data->ep_bound_work_q);
```
MarekPieta, line 1129: Maybe we could keep the workq common among devices? (then we should keep it global and initialize once) Alternatively we would need to define separate stacks for the workqueues (current implementation may lead to really tricky errors I think)

subsys/ipc/ipc_service/backends/ipc_icbmsg.c
```diff
@@ -110,6 +110,9 @@ LOG_MODULE_REGISTER(ipc_icbmsg,
 /** Registered endpoints count mask in flags. */
 #define FLAG_EPT_COUNT_MASK 0xFFFF
 
+/** Workqueue stack size for bounding processing. */
```
MarekPieta, line 113: Remove extra `)` from commit message: `bt_hci_driver: Endpoint binding failed with -11)`

subsys/ipc/ipc_service/backends/ipc_icbmsg.c
MarekPieta: With the common workq, we could also easily add a temporary Kconfig to avoid increasing memory usage if not needed (app that needs the workaround would explicitly enable it)

MarekPieta (changes requested):

alstrzebonski (approved):

zycz (approved):

kapi-no:
> This is acceptable as temporary solution. The long therm solution requires reusing dedicated work queue from the ICMsg module. I will implement it later in the upstream Zephyr.

Great, please make sure to revert this change when you bring the proper solution upstream (to NCS)

subsys/ipc/ipc_service/backends/ipc_icbmsg.c
MarekPieta: You could add an experimental Kconfig to avoid deprecation process I think (anyway I don't think it's necessary)

MarekPieta (approved):

carlescufi (changes requested):
why not `fromlist`?

kapi-no:
> why not `fromlist`?

Upstream PR is available here:

https://github.com/zephyrproject-rtos/zephyr/pull/72377

hubertmis:
> This is acceptable as temporary solution. The long therm solution requires reusing dedicated work queue from the ICMsg module. I will implement it later in the upstream Zephyr.

I don't think the ICMsg module should expose workqueues. It's not the responsibility of the ICMsg module. ICMsg should be about passing messages between processors.
If we want to have a shared IPC workqueue, we should extract it from ICMsg and create a new module like ipc_workqueue or icmsg_workqueue used by both ICMsg and icbmsg, and potentially other IPC users. We would need to clearly define when blocking of this workqueue is permitted to avoid deadlocks like the one being fixed here.

doki-nordic:
> I don't think the ICMsg module should expose workqueues. It's not the responsibility of the ICMsg module. ICMsg should be about passing messages between processors. If we want to have a shared IPC workqueue, we should extract it from ICMsg and create a new module like ipc_workqueue or icmsg_workqueue used by both ICMsg and icbmsg, and potentially other IPC users. We would need to clearly define when blocking of this workqueue is permitted to avoid deadlocks like the one being fixed here.

I agree that taking work queue into some common module is the best approach. Looking at the current implementation of ICMsg and ICBMsg, we should assume that this new shared work queue cannot be blocked.

With that assumption in mind, the "bounded" callback of ipc_service becomes problematic. If we call it from our new work queue, we must clearly state that user cannot block in the "bounded" callback of ipc_service. If we redirect the callback to system work queue, we will end up with the same dead lock, because "bounded" callback from ipc_hci cannot be called from system work queue as kapi-no pointed out at the beginning of this PR.

carlescufi (approved):

PR #1684, opened 2024-05-07
Allow disabling the nRF5340 network CPU by protocol drivers

jciupis:


ktaborowski (approved):

ankuns (approved):
I think that when we have so cool module like nrf53_cpunet_mgmt there should be no code that controls the cpunet directly by `nrf_reset_network_force_off`. Users willing to use the nrf53 cpunet services should call `nrf53_cpunet_enable(true)` and when they do not need it anymore call `nrf53_cpunet_enable(false)`. These calls should be indirect from hci_init and now in nrf_802154_spinel_backedn_init and who-knows-where-else. No board-specific code should then attempt to enable the cpunet, because what's the purpose? Just to have it enabled? 

But fully moving to the new paradigm is a different (I suspect quite a big) task than just ad-hoc fixing to make things work for now.

PR #1698, opened 2024-05-10
Bring in commits that enable RMC 

de-nordic:
Modification of hal_module CMakes and Kconfigs that will bring in nrfx_rramc.c into build in case when there is RRAM controller in DTS.


hubertmis (approved):

anangl (approved):

nvlsianpu:
This is Obsolete.

PR #1713, opened 2024-05-14
[nrf noup] scripts: west: build: Use sysbuild by default if in NCS dir

nordicjm:
Uses sysbuild by default when building an application which resides in an allowed NCS-based directory when the sysbuild config key is not set

### DIFF
--- a/scripts/west_commands/build.py
+++ b/scripts/west_commands/build.py
@@ -9,6 +9,11 @@ import shlex
 import sys
 import yaml
 
+from pathlib import Path
+from zephyr_ext_common import ZEPHYR_BASE
+sys.path.append(os.fspath(Path(__file__).parent.parent))
+import zephyr_module
+
 from west import log
 from west.configuration import config
 from zcmake import DEFAULT_CMAKE_GENERATOR, run_cmake, run_build, CMakeCache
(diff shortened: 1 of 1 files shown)


scripts/west_commands/build.py
```diff
@@ -557,7 +562,34 @@ def _run_cmake(self, board, origin, cmake_opts):
...
+        # Check if this is an ncs-repo directory
+        allow_list = ['zephyr', 'mcuboot', 'sidewalk', 'find-my', 'nrf']
+        in_ncs_repo_dir = False
+
+        for module in zephyr_module.parse_modules(ZEPHYR_BASE, self.manifest):
+            if module.meta['name'] in allow_list:
+                if self.source_dir[0:len(module.project)] == module.project:
+                    path_char = self.source_dir[len(module.project):len(module.project)+1]
+
+                    if path_char == '/' or path_char == '\\':
```
tejlmand, line 574: Just use the built-in pathlib features, like:
```
Path(source_dir).relative_to(module.project)
```

scripts/west_commands/build.py
```diff
@@ -9,6 +9,11 @@
 import sys
 import yaml
 
+from pathlib import Path
```
tejlmand, line 12: pathlib is already imported above: https://github.com/nrfconnect/sdk-zephyr/blob/c9eac8821f4cb4554f2ee301472aba704408455b/scripts/west_commands/build.py#L7

tejlmand (changes requested):
A quick glance.

I don't see the need to go though zephyr_module.py, just fetch the manifest projects directly.

scripts/west_commands/build.py
nordicjm: Updated

scripts/west_commands/build.py
nordicjm: It does not work without this line

scripts/west_commands/build.py
```diff
@@ -9,6 +9,11 @@
...
+from zephyr_ext_common import ZEPHYR_BASE
+sys.path.append(os.fspath(Path(__file__).parent.parent))
+import zephyr_module
```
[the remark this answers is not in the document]
nordicjm: Because this does not work with optional modules

nordicjm:
Has been moved to https://github.com/nrfconnect/sdk-zephyr/pull/1706

scripts/west_commands/build.py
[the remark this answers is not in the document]
tejlmand: what  does not work ?
Modules passed with `-DEXTRA_ZEPHYR_MODULES` ?
Or inactive Zephyr projects ?

The `manifest.get_projects([])` should return projects if they are active.

PR #1720, opened 2024-05-16
[nrf fromtree] ipc_service: backends: icbmsg: Silence maybe-uninitial…

zycz:
…ized warning

If compiling with optimizations (-O2/-Ofast) gcc emits a maybe-uninitialized warning for `size`.

As far as I can tell, `size` will always be initialized in the cases where it is used and only left uninitialized in error cases.

Reproduced on main with the multi_endpoint sample on nrf5340dk/nrf5340/cpuapp.

Signed-off-by: Herman Berget <herman.berget@nordicsemi.no>

(cherry picked from commit 9729651bb5f8cab445332e99cf548b75f67950e2)


zycz:
Link to upstream commit: https://github.com/zephyrproject-rtos/zephyr/commit/9729651bb5f8cab445332e99cf548b75f67950e2

kapi-no (approved):

hermabe (approved):

kapi-no:
Integrating this change as the Twister CI fails when it detects this warning. Any new test configuration that includes IPC for nRF54H20 fails as a result

PR #1721, opened 2024-05-16
Upmerge 16.05.2024

rlubos:
Remaining isses:
- [x] Reapply https://github.com/nrfconnect/sdk-zephyr/commit/15cc59d1c2507d330c1eb58c359d3fc681b961a6 after conflicts are resolved in the upstream PR

PR #1722, opened 2024-05-17
[nrf fromlist] dts: nordic: nrf54h20: Fix exmif node definition

anangl:
Remove the "snps,designware-spi" compatible from the EXMIF node in nRF54H20i, as the spi_dw driver cannot be used for this peripheral without Nordic-specific modifications that are not present upstream. An attempt to do so (just setting CONFIG_SPI=y will cause that, as the driver initialization function will be executed then) results in a bus fault.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/72928

The compatible is restored in the second [noup] commit, so that the EXMIF can be used with the modified spi_dw driver in NCS.

### DIFF
--- a/dts/common/nordic/nrf54h20.dtsi
+++ b/dts/common/nordic/nrf54h20.dtsi
@@ -347,7 +347,7 @@
 			};
 
 			exmif: spi@95000 {
-				compatible = "nordic,nrf-exmif", "snps,designware-spi";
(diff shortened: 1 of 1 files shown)


e-rk (approved):

masz-nordic (approved):

anangl:
The commits went in with upmerge in #1721.

PR #1725, opened 2024-05-21
[nrf noup] Bluetooth: Mesh: remove legacy adv support

HaavardRei:
Removes explicit support for the legacy advertiser due to incompatibility with SDC.

The legacy advertiser can be used (experimentally) with the Zephyr Link Layer enabled, but is not recommended.


subsys/bluetooth/mesh/Kconfig
```diff
@@ -59,12 +59,16 @@ choice BT_MESH_ADV
...
 	bool "Legacy advertising"
+	depends on BT_LL_SW_SPLIT
 	help
 	  Use legacy advertising commands for mesh sending. Legacy
-	  advertising is significantly slower than the extended advertising, but
-	  is supported by all controllers.
+	  advertising is significantly slower than the extended advertising.
 
-	  WARNING: The legacy advertiser can occasionally do more message
+	  WARNING: This feature is no longer supported in NCS. The
```
alxelax, line 67: Actually, this is not the truth. The feature is still supported. Please add technical reason here. I mean description of a real problem while it cannot be used.

subsys/bluetooth/mesh/Kconfig
```diff
@@ -59,12 +59,16 @@ choice BT_MESH_ADV
...
+	  legacy advertiser does not work with SDC, so the Zephyr Link Layer
+	  must be used.
+
+	  ADDITIONAL WARNING: The legacy advertiser can occasionally do more message
```
alxelax, line 71: Is this the syntax requirement to write "ADDITIONAL WARNING"?

alxelax (changes requested):

subsys/bluetooth/mesh/Kconfig
HaavardRei: Couldn't find another example of two separate warnings, I can just do a newline instead adding to the existing warning?

alxelax: I'd suggest to leave just "WARNING" if syntax allows.

subsys/bluetooth/mesh/Kconfig
HaavardRei: I mean, aren't we choosing to remove support for the feature?

For the last part, something like: 
> The legacy advertiser will not work with SDC, as attempting to start an advertisement during the scanner duty cycle will result in an error. In ZLL, a workaround is introduced to let Mesh advertise during scanning.

alxelax: AFAIK Legacy advertiser is still there.
I think we do not need to write here about ZLL workaround.

HaavardRei: But just because it is there does not mean that we will continue to support it? Isn't that the point of adding this dependency?

alxelax: Yep. What I want to see here is the real problem description (at least a pair of phrases) not a just message "this is not supported".

HaavardRei: Updated now

alxelax (approved):

ludvigsj (approved):

omkar3141 (approved):

PR #1732, opened 2024-05-21
[v2.6.99-cs2] dt-bindings: pinctrl: nrf: allow for up to 16 ports

jonathannilsen:
nRF54 SoC can have up to 16 ports. This patch adjusts the pinctrl_soc_pin_t bitmap so that ports 0..15 can be used. An extra 3 bits are needed compared to previous nRF platforms. Since 2 were reserved, only 1 bit from function has been taken. Number of available functions has been reduced from 65535 to 32767, a number we'll never reach.


carlescufi (approved):

gmarull (changes requested):
pull commit from https://github.com/nrfconnect/sdk-zephyr/pull/1734 (fromlist)

carlescufi (approved):

gmarull (approved):

PR #1735, opened 2024-05-21
Fixes to support 802.15.4 on nRF54H20

piotrkoziar:
Content:
1. cherry-picks https://github.com/zephyrproject-rtos/zephyr/pull/70222
2. other fixes - see https://github.com/zephyrproject-rtos/zephyr/pull/73156

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -126,7 +126,11 @@ static void nrf5_get_eui64(uint8_t *mac)
 	mac[index++] = (IEEE802154_NRF5_VENDOR_OUI >> 8) & 0xff;
 	mac[index++] = IEEE802154_NRF5_VENDOR_OUI & 0xff;
 
-#if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE) && defined(NRF_FICR_S)
+#if defined(NRF54H_SERIES)
+	/* Can't access SICR with device id on a radio core. Use BLE.ADDR. */
+	deviceid[0] = NRF_FICR->BLE.ADDR[0];
+	deviceid[1] = NRF_FICR->BLE.ADDR[1];
+#elif defined(CONFIG_TRUSTED_EXECUTION_NONSECURE) && defined(NRF_FICR_S)
 	soc_secure_read_deviceid(deviceid);
 #else
 	deviceid[0] = nrf_ficr_deviceid_get(NRF_FICR, 0);
--- a/modules/hal_nordic/Kconfig
+++ b/modules/hal_nordic/Kconfig
@@ -55,7 +55,7 @@ endchoice
 
 config NRF_802154_TEMPERATURE_UPDATE
 	bool "nRF 802.15.4 temperature update"
-	default y
+	default y if !SOC_NRF54H20
 	help
 	  Enable temperature update for nRF 802.15.4 driver
 
--- a/modules/hal_nordic/nrf_802154/sl_opensource/platform/nrf_802154_clock_zephyr.c
+++ b/modules/hal_nordic/nrf_802154/sl_opensource/platform/nrf_802154_clock_zephyr.c
@@ -4,18 +4,21 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <nrf_802154_config.h>
 #include <platform/nrf_802154_clock.h>
 
 #include <stddef.h>
 
 #include <compiler_abstraction.h>
+#include <zephyr/kernel.h>
+#if defined(CONFIG_CLOCK_CONTROL_NRF)
 #include <zephyr/drivers/clock_control/nrf_clock_control.h>
 #include <zephyr/drivers/clock_control.h>
+#elif !defined(NRF54H_SERIES)
+#error No implementation to start or stop HFCLK due to missing clock_control.
+#endif
 
 static bool hfclk_is_running;
-static bool lfclk_is_running;
-static struct onoff_client hfclk_cli;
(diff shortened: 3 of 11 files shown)


jciupis:
For all commits cherry-picked directly from Zephyr main, please do: `git cherry-pick -x <sha>` and then prefix the commit title with `[nrf fromtree]` sauce tag.

For all the remaining commits:
1. Fix commit messages so that every one has a title, body and sign-off and does not exceed maximum line length allowed in Zephyr.
2. Open PR(s) to Zephyr with those commits and cherry-pick them with `[nrf fromlist]` sauce tag and link to upstream PR in the cherry-picked commit message body. Alternatively, add `[nrf noup]` sauce tag to commits that were agreed to only go downstream.

Also, note that commits cherry-picked from https://github.com/zephyrproject-rtos/zephyr/pull/70222 might conflict with https://github.com/nrfconnect/sdk-zephyr/pull/1721 since they are pulled directly from upstream there.

modules/hal_nordic/nrf_802154/sl_opensource/platform/nrf_802154_clock_zephyr.c
jciupis, line 1: Personally, I think it would be cleaner to structure the code like this:
```c
#if defined(CONFIG_CLOCK_CONTROL_NRF)
/* The entire, unmodified implementation using clock_control APIs */
#elif NRF54H_SERIES
/* Only nrf_802154_clock_hfclk_start and nrf_802154_clock_hfclk_stop
 * implemented without any additional static functions */
#else
#error "Some descriptive error message"
#endif
```

But I'm not going to block the PR over this, just a suggestion. It would be cleaner to later remove the changes once clock control is implemented for nRF54H.

subsys/ipc/ipc_service/backends/ipc_icbmsg.c
```diff
@@ -541,7 +541,7 @@ static int send_release(struct backend_data *dev_data, const uint8_t *buffer,
  * @param[in] size		Actual size of the data, can be smaller than allocated,
  *				but it cannot change number of required blocks.
  *
- * @return			O or negative error code.
```
jciupis, line 544: Isn't it the number of sent _blocks_?

modules/hal_nordic/Kconfig
```diff
@@ -55,6 +55,7 @@ endchoice
 
 config NRF_802154_TEMPERATURE_UPDATE
 	bool "nRF 802.15.4 temperature update"
+	default n if SOC_NRF54H20
```
jciupis, line 58: These two lines could be expressed as:
```
default y if !SOC_NRF54H20
```

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -129,8 +129,10 @@ static void nrf5_get_eui64(uint8_t *mac)
 #if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE) && defined(NRF_FICR_S)
 	soc_secure_read_deviceid(deviceid);
 #else
-	deviceid[0] = nrf_ficr_deviceid_get(NRF_FICR, 0);
-	deviceid[1] = nrf_ficr_deviceid_get(NRF_FICR, 1);
+	/* Use BLE.ADDR for setting EUI64, as this is currently the only feasible
```
jciupis, line 132: Please provide this implementation as an nRF54H20-only option, not replace the existing implementation that seemingly worked well for other devices so far.

modules/hal_nordic/nrf_802154/sl_opensource/platform/nrf_802154_clock_zephyr.c
piotrkoziar: I applied changes inspired by this. Good idea with the #error

drivers/ieee802154/ieee802154_nrf5.c
```diff
@@ -126,11 +126,12 @@ static void nrf5_get_eui64(uint8_t *mac)
...
-	soc_secure_read_deviceid(deviceid);
+#if !defined(NRF54H_SERIES)
+	z_impl_hwinfo_get_device_id(deviceid, sizeof(deviceid));
 #else
-	deviceid[0] = nrf_ficr_deviceid_get(NRF_FICR, 0);
-	deviceid[1] = nrf_ficr_deviceid_get(NRF_FICR, 1);
+	/* Can't access SICR with device id on a radio core. Use BLE.ADDR. */
+	deviceid[0] = NRF_FICR->BLE.ADDR[0];
+	deviceid[1] = NRF_FICR->BLE.ADDR[1];
 #endif
```
ankuns, line 135: I think that we should provide the workaround code only when `NRF54H_SERIES` is defined while keeping old behavior while it is not defined, so:
```c
#if defined(NRF54H_SERIES)
	/* Can't access SICR with device id on a radio core. Use BLE.ADDR. */
	deviceid[0] = NRF_FICR->BLE.ADDR[0];
	deviceid[1] = NRF_FICR->BLE.ADDR[1];
#elif defined(CONFIG_TRUSTED_EXECUTION_NONSECURE) && defined(NRF_FICR_S)
	soc_secure_read_deviceid(deviceid);
#else
	deviceid[0] = nrf_ficr_deviceid_get(NRF_FICR, 0);
	deviceid[1] = nrf_ficr_deviceid_get(NRF_FICR, 1);
#endif
```

piotrkoziar: Removed the dependency to hwinfo and applied the suggestion.

ankuns (approved):

piotrkoziar:
The changes were pulled with the recent upmerge.

PR #1744, opened 2024-05-27
[nrf fromlist] bluetooth: tester: Adjust Broadcaster ext adv pars

Frodevan:
The Broadcaster setup has been hardcoded to include the BT_LE_PER_ADV_OPT_USE_TX_POWER parameter. This may cause incompabilities should the controller not support it, and since the feature may not be required for most broadcast tests it should not be forced.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/73361


Thalley:
This change should be done upstream. We do not need `BT_LE_PER_ADV_OPT_USE_TX_POWER` for LE Audio

Thalley (approved):

alexsven (approved):

PR #1748, opened 2024-05-28
[nrf noup] scripts: west: Fix default value for sysbuild

krish2718:
nrf-squash! [nrf noup] scripts: west: build: Use sysbuild by default if in NCS dir

For NCS directory sysbuild is enabled by default.


sachinthegreen (approved):

nordicjm (commented):
this is not true actually, it's only true for specific repos, for customers it is not enabled by default in their out of tree applications

krish2718:
> this is not true actually, it's only true for specific repos, for customers it is not enabled by default in their out of tree applications

ok, but in sdk-nrf, the help is wrong, which is being picked from this repo, so, this is still applicable, no?

nordicjm:
> > this is not true actually, it's only true for specific repos, for customers it is not enabled by default in their out of tree applications
> 
> ok, but in sdk-nrf, the help is wrong, which is being picked from this repo, so, this is still applicable, no?

This is the code from that file:

```
        if config_sysbuild is None:
            # Check if this is an ncs-repo directory
            allow_list = [ 'mcuboot', 'sidewalk', 'find-my', 'nrf', 'matter', 'suit-processor',
                           'memfault-firmware-sdk', 'zscilib', 'uoscore-uedhoc', 'zcbor',
                           'hal_nordic' ]
            config_sysbuild = False

            for module in zephyr_module.parse_modules(ZEPHYR_BASE, self.manifest):
                if module.meta['name'] in allow_list and Path(self.source_dir).is_relative_to(module.project):
                    config_sysbuild = True
                    break

            if config_sysbuild is False and Path(self.source_dir).is_relative_to(ZEPHYR_BASE):
                config_sysbuild = True
```

PR #1751, opened 2024-05-29
[nrf fromlist] drivers: ieee802154: nrf5: sleep if idle when RxOnWhen…

edmont:
…Idle=0

When RxOnWhenIdle is set to False, turn the radio off
if no operation is ongoing in order to save power.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/73471

### DIFF
--- a/drivers/ieee802154/ieee802154_nrf5.c
+++ b/drivers/ieee802154/ieee802154_nrf5.c
@@ -1009,6 +1009,10 @@ static int nrf5_configure(const struct device *dev,
 	case IEEE802154_CONFIG_RX_ON_WHEN_IDLE:
 		nrf_802154_rx_on_when_idle_set(config->rx_on_when_idle);
 		nrf5_data.rx_on_when_idle = config->rx_on_when_idle;
+
+		if (config->rx_on_when_idle == false) {
+			(void)nrf_802154_sleep_if_idle();
+		}
 		break;
 
 	default:


maciejbaczmanski (approved):

modules/openthread/platform/radio.c
```diff
@@ -920,11 +920,13 @@ void otPlatRadioSetRxOnWhenIdle(otInstance *aInstance, bool aRxOnWhenIdle)
...
 
-	ARG_UNUSED(aInstance);
-
 	LOG_DBG("RxOnWhenIdle=%d", aRxOnWhenIdle ? 1 : 0);
 
 	radio_api->configure(radio_dev, IEEE802154_CONFIG_RX_ON_WHEN_IDLE, &config);
+
+	if (!aRxOnWhenIdle) {
+		ARG_UNUSED(otPlatRadioSleep(aInstance));
+	}
```
abtink, line 929: Thanks @edmont.

The "RxOnWhenIdle" flag can be changed by various triggers within the OT stack. For instance, the `otThreadSetLinkMode()` API, called by higher-level code controlling the OT stack, can switch the device mode, making it sleepy or Rx-on at any point in time.

Therefore, we should consider that `otPlatRadioSetRxOnWhenIdle()` could also be invoked at any time, while the radio may be in different states. For example, the radio could be in the middle of frame transmission or performing an energy scan.

Would calling `otPlatRadioSleep()` in all cases be appropriate? For instance, if the radio is mid-transmission, the OT stack expects a "send done" callback after a successful `otPlatRadioTransmit()`. Will this behave correctly after calling `otPlatRadioSleep()` in such a case?

edmont: Discussion in upstream PR.

PR #1773, opened 2024-06-05
[nrf noup] dts: nordic: Add support for BICR generation

57300:
This will be replaced by a different solution upstream (NCSDK-27673).

Migrate the existing BICR bindings from NCS 2.4.99-cs3 to NCS 2.7.0. They are updated to remove obsoleted values. Enable generating bicr.hex for Application core by default.

### DIFF
--- a/dts/bindings/misc/nordic,nrf-bicr.yaml
+++ b/dts/bindings/misc/nordic,nrf-bicr.yaml
@@ -1,0 +1,117 @@
+# Copyright (c) 2024 Nordic Semiconductor ASA
+# SPDX-License-Identifier: LicenseRef-Nordic-4-Clause
+
+description: |
+  Nordic Board Information Configuration Registers (BICR)
+
+compatible: "nordic,nrf-bicr"
+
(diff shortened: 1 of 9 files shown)


dts/bindings/misc/nordic,nrf-bicr.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+  tamperswitch-enable:
+    type: boolean
+    description: Enable tamper protections.
+
+  tamper-activeshield-channels:
+    type: array
+    description: |
+      List of channels to enable for the tamper protection subsystem's active
+      shield detector.
```
jonathannilsen, line 127: As far as I know, these are unused and it might be best to leave them out of the noup solution to avoid confusion

57300: I think you're right. Dropped.

jonathannilsen (approved):

karl-nordic (approved):

anangl (approved):

masz-nordic (approved):

anangl:
Merged through #1777.

PR #1776, opened 2024-06-05
[nrf fromlist] Add features allowing twister to work with nrf54h and nrf54l

PerMac:
Nrf54h and nrf54l require sysbuild for every build, even for "single core" like upstream's hello_world. These commits add feature to twister allowing to always use sysbuild for platforms with `sysbuild: true` in their yamls. It also adds `sysbuild: true` to all nrf54h and nrf54l yamls


shanthanordic (approved):

anangl (approved):

nordic-segl (approved):

jaz1-nordic (approved):

ns-tolu (approved):

masz-nordic (approved):

anangl:
Merged through #1777.

PR #1788, opened 2024-06-11
[nrf fromtree] drivers: serial: nrfx_uarte: Use of legacy shim on RISCV

nordic-krch:
Legacy shim takes less flash so it should be a first choice on cores with less code memory (like RISCV cores on nrf54h20). Adding new instances support to the legacy shim.

(cherry picked from commit d07b13dbc203be8066e9f840edf917dd8969ce00)


kapi-no:
The NCS CoreMark sample requires this change so that the PPR core can correctly log the benchmark results.

kapi-no (approved):

nika-nordic (approved):

kl-cruz (approved):

masz-nordic (approved):

PR #1791, opened 2024-06-11
[nrf fromtree] drivers: timer: nrf_grtc_timer: Fix for random hangs

nordic-krch:
d599e2b6705eb removed early return from sys_clock_timeout_handler if current counter value is less than cc_val. It seems that this return is needed as after the removal some stress tests which were using system timers heavily started to hang.

(cherry picked from commit 1a8c5e067a384001f6df9c3b0025c085cccfc7d7)


nika-nordic (approved):

anangl (changes requested):
Invalid SHA in the commit message. See https://github.com/zephyrproject-rtos/zephyr/commit/af3c8b973ede41f34bef107a71c3bc9de676b20c.

kl-cruz (approved):

nordic-segl (approved):

nordic-krch:
@anangl SHA fixed

anangl (approved):

PR #1792, opened 2024-06-11
[nrf fromlist] samples: modules: canopennode: set bootloader in sampl…

gmarull:
…e.yaml

Set the bootloader choice for the
sample.modules.canopennode.program_download test case in the sample.yaml file instead of sysbuild.conf. The reason for this change is to avoid building MCUboot by default when using the --sysbuild option, as it may not be compatible with all platforms. This is currently causing issues on platforms like nRF54H20 where sysbuild is always used.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/74101

Signed-off-by: Gerard Marull-Paretas <gerard@teslabs.com>
(cherry picked from commit 4ca5c66b34e973a83f909d06ffde857e90f57aa6)


nordicjm (approved):

PerMac (approved):

anangl (approved):

anangl:
@gmarull A related manifest update PR is needed.

nordic-piks (approved):
Checked at CI at HW - working fine.

PR #1796, opened 2024-06-12
[nrf fromtree] Bluetooth: Host: Avoid processing "no change" encryption changes

cvinayak:
If the new encryption state is the same as the old one, there's no point in doing additional processing or callbacks. Simply log a warning and ignore the HCI event in such a case.

(cherry picked from commit bf363d7c3e4614ebc9b7e7a051657ae6fde6283b)

### DIFF
--- a/subsys/bluetooth/host/hci_core.c
+++ b/subsys/bluetooth/host/hci_core.c
@@ -2092,6 +2092,12 @@ static void hci_encrypt_change(struct net_buf *buf)
 		return;
 	}
 
+	if (conn->encrypt == evt->encrypt) {
+		LOG_WRN("No change to encryption state (encrypt 0x%02x)", evt->encrypt);
+		bt_conn_unref(conn);
+		return;
+	}
+
 	conn->encrypt = evt->encrypt;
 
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/host/hci_core.c
```diff
@@ -2092,6 +2092,12 @@ static void hci_encrypt_change(struct net_buf *buf)
 		return;
 	}
 
+	if (conn->encrypt == evt->encrypt) {
```
jostn, line 2095: Shouldn't this be above https://github.com/nrfconnect/sdk-zephyr/pull/1796/files#diff-774b3de86adb275da54becf2a0958a2d4dae03d63e47840037ecc575e9d658b4R2088 ?

cvinayak: If `status` != `0U` then `if (conn->encrypt == evt->encrypt)`  is not needed. Hence, the changes in this commit are corrrect.

jostn (approved):

jori-nordic (approved):

alwa-nordic (approved):

PR #1797, opened 2024-06-12
[nrf fromlist] hal_nordic: Fix reserved PPI faulty logic

jori-nordic:
The list of reserved PPIs by MPSL should be kept if MPSL is enabled. The 802154 radio driver is not the only user of MPSL.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/74139

Quoting original PR

> This is a blocker for Garmin's ANT stack that also uses MPSL.
Their application/stack doesn't start up at all without this fix.

### DIFF
--- a/modules/hal_nordic/nrfx/Kconfig
+++ b/modules/hal_nordic/nrfx/Kconfig
@@ -9,6 +9,10 @@ menu "nrfx drivers"
 
 rsource "Kconfig.logging"
 
+config MPSL
+	bool
+	# Dummy kconfig to allow compiling outside of NCS
+
 config NRFX_ADC
 	bool "ADC driver"
 	depends on $(dt_has_compat,$(DT_COMPAT_NORDIC_NRF_ADC))
--- a/modules/hal_nordic/nrfx/nrfx_glue.h
+++ b/modules/hal_nordic/nrfx/nrfx_glue.h
@@ -378,7 +378,7 @@ void nrfx_busy_wait(uint32_t usec_to_wait);
 #define NRFX_PPI_GROUPS_USED_BY_802154_DRV     0
 #endif // CONFIG_NRF_802154_RADIO_DRIVER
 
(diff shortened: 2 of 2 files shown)


anangl (approved):

martintv (approved):

olivier-le-sage (approved):

modules/hal_nordic/nrfx/Kconfig
```diff
@@ -9,6 +9,10 @@ menu "nrfx drivers"
 
 rsource "Kconfig.logging"
 
+config MPSL
+	bool
+	# Dummy kconfig to allow compiling outside of NCS
```
anangl, line 14: CI in https://github.com/nrfconnect/sdk-nrf/pull/15940 fails on building Zigbee samples that use overlays with `CONFIG_MPSL=n`.
I think that instead of the above you should introduce a Kconfig option like `MPSL_IN_USE` that will be selected by `MPSL`.

anangl (changes requested):

modules/hal_nordic/nrfx/Kconfig
jori-nordic: There shouldn't be a need for `CONFIG_MPSL=n` in APP-core overlays anyways. Do you know the reason?

anangl: I don't know. Perhaps the one who added such overlay could say something more.

jori-nordic: The CI is green. Could we get this in?
That fixes the immediate problem (samples not working at all).

We can make a nicer, generic solution after the release, so we can discuss it properly without the release stress.

anangl: CI fails in manifest update (https://github.com/nrfconnect/sdk-nrf/pull/15940) because of build failures, so this PR cannot be merged.

anangl (approved):

PR #1805, opened 2024-06-13
boards: nordic: Update PWM configuration to follow zephy upstream

kl-cruz:
Update nRF54 DKs with changes taken from zephyr upstream.


ludvigsj:
I believe you have missed the polarity of `pwm_led1` (in `nrf54l15pdk_nrf54l15-common.dtsi`) which has changed in upstream but is still `INVERTED` here in sdk-zephyr? Or am I not reading things right?

ludvigsj:
For info: without this we (BLE Mesh) have a sample with a non-functioning LED when running on nrf54L15, it would be nice if this could get in for NCS 2.7.0 (meaning: added to the milestone) so we have a fully functional sample, but this won't block the release for our case.

miha-nordic:
@kl-cruz is there sdk-nrf manifest update PR?

kl-cruz:
@miha-nordic I will prepare

carlescufi (changes requested):
why is this a `noup`? why do the pins change between up and downstream?

gmarull (changes requested):
can't be a noup

gmarull (approved):

carlescufi (approved):

PR #1810, opened 2024-06-13
[nrf fromtree] nvs: Add data CRC

RICCIARDI-Adrien:


anangl:
These commits have been already cherry picked in #1806.

PR #1819, opened 2024-06-14
[nrf noup] bluetooth: att: Allow ATT sent callback after data TX is done

MarekPieta:
By default, the BLE stack calls sent callback for ATT data when the data is passed to BLE controller for transmission. Enabling this Kconfig option delays calling the sent callback until data transmission is finished by BLE controller (the callback is delayed until receiving the num complete packets event).

PR also allows for an extra ref in bt_conn_send_cb (needed for the newly added feature).

Jira: NCSDK-27422

### DIFF
--- a/subsys/bluetooth/host/Kconfig.gatt
+++ b/subsys/bluetooth/host/Kconfig.gatt
@@ -30,6 +30,23 @@ config BT_ATT_RETRY_ON_SEC_ERR
 	  If an ATT request fails due to insufficient security, the host will
 	  try to elevate the security level and retry the ATT request.
 
+config BT_ATT_SENT_CB_AFTER_TX
+	bool "Delay ATT sent callback until data transmission is done by controller [EXPERIMENTAL]"
+	select EXPERIMENTAL
+	help
+	  By default, the BLE stack calls sent callback for ATT data when the
+	  data is passed to BLE controller for transmission. Enabling this
+	  Kconfig option delays calling the sent callback until data
+	  transmission is finished by BLE controller (the callback is called
+	  upon receiving the Number of Completed Packets HCI Event).
+
+	  The feature is not available in Zephyr RTOS (it's specific to NCS
+	  Zephyr fork). It is a temporary solution allowing to control flow of
+	  GATT notifications with HID reports for HID use-case.
+
+	  Enabling this option may require increasing CONFIG_BT_CONN_TX_MAX in
+	  configuration, because ATT would use additional TX contexts.
+
 config BT_EATT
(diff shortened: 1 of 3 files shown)


MarekPieta:
After discussion with @rugeGerritsen, I started to wonder if we should mark this feature as experimental/deprecated (it's a temporary fix that would be removed in one of the following NCS releases). On the other hand introducing a new feature as deprecated without providing an alternative solution may not be a good idea. An application that is going to use the workaround (nRF Desktop) has production quality since some time already, so enabling an experimental feature for all of the peripheral configurations there is also controversial.

Please let me know what's your opinion on this.

tejlmand:
Why is this a `[nrf noup]` ?

Can't this be written in a way which is beneficial for upstream.

jori-nordic:
> Can't this be written in a way which is beneficial for upstream

This is re-introducing coupling between the host and the controller that we're trying to remove upstream.
Having it [noup] allows us to patch the issue for the impeding release, while giving us some time to design a better way without the time pressure.

jori-nordic (changes requested):
there is an assert you also have to tweak.
See the bsim run in the upstream PR you opened.

tejlmand:
> Having it [noup] allows us to patch the issue for the impeding release, while giving us some time to design a better way without the time pressure.

thanks for the explanation.

subsys/bluetooth/host/Kconfig.gatt
```diff
@@ -30,6 +30,18 @@ config BT_ATT_RETRY_ON_SEC_ERR
...
 	  try to elevate the security level and retry the ATT request.
 
+config BT_ATT_SENT_CB_AFTER_TX
+	bool "Delay ATT sent callback until data transmission is done by BLE controller"
+	help
+	  By default, the BLE stack calls sent callback for ATT data when the
+	  data is passed to BLE controller for transmission. Enabling this
+	  Kconfig option delays calling the sent callback until data
+	  transmission is finished by BLE controller (the callback is called
+	  upon receiving the num complete packets event).
```
kapi-no, line 40: The event is called

```suggestion
	  upon receiving the Number of Completed Packets HCI Event).
```

kapi-no (approved):

mkapala-nordic (approved):

carlescufi (changes requested):
Why is this a noup?

carlescufi (approved):
Technical debt tracked here:
https://nordicsemi.atlassian.net/plugins/servlet/mobile?originPath=%2Fbrowse%2FNCSDK-28016#issue/NCSDK-28016

theob-pro (approved):

kapi-no (approved):

PR #1823, opened 2024-06-17
Pull in CMSIS-DAP support from upstream Zephyr

maxd-nordic:


drivers/dp/swdp_ll_pin.h
```diff
@@ -0,0 +1,84 @@
...
+#warning "Pin delay is not defined"
+#endif
+}
+
+static ALWAYS_INLINE void swdp_ll_pin_input(void *const base, uint8_t pin)
+{
+#if defined(CONFIG_SOC_SERIES_NRF52X) || defined(CONFIG_SOC_SERIES_NRF53X)
+	NRF_GPIO_Type * reg = base;
+
+	reg->PIN_CNF[pin] = 0b0000;
```
masz-nordic, line 43: HAL should be used here, and in code below.

masz-nordic (changes requested):

drivers/dp/swdp_ll_pin.h
maxd-nordic: this is from upstream zephyr - it's not the right place to change it here.

masz-nordic: It should be fixed in a follow-up PR in upstream, that would be cherry-picked as fromlist to this PR.

maxd-nordic: @masz-nordic Could you give me some pointers on how to do that? `nrf_gpio_cfg` uses quite a few more cycles then just setting the PIN_CNF register directly. Speed is of utmost importance for this part.

masz-nordic: Could you post disassembly here? Is `nrf_gpio_pin_port_decode` not getting optimized enough?
We can add new functions if needed.

drivers/dp/swdp_ll_pin.h
```diff
@@ -0,0 +1,84 @@
...
+#endif
+}
+
+
+static ALWAYS_INLINE void swdp_ll_pin_set(void *const base, uint8_t pin)
+{
+#if defined(CONFIG_SOC_SERIES_NRF52X) || defined(CONFIG_SOC_SERIES_NRF53X)
+	NRF_GPIO_Type * reg = base;
+
+	reg->OUTSET = BIT(pin);
```
masz-nordic, line 62: Can be replaced with `nrf_gpio_port_out_set`

drivers/dp/swdp_ll_pin.h
```diff
@@ -0,0 +1,84 @@
...
+#endif
+}
+
+static ALWAYS_INLINE void swdp_ll_pin_clr(void *const base, uint8_t pin)
+{
+#if defined(CONFIG_SOC_SERIES_NRF52X) || defined(CONFIG_SOC_SERIES_NRF53X)
+	NRF_GPIO_Type * reg = base;
+
+	reg->OUTCLR = BIT(pin);
```
masz-nordic, line 71: Can be replaced with `nrf_gpio_port_out_clear`

drivers/dp/swdp_ll_pin.h
```diff
@@ -0,0 +1,84 @@
...
+#endif
+}
+
+static ALWAYS_INLINE uint32_t swdp_ll_pin_get(void *const base, uint8_t pin)
+{
+#if defined(CONFIG_SOC_SERIES_NRF52X) || defined(CONFIG_SOC_SERIES_NRF53X)
+	NRF_GPIO_Type * reg = base;
+
+	return ((reg->IN >> pin) & 1);
```
masz-nordic, line 80: `nrf_gpio_pin_read`

tmon-nordic:
Is this going to be obsolete by upmerge?

maxd-nordic:
We have a version of this as a patch on hello.nrfcloud/firmware. Since this is upstream, it will reach us eventually.

PR #1824, opened 2024-06-17
[nrf noup] dts: common: nordic: nrf54l15 rram erase-block-size

michalek-no:
Adjusting erase block size to match resolution of
rram region protection.


masz-nordic:
Why is it a `noup`?

michalek-no:
> Why is it a `noup`?

that's temporary, need to check what fails in CI

PR #1828, opened 2024-06-17
drivers: adc: saadc: Disable burst mode on unused channels

msalau:
Hello,

I discovered that mixing channels with oversampling and without doesn't work right. It turned out that left-over burst mode on unused channels inhibit operation of channels combined into a sequence.

Discovered on nRF52832 (nRF52-DK) using the following approach:
  channels 0-3 are used for application purposes as a sequence
  channel 4 is used for battery measurements with oversampling

After few successful conversions the sequence (channels 0-3) freezes the thread while waiting for semaphore to end the conversion.

P.S.: Should I create a separate pull-request to the upstream zephyr repository? Or pull-request to  upstream should be done in the first place and you'll eventually pull the changes here?

### DIFF
--- a/drivers/adc/adc_nrfx_saadc.c
+++ b/drivers/adc/adc_nrfx_saadc.c
@@ -435,6 +435,10 @@ static int start_read(const struct device *dev,
 				m_data.positive_inputs[channel_id]);
 			++active_channels;
 		} else {
+			nrf_saadc_burst_set(
+				NRF_SAADC,
+				channel_id,
+				NRF_SAADC_BURST_DISABLED);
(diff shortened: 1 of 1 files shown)


anangl:
> P.S.: Should I create a separate pull-request to the upstream zephyr repository? Or pull-request to upstream should be done in the first place and you'll eventually pull the changes here?

@msalau Yes, this patch should be posted to upstream Zephyr. It can be then cherry-picked here, if needed urgently, or just left to be brought here with the next upmerge.

msalau:
@anangl Thanks! I'll create a pull-request to upstream Zephyr.

PR #1832, opened 2024-06-18
[nrf noup] Kconfig.nrf54l15pdk: Overwrite FLASH configs for PM

oyvindronningstad:
When using PM, overwrite FLASH_BASE_ADDRESS and FLASH_SIZE to point at the whole RRAM, not just the flpr partition, to allow partition manager to partition everything inside it.


carlescufi (changes requested):
Can't this live in sdk-nrf?

nordicjm (changes requested):
Sysbuild in ncs does not support building for nRF53 network core as the default with application core images added to it and must build for the application core then add network core images to it, see https://github.com/nrfconnect/sdk-nrf/pull/15845

> Building for nRF5340 network core as the primary and including an application core image
>     Projects should be updated to build for the application core first and add network images to them

Same thing with this, users should build for the main core first then add a flpr image to the project as a secondary

boards/nordic/nrf54l15pdk/Kconfig.nrf54l15pdk
```diff
@@ -5,3 +5,12 @@ config BOARD_NRF54L15PDK
...
 					    BOARD_NRF54L15PDK_NRF54L15_CPUFLPR_XIP
+
+if SOC_NRF54L15_ENGA_CPUFLPR && PARTITION_MANAGER_ENABLED
+config FLASH_BASE_ADDRESS
+	hex
+	default 0x0
+config FLASH_SIZE
+	int
+	default 1524
+endif
```
tejlmand, line 16: this seems like something which should be in the `Kconfig.defconfig`.
The `Kconfig.nrf54l15pdk` is shared with sysbuild. which doesn't want to know about `FLASH_SIZE` and `FLASH_BASE_ADDRESS`.

tejlmand (changes requested):

tejlmand:
> Can't this live in sdk-nrf?

probably not, if the defconfig of the board or SoC defines a default flash size before sdk-nrf Kconfig is loaded, but i'll leave final answer to @oyvindronningstad

oyvindronningstad:
@nordicjm Adding this limitation would solve a lot of these issues.

PR #1840, opened 2024-06-20
[nrf noup]: Do not use newlib nano for C++ unordered hash map test

tejlmand:
The hash map test: libraries.hash_map.newlib.cxx_unordered_map.djb2 is testing C++ Hash map implementation using newlib.

C++ select C++ exceptions which is incompatible with newlib nano. Newlib nano is default in NCS when newlib is selected, and thus building of the test will result in a Kconfig failure.

Support the test by disabling newlib nano but keeping newlib for the test.

Jira: NCSDK-27925
Jira: NCSDK-28096


tejlmand:
NCS Newlib nano default when newlib is selected is specified here: https://github.com/nrfconnect/sdk-nrf/blob/489ee63627ef46c62de164a4bace745299d4a63d/Kconfig.nrf#L47-L51

PerMac:
I am fine with such solution because I don't have to deal with [noups] ;) still, have to raise the question: can we allow for such noup? 
Does executing this test in NCS brings us any value? If not, we can just add it to the "won't fix" section of quarantine and forget about it.

tejlmand:
> Does executing this test in NCS brings us any value? If not, we can just add it to the "won't fix" section of quarantine and forget about it.

that's also fine with me.
Just opened this PR to get the discussion going 🙂

anangl (approved):

shanthanordic (approved):

shanthanordic:
I have approved but it is okay to quarantine and state wont fix. @PerMac  if you are ok we can ask Torsten to close this PR.

shanthanordic:
@PerMac  and @tejlmand  Even it is decided to get this in, It can go after NCS 2.7.0. Is that right ?

PerMac:
yes. All previous platforms were already sitting in the quarantine for a while, the new ones are also added there. There is no ruch in dealing with the issue now

tejlmand:
added to quarantine instead.
No need for this PR.

PR #1842, opened 2024-06-21
[nrf fromlist] drivers: serial: nrfx_uarte2: Fix handling of errors in interrupt driven mode 

nordic-krch:
Driver was expecting that interrupt driven API will call uart_err_check() soon after error is reported and was keeping error state until that happens and because of that it was not reporting new bytes. This is not always the case and if error interrupt is not enabled then error should be ignored.

Fixes NRFX-6032


shanthanordic:
cleared milestone too late for NCS 2.7.0

PR #1853, opened 2024-06-25
[nrf noup] soc: nrf54l: Add custom section for KMU

Vge0rge:
Add a custom section in the linker which should always be placed in the top of RAM. This will be used by the KMU to push keys into it. Since when you provision a key into the KMU you need to set specific a memory location for the PUSH operation we need to keep this memory location static across images/dfus.

This is a noup since the KMU is not supported upstream.

Ref: NCSDK-25121

### DIFF
--- a/soc/nordic/nrf54l/CMakeLists.txt
+++ b/soc/nordic/nrf54l/CMakeLists.txt
@@ -10,3 +10,12 @@ zephyr_include_directories(.)
 if(CONFIG_ELV_GRTC_LFXO_ALLOWED)
   message(WARNING "WARNING! ELV mode feature is EXPERIMENTAL and may brick your device!")
 endif()
+
+# We need a buffer in memory in a static location which can be used by
+# the KMU peripheral. The KMU has a static destination address, we chose
+# this address to be 0x20000000, which is the first address in the SRAM.
+if(NOT CONFIG_BUILD_WITH_TFM AND CONFIG_PSA_NEED_CRACEN_KMU_DRIVER)
+# Exclamation mark is printable character with the lowest number in ASCII table.
+# We are sure that this file will be included first.
+zephyr_linker_sources(RAM_SECTIONS SORT_KEY ! kmu_push_area_section.ld)
+endif()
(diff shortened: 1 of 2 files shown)


soc/nordic/nrf54l/CMakeLists.txt
```diff
@@ -10,3 +10,10 @@ zephyr_include_directories(.)
 if(CONFIG_ELV_GRTC_LFXO_ALLOWED)
   message(WARNING "WARNING! ELV mode feature is EXPERIMENTAL and may brick your device!")
 endif()
+
+# We need a buffer in memory in a static location which can be used by
+# the KMU peripheral. The KMU has a static destination address so we need
+# to make sure that this buffer is always at the start of RAM.
```
frkv, line 16: Related PR talks about "top" of RAM. Should it state "top" here as well or is "start" a better term overall

vili-nordic (approved):

soc/nordic/nrf54l/kmu_push_area_section.ld
```diff
@@ -0,0 +1,20 @@
...
+    __nrf_kmu_reserved_push_area = .;
+    *(.nrf_kmu_reserved_push_area)
+    __nrf_kmu_reserved_push_area_end = .;
+} GROUP_NOLOAD_LINK_IN(RAMABLE_REGION, RAMABLE_REGION)
+
+# It doesn't seem to be possible to enforce placing a section
+# at a specific address in memory using the Zephyr SECTION macros.
+# So this assert is necessary to avoid accidentatly moving this
+# section to a different address.
+ASSERT(__nrf_kmu_reserved_push_area == RAM_ADDR, "Error:  \
```
vili-nordic, line 15: Seems this assert is triggered in CI, so the approach here has some issues

soc/nordic/nrf54l/CMakeLists.txt
Vge0rge: I replaced it with the specific address to make it more clear.

soc/nordic/nrf54l/kmu_push_area_section.ld
Vge0rge: This should be fixed now

stephen-nordic (approved):

PR #1857, opened 2024-06-25
NAT64 implementation 

MarGasiorek:
This PR implements modules which allow to use NAT64 from OT stack.


rlubos:
If we're going public with this, why not send the patch upstream? I'm really not in favor of maintaining such a large noup patches.

jukkar:
I agree with @rlubos, this code is quite generic and the Nordic specific "secret sauce" is not really there, it would make sense to upstream this.

MarGasiorek:
PR in zephyr created: https://github.com/zephyrproject-rtos/zephyr/pull/75012

rlubos:
Thanks for opening upstream PR, commits should be marked with `[nrf fromlist]` now (or `[nrf fromtree]` if we can wait for the upstream PR to be merged).

MarGasiorek:
Tracked in separate PR.

PR #1863, opened 2024-06-27
[nrf fromlist] fs: nvs: Improve nvs_calc_free_space() result precision

RICCIARDI-Adrien:
The nvs_calc_free_space() function does not return 0 when the NVS is considered full, because some special ATEs are not taken into account.

This commit takes into account the ATE that is reserved for deletion in each sector, in addition of the 'GC done' ATE when present.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/74900


de-nordic:
This needs rebase.

RICCIARDI-Adrien:
Closing this PR because `sdk-zephyr` has been updated to a more recent Zephyr version already containing the feature.

PR #1868, opened 2024-06-28
[nrf noup] scripts: west_commands: build: Add ant to sysbuild list

nordicjm:
nrf-squash! [nrf noup] scripts: west: build: Use sysbuild by default if in NCS dir

Adds the ANT project to the list of repos that should be built using sysbuild by default


nordicjm:
@sarah-read

tejlmand (approved):

shanthanordic (approved):

PR #1894, opened 2024-07-17
[nrf noup] Revert unnecessary changes in BT_CTLR symbol

Damian-Nordic:
Adding dependency on BT_RPC_STACK symbol defined in NCS may be hard to contribute so it's better to override the default BT_CTLR value in the application that enables BT_RPC_STACK.


alxelax:
This PR means we need just to change zephyr manifest in ncs in this hash: a56fee6870e152b71a7baf5b85bfa0b9746286a0

alxelax (approved):

canisLupus1313 (approved):

PR #1903, opened 2024-07-22
[nrf noup] scripts: west: build: Enable sysbuild by default globally

nordicjm:
fixup! [nrf noup] scripts: west: build: Use sysbuild by default if in NCS dir

Enables sysbuild by default for all repositories, NCS or otherwise


MarekPieta (approved):
Looks ok, should not cause problems for our applications/samples

tejlmand (approved):

PR #1904, opened 2024-07-23
[nrf fromlist] tests: drivers: adc: add adc_error_cases tests.

nordic-babu:
Tests are checking error codes returned from adc_read() and adc_setup_channel() used with invalid configurations.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/76191

### DIFF
--- a/tests/drivers/adc/adc_error_cases/src/main.c
+++ b/tests/drivers/adc/adc_error_cases/src/main.c
@@ -1,0 +1,244 @@
+/*
+ * Copyright (c) 2024, Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/device.h>
+#include <zephyr/ztest.h>
+#include <zephyr/drivers/adc.h>
+
+static const struct device *dev_adc = DEVICE_DT_GET(DT_ALIAS(adc));
+#define BUFFER_LEN 8
+uint16_t m_sample_buffer[BUFFER_LEN];
+
+static struct adc_channel_cfg valid_channel_cfg = {
+	.gain = ADC_GAIN_1,
+	.channel_id = 0,
+	.reference = ADC_REF_INTERNAL,
+	.acquisition_time = ADC_ACQ_TIME_DEFAULT,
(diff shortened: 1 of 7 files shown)


tests/drivers/adc/adc_error_cases/src/main.c
```diff
@@ -0,0 +1,244 @@
...
+
+ZTEST(adc_error_cases, test_adc_read_invalid_resolution)
+{
+	int ret;
+
+	adc_channel_setup(dev_adc, &valid_channel_cfg);
+
+	struct adc_sequence invalid_seq = valid_seq;
+	/* Set resolution to invalid value */
+	invalid_seq.resolution = 99;
```
nordic-segl, line 76: Assuming that valid resolutions are 8-12 bits. I wonder if edge values like 7 or 13 would be better to check. However, this might be incompatible with other vendors (f.e. if their ADC supports higher/lower resolutions).

tests/drivers/adc/adc_error_cases/src/main.c
```diff
@@ -0,0 +1,244 @@
...
+
+	struct adc_sequence invalid_seq = valid_seq;
+	/* Set oversampling to invalid value */
+	invalid_seq.oversampling = 99;
+
+	ret = adc_read(dev_adc, &invalid_seq);
+
+	zassert_true(
+		ret == -EINVAL,
+		"adc_read() should return True,"
```
nordic-segl, line 56: Why 'True'?
Return values
    0	On success.

tests/drivers/adc/adc_error_cases/src/main.c
nordic-babu: yes, idea was to create generic TCs, as there is no restriction for resolution values in zephyr documentation 7 or 13 might be implemented by other vendors

tests/drivers/adc/adc_error_cases/src/main.c
nordic-babu: good catch, it was copy-paste error, fixed.

nordic-segl (approved):

nordic-piks (approved):

PR #1910, opened 2024-07-24
samples: basic: blinky: add mbox backend for eGPIO

jaz1-nordic:


samples/basic/blinky/sample.yaml
```diff
@@ -10,3 +10,30 @@ tests:
...
     integration_platforms:
       - frdm_k64f
+
+  sample.basic.blinky.egpio_icmsg:
+    tags:
+      - LED
+      - gpio
+    filter: dt_enabled_alias_with_parent_compat("led0", "gpio-leds")
+    depends_on: gpio
+    harness: led
```
nordic-piks, line 20: There is no such harness `led`, it is dummy one.
Please change to `console` and add expected regex from sample.

nordic-piks: You can do sth similar to this:
https://github.com/zephyrproject-rtos/zephyr/pull/72730/files

nordic-piks (changes requested):
Please allow verification via console logs for blinky sample.

samples/basic/blinky/sample.yaml
```diff
@@ -10,3 +10,45 @@ tests:
...
+      - SB_CONFIG_EGPIO_FLPR_APPLICATION=y
+
+  sample.basic.blinky.egpio_mbox:
+    tags:
+      - LED
+      - gpio
+    filter: dt_enabled_alias_with_parent_compat("led0", "gpio-leds")
+    depends_on: gpio
+    integration_platforms:
+      - nrf54l15pdk/nrf54l15/cpuapp
```
nordic-piks, line 42: Can you add DK target too?

samples/basic/blinky/sample.yaml
```diff
@@ -10,3 +10,45 @@ tests:
...
       - frdm_k64f
+
+  sample.basic.blinky.egpio_icmsg:
+    tags:
+      - LED
+      - gpio
+    filter: dt_enabled_alias_with_parent_compat("led0", "gpio-leds")
+    depends_on: gpio
+    integration_platforms:
+      - nrf54l15pdk/nrf54l15/cpuapp
```
nordic-piks, line 21: Can you add DK target too?

nordic-piks (approved):

samples/basic/blinky/sysbuild.conf
```diff
@@ -0,0 +1 @@
+SB_CONFIG_PARTITION_MANAGER=n
```
masz-nordic, line 1: This could be moved to `sample.yaml` to limit impact.

magp-nordic (approved):

masz-nordic (approved):

magp-nordic (approved):

nordic-segl (approved):

carlescufi (dismissed):
why is this a noup instead of a fromlist?

gmarull (commented):
I don't think we should be carrying these sort of patches downstream. They do not seem to follow best practices, and will likely keep increasing upmerge times. Why isn't this developed directly upstream?

carlescufi:
Please use the alt-root instead: https://github.com/nrfconnect/sdk-nrf/tree/main/scripts/twister/alt/zephyr
Check with @PerMac for additional info about the twister alt-root we have in Zephyr.

PR #1913, opened 2024-07-25
Updates for radio driver with nRF54H20

piotrkoziar:
* Adds missing device tree configs for dppic and ipct connections between the Radio core and the Global domains.
* Changes the ramp-up time from 1000us to 550us on nRF54H20. The time must fit inside general preconditions ramp up.
The hfclk time can be adjusted this way because the current solution is not precise until the clock_control is available.

manifest-pr-skip


ankuns (approved):

hubertmis (approved):
Configuring DPPI and IPCT is much too complicated in nRF54H. We should figure out (out of the scope of this PR) how to simplify it.

PR #1920, opened 2024-07-29
[nrf fromlist] drivers: clock_control: Add support for nRF54H20 clock controllers

anangl:
Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/76430


nordic-piks:
To be able to test this PR, those tests needs to be enable (and aligned for nrf54h20):
https://github.com/nrfconnect/sdk-zephyr/tree/main/tests/drivers/clock_control/clock_control_api
https://github.com/nrfconnect/sdk-zephyr/tree/main/tests/drivers/clock_control/nrf_lf_clock_start

https://github.com/nrfconnect/sdk-zephyr/tree/main/tests/drivers/clock_control/onoff

https://github.com/nrfconnect/sdk-zephyr/tree/main/tests/drivers/clock_control/nrf_clock_calibration


Possibly other tests needs to be added if existing are not sufficient to cover possible use cases for dvfs and clock_control.

anangl:
Rebased.

samples/boards/nrf/clock_control/sample.yaml
```diff
@@ -0,0 +1,7 @@
+sample:
+  name: Clock control sample for nRF
+tests:
```
nordic-piks, line 3: Please add harness console and expected regex to enable automatic verification of this sample, e.g:
https://github.com/nrfconnect/sdk-zephyr/blob/bc15a7e16a7266f66a229d1673177e45f0457652/samples/drivers/adc/adc_dt/sample.yaml#L38-L44

drivers/clock_control/clock_control_nrf2_hsfll.c
```diff
@@ -0,0 +1,283 @@
...
+}
+
+static int hsfll_init(const struct device *dev)
+{
+	const struct hsfll_dev_config *dev_config = dev->config;
+	nrf_hsfll_trim_t trim = {
+		.vsup   = sys_read32(dev_config->trim_vsup),
+		.coarse = sys_read32(dev_config->trim_coarse),
+		.fine   = sys_read32(dev_config->trim_fine),
+	};
```
karl-nordic, line 209: Probably TCOEF is missing here. It was added in Lilium FP1. Potentially is not supported by HAL yet.

drivers/clock_control/clock_control_nrf2_hsfll.c
```diff
@@ -0,0 +1,283 @@
...
+/* TODO: get from Kconfig */
+#define UPDATE_TIMEOUT K_MSEC(2)
+
+static const struct clock_options {
+	uint32_t frequency;
+	enum dvfs_frequency_setting setting;
+} clock_options[] = {
+	/* TODO: move to devicetree? */
+	{
+		.frequency = MHZ(64),
```
karl-nordic, line 34: Frequency is one thing. The other is mode (OPEN/CLOSED/BYPASS). Please confirm that mode is skipped intentionally here.

anangl:
Rebased.

anangl:
Rebased.

drivers/clock_control/clock_control_nrf2_lfclk.c
```diff
@@ -0,0 +1,217 @@
...
+	nrfs_clock_src_t src;
+} clock_options[] = {
+	/* TODO: use real accuracy values, handle other LFXO configurations */
+	{
+		.accuracy = 100,
+		.precision = 0,
+		.src = NRFS_CLOCK_SRC_LFCLK_XO_PIXO
+	},
+	{
+		.accuracy = 50,
```
karl-nordic, line 34: Please remember and be ready that some sources have the same accuracy. For instance.  Difference between NRFS_CLOCK_SRC_LFCLK_XO_PIERCE and NRFS_CLOCK_SRC_LFCLK_XO_PIERCE_HP is that  HP have had better precision. Accuracy and frequance is the same.

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20-common.dtsi
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+#include "nrf54h20dk_nrf54h20-memory_map.dtsi"
+#include "nrf54h20dk_nrf54h20-ipc_conf.dtsi"
+#include "nrf54h20dk_nrf54h20-pinctrl.dtsi"
+
+&hfxo {
+	status = "okay";
+	accuracy-ppm = <30>;
+	start-up-time-us = <850>;
```
karl-nordic, line 14: My understanding is that such information will be read out from BICR and not fixed in DT. Is my understanding correct ?

hubertmis: BICR must be generated based on data from .dts. I understand this `start-up-time-us` property should be used to generate BICR.

Then the Application CPU can use both sources because both .dts and BICR should contain the same value for the application build.
Other software components built separately from the Application must use BICR because they don't have access to the Application's .dts including HFXO properties.

bjarki-andreasen:
This PR continues from #1963

PR #1925, opened 2024-07-31
hal: nrf-regtool: Use sysbuild to merge UICRs

tomchy:
Use sysbuild functionalities to merge UICRs with local domain firmware.

Ref: NCSDK-NONE


tomchy:
After some thought - it'd be better if the `ncs_merge_file` was capable of handling HEX files merged using `runners_yaml_props_target` override instead.

PR #1944, opened 2024-08-06
[nrf noup] bluetooth: conn: Skip buffer ref count check in send_buf

MarekPieta:
If ATT sent callback is delayed until data transmission is done by BLE controller, the transmitted buffer may have an additional reference. The reference is used to extend lifetime of the net buffer until the data transmission is confirmed by ACK of the remote.

send_buf function can be called multiple times, if buffer has to be fragmented over HCI. In that case, the callback is provided as an argument only for the last transmitted fragment. The `buf->ref == 1` check is skipped because it's impossible to properly validate number of references for the sent fragments if buffers may have the additional reference.

Jira: NCSDK-28624

### DIFF
--- a/subsys/bluetooth/host/conn.c
+++ b/subsys/bluetooth/host/conn.c
@@ -654,13 +654,17 @@ static int send_buf(struct bt_conn *conn, struct net_buf *buf,
 
 	uint16_t frag_len = MIN(conn_mtu(conn), len);
 
-	if (buf->ref > 1 + (cb ? 1 : 0)) {
-		/* Allow for an additional buffer reference if callback is provided.
-		 * This can be used to extend lifetime of the net buffer until the
-		 * data transmission is confirmed by ACK of the remote.
-		 */
-		__ASSERT_NO_MSG(false);
-	}
+	/* If ATT sent callback is delayed until data transmission is done by BLE controller, the
+	 * transmitted buffer may have an additional reference. The reference is used to extend
+	 * lifetime of the net buffer until the data transmission is confirmed by ACK of the remote.
+	 *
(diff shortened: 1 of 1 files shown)


zycz (approved):

jori-nordic (changes requested):
Writing "to prevent spurious assertion failures" right above the assert in question looks like are "fixing the test" instead of fixing the root cause. 

Since this is a `noup`, I suggest removing the assert altogether and having this sentence in the commit message
 
> `send_buf()` can be called multiple times, if `buf` has to be fragmented over HCI. `buf` will have an additional reference taken. We cannot know if the application has provided a callback until the very last fragment. That means we can't assert on the amount of references for fragments here.

MarekPieta:
> Writing "to prevent spurious assertion failures" right above the assert in question looks like are "fixing the test" instead of fixing the root cause.
> 
> Since this is a `noup`, I suggest removing the assert altogether and having this sentence in the commit message
> 
> > `send_buf()` can be called multiple times, if `buf` has to be fragmented over HCI. `buf` will have an additional reference taken. We cannot know if the application has provided a callback until the very last fragment. That means we can't assert on the amount of references for fragments here.

Actually the commit aligns the assertion to updated implementation (fixing test condition in the assertion). If you insist, I will remove the assertion, but I would suggest to keep it (it validates ref counter properly for the last fragment). I rephrased my comments taking into account your suggestions.

jori-nordic:
> it validates ref counter properly for the last fragment

yes but the assert was supposed to validate it for every frag. It is weakened enough that we should just remove it altogether.

jori-nordic (approved):
🚀

zycz (approved):

alwa-nordic (approved):

PR #1956, opened 2024-08-19
[nrf fromlist] net: openthread: Fix missing dependency for OT `diag` …

canisLupus1313:
Commit fixes missing dependency for `ot diag cw` command for nrf5xx by adding default vale to `NRF_802154_CARRIER_FUNCTIONS`.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/77001

### DIFF
--- a/modules/hal_nordic/Kconfig
+++ b/modules/hal_nordic/Kconfig
@@ -204,6 +204,7 @@ config NRF_802154_SECURITY_KEY_STORAGE_SIZE
 
 config NRF_802154_CARRIER_FUNCTIONS
 	bool "nRF 802.15.4 carrier functions"
+	default y if OPENTHREAD_DIAG
 	help
 	  This option enables functions such as modulated carrier and continuous carrier.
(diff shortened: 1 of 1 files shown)


modules/hal_nordic/Kconfig
```diff
@@ -204,6 +204,7 @@ config NRF_802154_SECURITY_KEY_STORAGE_SIZE
 
 config NRF_802154_CARRIER_FUNCTIONS
 	bool "nRF 802.15.4 carrier functions"
+	default y if OPENTHREAD_DIAG
```
hubertmis, line 207: Should `OPENTHREAD_DIAG` `depend on NRF_802154_CARRIER_FUNCTIONS`? Someone could disable this default which, if I understand correctly, would break the build

hubertmis (approved):
Looks good, but for a proper fix more logic could be needed.

alxelax (approved):

PR #1966, opened 2024-08-22
west.yml: Update zcbor to 0.9.0

oyvindronningstad:
from 0.8.2


oyvindronningstad:
Now included in the upmerge

PR #1968, opened 2024-08-23
[nrf noup] board: nordic_ thingy53: Enable QSPI by default

hellesvik-nordic:
The Thingy:53 enabled MCUboot and external flash by default. Therefore, it should also enable drivers for its external flash by default.


boards/nordic/thingy53/Kconfig.defconfig
```diff
@@ -145,6 +145,10 @@ config MCUBOOT_USB_SUPPORT
 	bool
 	default n
 
+config NORDIC_QSPI_NOR
+	bool
```
gmarull, line 149:
```suggestion
```

nordicjm (approved):

PR #1971, opened 2024-08-26
[nrf fromlist] tests: bluetooth: tester: refactor CSIP btp command

Frodevan:
_Note: The corresponding changes from this are already in our auto-pts repo, so this should fix a discrepancy..._

Refactor CSIP btp commands by removing the 'count' parameter from set_coordinator_lock and set_coordinator_release functions. If we want to support lock/release procedure on subset of set members in the future, 'count' param doesn't indicate a specific set member. Instead, introduce 'address_count' and 'addr' array params, although their usage is not yet implemented.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/71059


Thalley (approved):

koffes (approved):

Frodevan:
Rebasing manifest ci branch

PR #1975, opened 2024-08-27
nRF2 clock control sample patch

bjarki-andreasen:
Adjust clock frequencies in sample to match the 16MHz produced by the FLL16M.

This PR reverts the [nrf fromlist] commit which adds the test suite, and re-applies it with the changes to match the upstream commit.


carlescufi:
Superseed #1981

PR #1979, opened 2024-08-27
[nrf noup] wifi_mgmt: Add support for PS data retrieval mechanism config

ajayparida:
In PS mode, buffered data can be retrieved in two different ways, add an API to configure the mechanism at runtime.


krish2718 (approved):

sachinthegreen (approved):

ajayparida:
This chnage is already part of Zephyr(pulled in in some other commit). So closing this.

PR #1983, opened 2024-08-29
[nrf noup] soc: nrf54l: Disable CRACEN RAMs on system off

Vge0rge:
Configure the CRACEN internal RAMs to avoid being retained during system off. Exiting system off triggers a system level reset so there is no need for the CRACEN RAMs to be retained during that.

Noup since the upstream code is not TFM aware yet and the edited function is not upstream.


soc/nordic/nrf54l/soc.c
```diff
@@ -171,6 +172,17 @@ int nordicsemi_nrf54l_init(void)
 #if !defined(CONFIG_TRUSTED_EXECUTION_NONSECURE) || defined(__NRF_TFM__)
 	/* Currently not supported for non-secure */
 	SystemCoreClockUpdate();
+
+	/* Don't retain CRACEN RAMs during System OFF */
+	nrf_memconf_ramblock_ret_mask_enable_set(NRF_MEMCONF, 1, (1 << 2), false);
+	nrf_memconf_ramblock_ret_mask_enable_set(NRF_MEMCONF, 1, (1 << 3), false);
+
+	/* Don't retain CRACEN RAMs during System ON when CRACEN is not used */
```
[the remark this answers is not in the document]
Vge0rge: I am not sure if this is correct actually. The Cracen RAMs are never powered (thats the term in the document which is similar to enabled I guess) during the System on idle or System off no matter what is the configuration.  They can be retained while they are powered off depended on the configuration.

soc/nordic/nrf54l/soc.c
```diff
@@ -171,6 +172,17 @@ int nordicsemi_nrf54l_init(void)
 #if !defined(CONFIG_TRUSTED_EXECUTION_NONSECURE) || defined(__NRF_TFM__)
 	/* Currently not supported for non-secure */
 	SystemCoreClockUpdate();
+
+	/* Don't retain CRACEN RAMs during System OFF */
+	nrf_memconf_ramblock_ret_mask_enable_set(NRF_MEMCONF, 1, (1 << 2), false);
+	nrf_memconf_ramblock_ret_mask_enable_set(NRF_MEMCONF, 1, (1 << 3), false);
```
tomi-font, line 178: Could those `(1 << X)` be defines to explain what they are?

tomi-font (approved):

soc/nordic/nrf54l/soc.c
MarkusLassila: Could MEMCONF_POWER_RET_MEM2_Msk, MEMCONF_POWER_RET_MEM3_Msk and MEMCONF_POWER_CONTROL_MEM2_Msk and MEMCONF_POWER_CONTROL_MEM3_Msk be utilized for this?

masz-nordic: Please don't mix MDK defines with HAL functions, if you think those masks are needed we can add them to HAL.

SeppoTakalo (approved):

soc/nordic/nrf54l/soc.c
Vge0rge: Yeah, I will not mix these up indeed. But I will write some  more description on the bits that I am setting now that I got a confirmation from the spec writers that it is ok to do so.

Vge0rge:
> @Vge0rge is this PR still valid?

It is but I didn't have a chance to work on it for some time now :/

tomi-font:
@Vge0rge Is this still relevant?

PR #1986, opened 2024-08-30
Remove Breathe

gmarull:


nordicjm (approved):

doc/conf.py
```diff
@@ -73,7 +73,6 @@
 # -- General configuration ------------------------------------------------
 
 extensions = [
-    "breathe",
```
thst-nordic, line 76: why was breathe removed?  
Will it affect nordic's documentation team?

de-nordic (approved):
MCUmgr doc OK

doc/conf.py
gmarull: Breathe was a resource hog for the project, so an alternative has been introduced. We are all aware and already have a working solution that is to serve Doxygen content as upstream does. Btw, documentation CI will be much faster, so no workarounds to make it _fast_ will be needed.

PR #1989, opened 2024-09-01
[nrf fromtree] tests: bluetooth: tester: add nrf54h20 configs

AyturkDuzen:
Added:
nrf54h20dk_nrf54h20_cpuapp.overlay: Choose uart136 hwfc 115200 nrf54h20dk_nrf54h20_cpuapp.conf


hermabe (approved):
Not sure if the "cherry picked from" line is correct, but otherwise LGTM

EDIT: Should be https://github.com/zephyrproject-rtos/zephyr/commit/5f00c18005ac337b6daee94b2d3d8eabb8bb50a0 instead

Thalley:
I'd suggest adding more reviewers as well. I'm not familiar with the change nor this process

Thalley:
@hermabe can you help adding the right reviewers for this? I don't think I nor @asbjornsabo are the right ones

rugeGerritsen (approved):

theob-pro:
I missed the Zephyr PR but I see that it's pretty much the same as the 53 overlay/config. But why `CONFIG_BT_TINYCRYPT_ECC` is not selected here? Is it because 54h20 use the PSA Crypto backend?

AyturkDuzen:
> I missed the Zephyr PR but I see that it's pretty much the same as the 53 overlay/config. But why `CONFIG_BT_TINYCRYPT_ECC` is not selected here? Is it because 54h20 use the PSA Crypto backend?

CONFIG_BT_TINYCRYPT_ECC is selected as part of default prj.conf here https://github.com/nrfconnect/sdk-zephyr/blob/main/tests/bluetooth/tester/prj.conf#L39

theob-pro (approved):

PR #1995, opened 2024-09-03
[nrf noup] soc: nordic: add delay after wake up

adamkondraciuk:
An additional delay after wake-up is needed.

### DIFF
--- a/soc/nordic/nrf54h/power.c
+++ b/soc/nordic/nrf54h/power.c
@@ -63,6 +63,7 @@ void nrf_poweroff(void)
 /* Resume domain after local suspend to RAM. */
 static void sys_resume(void)
 {
+	k_busy_wait(100);
 	if (IS_ENABLED(CONFIG_ICACHE)) {
 		/* Power up and re-enable ICACHE */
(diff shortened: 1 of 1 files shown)


nordic-piks (approved):

masz-nordic:
Can you add a comment with why this is needed?

anangl (changes requested):
It should be justified why this delay is needed and why it is not applied upstream.

soc/nordic/nrf54h/power.c
```diff
@@ -33,7 +33,6 @@ static void suspend_common(void)
 
 	/* Disable retention */
 	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
```
karstenkoenig, line 36: This change looks unrelated, and neither change has a comment that marks it as workaround or temporary when a better solution is in place

karstenkoenig:
To my understanding we shouldn't do the delay method but just wait for a proper fix in sysctrl that doesn't turn on closed loop for fll16m. or is this supposed to fix something else?

PR #1996, opened 2024-09-04
[nrf fromtree] Bluetooth: Mesh: Clarify log to run out of retransmiss…

tot0rokr:
…ion attempt

The logs for retransmission attempts for sending segments and ack retransmission attempts for receiving segments were the same, making it difficult to distinguish between them. This can be solved by adding information to the log of ack retransmission.

Signed-off-by: Junho Lee <tot0roprog@gmail.com>
(cherry picked from commit fe24cac70b5d6e587b051fe4b945696a9cef5e80)

### DIFF
--- a/subsys/bluetooth/mesh/transport.c
+++ b/subsys/bluetooth/mesh/transport.c
@@ -1220,7 +1220,7 @@ static void seg_ack(struct k_work *work)
 	rx->last_ack = k_uptime_get_32();
 
 	if (rx->attempts_left == 0) {
-		LOG_DBG("Ran out of retransmit attempts");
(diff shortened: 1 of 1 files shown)


PavelVPV:
Hi @tot0rokr , thanks for your contribution! I wonder if we really need this now and can't wait for next upmerge. Or, at least, when we manually upmerging changes from upstream I think we need to pull other intermediate commits as well so to avoid potential conflicts between commits.

tot0rokr:
@PavelVPV I see what you're saying.
It's not a critical commit right now, so I'll look forward to seeing it merged together when other important commits are merged in the future.

PR #2010, opened 2024-09-06
[nrf fromlist] drivers: timer: grtc: Switch GRTC clock source to LFXO

adamkondraciuk:
GRTC needs to use direct clock source path instead of system clock path to support ELV mode for nRF54L targets.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/78082


martintv (approved):
tested and works with lfxo

PR #2012, opened 2024-09-09
[nrf noup] mgmt: mcumgr: grp: img_mgmt: Add handling for QSPI XIP

nordicjm:
Adds handling code to allow selecting the correct image slot when using QSPI XIP in DirectXIP mode

manifest-pr-skip


subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt.c
```diff
@@ -159,6 +159,12 @@ int img_mgmt_active_slot(int image)
 	/* Multi image does not support DirectXIP currently */
```
gchwier, line 159: so this comment should be removed now

nordicjm: The comment is still accurate, this is a "hack" for one specific mode

gchwier (approved):

PR #2003, opened 2024-09-04
Nrfx 6118 icbmsg align to no-multithreading

magp-nordic:
Align icbmsg to work in no-multithreading.
~Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/78188~ merged already
Upstream PR for the last commit: https://github.com/zephyrproject-rtos/zephyr/pull/79382

PR #2025, opened 2024-09-13
Nrfx 6118 add icbmsg to egpio

magp-nordic:
Based on #2003 and #1910, only the last commit is relevant to this PR.


magp-nordic:
Changes moved to: https://github.com/nrfconnect/sdk-nrf/pull/17321/commits

PR #2030, opened 2024-09-18
soc: nordic: nrf54l: integrate normal voltage mode fixes

nika-nordic:
MDK 8.67.0 deprecates normal voltage mode for nRF54L, so it should no longer be used by the end users.


nika-nordic:
Already merged as a part of another PR

PR #2032, opened 2024-09-18
[nrf fromlist] soc: nordic: nrf54h20: Add LRC around idle

adamkondraciuk:
Add LRC domain power management around idle. The Fast Clock Domain needs to be forced-on during CPU idle. Also the force needs to be disabled on idle-exit.

### DIFF
--- a/soc/nordic/common/soc_lrcconf.c
+++ b/soc/nordic/common/soc_lrcconf.c
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <soc_lrcconf.h>
+#include <zephyr/kernel.h>
+
+static struct k_spinlock lock;
+static sys_slist_t poweron_main_list;
+static sys_slist_t poweron_active_list;
+
+void soc_lrcconf_poweron_request(sys_snode_t *node, nrf_lrcconf_power_domain_mask_t domain)
+{
+	__ASSERT(is_power_of_two(domain), "Only one bit can be set for the domain parameter");
+
+	sys_slist_t *poweron_list;
+
+	if (domain == NRF_LRCCONF_POWER_MAIN) {
+		poweron_list = &poweron_main_list;
+	} else if (domain == NRF_LRCCONF_POWER_DOMAIN_0) {
+		poweron_list = &poweron_active_list;
+	} else {
+		return;
+	}
+	K_SPINLOCK(&lock) {
+		if (sys_slist_len(poweron_list) == 0) {
+			nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, domain, true);
+		}
+
+		sys_slist_find_and_remove(poweron_list, node);
+		sys_slist_append(poweron_list, node);
+	}
+}
+
+void soc_lrcconf_poweron_release(sys_snode_t *node, nrf_lrcconf_power_domain_mask_t domain)
+{
+	__ASSERT(is_power_of_two(domain), "Only one bit can be set for the domain parameter");
+
+	sys_slist_t *poweron_list;
+
+	if (domain == NRF_LRCCONF_POWER_MAIN) {
+		poweron_list = &poweron_main_list;
+	} else if (domain == NRF_LRCCONF_POWER_DOMAIN_0) {
+		poweron_list = &poweron_active_list;
+	} else {
+		return;
+	}
+
+	K_SPINLOCK(&lock) {
+		if (!sys_slist_find_and_remove(poweron_list, node)) {
+			K_SPINLOCK_BREAK;
+		}
+
+		if (sys_slist_len(poweron_list) > 0) {
+			K_SPINLOCK_BREAK;
+		}
+		nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, domain, false);
+	}
+}
--- a/soc/nordic/common/soc_lrcconf.h
+++ b/soc/nordic/common/soc_lrcconf.h
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file nRF SoC specific helpers for lrcconf management
+ */
+
+#ifndef ZEPHYR_SOC_NORDIC_COMMON_LRCCONF_H_
+#define ZEPHYR_SOC_NORDIC_COMMON_LRCCONF_H_
+
+#include <hal/nrf_lrcconf.h>
+
+/**
+ * @brief Request lrcconf power domain
+ *
+ * @param node	 Pointer to the @ref sys_snode_t structure which is the ID of the
+ *		 requesting module.
+ * @param domain The mask that represents the power domain ID.
+ */
+void soc_lrcconf_poweron_request(sys_snode_t *node, nrf_lrcconf_power_domain_mask_t domain);
+
+/**
+ * @brief Release lrcconf power domain
+ *
+ * @param node	 Pointer to the @ref sys_snode_t structure which is the ID of the
+ *		 requesting module.
+ * @param domain The mask that represents the power domain ID.
+ */
+void soc_lrcconf_poweron_release(sys_snode_t *node, nrf_lrcconf_power_domain_mask_t domain);
+
+#endif /* ZEPHYR_SOC_NORDIC_COMMON_LRCCONF_H_ */
--- a/soc/nordic/nrf54h/power.c
+++ b/soc/nordic/nrf54h/power.c
@@ -9,20 +9,26 @@
 #include <zephyr/pm/policy.h>
 #include <zephyr/arch/common/pm_s2ram.h>
 #include <hal/nrf_resetinfo.h>
-#include <hal/nrf_lrcconf.h>
 #include <hal/nrf_memconf.h>
 #include <zephyr/cache.h>
 #include <power.h>
+#include <soc_lrcconf.h>
+#include "soc.h"
 #include "pm_s2ram.h"
 
-static void suspend_common(void)
+static void common_suspend(void)
 {
+	sys_snode_t *node;
 
-	/* Flush, disable and power down DCACHE */
-	sys_cache_data_flush_all();
-	sys_cache_data_disable();
-	nrf_memconf_ramblock_control_enable_set(NRF_MEMCONF, RAMBLOCK_POWER_ID,
-						RAMBLOCK_CONTROL_BIT_DCACHE, false);
+	node = soc_pd_sys_snode_get();
+
+	if (IS_ENABLED(CONFIG_DCACHE)) {
+		/* Flush, disable and power down DCACHE */
+		sys_cache_data_flush_all();
+		sys_cache_data_disable();
+		nrf_memconf_ramblock_control_enable_set(NRF_MEMCONF, RAMBLOCK_POWER_ID,
+							RAMBLOCK_CONTROL_BIT_DCACHE, false);
+	}
 
 	if (IS_ENABLED(CONFIG_ICACHE)) {
 		/* Disable and power down ICACHE */
@@ -31,38 +37,15 @@ static void suspend_common(void)
 							RAMBLOCK_CONTROL_BIT_ICACHE, false);
 	}
 
-	/* Disable retention */
-	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
+	soc_lrcconf_poweron_release(node, NRF_LRCCONF_POWER_DOMAIN_0);
 }
 
-void nrf_poweroff(void)
+static void common_resume(void)
 {
-	nrf_resetinfo_resetreas_local_set(NRF_RESETINFO, 0);
-	nrf_resetinfo_restore_valid_set(NRF_RESETINFO, false);
+	sys_snode_t *node;
 
-	nrf_lrcconf_retain_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, false);
+	node = soc_pd_sys_snode_get();
 
-	/* TODO: Move it around k_cpu_idle() implementation. */
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN, false);
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0, false);
-
-	suspend_common();
-
-	nrf_lrcconf_task_trigger(NRF_LRCCONF010, NRF_LRCCONF_TASK_SYSTEMOFFREADY);
-
-	__set_BASEPRI(0);
-	__ISB();
-	__DSB();
-	__WFI();
-
-	CODE_UNREACHABLE;
-}
-
-#if IS_ENABLED(CONFIG_PM_S2RAM)
-/* Resume domain after local suspend to RAM. */
-static void sys_resume(void)
-{
 	if (IS_ENABLED(CONFIG_ICACHE)) {
 		/* Power up and re-enable ICACHE */
 		nrf_memconf_ramblock_control_enable_set(NRF_MEMCONF, RAMBLOCK_POWER_ID,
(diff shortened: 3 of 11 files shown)


soc/nordic/nrf54h/soc.c
```diff
@@ -122,6 +122,8 @@ bool z_arm_on_enter_cpu_idle(void)
 #ifdef CONFIG_LOG_FRONTEND_STMESP
 	log_frontend_stmesp_pre_sleep();
 #endif
+	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0,
+			!IS_ENABLED(CONFIG_SOC_NRF54H20_CPURAD));
```
hubertmis, line 126: This should be forced also for CPURAD.

I think we should have symmetry here. If restoring this register is done in `pm_state_exit_post_ops()`, can we configure it in `pm_state_enter_pre_ops()` or something like this?

hubertmis: Also, this would conflict with clock_control configuring the same registers: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/clock_control/clock_control_nrf2_common.c#L172-L173

soc/nordic/nrf54h/Kconfig.soc
```diff
@@ -6,6 +6,7 @@
 config SOC_NRF54H20
 	bool
 	select SOC_SERIES_NRF54HX
+	select ARM_ON_ENTER_CPU_IDLE_HOOK if PM_S2RAM
```
hubertmis, line 9:
```suggestion
	select ARM_ON_ENTER_CPU_IDLE_HOOK
```

The hook is not required for S2RAM, but for S2IDLE

soc/nordic/nrf54h/soc.c
hubertmis: Just checked with PAC. In Radio CPU we don't need to `force_set()` to enter S2IDLE. So this condition is fine after all.

soc/nordic/nrf54h/power.c
```diff
@@ -141,5 +138,8 @@ void pm_state_set(enum pm_state state, uint8_t substate_id)
 
 void pm_state_exit_post_ops(enum pm_state state, uint8_t substate_id)
 {
+	if (state == PM_STATE_SUSPEND_TO_IDLE) {
```
hubertmis, line 141: It turned out, forcing POWERON is not required in the Radio CPU.

```suggestion
	if (state == PM_STATE_SUSPEND_TO_IDLE && !IS_ENABLED(CONFIG_SOC_NRF54H20_CPURAD)) {
```

soc/nordic/nrf54h/soc.c
adamkondraciuk: > ~This should be forced also for CPURAD.~
> 
> I think we should have symmetry here. If restoring this register is done in `pm_state_exit_post_ops()`, can we configure it in `pm_state_enter_pre_ops()` or something like this?
The only "symmetric" function I've found is [pm_state_set](https://github.com/nrfconnect/sdk-zephyr/blob/main/include/zephyr/pm/pm.h#L139) which is defined for nRF54H20 [here](https://github.com/nrfconnect/sdk-zephyr/blob/main/soc/nordic/nrf54h/power.c#L131)

hubertmis: Ok, then we should refactor k_cpu_idle() in such a way that it has symmetric hooks called from within its critical section. And we should aim to put this code in these hooks.

But until we get there I'm fine with currently used hooks. Hopefully, it unblocks progress with Bluetooth and other cases.

soc/nordic/nrf54h/Kconfig.soc
adamkondraciuk: > The hook is not required for S2RAM, but for S2IDLE

The default configuration ensures us, that for APP the DOMAIN_0 is powered on : https://github.com/nrfconnect/sdk-zephyr/blob/main/soc/nordic/nrf54h/soc.c#L55-L58
For RADIO it is not required as you said - do we assume, that something will change this setting on the fly then? Or the initial setting is invalid and should be changed?

hubertmis (changes requested):
The resource usage conflict with clock_control seems to be still in place. Have a look at these functions:
https://github.com/nrfconnect/sdk-zephyr/blob/main/drivers/clock_control/clock_control_nrf2_common.c#L169-L198

soc/nordic/nrf54h/soc.c
ppryga-nordic: This code makes `NRF_LRCCONF_POWER_DOMAIN_0` in Radio core to be always set to auto, no matter of its original state. That could collide with any other user of LCRCONF010 like clock control. Or am I wrong here?

soc/nordic/common/soc_lrcconf.h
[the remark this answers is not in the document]
adamkondraciuk: for now  - yes, but some of other devices should
use it to

hubertmis (approved):

soc/nordic/nrf54h/soc.c
karstenkoenig: > Ok, then we should refactor k_cpu_idle() in such a way that it has symmetric hooks called from within its critical section. And we should aim to put this code in these hooks.
> 
> But until we get there I'm fine with currently used hooks. Hopefully, it unblocks progress with Bluetooth and other cases.

If only suspend-to-ram state is defined as power state and nothing else the system will still call `k_cpu_idle()` 
https://github.com/nrfconnect/sdk-zephyr/blob/main/kernel/idle.c#L85 and request poweron and never end up releasing it. This might still be acceptable to unblock the issue at hand but should still have a solution soon, otherwise we'll see different power consumptions based on how idle is used in combination with the pm subsystem.

karstenkoenig: It will even happen when idle is defined as a pm state, as it might have an expected idle time below the threshold, such as in the case of our example overlay file here
https://github.com/nrfconnect/sdk-nrf/blob/84492d6d9264745050a018d01f4f333b285864c1/tests/benchmarks/multicore/idle/boards/nrf54h20dk_nrf54h20_cpuapp_s2ram.overlay#L12C39-L12C39
so a `k_msleep(10)` will not release the request

soc/nordic/nrf54h/soc.c
```diff
@@ -52,13 +60,22 @@ static void power_domain_init(void)
...
-
+#if !IS_ENABLED(CONFIG_SOC_NRF54H20_CPURAD)
+	sys_snode_t *node;
+#if IS_ENABLED(CONFIG_PM) || IS_ENABLED(CONFIG_POWEROFF)
+	node = pm_sys_node_id_get();
+#else
+	node = &soc_node;
+#endif
+	soc_lrcconf_poweron_request(node, NRF_LRCCONF_POWER_DOMAIN_0);
+	soc_lrcconf_poweron_request(node, NRF_LRCCONF_POWER_MAIN);
```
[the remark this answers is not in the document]
adamkondraciuk: Don't we need those when `CONFIG_PM=n` and `CONFIG_POWEROFF=n`? In such case there is no LRCCONF management when the CPU goes to idle (maybe should be?) . If the `clock_control` module will call `soc_lrcconf_poweron_release()` then the domain loses its power when idle... Correct?

soc/nordic/nrf54h/Kconfig
```diff
@@ -28,6 +28,7 @@ config SOC_NRF54H20_CPUAPP
 	select NRFS_HAS_VBUS_DETECTOR_SERVICE
 	select HAS_PM
 	select HAS_POWEROFF
+	select ARM_ON_ENTER_CPU_IDLE_HOOK if PM || POWEROFF
```
hubertmis, line 31: Is this still required considering we changed the default to POWERON.ACT=1

soc/nordic/nrf54h/soc.c
```diff
@@ -29,6 +31,12 @@ LOG_MODULE_REGISTER(soc, CONFIG_SOC_LOG_LEVEL);
 #define HSFLL_NODE DT_NODELABEL(cpurad_hsfll)
 #endif
 
+#if !(defined(CONFIG_PM) || defined(CONFIG_POWEROFF))
+#if !defined(CONFIG_SOC_NRF54H20_CPURAD)
+static sys_snode_t soc_node;
+#endif
+#endif
+
```
hubertmis, line 39: Instead of these `#if defined`s, how about a function like `const * sys_snode_t soc_act_pd_sys_snode_get(void)` and define this sys_snode_t always? This would also invert the dependencies: from pm to soc - like it should be.

soc/nordic/nrf54h/soc.c
```diff
@@ -52,13 +60,17 @@ static void power_domain_init(void)
...
 
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_DOMAIN_0,
-			!IS_ENABLED(CONFIG_SOC_NRF54H20_CPURAD));
-	nrf_lrcconf_poweron_force_set(NRF_LRCCONF010, NRF_LRCCONF_POWER_MAIN,
-			!IS_ENABLED(CONFIG_SOC_NRF54H20_CPURAD));
-
+#if !defined(CONFIG_SOC_NRF54H20_CPURAD)
+#if !(defined(CONFIG_PM) || defined(CONFIG_POWEROFF))
+	soc_lrcconf_poweron_request(&soc_node, NRF_LRCCONF_POWER_DOMAIN_0);
+#endif /* !(defined(CONFIG_PM) || defined(CONFIG_POWEROFF)) */
```
hubertmis, line 66: This should apply to the Radio CPU as well to make sure the cache is properly handled in there.

soc/nordic/nrf54h/soc.c
```diff
@@ -122,6 +134,9 @@ bool z_arm_on_enter_cpu_idle(void)
 #ifdef CONFIG_LOG_FRONTEND_STMESP
 	log_frontend_stmesp_pre_sleep();
 #endif
+#if defined(CONFIG_PM) || defined(CONFIG_POWEROFF)
+	pm_state_enter_idle_prepare();
+#endif
```
hubertmis, line 139: Assuming, the `soc_node` getter function is created, this can be removed.

soc/nordic/nrf54h/power.h
```diff
@@ -10,11 +10,20 @@
 #ifndef _ZEPHYR_SOC_ARM_NORDIC_NRF_POWER_H_
 #define _ZEPHYR_SOC_ARM_NORDIC_NRF_POWER_H_
 
+#include <zephyr/sys/slist.h>
+
```
hubertmis, line 14: This include looks unused in this header file

soc/nordic/nrf54h/power.c
```diff
@@ -126,20 +125,32 @@ static void do_suspend_to_ram(void)
...
 #endif
 }
 
 void pm_state_exit_post_ops(enum pm_state state, uint8_t substate_id)
 {
+#if !defined(CONFIG_SOC_NRF54H20_CPURAD)
+	if (state == PM_STATE_SUSPEND_TO_IDLE) {
+		soc_lrcconf_poweron_release(&pm_node, NRF_LRCCONF_POWER_DOMAIN_0);
+	}
+#endif
```
hubertmis, line 147: This seems unneeded now. When exiting suspend to idle, the ACT PD should remain active.

soc/nordic/nrf54h/power.c
```diff
@@ -31,21 +35,20 @@ static void suspend_common(void)
 						RAMBLOCK_CONTROL_BIT_ICACHE, false);
 	}
```
hubertmis, line 37: I'm missing `soc_lrcconf_poweron_release(..., NRF_LRCCONF_POWER_DOMAIN_0);` around here

soc/nordic/nrf54h/power.c
```diff
@@ -77,12 +80,10 @@ static void sys_resume(void)
 		sys_cache_data_enable();
 	}
```
hubertmis, line 82: I'm missing `soc_lrcconf_poweron_request(..., NRF_LRCCONF_POWER_DOMAIN_0);` around here

hubertmis (changes requested):

soc/nordic/nrf54h/soc.c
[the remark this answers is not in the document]
hubertmis: The idea was to move it to idle hooks. But recently, we changed the decision about the default LRCCONFIG settings around the active power domains. So this is irrelevant. I didn't another review iteration with the new design in mind.

hubertmis (changes requested):

soc/nordic/nrf54h/soc.c
hubertmis: According to my understanding, you don't need to enable retention in radio lrcconfs. However, you need to force the domain on to keep cache powered. If not, the cace would lose its state when cpu gets retained

soc/nordic/nrf54h/Kconfig
adamkondraciuk: Yes - the hook is applied when DOMAIN_0 is not powered all the time - for proper idle execution. When no PM API is used, the domain is always on and no hook is required

soc/nordic/nrf54h/power.c
karstenkoenig: Couldn't we just simplify this and turn on the ACT PD on startup and just turn it off when entering s2ram or softoff? No need for idle hooks etc? You are right the active power domain needs to be up anyways except for entering suspend to ram or soft off

soc/nordic/nrf54h/power.c
```diff
@@ -59,30 +85,59 @@ void nrf_poweroff(void)
...
-		/* Power up and re-enable ICACHE */
-		nrf_memconf_ramblock_control_enable_set(NRF_MEMCONF, RAMBLOCK_POWER_ID,
-							RAMBLOCK_CONTROL_BIT_ICACHE, true);
-		sys_cache_instr_enable();
+	sys_snode_t *node;
+
+	node = soc_pd_sys_snode_get();
+
+	switch (substate_id) {
+	case 0: /* Substate for idle with cache retained - not implemented yet. */
```
hubertmis, line 95: 0 should be for "Idle" - like calling `k_cpu_idle()`
1 should be reserved for the cache retained in RAM
2 for the cache disabled

hubertmis (changes requested):
Just please align the sub-states ids with the design. Other than that it looks good!

soc/nordic/nrf54h/power.c
adamkondraciuk: Isn't it like, that "idle like calling `k_cpu_idle()` " is the default state performed when no circumstances for other states to be used. I noticed that when I call `k_sleep(t)`, where `t < [any of residency times defined]`, function `s2idle_enter()` is not executed, but regular idle thread is called (so [here](https://github.com/nrfconnect/sdk-zephyr/blob/main/arch/arm/core/cortex_m/cpu_idle.c#L94) So does it make sense to define such state (or substate)?

adamkondraciuk:
> Just please align the sub-states ids with the design. Other than that it looks good!

How about overlays [here](https://github.com/nrfconnect/sdk-nrf/pull/17512/files#diff-beafb3e3aa8a3c86dc6513c1dcabafce9c0b0246bda1515309c98ba19d1b6285) ? Maybe those states could be defined in DTS instead?

soc/nordic/nrf54h/power.c
hubertmis: I don't know internals of zephyr pm, if it bypasses the policy module and calls k_cpu_idle by itself in some circumstances. By still, to align the implementation with the design, I would prefer to change the substates enumeration. Even if the substate 0 would be unused

karstenkoenig: The idle thread doesn't bypass the power module strictly speaking, but it will call `k_cpu_idle()` when `pm_suspend(ticks)` returns false, and that is if ticks is below any of the defined thresholds of the power states. See [here](https://github.com/nrfconnect/sdk-zephyr/blob/main/kernel/idle.c#L85)

hubertmis: Ok, so the problem is that the "Idle" power state is not defined and not handled by the PM subsystem and because of this the idle thread falls back to `k_cpu_idle()`. So then for clear implementation, I would suggest implementing all power states in the PM so that `pm_suspend(ticks)` always returns `true`. For very short sleep, it can even select the `ACTIVE` state and return `true`.

hubertmis:
> How about overlays [here](https://github.com/nrfconnect/sdk-nrf/pull/17512/files#diff-beafb3e3aa8a3c86dc6513c1dcabafce9c0b0246bda1515309c98ba19d1b6285) ? Maybe those states could be defined in DTS instead?

Yes, all power states should be defined in the .soc device tree files. Not in the sample overlays.

karstenkoenig:
> > How about overlays [here](https://github.com/nrfconnect/sdk-nrf/pull/17512/files#diff-beafb3e3aa8a3c86dc6513c1dcabafce9c0b0246bda1515309c98ba19d1b6285) ? Maybe those states could be defined in DTS instead?
> 
> Yes, all power states should be defined in the .soc device tree files. Not in the sample overlays.

@nordic-segl did that here and has his name forever tied to nrf54h20 power management :heart: 
https://github.com/zephyrproject-rtos/zephyr/pull/79098

The way the idle state is defined there will mean that any `k_sleep()` with a wakeup under 100ms will not go via the pm defined idle state but handled by the idle thread itself via a direct call to `k_cpu_idle()`.

soc/nordic/nrf54h/power.c
karstenkoenig: I don't think that'll work either, when the pm system picks the `ACTIVE` state as next state `pm_suspend()` returns `false`
https://github.com/zephyrproject-rtos/zephyr/blob/06bdd1001c7989f36731d34a5809a1ba22359d4a/subsys/pm/pm.c#L215

hubertmis:
> The way the idle state is defined there will mean that any `k_sleep()` with a wakeup under 100ms will not go via the pm defined idle state but handled by the idle thread itself via a direct call to `k_cpu_idle()`.

Then I would consider modifying this a little:

```
power-states {
	active: active {
		compatible = "zephyr,power-state";
		power-state-name = "active";
		min-residency-us = <0>;
	};
	idle: idle {
		compatible = "zephyr,power-state";
		power-state-name = "suspend-to-idle";
		substate-id = <0>;
		min-residency-us = <100>; // Guessing. TBD
	};
	// substate-id = <1>; is reserved for "idle-cache-retained"
	idle-cache-disabled: idle-cache-disabled {
		compatible = "zephyr,power-state";
		power-state-name = "suspend-to-idle";
		substate-id = <2>;
		min-residency-us = <100000>; // Guessing. TBD
	};
};
```

adamkondraciuk:
> > The way the idle state is defined there will mean that any `k_sleep()` with a wakeup under 100ms will not go via the pm defined idle state but handled by the idle thread itself via a direct call to `k_cpu_idle()`.
> 
> Then I would consider modifying this a little:
> 
> ```
> power-states {
> 	active: active {
> 		compatible = "zephyr,power-state";
> 		power-state-name = "active";
> 		min-residency-us = <0>;
> 	};
> 	idle: idle {
> 		compatible = "zephyr,power-state";
> 		power-state-name = "suspend-to-idle";
> 		substate-id = <0>;
> 		min-residency-us = <100>; // Guessing. TBD
> 	};
> 	// substate-id = <1>; is reserved for "idle-cache-retained"
> 	idle-cache-disabled: idle-cache-disabled {
> 		compatible = "zephyr,power-state";
> 		power-state-name = "suspend-to-idle";
> 		substate-id = <2>;
> 		min-residency-us = <100000>; // Guessing. TBD
> 	};
> };
> ```
> 
> This way the power management decisions are explicit and clear. You don't need to get expertise "What happens when the current residency is lower than defined for any power state?"

Zephyr defines such behavior very clearly [here](https://docs.zephyrproject.org/latest/services/pm/system.html) . Does it make sense to redefine those states which would behave exactly as described in Zephyr doc? That is a bit confusing. Also it introduces additional overhead for the simplest power states.

hubertmis (approved):

nordic-segl (approved):

karstenkoenig (approved):

kl-cruz (approved):

nordic-piks (changes requested):
Duplicated power states definition needs to be remove from https://github.com/nrfconnect/sdk-nrf/tree/main/tests/benchmarks/multicore
at PR which updates into sdk-nrf.

nordic-piks (approved):

nordicjm (changes requested):

soc/nordic/common/soc_lrcconf.c
```diff
@@ -0,0 +1,61 @@
...
+#include <soc_lrcconf.h>
+#include <zephyr/kernel.h>
+
+static struct k_spinlock lock;
+static sys_slist_t poweron_main_list;
+static sys_slist_t poweron_active_list;
+
+void soc_lrcconf_poweron_request(sys_snode_t *node, nrf_lrcconf_power_domain_mask_t domain)
+{
+	__ASSERT(is_power_of_two(domain), "Only one bit can be set for the domain parameter");
```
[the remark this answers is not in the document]
adamkondraciuk: fixed

nordicjm (approved):

PR #2033, opened 2024-09-18
sample: smp_svr: overlay-bt: Increase the conn param int and timeout

laxiLang:
For BLE DFU integration test.


nordicjm (changes requested):
If this is needed for a test then it should be put in the test repo and the twister functionality for using an alternative configuration root should be used

PR #2036, opened 2024-09-20
[nrf-noup] runners: add DFU runner

maxd-nordic:
This patch adds a runner to utilize nrfutil for DFU. It also adds the corresponding ZIP file type to the runner system.


gmarull (changes requested):
why is this required as a noup?

maxd-nordic:
> why is this required as a noup?

The runner is specific to a device that is not supported upstream. Do we have any chance of getting this upstream?

gmarull:
> > why is this required as a noup?
> 
> The runner is specific to a device that is not supported upstream. Do we have any chance of getting this upstream?

it seems to handle old socs?

maxd-nordic:
> > > why is this required as a noup?
> > 
> > 
> > The runner is specific to a device that is not supported upstream. Do we have any chance of getting this upstream?
> 
> it seems to handle old socs?

Yes, but nrfutil will only do mcuboot dfu on devices it explicitly recognizes as mcuboot. And these are only Thingy:91 and Thingy:91 X AFAIK.

nordicjm (commented):
No real preference on it being here or upstream, one negative for it being upstream is that no-one in upstream can use this since dfu_application.zip is an ncs-generated file, it cannot be generated in just upstream zephyr so would essentially be dead code and the first sign of something failing with it would be in an upmerge

maxd-nordic:
@nordicjm @gmarull What's the verdict? Should I open an upstream PR instead?

gmarull (commented):
For the runner I think yes, for the zip part I'm not sure if this is really a proper way to do this @tejlmand ?

tejlmand (changes requested):
So far, this mostly looks like something to have in sdk-nrf, and not a noup patch.

PR #2038, opened 2024-09-23
USB cherry pick series for NCS 2.7.99-CS2

tmon-nordic:
Bring the features that are supposed to be in NCS 2.7.99-CS2


tmon-nordic:
Descoped due to issues that are not present in upmerge branch. No point in trying to figure out the root cause which most likely is fixed in upstream Zephyr (and hence it does not reproduce in upmerge).

PR #2039, opened 2024-09-23
[nrf fromtree] dts: bindings: arm: nordic: Add TDDCONF sources

karstenkoenig:
For nrf54h20 a range of combinations exist to configure the test and debug domains data sources and sinks. Expose them in DTS to allow configuring them. Also drop the previous style which was too rigid to extend to cover all cases cleanly. The old style was only used in a single sample application so far.

Signed-off-by: Karsten Koenig <karsten.koenig@nordicsemi.no>

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/78831


masz-nordic (approved):

57300 (approved):

nordic-krch (approved):

nordicjm (changes requested):
a42e290785f8dc37c7e56e1a43567ad02b0c864d is not the same as 8fdcf132716136935c990c3496e232a9bec8ce20, a fromlist or fromtree must cleanly apply without any modifications

karstenkoenig:
Hi @anangl, you are listed as codeowner. There is a commit upstream that can't apply cleanly downstream because the file diverged downstream. I tried aligning up- and downstream when resolving the conflict in the cherry-pick, so afterwards they will be both aligned.

upstream commit: https://github.com/zephyrproject-rtos/zephyr/pull/78831/commits/8fdcf132716136935c990c3496e232a9bec8ce20
downstream commit: https://github.com/nrfconnect/sdk-zephyr/pull/2039/commits/a42e290785f8dc37c7e56e1a43567ad02b0c864d

The downstream solution of using `EXACT` version for nrf-regtool seems to be the correct choice, the previous commit to upstream even says 'enforce exact version match'
https://github.com/zephyrproject-rtos/zephyr/commit/9c23effb66e73022ef43413ecdae1d3cd353e0a1

anangl:
> The downstream solution of using `EXACT` version for nrf-regtool seems to be the correct choice, the previous commit to upstream even says 'enforce exact version match' [zephyrproject-rtos/zephyr@9c23eff](https://github.com/zephyrproject-rtos/zephyr/commit/9c23effb66e73022ef43413ecdae1d3cd353e0a1)

Not really. See https://github.com/zephyrproject-rtos/zephyr/pull/69633#discussion_r1508045111. `EXACT` was  intentionally not added. And I don't think we need it downstream (468d06db5542594fe9e6420fd2e1fe30d9209699 does not really explain why it was added). I think you should add a noup commit that removes `EXACT` downstream and modify the upstream commit to not add this modifier. Then it will apply cleanly.

karstenkoenig:
@anangl uh good catch, thanks for that. Given this cmake check fails upstream due to the outdated nrf-regtool version in the upstream docker image I'll just drop the version change upstream and only do the suggested noup here to drop the exact. 
Then we can bump this cmake check in a separate PR after the upstream docker image has the new nrf-regtool version

karstenkoenig:
@nordicjm Could you please have another look if this solution works for you?

nordicjm (approved):
Build system change is OK

nordic-krch:
Because we dropped `EXACT` STM logging does not work silently and user does not know that it because regtool update is needed. I think we should bring back `EXACT`.

karstenkoenig:
Is it not rather that 6.0.0 is needed now and not checked via CMake yet? So `nrf-regtool 6.0.0 REQUIRED`?

I tried kicking off the change upstream
https://github.com/zephyrproject-rtos/docker-image/commit/eb419f103940f8e4f0e3fdf0480bb6ef5841e13a
https://github.com/zephyrproject-rtos/zephyr/pull/79244

But am a bit lost with the upstream docker image, the zephyr builds requiring (at least) nrf-regtool 6.0.0 still fail with 6.0.0 not present in the image.

PR #2041, opened 2024-09-24
drivers: serial: Rework nrfx_uarte shim to support nrf54x platforms

nordic-krch:
Set of commits cherry-picked from upstream and from upstream PR (https://github.com/zephyrproject-rtos/zephyr/pull/75462) which enable use of legacy shim on nrf54x.
PR introduces:
- support for new UARTE HW features - frame timeout, ENDTX_STOPTX short
- support for memory and cache management (DMM)
- support for fast instance (UARTE120)


nordic-krch:
superseded by #2146

PR #2053, opened 2024-09-27
[nrf noup] test-spec: update CI-rs-test and Ci-fem-test

dawidprzybylo:
Narrow down paths base on compilation dependencies.


nordic-piks (approved):

.github/test-spec.yml
```diff
@@ -151,40 +151,54 @@
...
+  - "share/**/*"
+  - "soc/nordic/**/*"
+  - "subsys/net/**/*"
+  - "subsys/settings/**/*"
+  - "subsys/bluetooth/shell/**/*"
+  - "subsys/ipc/**/*"
+  - "Kconfig"
+  - "CMakeLists.txt"
+  - "!**/*.rst"
+  - "!**/*.md"
```
mariusz-nordicsemi, line 176: With such pattern an update to any file not matching *.rst or *.md would trigger the build. I believe this is not a goal.
Perhaps you should just remove these negations.

dawidprzybylo:
> The negation patterns are rather invalid.

test-spec updated

mariusz-nordicsemi (approved):

PR #2054, opened 2024-09-30
[nrf fromlist] manifest: openthread: Regular OpenThread upmerge

maciejbaczmanski:
Update OpenThread up to `2aeb8b8` and align KConfig options.

Additionally add `OPENTHREAD_STORE_FRAME_COUNTER_AHEAD` Kconfig option
and set it to 100000, as after calculations it appears to
be a more suitable value.


doublemis1 (approved):
LGTM

PR #2056, opened 2024-10-01
[nrf noup] Sysbuild support for MCUboot crypto backend selection

de-nordic:
The commit adds support for selecting MCUboot crypto backend, when building with sysbuild.

This currently adds two options:
 - SB_CONFIG_BOOT_CRYPTO_LIBRARY_DEFAULT -- which keeps current behaviour where a library is selected depending on preferred signature method
 - SB_CONFIG_BOOT_CRYPTO_LIBRARY_PSA -- enforces usage of PSA, regardless of signature algorithm; the option currently supports the ED25519 only.

### DIFF
--- a/share/sysbuild/images/bootloader/Kconfig
+++ b/share/sysbuild/images/bootloader/Kconfig
@@ -55,6 +55,35 @@ config BOOT_SIGNATURE_TYPE_ED25519
 
 endchoice
 
+choice BOOT_CRYPTO_LIBRARY
+	prompt "Library used by cryptographic functions of MCUboot"
+	default BOOT_CRYPTO_LIBRARY_DEFAULT
+
+config BOOT_CRYPTO_LIBRARY_PSA
+	bool "Enforce use of PSA as cryptographic library"
+
+config BOOT_CRYPTO_LIBRARY_DEFAULT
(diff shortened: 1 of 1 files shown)


share/sysbuild/images/bootloader/Kconfig
```diff
@@ -55,6 +55,35 @@ config BOOT_SIGNATURE_TYPE_ED25519
 
 endchoice
 
+choice BOOT_CRYPTO_LIBRARY
```
nordicjm, line 58: this goes in sdk-nrf under sysbuild/Kconfig.sysbuild

nordicjm (changes requested):

share/sysbuild/images/bootloader/Kconfig
```diff
@@ -55,6 +55,35 @@ config BOOT_SIGNATURE_TYPE_ED25519
...
+	bool
+	default n
+
+config MBEDTLS_PSA_CRYPTO_C
+	bool
+	default y
+
+config NRF_SECURITY
+	bool
+	default y
```
nordicjm, line 83: this does not do anything to the target image, these can go

PR #2070, opened 2024-10-07
logging: Fix case where logging strings were not stripped from the binary when dictionary logging was used

nordic-krch:
Cherry-picked from https://github.com/zephyrproject-rtos/zephyr/pull/79506

### DIFF
--- a/include/zephyr/logging/log_output.h
+++ b/include/zephyr/logging/log_output.h
@@ -190,11 +190,34 @@ void log_output_msg_syst_process(const struct log_output *log_output,
  */
 void log_output_dropped_process(const struct log_output *output, uint32_t cnt);
 
+/** @brief Write to the output buffer.
+ *
+ * @param outf Output function.
+ * @param buf  Buffer.
+ * @param len  Buffer length.
+ * @param ctx  Context passed to the %p outf.
+ */
+static inline void log_output_write(log_output_func_t outf, uint8_t *buf, size_t len, void *ctx)
+{
+	int processed;
+
+	while (len != 0) {
+		processed = outf(buf, len, ctx);
+		len -= processed;
+		buf += processed;
+	}
+}
+
 /** @brief Flush output buffer.
  *
  * @param output Pointer to the log output instance.
  */
-void log_output_flush(const struct log_output *output);
+static inline void log_output_flush(const struct log_output *output)
+{
+	log_output_write(output->func, output->buf, output->control_block->offset,
+			 output->control_block->ctx);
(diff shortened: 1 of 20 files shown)


karstenkoenig (approved):

include/zephyr/logging/log_output.h
```diff
@@ -190,11 +190,35 @@ void log_output_msg_syst_process(const struct log_output *log_output,
...
  * @param output Pointer to the log output instance.
  */
-void log_output_flush(const struct log_output *output);
+static inline void log_output_flush(const struct log_output *output)
+{
+	log_output_write(output->func, output->buf,
+		     output->control_block->offset,
+		     output->control_block->ctx);
+	output->control_block->offset = 0;
+}
```
nordicjm, line 221: differs from upstream PR

nordicjm (changes requested):

include/zephyr/logging/log_output.h
nordic-krch: updated.

subsys/logging/Kconfig.template.log_format_config
```diff
@@ -7,6 +7,7 @@ choice "LOG_BACKEND_$(backend)_OUTPUT"
 
 config LOG_BACKEND_$(backend)_OUTPUT_TEXT
 	bool "Text"
+	select LOG_OUTPUT
```
nordicjm, line 10: Commit is not valid as it differs from upstream. Should be updated to fromtree, and patches must match 1:1

nordicjm (changes requested):

subsys/logging/Kconfig.template.log_format_config
nordic-krch: updated to fromtree

nordic-krch: however, it does not match 1:1 because in upstream there is one more file that was changed but this file do not exist in sdk-zephyr (it came after the upmerge). What is the approach in that case?

nordicjm: Pull in the commits that are needed for this to apply cleanly, otherwise you can't bring it in as fromlist/fromtree and you would need to noup it

nordic-krch: done, had to fetch one more commit.

nordic-krch: @nordicjm can you take another look so we could merge it soon?

PR #2072, opened 2024-10-07
samples: net: Cherry-pick Wi-Fi snippet && fixes

rlubos:
manifest-pr-skip


rlubos:
@nordicjm Is there a smarter way to enable Wi-Fi driver globally in sysbuid-enabled builds other than https://github.com/nrfconnect/sdk-zephyr/pull/2072/commits/94d4de4fc7f50668fe44b141f5919d5212a67ec7? In upstream we just get the driver enabled automatically.

glarsennordic (approved):

nordicjm:
> @nordicjm Is there a smarter way to enable Wi-Fi driver globally in sysbuid-enabled builds other than [94d4de4](https://github.com/nrfconnect/sdk-zephyr/commit/94d4de4fc7f50668fe44b141f5919d5212a67ec7)? In upstream we just get the driver enabled automatically.

That is the correct way to do it. The driver in upstream is most likely enabled because it has a dts entry and the sample enables NETWORKING, there is no devicetree processing in sysbuild and because partition manager runs from a sysbuild context, it needs to be in control of wifi enablement (though maybe once PM goes it can be moved to application)

jukkar (approved):

samples/net/wifi/Kconfig.sysbuild
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+config WIFI_NRF70
+	default y if BOARD_NRF7002DK_NRF5340_CPUAPP || \
+		     BOARD_NRF7002DK_NRF5340_CPUAPP_NS
```
krish2718, line 9: We also need to do this for all Wi-Fi shields, else those combinations will fail :-(.

rlubos: For shields, do we have a Kconfig symbol we can depend on here?

krish2718: We have a Kconfig symbol but it's not working for sysbuild for some reason, so, I have directly updated `sample.yaml` to get twister green.

PR #2076, opened 2024-10-08
Update to nrfx 3.7.0 and add support for nRF54H20 EngB

gmarull:
Update to nrfx 3.7.0 and add support for nRF54H20 EngB


nika-nordic (approved):

anangl (changes requested):
e156e76aa119a5c7840e18624515f727dfd03dca is not a valid fromtree commit.

west.yml
```diff
@@ -188,7 +188,7 @@ manifest:
       groups:
         - hal
     - name: hal_nordic
-      revision: af7b21c4f875bea26689307daa8e52e9a8e8323c
+      revision: ee6213b6cfcd7fdc10f7ec3a4f7f37f35dfa5343
```
gmarull: note: this is still tuned, as upstream moved from `6d4acb8dbd9e92c7c5d6e18724197e3f007ae45e`, not `af7b21c4f875bea26689307daa8e52e9a8e8323c`.

anangl: So it is still wrong (https://github.com/zephyrproject-rtos/zephyr/commit/54aca3ef58af4b4d640aa249c3260e0440d1362e should be cherry-picked first instead of this tuning).
Similar story with d571a8f885f54d24b7740ec4a49bb281a8e67e63 (https://github.com/zephyrproject-rtos/zephyr/commit/942fa67a45bccd1abdfb5317a88b661d325c2d45 is missing).

anangl: Could you also squash those noup commits into one?

gmarull: sure, let me do things properly.

anangl (approved):

kl-cruz (approved):

PR #2077, opened 2024-10-08
[nrf noup] modules: mbedtls: Remove Nordic only PSA wants

Vge0rge:
Remove Nordic only PSA want Kconfigs fro Zephyr to nrf_security.


tomi-font (approved):

tejlmand (changes requested):
we should not be creating `noup` patches for things that can line in downstream repos.

Why is this being moved to a noup patch ?
it will just make future upmerges harder to handle of this principle spreads.

This is why we try to make Zephyr modular so downstream stuff can stay in the downstream repos.

Vge0rge:
> we should not be creating `noup` patches for things that can line in downstream repos.
> 
> Why is this being moved to a noup patch ? it will just make future upmerges harder to handle of this principle spreads.
> 
> This is why we try to make Zephyr modular so downstream stuff can stay in the downstream repos.

What you say make sense to me @tejlmand , so I just reverted the noup here and I moved the file in nrf_security.

frkv (approved):
Good improvement that I didn't dare to do in the upmerge

LGTM and hoping we don't have any dependency for the nordic-only configurations

Fingers crossed for clean CI!

PR #2086, opened 2024-10-09
[nrf noup] ci: Fix manifest-pr-skip

57300:
fixup! [nrf noup] ci: Enable action-manifest-pr

When the PR title or body contains "manifest-pr-skip", this action will cancel itself, which results in a failure status. Use the same condition to skip the whole workflow instead.

### DIFF
--- a/.github/workflows/manifest-PR.yml
+++ b/.github/workflows/manifest-PR.yml
@@ -11,6 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: handle manifest PR
+        if: >
+          !(contains(github.event.pull_request.title, 'manifest-pr-skip') ||
+            contains(github.event.pull_request.body, 'manifest-pr-skip'))
         uses: nrfconnect/action-manifest-pr@main
         with:
           token: ${{ secrets.NCS_GITHUB_TOKEN }}


.github/workflows/manifest-PR.yml
```diff
@@ -11,6 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: handle manifest PR
+        if: >
+          !(contains(github.event.pull_request.title, 'manifest-pr-skip') ||
+            contains(github.event.pull_request.body, 'manifest-pr-skip'))
```
57300, line 16: Hopefully it works, but I don't know how to verify it in advance.

rugeGerritsen: @57300 . I believe this should rather be fixed in the action itself, so that it also works for other users of this action. @karhama , do you agree?

karhama: Yes I would recommend changing action itself if cancelled status is causing problems. However I did not fully understand in which case cancelled becomes the problem?  Idea in action side was roughly : success == manifest PR created, failure == failed to create manifest PR (when it should have been created) and  cancelled == "neutral status" i.e. no need to create manifest PR.

57300: Updating the action itself sounds reasonable. I'm just unsure what's the best way to do it, seems to me like it could make the code more convoluted.

The only problem with the cancel is that it makes the PR red, which is misleading. It also shows up when filtering PRs by "status:failure". I think a code owner would still be able to merge it though, so I guess it's minor.

57300:
Resolved via https://github.com/nrfconnect/action-manifest-pr/pull/13

PR #2101, opened 2024-10-14
[nrf fromtree] drivers: sensor: qdec: fix QDEC overflow handling

mstasiaknordic:
Fixed overflow handling in nrf QDEC sensor driver. It failed to inform user of an overflow in the ACC register, that caused the fetched sample to be invalid because of discarding data that would cause the overflow.

Now, `qdec_nrfx_sample_fetch()` returns error code if an overflow occured, based on overflow field in qdec data structure.

Also, driver is now more adjusted to sensor driver API. It is guaranteed that two subsequent calls of `sensor_channel_get()` function for the same channels will yield the same value, if `sensor_sample_fetch()` has not been called in the meantime.

### DIFF
--- a/drivers/sensor/nordic/qdec_nrfx/qdec_nrfx.c
+++ b/drivers/sensor/nordic/qdec_nrfx/qdec_nrfx.c
@@ -26,7 +26,9 @@ LOG_MODULE_REGISTER(qdec_nrfx, CONFIG_SENSOR_LOG_LEVEL);
 
 
 struct qdec_nrfx_data {
+	int32_t fetched_acc;
 	int32_t acc;
+	bool overflow;
 	sensor_trigger_handler_t data_ready_handler;
 	const struct sensor_trigger *data_ready_trigger;
 };
@@ -49,6 +51,8 @@ static void accumulate(struct qdec_nrfx_data *data, int32_t acc)
 
 	if (!overflow) {
 		data->acc += acc;
+	} else {
+		data->overflow = true;
 	}
 
 	irq_unlock(key);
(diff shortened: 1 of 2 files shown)


drivers/sensor/nordic/qdec_nrfx/qdec_nrfx.c
```diff
@@ -87,8 +103,7 @@ static int qdec_nrfx_channel_get(const struct device *dev,
 	}
 
 	key = irq_lock();
-	acc = data->acc;
-	data->acc = 0;
+	acc = data->fetched_acc;
```
pdunaj, line 106: This change will break https://github.com/nrfconnect/sdk-nrf/blob/main/applications/nrf_desktop/src/hw_interface/wheel.c#L119

Fetch is not called there - probably an overlook.

pdunaj (changes requested):
When bringing the change to sdk-nrf please also fix the nrf_desktop code and look if there are no other dependencies that may get broken.

drivers/sensor/nordic/qdec_nrfx/qdec_nrfx.c
mstasiaknordic: Thanks for pointing this out, wheel updated. Could not find any other occurrences beside the one you mentioned.

MarekPieta (approved):
The cherry-pick looks good

mstasiaknordic:
@pdunaj I believe nrf_desktop issue has been resolved

mstasiaknordic:
@carlescufi Do you know why Commit Tags job fails? I have seen such issue on other PRs and I am unable to resolve it here. I have done no changes to the commits during the cherry-picking. It has occured after a rebase

anangl:
> @carlescufi Do you know why Commit Tags job fails? I have seen such issue on other PRs and I am unable to resolve it here. I have done no changes to the commits during the cherry-picking. It has occured after a rebase

You have cherry-picked commits from your upstream PR instead of those actually merged ones and your commit messages refer to wrong SHAs.
Use `git cherry-pick -ex 9b5260de9a31738f5632a0f7240f1f70bc1d4cd4~2..9b5260de9a31738f5632a0f7240f1f70bc1d4cd4`.

mstasiaknordic:
>You have cherry-picked commits from your upstream PR instead of those actually merged ones and your commit messages refer to wrong SHAs.

Thanks for pointing that out!

anangl (approved):

pdunaj (approved):

PR #2103, opened 2024-10-14
[nrf fromtree]  cmake: mcuboot: Use imgtool instead of west for signing

nordicjm:
Copies commits from upstream


de-nordic (approved):

tejlmand (approved):

nordicjm:
Will be brought in via upmerge

PR #2105, opened 2024-10-14
[nrf noup] drivers: entropy: Switch entropy driver in nRF54L20

e-rk:
nrf-squash! [nrf noup] entropy: Add fake entropy nRF PRNG driver

Change psa driver to nordic entropy. nRF54L20 needs it as well as nRF54H20.

### DIFF
--- a/drivers/entropy/Kconfig.nrf_prng
+++ b/drivers/entropy/Kconfig.nrf_prng
@@ -9,7 +9,7 @@ config FAKE_ENTROPY_NRF_PRNG
 	bool "A fake nRF entropy driver"
 	default y
 	depends on DT_HAS_NORDIC_ENTROPY_PRNG_ENABLED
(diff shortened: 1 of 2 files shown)


masz-nordic (approved):

pdunaj (changes requested):
Why would nrf54h20 use fake entropy?

pdunaj (approved):
This is L20 not H20.

e-rk:
> Why would nrf54h20 use fake entropy?

This is nRF54L20, not nRF54H20. Also PSA entropy does not work on nRF54L20 yet IIRC. Without this, many samples we need for testing don't build.

pdunaj:
> > Why would nrf54h20 use fake entropy?
> 
> This is nRF54L20, not nRF54H20. Also PSA entropy does not work on nRF54L20 yet IIRC. Without this, many samples we need for testing don't build.

got it :)

Vge0rge (approved):

PR #1856, opened 2024-06-25
[nrf noup] boards: nordic: Turn on NRFS globally

zycz:
Turn on NRFS globaly so turning off
suspending MRAM for NRF54H20 DK will work.

PR #2115, opened 2024-10-15
boards: nrf54h20: refactor MRAM board level code

gmarull:
Move code to soc level, also refactor it a bit. This makes sure this setting is applied regardless of the board.


modules/hal_nordic/nrfs/Kconfig
```diff
@@ -38,8 +38,6 @@ config NRFS
 	bool "nRF Services Support"
 	select NRFS_LOCAL_DOMAIN if (SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD || SOC_NRF9280_CPUAPP || SOC_NRF9280_CPURAD)
 	depends on HAS_NRFS
-	depends on !MISRA_SANE
```
57300, line 41: Do you know if this is addressed? https://github.com/nrfconnect/sdk-zephyr/pull/1856#issuecomment-2188854151

soc/nordic/nrf54h/Kconfig
```diff
@@ -53,3 +53,8 @@ config SOC_NRF54H20_CPUPPR
 
 config SOC_NRF54H20_CPUFLPR
 	depends on RISCV_CORE_NORDIC_VPR
+
+config SOC_NRF54H20_NO_MRAM_LATENCY
+	bool "Disallow MRAM latency"
+	depends on NRFS && NRFS_MRAM_SERVICE_ENABLED
```
57300, line 59: Would it be a good idea to change this to:
```suggestion
	imply NRFS
	imply NRFS_MRAM_SERVICE_ENABLED
```
if we can avoid changing the defconfig?

soc/nordic/nrf54h/mram.c
```diff
@@ -5,54 +5,55 @@
...
 	case NRFS_MRAM_LATENCY_REQ_REJECTED:
-		LOG_ERR("MRAM latency handler - request rejected!");
-		break;
+		LOG_ERR("MRAM latency not allowed setting rejected");
+		k_panic();
+		__fallthrough;
 	default:
-		LOG_ERR("MRAM latency handler - unexpected event: 0x%x", p_evt->type);
-		break;
+		CODE_UNREACHABLE;
```
57300, line 30: This could prevent the handler from returning properly if a new event is ever added. I think we should just break here, and maybe keep the "unexpected event" print as a `LOG_DBG`.

modules/hal_nordic/nrfs/Kconfig
gmarull: If that's the case then this needs to be upstreamed.

soc/nordic/nrf54h/Kconfig
gmarull: right, will change

modules/hal_nordic/nrfs/Kconfig
gmarull: I've tested with the option and I'm able to compile successfully, maybe it's not a problem anymore?

57300 (approved):

nordicjm (approved):

bjarki-andreasen (approved):

soc/nordic/nrf54h/mram.c
```diff
@@ -5,54 +5,55 @@
...
-LOG_MODULE_REGISTER(MODULE);
+#include <zephyr/toolchain.h>
 
 #include <services/nrfs_mram.h>
 #include <nrfs_backend_ipc_service.h>
 
-#define MRAM_SUSPEND_OFF_INIT_PRIO 90
+LOG_MODULE_REGISTER(mram, CONFIG_SOC_LOG_LEVEL);
 
 void mram_latency_handler(nrfs_mram_latency_evt_t const *p_evt, void *context)
```
anangl, line 17:
```suggestion
static void mram_latency_handler(nrfs_mram_latency_evt_t const *p_evt, void *context)
```

anangl (approved):

zycz (approved):

hubertmis (approved):
This module should be hooked up with `pm_policy_latency_changed_subscribe()` instead of `SYS_INIT()`. But I understand, this comes in future PRs.

PR #2116, opened 2024-10-15
[nrf fromtree] soc: nordic: nrf53: SOC_NRF53_CPUNET_ENABLE should not…

sean-madigan:
… depend on !BT

This causes issues as we still want the net core to be set up on initialisation if CONFIG_BT is enabled in many cases.

The previous changes in
https://github.com/zephyrproject-rtos/zephyr/pull/74304 assumed that because this is also handled in
`bt_hci_transport_setup` that it shouldn't be done on initialisation too.

Also, I think that
`bt_hci_transport_setup` should not be even controlling the network core state like it is currently and should be looked at being changed in a future PR.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/79875

### DIFF
--- a/soc/nordic/nrf53/Kconfig
+++ b/soc/nordic/nrf53/Kconfig
@@ -197,7 +197,6 @@ config SOC_NRF53_CPUNET_MGMT
 
 config SOC_NRF53_CPUNET_ENABLE
 	bool "NRF53 Network MCU is enabled at boot time"
-	depends on !BT
 	default y if NRF_802154_SER_HOST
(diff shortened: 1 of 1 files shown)


anangl (approved):

masz-nordic (approved):

kl-cruz (approved):

anangl (approved):

laxiLang:
@de-nordic @nordic-krch @carlescufi what is the conclusion for this actual fromtree PR (upstream merged)? It staled for 2 weeks from the last action.

PR #2127, opened 2024-10-17
Gd pwr service

Rafal-Nordic:


de-nordic:
I am not sure why I am here, I think this is something for USB expert? @tmon-nordic ?

PR #2128, opened 2024-10-17
[nrf fromtree] boards: nordic: ipc node added dcache alignement

lstnl:
dcache-alignement needs to be defined for ipc to have consistent memory organization on both endpoints, when shared memory is cacheable. nrf54h20 and nrf9280 are using cacheable shared memory.
This is applied for ipc with icmsg backend.

Signed-off-by: Lukasz Stepnicki <lukasz.stepnicki@nordicsemi.no>
(cherry picked from commit 022122d659a7c10180b8a91c6a0b49874f2850fa)


lstnl:
This was cherry picked to other pr and is already in main.

PR #2129, opened 2024-10-17
Nrfx 6553 fix tls handling for s2ram

adamkondraciuk:
The S2RAM procedure requires marker checking after reset. Such checking is performed on very early stage of the system initialization and must ensure that the stack is not used due to the TLS pointer which is not initialized yet.


nordic-piks (approved):
Basic testing fine.

nordic-krch (approved):

hubertmis (approved):

PR #2135, opened 2024-10-18
[nrf noup] boards: nordic: Enable PSA RNG for secdom devices

frkv:
Noup since secdom is not available upstream.
This enables the PSA RNG as the default Zephyr entropy provider for the following secdom enabled targets:
nrf54h20_cpuapp and cpurad
nrf9280_cpuapp

### DIFF
--- a/boards/nordic/nrf9280pdk/nrf9280pdk_nrf9280_cpuapp.dts
+++ b/boards/nordic/nrf9280pdk/nrf9280pdk_nrf9280_cpuapp.dts
@@ -28,7 +28,7 @@
 		zephyr,bt-hci = &bt_hci_ipc0;
 		nordic,802154-spinel-ipc = &ipc0;
 		zephyr,canbus = &can120;
-		zephyr,entropy = &prng;
+		zephyr,entropy = &psa_rng;
 	};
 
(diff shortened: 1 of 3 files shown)


tomi-font (approved):

boards/nordic/nrf9280pdk/nrf9280pdk_nrf9280_cpuapp.dts
```diff
@@ -28,7 +28,7 @@
 		zephyr,bt-hci = &bt_hci_ipc0;
 		nordic,802154-spinel-ipc = &ipc0;
 		zephyr,canbus = &can120;
-		zephyr,entropy = &prng;
+		zephyr,entropy = &psa_rng;
```
Vge0rge, line 31: Just a suggestion and not a blocker but you can even just leave that as it was with the prng. I removed that before because I was trying to remove the prng all together. But now the prng cannot be removed since there are more platforms using it, so this can stay as it was before if you want.

Vge0rge (approved):

shanthanordic:

masz-nordic:
Needs rebase if still valid.

PR #2154, opened 2024-10-23
[nrf fromtree] boards: nordic: ipc: added dcache alignement

doki-nordic:
The nRF54 and nRF92 chips has data cache, which means the ICMsg and ICBMsg must be configured to follow required cache alignment of the shared memory.
The `dcache-alignement` needs to be defined for that.

Upstream commit: https://github.com/zephyrproject-rtos/zephyr/commit/cbaafe209c17f03feea3760be855dc76ef5612cc

PR #2155, opened 2024-10-23
[nrf fromlist] ipc: icbmsg: Reduce block alignment to 32-bits

doki-nordic:
[nrf fromlist] ipc: icbmsg: Reduce block alignment to 32-bits

The ICBMsg backend divides its memory into
blocks. Each block is aligned to data cache
alignment. Is it not required, since adjacent
blocks has the same data flow direction (either
read-only or write-only). This commit changes
it to 32-bits making wasted memory significantly
reduced.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/79604

Depends on: https://github.com/nrfconnect/sdk-zephyr/pull/2154


dchat-nordic (approved):

maje-emb (approved):

shanthanordic:

PR #2159, opened 2024-10-24
[nrf fromtree] picolibc: Update module to version 1.8.8

aescolar:
Switch to released version with all of the Zephyr build fixes

(cherry picked from commit d5b653961487ce516dc0eefca65d19dfc72a1497)

1.8.8 came just 2 weeks after 1.8.7 but includes several bugfixes, including a fix for a bug that prevented building with gcc9 which is the default in Ubuntu 20.04.
https://github.com/picolibc/picolibc/releases/tag/1.8.8

As requested by @PerMac , due to failures seen in a CI job


PerMac:
@ns-tolu This is responsible for some failures in CI testing bleutooth/mesh using tests from upstream but with softdevice controller.  Since those test don't work now due to other reasons and probably won't be resolved for 2.8. I don't know if the error seen there is also visible anywhere else. As a release manager, please decide if this should enter the release or not.

PerMac:
@aescolar @carlescufi what is your opinion on priority for this?

aescolar:
> Since those test don't work now

A clarification: Not only those tests. Also some others which would pass otherwise.
This particular bug prevents building anything with picolibc from the source module with gcc 9. Which includes also building using just plain native_sim with picolibc with this oldish gcc.
Note gcc 9 is the default in Ubuntu 20.04 LTS (which I'd expect some people will still use, just like we do in some of our CI jobs).

Apart from this, Keith's release notes was not too clear on what else is broken with pico 1.8.7, but given that he made a new release right away to fix those I presume they are worthwhile.

So I would quite inclined to update pico to 1.8.8.

carlescufi:
Given that we use GCC 12 in our toolchain, and that this only affects if building picolib from source, I suggest we do not include this in 2.8.0.

aescolar:
>  I suggest we do not include this in 2.8.0.

Ok, I will close this PR then. The update can come in the next upmerge, or if somebody else requests it.

PerMac:
I am reopening the PR. Since we have 2.8 release branch separated, this can go to the main without affecting the release (without backporting)

PerMac:
@nrfconnect/ncs-co-build-system @carles I think it is worth fixing now since AFAIK there won't be more upmerges this year

aescolar:
Closing, as this already made it in thru an upmerge

carles:
Hi there, I have nothing to do with this project but I keep getting notifications because you keep tagging my GitHub account (@carles). I suspect you are actually trying to tag some other Carles who is a contributor to this project. Please, make sure to tag the right account, thank you :)

PR #2199, opened 2024-10-30
[nrf fromlist] soc: nordic: nrf54l15: fix APPROTECT handling

nika-nordic:
To configure APPROTECT on nRF54L15 different set of MDK symbols must be used. Additionally, nRF54L15 does not support loading APPROTECT configuration from the UICR in runtime.

Upstream PR #: 80366

This PR was reverted in https://github.com/nrfconnect/sdk-zephyr/pull/2193 because TF-M alignments were missing. They are added now in sdk-nrf PR

### DIFF
--- a/soc/nordic/Kconfig
+++ b/soc/nordic/Kconfig
@@ -98,15 +98,24 @@ config NFCT_PINS_AS_GPIOS
 
 choice NRF_APPROTECT_HANDLING
 	bool "APPROTECT handling"
-	depends on SOC_SERIES_NRF52X || SOC_NRF5340_CPUNET || \
-		   SOC_NRF5340_CPUAPP || SOC_SERIES_NRF91X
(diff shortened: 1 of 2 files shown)


soc/nordic/Kconfig
```diff
@@ -98,15 +98,24 @@ config NFCT_PINS_AS_GPIOS
...
 	help
 	  Specifies how the SystemInit() function should handle the APPROTECT
 	  mechanism.
 
+config NRF_APPROTECT_DISABLE
+	bool "Disable"
+	depends on SOC_NRF54L15_CPUAPP
+	help
+	  When this option is selected, the SystemInit() disables
+	  the APPROTECT mechanism.
```
MarkusLassila, line 114: Does this then mean that the debugger can be attached?

MarkusLassila (approved):

soc/nordic/Kconfig
nika-nordic: yes

anangl (approved):

PR #2217, opened 2024-11-04
[nrf fromlist] soc: nordic: nrf54l: tune configuration of DCDC regulator

nika-nordic:
DCDC regulator on nRF54L may not always works as intended. Tune the fix addressing that.

Upstream PR #: 80635


miha-nordic:
Checked, issue fixed no side effects observed.

miha-nordic (approved):

anangl (approved):

PR #2229, opened 2024-11-05
[nrf noup] sysbuild: Support for ED25519 with SHA512 and PureEdDSA

de-nordic:
The commit adds imgtool parameters that allow to sign image with SHA512, for ED25519, or use PureEdDSA variant, where signature is calculated over entire image rather than SHA of that image.


share/sysbuild/image_configurations/MAIN_image_default.cmake
```diff
@@ -37,4 +37,16 @@ if(SB_CONFIG_BOOTLOADER_MCUBOOT)
...
+
+  set(imgtool_extra_args "${CONFIG_MCUBOOT_EXTRA_IMGTOOL_ARGS}")
+  if(SB_CONFIG_BOOT_SIGNATURE_TYPE_ED25519)
+    if(SB_CONFIG_BOOT_SIGNATURE_TYPE_PURE)
+      string(APPEND imgtool_extra_args " --pure")
+    else()
+      string(APPEND imgtool_extra_args " --sha 512")
+    endif()
+  endif()
+
```
nordicjm, line 49: this should not be in sdk-zephyr at all, move it to sdk-nrf, as part of sysbuild, and set Kconfigs in the image which then enable these options, do not just the extra args - that field is for users to change as they want

nordicjm (changes requested):

PR #2230, opened 2024-11-06
[nrf fromlist] drivers: serial: nrfx_uarte: Fix RX path without low p…

nordic-krch:
…ower modes

RX FIFO flushing on RXTO event should only be performed when UARTE peripheral might be disable during inactivity and that happens when low power modes is enabled or when device runtime PM is used. Flushing was incrementing flush_cnt which was not used (flushed data is not copied to the next buffer) which was causing data loss and invalid RX data length reporting.

Upstream PR #: 80968


MarkusLassila (approved):

MarkusLassila:
@nordic-krch: This is needed for 2.8 branch as well.

anangl (approved):

PR #2253, opened 2024-11-11
wifi: hostap: Pull external crypto support

krish2718:
Pulls in external crypto support that will be used in NCS to manage hostap dependencies and help use revert the `noup` patches (they will be removed during history rewrite in upcoming upmerge).


VivekUppunda (approved):

nordicjm (approved):

rlubos (approved):

modules/hostap/Kconfig
```diff
@@ -109,9 +109,14 @@ config WIFI_NM_WPA_SUPPLICANT_WEP
 
 choice WIFI_NM_WPA_SUPPLICANT_CRYPTO_BACKEND
 	prompt "WPA supplicant crypto implementation"
+<<<<<<< HEAD
```
rado17: merge conflict?

rado17: Jumped the gun there. Saw that this is resolved in next commit. Can we squash them

rado17 (approved):

krish2718:
@carlescufi @rlubos the OSS history on the manifest PR is failing due to a bug in the history, I don't see how it can be fixed without a rewrite, WDYT?

rlubos:
Do I understand correctly that the problem here is that we have two commits with the same title `[nrf noup] modules: hostap: Add NCS PSA for hostap crypto ALT`? I see that one of the commits (https://github.com/nrfconnect/sdk-zephyr/commit/33919fc981c12469ea96f987c23b90bfa5a8ebea) is still listed in ncs loot even after reverting both, which causes oss-history to fail when trying to reapply it (empty commit...).

Would like to hear from @carlescufi first, but I don't really see a way around this other than history rewrite... For sure we cannot merge like this as it'll break the check for every other PR.

krish2718:
> Do I understand correctly that the problem here is that we have two commits with the same title `[nrf noup] modules: hostap: Add NCS PSA for hostap crypto ALT`? I see that one of the commits ([33919fc](https://github.com/nrfconnect/sdk-zephyr/commit/33919fc981c12469ea96f987c23b90bfa5a8ebea)) is still listed in ncs loot even after reverting both, which causes oss-history to fail when trying to reapply it (empty commit...).

Yes that is the issue.
 
> Would like to hear from @carlescufi first, but I don't really see a way around this other than history rewrite... For sure we cannot merge like this as it'll break the check for every other PR.

Sure, let's resolve this now.

PR #2255, opened 2024-11-12
[nrf noup] drivers: spi: dw: disable pin retention after enabling

gmarull:
This is required so that device stays functional while enabled.

### DIFF
--- a/drivers/spi/spi_dw.c
+++ b/drivers/spi/spi_dw.c
@@ -45,6 +45,10 @@ LOG_MODULE_REGISTER(spi_dw);
 #include <nrfx.h>
 #endif
 
+#ifdef CONFIG_SOC_NRF54H20_GPD
+#include <nrf/gpd.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


ahasztag (changes requested):
This does not seem to be enough to fix the issue;

It seems that commenting out https://github.com/nrfconnect/sdk-zephyr/commit/f443d32e8bb9e1f27febfae32760b3b95a7c6857#diff-77c81f9bdc8e780abae508eb5aaf73b663975f3f9e98943decc8f0e4f1aee121R240 
helps - If I'm not wrong the `pinctrl` API does not use this.

It seems that the only thing which uses the `gpio_nrfx_pin_configure` in the `spi_dw` driver is the CS pin - https://github.com/nrfconnect/sdk-zephyr/blob/a70d6bd9660a3450f264d1d7017828196a964b3f/drivers/spi/spi_dw.c#L573 - maybe the problem is the CS pins, which still have retention on?

PR #2256, opened 2024-11-12
[nrf noup] soc:nordic:nrf54l Fix for arch_busy_wait on 54l with tfm

degjorva:
Noup as it patches code that does not exist in zephyr.
arch_busy_wait waited half the expected time due to SystemCoreClock
being set back to the default value when ns builds. 
Changed to rely on device tree as to avoid hardcoded clock frequencies

### DIFF
--- a/drivers/audio/dmic_nrfx_pdm.c
+++ b/drivers/audio/dmic_nrfx_pdm.c
@@ -8,17 +8,25 @@
 #include <zephyr/drivers/clock_control/nrf_clock_control.h>
 #include <zephyr/drivers/pinctrl.h>
 #include <soc.h>
+#include <dmm.h>
 #include <nrfx_pdm.h>
 
 #include <zephyr/logging/log.h>
 #include <zephyr/irq.h>
 LOG_MODULE_REGISTER(dmic_nrfx_pdm, CONFIG_AUDIO_DMIC_LOG_LEVEL);
 
+#if CONFIG_SOC_SERIES_NRF54HX
+#define DMIC_NRFX_CLOCK_FREQ 8*1000*1000UL
+#else
+#define DMIC_NRFX_CLOCK_FREQ 32*1000*1000UL
+#endif
+
 struct dmic_nrfx_pdm_drv_data {
 	const nrfx_pdm_t *pdm;
 	struct onoff_manager *clk_mgr;
 	struct onoff_client clk_cli;
 	struct k_mem_slab *mem_slab;
+	void *mem_slab_buffer;
 	uint32_t block_size;
 	struct k_msgq rx_queue;
 	bool request_clock : 1;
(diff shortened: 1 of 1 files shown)


soc/nordic/nrf54l/soc.c
```diff
@@ -170,19 +176,16 @@ static inline void power_and_clock_configuration(void)
...
-	SystemCoreClockUpdate();
-#endif
-
 #ifdef __NRF_TFM__
 	/* TF-M enables the instruction cache from target_cfg.c, so we
 	 * don't need to enable it here.
 	 */
 #else
+	/* Update SystemCoreClock in Zephyr based on device tree to avoid SystemCoreClock
+	 * being overwritten with default value when initializing with TFM
```
tomi-font, line 185:
```suggestion
	 * being overwritten with default value when initializing with TF-M
```

soc/nordic/nrf54l/soc.c
```diff
@@ -47,6 +47,12 @@ LOG_MODULE_REGISTER(soc, CONFIG_SOC_LOG_LEVEL);
 #define HFXO_NODE DT_NODELABEL(hfxo)
 #endif
 
+#if DT_PROP(DT_PATH(cpus, cpu_0), clock_frequency)
+#define DEVICE_DT_CLOCK_FREQ DT_PROP(DT_PATH(cpus, cpu_0), clock_frequency)
+#elif DT_PROP(DT_PATH(cpus, cpu_1), clock_frequency)
```
tomi-font, line 52: What is the logic for prioritizing `cpu_0` over `cpu_1`? (I don't know what those refer to.)

degjorva: It is just to handle the difference between cpuapp and cpuflpr. cpuapp uses cpu_0 to set clock-frequency while cpuflpr uses cpu_1. Will add a comment

soc/nordic/nrf54l/soc.c
```diff
@@ -47,6 +47,13 @@ LOG_MODULE_REGISTER(soc, CONFIG_SOC_LOG_LEVEL);
 #define HFXO_NODE DT_NODELABEL(hfxo)
 #endif
 
+/* Different build images uses different CPU-names. That is handled here */
```
tomi-font, line 50:
```suggestion
/* Different build images uses different CPU names. That is handled here */
```

soc/nordic/nrf54l/soc.c
```diff
@@ -47,6 +47,13 @@ LOG_MODULE_REGISTER(soc, CONFIG_SOC_LOG_LEVEL);
...
+#if DT_PROP(DT_PATH(cpus, cpu_0), clock_frequency)
+#define DEVICE_DT_CLOCK_FREQ DT_PROP(DT_PATH(cpus, cpu_0), clock_frequency)
```
tomi-font, line 52: A bit weird to read this name: `device devicetree clock frequency`

PR #2258, opened 2024-11-12
soc: nordic: nrf54h: disable IRQ before PM config

adamkondraciuk:
IRQs must be disabled before starting any procedures to prepare for low-power states.


arch/arm/core/cortex_m/pm_s2ram.S
```diff
@@ -85,14 +85,19 @@ SECTION_FUNC(TEXT, arch_pm_s2ram_suspend)
 	 * not successful (in r0 the return value).
 	 */
 
+	/* Move return value of system_off to callee-saved register. */
+	mov 	r4, r0
```
karstenkoenig, line 89: nice job picking that fix up

karstenkoenig (approved):

ns-tolu (approved):

PR #2261, opened 2024-11-13
[nrf noup] drivers: spi: dw: turn on FAST_ACTIVE1 domain

gmarull:
FAST_ACTIVE1 domain needs to be kept ON while the peripheral is active (always in this driver).

### DIFF
--- a/drivers/spi/spi_dw.c
+++ b/drivers/spi/spi_dw.c
@@ -45,6 +45,10 @@ LOG_MODULE_REGISTER(spi_dw);
 #include <nrfx.h>
 #endif
 
+#ifdef CONFIG_SOC_NRF54H20_GPD
+#include <nrf/gpd.h>
+#endif
+
 static inline bool spi_dw_is_slave(struct spi_dw_data *spi)
 {
 	return (IS_ENABLED(CONFIG_SPI_SLAVE) &&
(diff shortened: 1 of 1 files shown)


ahasztag (commented):
Perhaps we should also enable the FAST_DOMAIN in https://github.com/nrfconnect/sdk-zephyr/blob/9ff506841d0ef3c3ada27f1e65c92385093c5ee3/drivers/spi/spi_dw.c#L80 - I am not sure how this will behave without that - we might not be turning off EXMIF and not notice it

karstenkoenig:
> Perhaps we should also enable the FAST_DOMAIN in
> 
> https://github.com/nrfconnect/sdk-zephyr/blob/9ff506841d0ef3c3ada27f1e65c92385093c5ee3/drivers/spi/spi_dw.c#L80
> - I am not sure how this will behave without that - we might not be turning off EXMIF and not notice it

I'd say we should wrap the configure section with it, to make sure everything is powered at least for the whole time that it gets configured.

karstenkoenig:
What happened to clearing retain for dq0, dq1 and clock pins? Is that not needed as well or is this elsewhere?

ahasztag (approved):

ahasztag:
Our tests pass on this PR

robertstypa (approved):

karstenkoenig (approved):

karstenkoenig:
> What happened to clearing retain for dq0, dq1 and clock pins? Is that not needed as well or is this elsewhere?

solved in the chat, retain never gets set. somewhat unintentionally

PR #2267, opened 2024-11-13
cmake: dts: Use temporary file for dts.cmake with sysbuild

nordicjm:
Uses a temporary file when sysbuild is being used for images then uses CMake to copy to the correct file if it has changed. This prevents a ping-pong of configuring and building cycle when nothing has changed and there is sysbuild code which loads in the devicetree data from an image

Upstream PR #: 81340

### DIFF
--- a/cmake/modules/dts.cmake
+++ b/cmake/modules/dts.cmake
@@ -321,17 +321,26 @@ endif()
 #
 # Run GEN_DTS_CMAKE_SCRIPT.
 #
+# A temporary file is copied to the original file if it differs. This prevents issue such as a
+# cycle when sysbuild is used of configuring and building multiple times due to the dts.cmake file
+# of images having a newer modification time than the sysbuild build.ninja file, despite the
+# output having not changed
+#
+set(dts_cmake_tmp ${DTS_CMAKE}.new)
 
 execute_process(
   COMMAND ${PYTHON_EXECUTABLE} ${GEN_DTS_CMAKE_SCRIPT}
   --edt-pickle ${EDT_PICKLE}
-  --cmake-out ${DTS_CMAKE}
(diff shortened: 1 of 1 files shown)


cmake/modules/dts.cmake
```diff
@@ -321,17 +321,26 @@ endif()
 #
 # Run GEN_DTS_CMAKE_SCRIPT.
 #
+# A temporary file is used copied to the original file if it differs, this prevents issue such as
```
trantanen, line 324: Should probably be:
```suggestion
# A temporary file is copied to the original file if it differs. This prevents issue such as
```

trantanen (approved):
This is fixing rebuilding when doing `west flash` after `west build` at least with asset_tracker_v2 and modem_shell when there are multiple images (FOTA configurations having mcuboot).

cmake/modules/dts.cmake
nordicjm: Thanks, fixed upstream and here

tejlmand:
@trantanen when a PR in sdk-zephyr is a fromlist, then please give review comments in the linked upstream PR: zephyr:81340
and not in the downstream `fromlist`.

tejlmand (approved):

trantanen:
> @trantanen when a PR in sdk-zephyr is a fromlist, then please give review comments in the linked upstream PR: zephyr:81340 and not in the downstream `fromlist`.

Sorry about that. I noticed there was upstream PR only after reviewing. I didn't pay attention this was `fromlist` as I don't tend to look at commits but just the changes of the PR as a whole. And there was no link to upstream but later realized the PR number is there but you don't link them anymore to avoid spamming in upstream.

PR #2269, opened 2024-11-13
[nrf noup] samples: mgmt: mcumgr: smp_svr: Add nrf5340dk-optimised sample

nordicjm:
Adds a configuration to this sample that is optimised for testing/usage with the nRF5340 which enables the Bluetooth transport, enables most management groups, has a static partition manager file with secondary partitions on QSPI flash, configures MCUboot in overwrite only mode with network core update support


nvlsianpu (changes requested):
Multiple board configuration/overlay files was added in the sample's sysbuild/mcuboot/boards directory as well (not only nRF540dk).
I see that many of the can be involved into the build configuration as I'm recognizing them as their copies from MCUBoot's repository.

1. What intention behind that?
2. Why 3-rd party board files are added?
3. Should we remove original files from MCUboot's repository?

nordicjm:
>     1. What intention behind that?

So that this sample can be used with any (nordic-based) board still, as the files in MCUboot will not be used

>     2. Why 3-rd party board files are added?

As above, only nordic-based boards are added

>     3. Should we remove original files from MCUboot's repository?

No

nordicjm:
Would add that this is only need due to, yet again, partition manager

nvlsianpu (approved):

gchwier (approved):
nit: nrf54l15PDK can be removed, only DK is now supported
nit2: do we need pm_static files? Partition Manager should generate pm_static

and it is not fine that we need to copy all those configs from MCUboot repo when using sysbuild. As I remember, for child images those configs were taken as a default parameters, and overwritten by sample configs, now if we have any file/folder in sysbuild dir we must copy defaults to the project dir :(

samples/subsys/mgmt/mcumgr/smp_svr/sysbuild/mcuboot/app.overlay
nvlsianpu, line 1: Will we can avoid many of *.conf files copies If this file will be removed?

nordicjm: Have reworked PR, all these files are now gone

nvlsianpu:
@nordicjm Can you rebase this?

nvlsianpu:
@nordicjm Can be merged.

de-nordic (approved):

nvlsianpu:
@nrfconnect/ncs-test-leads Please give tick.

miha-nordic:
> @nrfconnect/ncs-test-leads Please give tick.

CI is red. Please take a look on DFU tests.

PR #2277, opened 2024-11-18
modules: hal_nordic: nrfs: add GDFS Service

ppelikan-nordic:
Adding the implementation for the GDFS service


soc/nordic/nrf54h/Kconfig
```diff
@@ -77,9 +79,3 @@ config SOC_NRF54H20_ENGB_CPUFLPR
 	depends on RISCV_CORE_NORDIC_VPR
 
 rsource "gpd/Kconfig"
-
```
lstnl, line 80: Probably not intendetd

PR #2279, opened 2024-11-19
[nrf fromlist] tests: drivers: watchdog: disable DCACHE for nRF54H20dk

mstasiaknordic:
Disabled data caching for nRF54H20dk in wdt_basic_api test. It caused variables stored in noinit section to be cached, resulting in reset loop.

Upstream PR #: 81438


nordic-segl (changes requested):
Don't extend sample.yaml as test `drivers.watchdog` is keep failing for nrf54h20.
Add ./boards/nrf54h20dk_nrf54h20_cpuapp.conf with
CONFIG_DCAHE=n
in it.

nordic-segl (approved):

nika-nordic (approved):

kl-cruz (approved):

PR #2282, opened 2024-11-20
[nrf fromlist] drivers: serial: nrfx_uarte: Fix low power mode for interrupt driven API

nordic-krch:
Fix for UARTE driver. Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/81631


kl-cruz (approved):

nordic-krch:
Came with upmerge

PR #2300, opened 2024-11-26
GDFS and GDPWR NRFS services and Zephyr implementations

bjarki-andreasen:
Blocked until WiFi changes have been merged https://github.com/nrfconnect/sdk-nrf/pull/19147, and SDSC bundle is ready with GDSF and GDPWR services.


soc/nordic/nrf54h/Kconfig
```diff
@@ -71,3 +73,23 @@ config SOC_NRF54H20_NO_MRAM_LATENCY
...
+	select NRFS_GDFS_SERVICE_ENABLED
+	select CLOCK_CONTROL
+	default y if SOC_NRF54H20_CPUAPP
+
+if SOC_NRF54H20_GLOBAL_HSFLL_RESTRICT_MIN_FREQ
+
+config SOC_NRF54H20_GLOBAL_HSFLL_MIN_FREQ_HZ
+	int "Minimum global HSFLL clock frequency in Hertz"
+	range 64000000 320000000
+	default 320000000
```
tmon-nordic, line 89: Having a range here is pretty pointless IMHO because the only allowed frequencies are: 64, 128, 256 and 320 MHz. Please change it into a choice of one of the allowed values.

bjarki-andreasen: will do

bjarki-andreasen: done

tmon-nordic: How to make `SOC_NRF54H20_GLOBAL_HSFLL_MIN_FREQ_64_MHZ` selected with prj.conf? No matter what I try it always defaults to 320 MHz.

bjarki-andreasen: `CONFIG_SOC_NRF54H20_GLOBAL_HSFLL_MIN_FREQ_64_MHZ=y`

tmon-nordic: Thanks, turns out I was modifying wrong `prj.conf` file (I should have modified `usbd_next_prj.conf`).

soc/nordic/nrf54h/Kconfig
```diff
@@ -67,3 +69,36 @@ config SOC_NRF54H20_CPUFLPR
 	depends on RISCV_CORE_NORDIC_VPR
 
 rsource "gpd/Kconfig"
+
+config SOC_NRF54H20_GLOBAL_HSFLL_RESTRICT_MIN_FREQ
+	bool "Restrict minimum global HSFLL clock frequency"
```
karstenkoenig, line 74: The name is unfortunate as there is three global hsflls, 120, 121 and 200, I'd recommend renaming the GLOBAL_HSFLL to HSFLL120 throughout. I know it creates a detail level that most wouldn't want to be bothered with, but it will just make things confusing long term when it comes to dealing with the other global hsflls

soc/nordic/nrf54h/Kconfig
```diff
@@ -67,3 +69,36 @@ config SOC_NRF54H20_CPUFLPR
...
+	select NRFS
+	select NRFS_GDFS_SERVICE_ENABLED
+	select CLOCK_CONTROL
+	default y if SOC_NRF54H20_CPUAPP
```
ppryga-nordic, line 78: Why this is enabled by default when APP core is selected? What if there is an application that uses Radio core only?

drivers/clock_control/clock_control_nrf2_global_hsfll.c
```diff
@@ -0,0 +1,291 @@
...
+	rc = k_sem_take(&dev_data->evt_sem, GLOBAL_HSFLL_FREQ_REQ_TIMEOUT);
+	if (rc) {
+		return -EIO;
+	}
+
+	if (dev_data->evt != NRFS_GDFS_EVT_FREQ_CONFIRMED) {
+		return -EIO;
+	}
+
+	nrfs_gdfs_uninit();
```
lstnl, line 257: Why do you uninit here ?

hakonfam: I guess its to have the same state after the driver has initialized regardless off whether or not you have enabled `GLOBAL_HSFLL_INIT_LOW_REQ` ?

we
hakonfam, line 1: This should be removed

hakonfam (changes requested):

drivers/clock_control/clock_control_nrf2_global_hsfll.c
lstnl: ok I see this is when GLOBAL_HSFLL_INIT_LOW_REQ is enabled

lstnl (approved):
Just remove this we file.

soc/nordic/nrf54h/Kconfig
bjarki-andreasen: This has been removed :)

we
bjarki-andreasen: has it been? github is showing no line change?

hakonfam (approved):

we
bjarki-andreasen: removed, no idea how that ended there :D

bjarki-andreasen: ok, it was part of upstream, added upstream commit which removes it :)

PR #2310, opened 2024-11-28
[nrf noup] twister: fix linter issue in blackbox tests

gchwier:
Change also in upstream in PR #78716, but with many other changes.


gchwier:
fix is in Upstream, will be added with next upmerge

PR #2317, opened 2024-11-28
[nrf fromlist] soc: nordic: nrf54l: remove configuration of DCDC regulator

nika-nordic:
Since nrfx 3.9 integration, configuration is executed in MDK SystemInit().

Additionally, add workaround for nRF54L anomaly 31.

Upstream PR #: 82268

Link: https://github.com/zephyrproject-rtos/zephyr/pull/82268


nika-nordic:
Checked the SystemInit() MDK code locally on actual DK and there is no regression between previous Zephyr-based approach and current MDK-based approach

kl-cruz (approved):

PR #2319, opened 2024-11-29
Set status okay to DPPI and PPIB nodes on nrf54l

e-rk:
The DPPI and PPIB peripheral nodes must be enabled to allow the
CONFIG_HAS_HW_NRF_DPPIC to be set. This change is consistent with what
was done on nRF5340 and does not introduce any additional memory
overhead, because there is no Zephyr driver behind the nrf-dppic and
nrf-ppib bindings.


kl-cruz (approved):

rugeGerritsen (approved):

e-rk:
https://github.com/nrfconnect/sdk-zephyr/pull/2357 needs to be pulled first

PR #2324, opened 2024-12-02
Add RPU recovery info

kapbh:
Add RPU recovery information in wifi utils.

### DIFF
--- a/drivers/wifi/nrf_wifi/src/wifi_util.c
+++ b/drivers/wifi/nrf_wifi/src/wifi_util.c
@@ -915,6 +915,52 @@ unlock:
 	k_mutex_unlock(&ctx->rpu_lock);
 	return ret;
 }
+
+static int nrf_wifi_util_rpu_recovery_info(const struct shell *sh,
+					   size_t argc,
+					   const char *argv[])
+{
+	struct nrf_wifi_fmac_dev_ctx *fmac_dev_ctx = NULL;
+	struct nrf_wifi_hal_dev_ctx *hal_dev_ctx = NULL;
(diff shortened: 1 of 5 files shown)


sachinthegreen (approved):

krish2718 (approved):

drivers/wifi/nrf_wifi/src/wifi_util.c
```diff
@@ -915,6 +915,52 @@ static int nrf_wifi_util_trigger_rpu_recovery(const struct shell *sh,
...
+static int nrf_wifi_util_rpu_recovery_info(const struct shell *sh,
+					   size_t argc,
+					   const char *argv[])
+{
+	struct nrf_wifi_fmac_dev_ctx *fmac_dev_ctx = NULL;
+	struct nrf_wifi_hal_dev_ctx *hal_dev_ctx = NULL;
+	unsigned long current_time_ms = nrf_wifi_osal_time_get_curr_ms();
+	int ret;
+
+	k_mutex_lock(&ctx->rpu_lock, K_FOREVER);
```
VivekUppunda, line 928: This seems wrong. You are taking the lock which is a part of `ctx` and later checking whether `ctx` is NULL or not? you should first check if `ctx` is NULL and then take the lock.

kapbh: Yes, need to check before taking lock. It seems it contains everywhere in the file. I will take care of this in separate PR for all change in wifi_utils.

VivekUppunda (approved):

krish2718 (approved):

PR #2354, opened 2024-12-10
[nrf noup] use -DUSE_PARTITION_MANAGER=1 for lld linker script template

tejlmand:
fixup! [nrf noup] tree-wide: support NCS Partition Manager (PM) definitions

### DIFF
--- a/cmake/linker/lld/target.cmake
+++ b/cmake/linker/lld/target.cmake
@@ -52,6 +52,7 @@ macro(configure_linker_script linker_script_gen linker_pass_define)
     -imacros ${AUTOCONF_H}
     ${current_includes}
     ${template_script_defines}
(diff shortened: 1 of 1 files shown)


cmake/linker/lld/target.cmake
```diff
@@ -52,6 +52,7 @@ macro(configure_linker_script linker_script_gen linker_pass_define)
     -imacros ${AUTOCONF_H}
     ${current_includes}
     ${template_script_defines}
+    -DUSE_PARTITION_MANAGER=$<OR:$<BOOL:${CONFIG_PARTITION_MANAGER_ENABLED}>,$<BOOL:${IMAGE_NAME}>,$<TARGET_EXISTS:partition_manager>>
```
nordicjm, line 55: can optionally be condensed to `-DUSE_PARTITION_MANAGER=$<BOOL:${CONFIG_PARTITION_MANAGER_ENABLED}>` since this doesn't need to support child/parent and it's currently being removed

tejlmand: as you said yourself, parent/child is currently being removed so it's anyhow a short-lived solution, and for that I simply prefer to keep this line aligned with the construct used already in the `ld/target.cmake`.

PR #2360, opened 2024-12-11
[nrf fromlist] soc: nordic: poweroff: disable remaining RAM on emul l05/l10 targets

nika-nordic:
nRF54L05 and nRF54L10 devices that are emulated on nRF54L15 needs to have manually disabled remaining RAM blocks that are outside of their RAM region definitions.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/82262


nordic-piks (approved):
It fixes increased current during system off for L05 and L10.

miha-nordic (approved):

PR #2363, opened 2024-12-12
[nrf noup] soc: nordic: nrf54l Fix for wait lasting half expected time

degjorva:
Noup as it patches code that does not exist in zephyr. arch_busy_wait and k_busy_wait waited half the expected time due to SystemCoreClock being set back to the default value when ns builds. Changed to rely on device tree as to avoid hardcoded clock frequencies


tomi-font (approved):

Vge0rge (approved):

RobertGalatNordic:
verifeid that the change fixes the issue.

RobertGalatNordic (approved):

PR #2367, opened 2024-12-12
tests: tls_configurations: adjust Kconfig in overlay-ec

frkv:
CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_BASIC is promptless so it cannot be selected. Moreover it's also automatically enabled by CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE in the same overlay file so there would be no need to explicitly enable it.

As for the IMPORT, EXPORT, DERIVE they are needed for the TLS connection to work properly. Previously it was working because at least IMPORT and EXPORT are internally enabled by Mbed TLS at build time. So here we are basically doing the same enablements with Kconfigs in clear.


shanthanordic:
Cleared milestone

PR #2379, opened 2024-12-13
drivers: counter: nrfx_timer: Adjust timer to request global hsfll

nordic-krch:
It is based on #2377.


bjarki-andreasen:
Merged as part of https://github.com/nrfconnect/sdk-zephyr/pull/2385

PR #2408, opened 2024-12-20
[nrf fromtree] net: lib: tls_credentials: return size required

plskeggs:
If either no buffer is provided or the size of it is too small, return the required length.

(cherry picked from commit 6ec5729a3dd80ef8bd4b2aa02263e6d4b75e5e72)


plskeggs:
Not required -- is part of next upmerge.

PR #2417, opened 2025-01-10
[nrf noup] drivers: clock_control: nrf: skip nrf_clock.h on haltium

jonathannilsen:
Don't include hal/nrf_clock.h on Haltium platforms as that HAL currently does not compile for those. Since this issue only affects projects that are downstream of NCS, and should be fixed soon by NRFX-6892, this patch is made as a temporary no-up that can be reverted along with the introduction of the fix.

Ref: NCSDK-31213


nika-nordic:
duplicate of: https://github.com/nrfconnect/sdk-zephyr/pull/2397 , which can be actually upstreamed (just as this PR)

jonathannilsen:
> duplicate of: #2397 , which can be actually upstreamed (just as this PR)

Ah, thanks for letting me know. Closing this then

PR #2431, opened 2025-01-16
[nrf fromlist] arch: riscv: option to init custom hw stacked esf memb…

lstnl:
…ers.

When RISCV_SOC_HAS_ISR_STACKING is used, it may
be needed to initialize custom hw stacked esf members. Some initial values may need to be aligned with
hw stacking mechanism to avoid any side effects.

Upstream PR #: 84076

This PR fixes issue https://nordicsemi.atlassian.net/browse/NCSDK-31064

### DIFF
--- a/arch/riscv/Kconfig
+++ b/arch/riscv/Kconfig
@@ -83,6 +83,11 @@ config RISCV_SOC_HAS_ISR_STACKING
 	    saved on the stack by the hardware, and the registers saved by the
 	    software macros. The structure must be called 'struct arch_esf'.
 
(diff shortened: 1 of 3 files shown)


masz-nordic (approved):

karstenkoenig (approved):

karstenkoenig:
Also affects the other vpr cores so even if sysctrl gets built off main sdk-zephyr/sdk-nrf it will need to be backported to nrf54h20 branch, so added the backport tag
Guess we should add it the the 54L products too?

PR #2433, opened 2025-01-16
nrf54h20 flpr fixes

hubertmis:
Write-back cache after copying VPR's code and data to the destination memory. It allows VPRs to run from RAM regions cacheable by its host.

It allows to use FLPR in nRF54H20.

Upstream PR #: 84104

### DIFF
--- a/drivers/misc/nordic_vpr_launcher/nordic_vpr_launcher.c
+++ b/drivers/misc/nordic_vpr_launcher/nordic_vpr_launcher.c
@@ -7,6 +7,7 @@
 
 #include <string.h>
 
+#include <zephyr/cache.h>
 #include <zephyr/devicetree.h>
 #include <zephyr/init.h>
 #include <zephyr/logging/log.h>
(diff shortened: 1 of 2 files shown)


57300 (approved):

nordic-piks (approved):

karstenkoenig:
Can you please add in this here to fix the stacking sequence?
https://github.com/zephyrproject-rtos/zephyr/pull/83997/commits/7523ebd35a39d9afbd776430e91656820c367663

karstenkoenig (changes requested):
There is a bunch of vpr related fixes now that are all relevant for nrf54h20 and so will also need a backport. We could avoid long CI runtimes and a big number of PRs by just collecting them all here.
Mandatory
https://github.com/zephyrproject-rtos/zephyr/pull/84104
https://github.com/zephyrproject-rtos/zephyr/pull/83997
https://github.com/zephyrproject-rtos/zephyr/pull/84076
Optional
https://github.com/zephyrproject-rtos/zephyr/pull/83940

karstenkoenig:
Note @lstnl  there is more vpr related fixes, this here is to fix missing cache handling. I think we should group them into a single PR including your esf fix

karstenkoenig (approved):
After discussion agreed to keep https://github.com/zephyrproject-rtos/zephyr/pull/84076 out due to criticality of that. Thanks for cherry-picking the stacking order change in!

PR #2457, opened 2025-01-31
[Backport v2.2-branch] Bluetooth: Controller: Fix device address in Periodic Adv Sync

cvinayak:
Fix incorrect device address reported in the LE Periodic
Advertising Sync Established event when using Periodic
Advertiser List.

During Extended Scanning there can be an ADV_EXT_IND PDU
received between currently being received ADV_EXT_IND PDU
and AUX_ADV_IND PDU; if the one received between has an
address match then incorrectly the Periodic Synchronization
was established to the device whos AUX_ADV_IND PDU is being
received. Fix by storing the auxiliary context that has the
address match and compare with it when matching the SID in
SyncInfo of AUX_ADV_IND PDU being received prior to creating
the synchronization.

Signed-off-by: Vinayak Kariappa Chettimada <vich@nordicsemi.no>

### DIFF
--- a/subsys/bluetooth/controller/ll_sw/ull_scan_aux.c
+++ b/subsys/bluetooth/controller/ll_sw/ull_scan_aux.c
@@ -361,19 +361,30 @@ void ull_scan_aux_setup(memq_link_t *link, struct node_rx_hdr *rx)
 
 	ptr = h->data;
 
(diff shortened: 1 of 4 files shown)


cvinayak:
Closing, not active for some time now.

PR #2479, opened 2025-02-10
[nrf fromlist] drivers: flash: mspi_nor: support MODE_SINGLE and MODE_QUAD_1_4_4

masz-nordic:
drivers: flash: mspi_nor: support MODE_SINGLE and MODE_QUAD_1_4_4

### DIFF
--- a/drivers/flash/flash_mspi_nor.c
+++ b/drivers/flash/flash_mspi_nor.c
@@ -28,6 +28,177 @@ struct flash_mspi_nor_data {
 	struct mspi_xfer xfer;
 };
 
+struct flash_mspi_nor_cmd {
+	enum mspi_xfer_direction dir;
+	uint32_t cmd;
+	uint16_t tx_dummy;
+	uint16_t rx_dummy;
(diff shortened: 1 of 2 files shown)


nika-nordic (approved):

drivers/flash/flash_mspi_nor.c
```diff
@@ -450,53 +619,39 @@ static int flash_chip_init(const struct device *dev)
...
-		LOG_ERR("Failed to read JEDEC ID in single line mode: %d", rc);
+		LOG_ERR("Failed to read JEDEC ID in initial line mode: %d", rc);
 		return rc;
 	}
 
-	/*
+	/* For octal mode:
 	 * If the read ID does not match the one from DTS, assume the flash
 	 * is already in the Octa I/O mode, so switching it is not needed.
 	 */
```
magp-nordic, line 636: Maybe there should be a note about QUAD mode too. Right now it is easy to overlook that for QUAD this if is true.

magp-nordic (approved):

PR #2489, opened 2025-02-13
[nrf noup] lib: net_buf: buf: Revert alloc DBG to WRN change

sean-madigan:
This warning is raised in many places when using bluetooth. However, it is not very useful for the user to see.

This is a noup PR as there is currently more changes coming in around this code and how to handle deadlocks in zephyr.


alwa-nordic:
Note on noup resolution: This code is removed upstream in https://github.com/zephyrproject-rtos/zephyr/commit/e13e893dfcf5754b002421ad9af83eab09f0c559, so this noup should be dropped when we rebase main on upstream (upmerge).

PR #2504, opened 2025-02-17
Remove leftover socket options

MirkoCovizzi:
* Removes leftover socket options `TLS_SESSION_CACHE_DISABLED`
and `TLS_SESSION_CACHE_ENABLED`. These exist already
in `socket.h`.

* Adds a base value for NCS specific socket options to
prevent overlaps with Zephyr's socket options.

### DIFF
--- a/include/zephyr/net/socket_ncs.h
+++ b/include/zephyr/net/socket_ncs.h
@@ -49,10 +49,6 @@ extern "C" {
  */
 #define TLS_DTLS_HANDSHAKE_STATUS 21
 
-/* Valid values for TLS_SESSION_CACHE option */
-#define TLS_SESSION_CACHE_DISABLED 0 /**< Disable TLS session caching. */
-#define TLS_SESSION_CACHE_ENABLED 1 /**< Enable TLS session caching. */
-
 /* Valid values for TLS_DTLS_HANDSHAKE_TIMEO option */
 #define TLS_DTLS_HANDSHAKE_TIMEO_NONE 0 /**< No timeout */
 #define TLS_DTLS_HANDSHAKE_TIMEO_1S 1 /**< 1 second */
--- a/include/zephyr/net/socket_ncs.h
+++ b/include/zephyr/net/socket_ncs.h
@@ -25,6 +25,10 @@ extern "C" {
 #define SOCK_NATIVE 0x80000000
 #define SOCK_NATIVE_TLS 0x40000000
 
+/** Define a base for NCS specific socket options to prevent overlaps with Zephyr's socket options.
+ */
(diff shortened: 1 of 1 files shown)


eivindj-nordic (approved):

rlubos (approved):

include/zephyr/net/socket_ncs.h
```diff
@@ -25,33 +25,34 @@ extern "C" {
 #define SOCK_NATIVE 0x80000000
 #define SOCK_NATIVE_TLS 0x40000000
 
+/** Define a base for NCS specific socket options to prevent overlaps with Zephyr's socket options.
+ *  Allocates to NCS specific socket options the signed integer range 1073741824 to INT32_MAX.
+ */
+#define NET_SOCKET_NCS_BASE 0x4000000
```
lemrey, line 31: Could we use a round integer number here? It is not very useful to have a round Hex number in this case

include/zephyr/net/socket_ncs.h
```diff
@@ -25,33 +25,34 @@ extern "C" {
...
+
 /* NCS specific TLS level socket options */
 
 /** Socket option to set DTLS handshake timeout, specifically for nRF sockets.
  *  The option accepts an integer, indicating the total handshake timeout,
  *  including retransmissions, in seconds.
  *  Accepted values for the option are: 1, 3, 7, 15, 31, 63, 123.
  */
-#define TLS_DTLS_HANDSHAKE_TIMEO 18
+#define TLS_DTLS_HANDSHAKE_TIMEO (NET_SOCKET_NCS_BASE + 18)
```
lemrey, line 40: side note, did we not have a `SO_` prefix or similar for these?

include/zephyr/net/socket_ncs.h
MirkoCovizzi: This number is exactly halfway through the positive integer range. Do you have a different suggestion?

lemrey: Yes, to use a round integer number like e.g. `100000`. The reason is that this number is `1073741824` in decimal, and combined with decimal definitions of socket option values makes it is hard to debug this values when we need to. For example, which socket option is `1073741842`?

When you print it in hex, it's `0x40000012` and note that `12` is in hex, so we need to convert to integer to figure out it's actually socket option `18` that is, `TLS_DTLS_HANDSHAKE_TIMEO`. If we saw `100018` it would be easier to tell what we are looking at.

MirkoCovizzi: Thanks for the clarification.

lemrey (approved):

PR #2510, opened 2025-02-17
soc: nordic: nrf54l: soc: Rework to remove noup commits.

nordic-krch:
`soc.c` diverged with few `noup` commits. This is an attempt to make soc.c again purely upstream based.


nordicjm (approved):

mstasiaknordic (approved):

Vge0rge:
Looks good and I tested it with TFM and seems to work fine.

Vge0rge (approved):

PR #2522, opened 2025-02-18
[nrf noup] entropy: Add fake entropy nRF PRNG driver

e-rk:
This adds temporary entropy driver simulation for
nRF54l09 device since final entropy source is not
available yet.

TODO: Remove this commit when proper solution will be available.


e-rk:
> For NCS we have the nrf_security diver, but I presume @e-rk wants to enable this one because the nrf_security driver also does not support an L09 device yet.
Yes, we need some entropy source, because many applications won't build without it.

piotrkoziar (approved):

nordic-krch (approved):

PR #2527, opened 2025-02-19
[nrf noup] dts: Select SoftDevice Controller on nRF54L09

e-rk:
The SDC HCI controller is defined as a device tree node. A node representing the SDC controller is added and selected as the default over the open source link layer. This is consistent with other SoCs.


rugeGerritsen (approved):
Going forward we should consider finding some more scalable way which doesn't require noup-commits.

PR #2531, opened 2025-02-20
[nrf noup] mbox: Temporary workaround for SCFW not capturing some ic-…

meijemac:
Woraround that sends a second vevif SW trigger. It seems that SCFW does not always catches each vevif IRQ - therefore a second time the vevif task will be triggered as a temporary workaround.

### DIFF
--- a/drivers/mbox/mbox_nrf_vevif_task_tx.c
+++ b/drivers/mbox/mbox_nrf_vevif_task_tx.c
@@ -5,12 +5,14 @@
 
 #define DT_DRV_COMPAT nordic_nrf_vevif_task_tx
 
+#include <zephyr/kernel.h>
 #include <zephyr/devicetree.h>
 #include <zephyr/drivers/mbox.h>
 
 #include <haly/nrfy_vpr.h>
 
 #define TASKS_IDX_MAX NRF_VPR_TASKS_TRIGGER_MAX
+#define VEVIF_RETRIGGER_DELAY	12
 
 struct mbox_vevif_task_tx_conf {
 	NRF_VPR_Type *vpr;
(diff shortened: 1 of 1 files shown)


drivers/mbox/mbox_nrf_vevif_task_tx.c
```diff
@@ -39,6 +40,10 @@ static int vevif_task_tx_send(const struct device *dev, uint32_t id, const struc
 
 	nrfy_vpr_task_trigger(config->vpr, nrfy_vpr_trigger_task_get(id));
 
+	k_busy_wait(VEVIF_RETRIGGER_DELAY);
```
lstnl, line 43: #include <zephyr/kernel.h> missing for k_busy wait

gmarull (dismissed):
why is this a noup?

meijemac:
> why is this a noup?

Don't know what it should be - for the moment this is a workaround so that we can make a SDSC release. We are still investigating and we hope to find the root cause - so that we can elimate this workaround.

drivers/mbox/mbox_nrf_vevif_task_tx.c
```diff
@@ -5,12 +5,14 @@
...
 #define DT_DRV_COMPAT nordic_nrf_vevif_task_tx
 
+#include <zephyr/kernel.h>
 #include <zephyr/devicetree.h>
 #include <zephyr/drivers/mbox.h>
 
 #include <haly/nrfy_vpr.h>
 
 #define TASKS_IDX_MAX NRF_VPR_TASKS_TRIGGER_MAX
+#define VEVIF_RETRIGGER_DELAY	12
```
57300, line 15: Do you know if this is needed for all VPRs?
Should we have this as a Kconfig, where value 0 means "no workaround needed"? Then we could set it out-of-tree for SysCtrl, and there wouldn't be a problem with sending this upstream.

meijemac: We don't know

57300 (approved):
LGTM, assuming a noup is acceptable until the root cause of HM-25973 is found

lstnl (approved):

tmon-nordic (approved):

PR #2541, opened 2025-02-20
[nrf noup] tests: drivers: flash: common: run on SDP MSPI

jaz1-nordic:
Only last commit.
Based on #2479.

### DIFF
--- a/tests/drivers/flash/common/src/main.c
+++ b/tests/drivers/flash/common/src/main.c
@@ -32,7 +32,9 @@
 #elif defined(TEST_AREA_DEV_NODE)
 
 #define TEST_AREA_DEVICE	DEVICE_DT_GET(TEST_AREA_DEV_NODE)
+#ifndef TEST_AREA_OFFSET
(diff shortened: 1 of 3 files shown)


magp-nordic (approved):

tests/drivers/flash/common/src/main.c
```diff
@@ -32,7 +32,11 @@
 #elif defined(TEST_AREA_DEV_NODE)
 
 #define TEST_AREA_DEVICE	DEVICE_DT_GET(TEST_AREA_DEV_NODE)
+#if defined(CONFIG_SOC_NRF54L15_CPUAPP) && defined(CONFIG_MSPI_NRFE)
```
nordicjm, line 35: this is kind of yuck really modifying an upstream test file like this, would be better to have a test in sdk-nrf instead and reference upstream files or new files so that it doesn't break in an upmerge, or even making these Kconfigs in the test upstream then adding a no-up conf for the board

masz-nordic:
Test replicated in sdk-nrf. Closing.

PR #2545, opened 2025-02-21
[nrf noup] samples: code_relocation_nocopy: Support NS variant of nRF5340 DK

anangl:
Add Kconfig fragment that allows using the sample with the non-secure variant of the board.
Applied as a noup because upstream the build fails on signing the image.


nordicjm:
I don't know what the intention is here but this does absolutely nothing for QSPI split image support, as can be seen by trying to build https://github.com/nrfconnect/sdk-nrf/tree/main/samples/nrf5340/extxip_smp_svr for an ns board target:
```
CMake Error at /tmp/bb/nrf/cmake/sysbuild/image_signing_split.cmake:147 (message):
  TFM is not currently supported with QSPI XIP
```
All this does is allow a single sample to build which does not support updates in any way

bjarki-andreasen:
Requires rebase since upmerge :)

carlescufi:
PoC, can be closed.

PR #2548, opened 2025-02-21
[nrf noup] modules: hal_nordic: cleanup nrfx_config

masz-nordic:
- Remove temporary workaround for SDC in nrfx's reserved resources.
- Modify NCS reserved resources to use upstream version with additions instead of creating new file
- Align filename


bjarki-andreasen:
Requires rebase since upmerge :)

PR #2576, opened 2025-03-05
include: common: sys_bitops: Specify sign when bitshifting

rugeGerritsen:
When left-shifting '1' by 31, the result is undefined. This is something ASAN detects.

Solve this by explicitly defining that the integer is unsigned.

Upstream PR #: 86624

### DIFF
--- a/include/zephyr/arch/common/sys_bitops.h
+++ b/include/zephyr/arch/common/sys_bitops.h
@@ -25,21 +25,21 @@ static ALWAYS_INLINE void sys_set_bit(mem_addr_t addr, unsigned int bit)
 {
 	uint32_t temp = *(volatile uint32_t *)addr;
 
-	*(volatile uint32_t *)addr = temp | (1 << bit);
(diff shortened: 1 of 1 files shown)


include/zephyr/arch/common/sys_bitops.h
```diff
@@ -62,19 +62,19 @@ static ALWAYS_INLINE
 	/* Doing memory offsets in terms of 32-bit values to prevent
 	 * alignment issues
 	 */
-	sys_set_bit(addr + ((bit >> 5) << 2), bit & 0x1F);
+	sys_set_bit(addr + ((bit >> 5) << 2U), bit & 0x1F);
```
jostn, line 65: Is it needed to have 2U here? Is this implementation defined?

rugeGerritsen: There is no need to use `2U` here, I just added it for consistency. Same question was asked in the upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/86624

jostn (approved):

PR #2595, opened 2025-03-10
scripts: support number of jobs and verbosity in west build.

tejlmand:
Calling `west build --build-opt="-v" --build-opt="-j=<n>"`
passes the build options to `cmake --build ... -- <build-opt>` which
again is passed to the native build tool, such as ninja or make.
    
However, when ExternalProjects are used in CMake, such as in TF-M or
sysbuild builds then those extra build options are only passed to the
first image build and not those build as external projects.
    
CMake supports environment variables for those flags, so translate
verbosity and number of jobs to those environment variables and there


nordicjm (approved):

kylebonnici (approved):
Needed for ~~https://github.com/nrfconnect/sdk-nrf/pull/20996~~ https://github.com/nrfconnect/sdk-zephyr/pull/2632

PR #2596, opened 2025-03-10
DNM: add zms_get_num_cycles()

Damian-Nordic:
...which calculates the maximum number of times a single
ZMS sector has been recycled. This is to enable estimation
of an RRAM lifetime based on a set of tests.

manifest-pr-skip


subsys/fs/zms/zms.c
```diff
@@ -1828,3 +1853,26 @@ int zms_sector_use_next(struct zms_fs *fs)
 	k_mutex_unlock(&fs->zms_lock);
 	return ret;
 }
+
+uint32_t zms_get_num_cycles(struct zms_fs *fs)
```
rghaddab, line 1857: Why do you need to read all cycle_counts for all sectors ?
Reading the full_cycle_cnt of the current sector should be enough as it holds the maximum number of cycle_cnt.
I don't see scenarios where some closed sectors will have a full_cycle_cnt bigger than the current open sector.

Damian-Nordic: I think it may happen if GC is interrupted and must be restarted.

rghaddab: We increase the cycle only after erasing a sector which is the last operation after the GC has finished.
If the GC got interrupted, next time we won't find a GC_DONE ATE and we restart the GC, the cycle_cnt is increased only at the end.

Damian-Nordic: But if you start copying entries from sector `N+1` to sector `N`, and GC is interrupted, then, after reboot, you must erase `N` again (even though this was erased before GC). This is what https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/fs/zms/zms.c#L1284 implements, no?

rghaddab: you are right, we do this for the current sector and not the next one, only in this case.
I will cherry-pick yours in my PR.
Please go ahead and push it upstream along with its documentation. It is a nice feature to have

omkar3141:
This is great small feature addition that will have long lasting effect, to benchmark and predict the application's usage of Flash/RRAM.

subsys/fs/zms/zms_priv.h
```diff
@@ -57,8 +57,12 @@ struct zms_ate {
 		/** data field used to store small sized data */
 		uint8_t data[8];
 		struct {
-			/** data offset within sector */
-			uint32_t offset;
+			union {
+				/** data offset within sector */
+				uint32_t offset;
+				/** full cycle count (for empty ATE) */
```
omkar3141, line 63: what does "full" indicate here? I think this is causing a bit of confusion for me.

Damian-Nordic: full as opposed to "module 256" that is normally stored in each ATE

rghaddab: the cycle_cnt is truncated (only 8 bits)
The full cycle_cnt is uint32_t

PR #2608, opened 2025-03-12
Fix null pointer to strcmp in posix mqueue

tervonenja:
Pull posix mqueue related fix from upstream.


tervonenja:
I did not quite get it by reading the docs on how you're supposed to get these commits merged from the upstream. But here's a draft. Let's make it happen together, shall we?

PR #2613, opened 2025-03-12
drivers: nrf_rtc_timer and nrfx_uarte

lemrey:
One could be merged to upstream; the other is a temporary solution until a logging backend using `nrfx` is available.


carlescufi (changes requested):
@lemrey I am quite sure we can fix `[nrf noup] drivers: serial: nrfx_uarte: Use Softdevice critical section` so that it can be upstream except for the actual `sd_` calls. @nordic-krch any good ideas?

EDIT: why not override `irq_lock()` globally?

lemrey:
One of the commits was merged separately, the other I do not think we need given that IRQs for peripherals owned by the SoftDevice are connected as `IRQ_ZERO_LATENCY`, which are not locked when `irq_lock()` is called.

PR #2615, opened 2025-03-13
Settings ZMS backend optimizations

rghaddab:
These are series of optimization for the new backend + some commits to syncronize ZMS to the upstream version


subsys/settings/src/settings_zms.c
```diff
@@ -207,7 +207,41 @@ static int settings_zms_load(struct settings_store *cs, const struct settings_lo
...
+				 */
+				continue;
+			}
+			/* At this steps the names are equal, let's set the handler */
+			read_fn_arg.fs = &cf->cf_zms;
+			read_fn_arg.id = ZMS_NAME_ID_FROM_HASH(name_hash) + ZMS_DATA_ID_OFFSET;
 
+			ret = settings_call_set_handler(arg->subtree, rc2, settings_zms_read_fn,
+							&read_fn_arg, (void *)arg);
+			return ret;
```
Damian-Nordic, line 244: Unfortunately, you cannot return here. If `subtree == "abc"`, the code is expected to return not only "abc" but also "abc/def" etc. You can only break if `ret != 0`.

rghaddab: @Damian-Nordic you are right I will change that.
I actually didn't see any application using this feature of settings where you can load subtree "abc" and it will load abc/*

omkar3141 (dismissed):
Could you check sonar cloud issue and see if you can resolve as many as possible if implementation is settled?

I haven't re-read the full PR yet.

rghaddab:
> Could you check sonar cloud issue and see if you can resolve as many as possible if implementation is settled?
> 
> I haven't re-read the full PR yet.

I resolved as many Sonar Cloud Issues as possible.

carlescufi:
@omkar3141 @Damian-Nordic please take another look

carlescufi (approved):

PR #2585, opened 2025-03-06
[nrf fromtree] drivers: i2c_nrfx_twim: add exlusive access API

ankuns:
This commit provides an extension to the i2c_nrfx_twim driver. It introduces possibility to acquire/release exclusive access to the i2c bus

The provided API is necessary for use cases when, i2c transfers must be started from interrupt context, while still allowing to share the i2c bus in multi-tasking environment.

PR #2624, opened 2025-03-17
[nrf noup] drivers: i2c_nrfx_twim: add asynchronous write and exclusive access API

ankuns:
This commit provides an extension to the i2c_nrfx_twim driver. It introduces possibility to:
- acquire/release exclusive access to the i2c bus
- perform asynchronous i2c write transfers.

The provided API is necessary for use cases when, i2c transfers must be started from interrupt context, while still allowing to share the i2c bus in multi-tasking environment.

This PR is necessary to implement the temperature compensation feature for the nRF2220 Front-End Module that works with MPSL and still allows to share i2c bus with drivers implementing the ordinary i2c.h API.
This is covered by PR https://github.com/nrfconnect/sdk-nrf/pull/20960

The extension is not welcome in upstream
https://github.com/zephyrproject-rtos/zephyr/pull/86721
It works only for Nordic SoCs, but the MPSL also works only with Nordic SoCs.
Upstream-first approach can't be applied in this case.

I do not encourage anyone to use the proposed extension API, however it is necessary to implement the nRF2220 temperature compensation feature.

Usage:
Will be shortly presented with introduction of the nRF2220 FEM temperature compensation.
From a thread context call `i2c_nrfx_twim_exclusive_access_acquire`
We have the exclusive access to the i2c bus, which prevents other users of the i2c from messing with the underlying hardware driver.
Then we can perform write transfers over the i2c bus to an i2c slave device, possibly from interrupt context by calling i2c_nrfx_twim_async_transfer_begin.
When we are done we call i2c_nrfx_twim_exclusive_access_release.

The transfers `i2c_nrfx_twim_async_transfer_begin` for the nRF2220 temperature compensation must happen within a timeslot negotiated with MPSL scheduler.


nordic-piks:
Please add tests/samples for this new features at driver.

masz-nordic:
>The extension is not welcome in upstream

How so? I do not see change requests, just discussion.

ankuns:
@masz-nordic 

> Does that mean the TWIM peripheral is used at the same time in Zephyr driver and a baremetal driver?

No, the Zephyr driver is used only. The nrf2220 FEM from driver from MPSL does not have it's own i2c driver. It requires to provide functions operating on i2c bus by an interface structure containing function pointers to actual implementation. In NCS these pointers point to a thin wrappers over the i2c driver. The bare metal is not supported, but in case of need such structure pointing to appropriate i2c implementation can be provided in bare metal application.

This approach was enough while nrf2220 needed i2c only at bootup stage, to write initial set of registers. The i2c was not used never again. But the nrf2220 temperature compensation feature is different. The i2c needs to be used also after bootup while device is operating. It needs to be able to perform an i2c write transfer from an interrupt given by the MPSL scheduler (it governs access to the radio and thus the FEM, the transfers may be perfromed while the nrf2220 is in very particular state, "bypass"). So the idea is to first take the exclusive access to the i2c bus, then request MPSL scheduler for a timeslot (yes in this time we prevent other i2c transfers from happening because we hold exclusive access, that's intended). When MPSL scheduler gives us a timeslot we have to perfrom i2c write, without delays (the MPSL scheduler requirement). We need asynchronous write transfers to perform that. 

This approach has the benefit that the i2c bus can still be shared with other ordinary Zephyr devices/features following the ordinary zephyr/drivers/i2c.h api (at least for I2C_SHELL feature that is used for testing). The only limitations imposed by the nrf2220 temperature compensation feature are:
- rarely (when temperature compensation registers are to be written to the nrf2220), we take exclusive access and negotiate mpsl timeslot. 
- we must use i2c_nrfx_twim implementation of the driver, but MPSL works only on nRF devices, so not really a limitation.

ankuns:
@nordic-piks 
> Please add tests/samples for this new features at driver.

Added `drivers.i2c.nrfx_twim_async_api`

carlescufi (changes requested):
This is too big for a `noup`, it needs discussion for better integration.

ankuns:
@carlescufi I agree, that why there is
https://github.com/nrfconnect/sdk-zephyr/pull/2585
and upstream
https://github.com/zephyrproject-rtos/zephyr/pull/86721
Added you to a chat on Teams.

ankuns:
With this PR https://github.com/nrfconnect/sdk-zephyr/pull/2585
only the exclusive access acquire/release went in. The asynchronous transfers will be done differently not on the Zephyr level.

PR #2632, opened 2025-03-18
[nrf fromlist] cmake: add tfm / tfa image information to build info

tejlmand:
Support for list of images in build info was added with commit 4061311da3ee2ad8f048bf3e94cfb42a8681b6b1 and is used by sysbuild.

Zephyr itself also uses CMake's External Project feature when including TF-M or TF-A in a Zephyr build.

Populate build info with TF-M / TF-A information when said image is included in the build.

Upstream PR #: 86881

### DIFF
--- a/modules/trusted-firmware-a/CMakeLists.txt
+++ b/modules/trusted-firmware-a/CMakeLists.txt
@@ -8,6 +8,9 @@ if (CONFIG_BUILD_WITH_TFA)
 
   include(ExternalProject)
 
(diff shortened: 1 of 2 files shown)


nordicjm (approved):

kylebonnici (approved):

kylebonnici:
@tejlmand 

Just made an observation 

image

type is not set like we do at sysbuild level

image

I think type should be set to allow us to differentiate

PR #2634, opened 2025-03-18
[nrf fromlist] cmake: support directory as argument to zephyr_library…

tejlmand:
…_amend().

Extend `zephyr_library_amend()` to support an optional directory argument.
The current `zephyr_library_amend()` works well when used inside a Zephyr module with same structure, but fails when the macro is called from Zephyr module integration code is located in a Zephyr `MODULE_EXT_ROOT` because in this case the CMake code being executed is not present in the Zephyr module itself, in which case the dir name creation based on relative to module dir give wrong result.

For this use-case then support a base directory.
This also allows for use-cases in Zephyr modules where the directory structure matching Zephyr's own structure is placed in a sub-folder.

Upstream PR#: 87274

### DIFF
--- a/cmake/modules/extensions.cmake
+++ b/cmake/modules/extensions.cmake
@@ -463,8 +463,17 @@ endmacro()
 
 # Provides amend functionality to a Zephyr library for out-of-tree usage.
 #
+# Usage:
+#   zephyr_library_amend([<dir>])
+#
 # When called from a Zephyr module, the corresponding zephyr library defined
 # within Zephyr will be looked up.
+#
+# <dir>: Use '<dir>' as out-of-tree base directory from where the Zephyr
+#        library name shall be generated.
+#        <dir> can be used in cases where the structure for the library is not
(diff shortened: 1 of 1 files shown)


nordicjm (approved):

cmake/modules/extensions.cmake
```diff
@@ -463,9 +463,18 @@ endmacro()
 
 # Provides amend functionality to a Zephyr library for out-of-tree usage.
 #
+# Usage:
+#   zephyr_library_amend([<dir>])
+#
 # When called from a Zephyr module, the corresponding zephyr library defined
 # within Zephyr will be looked up.
 #
+# <dir>: Use '<dir>' as base directory from where the Zephyr library name shall
```
Hi-Im-David, line 472:
```suggestion
# <dir>: Use '<dir>' as  out-of-tree base directory from where the Zephyr library name shall be generated.
```

Hi-Im-David (approved):

PR #2637, opened 2025-03-18
DNM: PSA core lite additions

frkv:
Macro additions for PSA core lite.

For PSA core lite, see PRs:
- https://github.com/nrfconnect/sdk-nrf/pull/20812
- https://github.com/nrfconnect/sdk-nrf/pull/20813 (permanently a draft PR, not intended to be taken in)


bjarki-andreasen:
> the changes in `include/zephyr/sys` are way too invasive with no current agreement upstream. @bjarki-andreasen and @nordic-krch, what do you think we should do here?

I already stated I don't think this amount of logic should be in macros at all, kconfig should be used for conditional and programmatic defines, not macros.

PR #2638, opened 2025-03-18
Nrfx none revert s2ram procedures

adamkondraciuk:
Restore pm s2ram procedures to C code form and use temporarily interrupt stack at the early stage.

### DIFF
--- a/soc/nordic/nrf54h/pm_s2ram.c
+++ b/soc/nordic/nrf54h/pm_s2ram.c
@@ -184,18 +184,16 @@ void pm_s2ram_mark_set(void)
 
 bool pm_s2ram_mark_check_and_clear(void)
 {
-	bool unretained_wake;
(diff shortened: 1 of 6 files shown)


soc/nordic/nrf54h/pm_s2ram.c
```diff
@@ -185,8 +185,6 @@ void __attribute__((naked)) pm_s2ram_mark_set(void)
 
 bool __attribute__((naked)) pm_s2ram_mark_check_and_clear(void)
 {
-	register uint32_t link_reg __asm__("r14");
```
nordicjm, line 188: do not `[nrf noup]` a revert

nordicjm (changes requested):

kl-cruz (approved):

nordicjm (approved):

PR #2645, opened 2025-03-19
drivers: flash: mspi_nor: initial config, high performance and fast read

masz-nordic:
Pull in changes from upstream that switch to single lane when required, enable high performance bit when needed and use fast read in single.

### DIFF
--- a/drivers/flash/flash_mspi_nor.c
+++ b/drivers/flash/flash_mspi_nor.c
@@ -22,18 +22,206 @@ LOG_MODULE_REGISTER(flash_mspi_nor, CONFIG_FLASH_LOG_LEVEL);
 #define WITH_RESET_GPIO 1
 #endif
 
+#define FLASH_MX25R_LH_MASK BIT(1)
+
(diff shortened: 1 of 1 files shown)


drivers/flash/flash_mspi_nor.c
```diff
@@ -342,25 +364,54 @@ static int api_read(const struct device *dev, off_t addr, void *dest,
...
-	return mspi_transceive(dev_config->bus, &dev_config->mspi_id, &dev_data->xfer);
+	rc = mspi_transceive(dev_config->bus, &dev_config->mspi_id, &dev_data->xfer);
+
+	if (rc < 0) {
+		LOG_ERR("Status xfer failed: %d", rc);
+		return rc;
+	}
+
+	/* Exit command mode */
+	if (curr_cfg != &dev_config->mspi_nor_init_cfg) {
```
jaz1-nordic, line 396:
```suggestion
	if (curr_cfg != &dev_config->mspi_nor_cfg) {
```
?

masz-nordic: Entering and exiting the command mode needs to be done only when device is not currently using `mspi_nor_init_cfg`.

jaz1-nordic (approved):

PR #2652, opened 2025-03-20
[nrf noup] drivers: flash: align handling reset pin to SDP MSPI

magp-nordic:
Align handling reset pin to SDP MSPI so that it would work with nRF54L15 DK flash, which has RESET pin the same as one of data pins (D3).

Need changes from https://github.com/nrfconnect/sdk-nrf/pull/21076 to work.

### DIFF
--- a/drivers/flash/flash_mspi_nor.c
+++ b/drivers/flash/flash_mspi_nor.c
@@ -13,6 +13,10 @@
 #include <zephyr/pm/device.h>
 #include <zephyr/pm/device_runtime.h>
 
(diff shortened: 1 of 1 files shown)


magp-nordic:
Closing, after discussion with @anangl we came to the conclusion that these changes are not needed, since reset gpio defined in dts should not overlay if any MSPI data pin. If it does, it should be handled internally in MSPI driver as pull-up or holding high state until QUAD mode is enabled in flash, which is the current solution in SDP MSPI.

PR #2656, opened 2025-03-21
[nrf fromtree] boards: nordic: nrf54l09pdk: Change pin used for led0

mstasiaknordic:
P1.7 used for led 0 was not passing gpio_api_1pin test (probably shortened with another pin in the test setup. Use different pin that passes the gpio_api_1pin test. At current stage this PDK is 'virtual' so this pin is not attached to any LED and it can be changed.

(cherry picked from commit b298faa5e35e48dd6137f260946bc0eff1b1bbd4)

### DIFF
--- a/boards/nordic/nrf54l09pdk/nrf54l09pdk_nrf54l09-common.dtsi
+++ b/boards/nordic/nrf54l09pdk/nrf54l09pdk_nrf54l09-common.dtsi
@@ -10,7 +10,7 @@
 	leds {
 		compatible = "gpio-leds";
 		led0: led_0 {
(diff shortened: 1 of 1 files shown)


nordic-piks (approved):

nordic-segl (approved):

adamkondraciuk (approved):

nordic-segl:
FYI - P1.07 seems to be UART CTS signal. This could impact "led" behaviour.

mstasiaknordic:
> FYI - P1.07 seems to be UART CTS signal. This could impact "led" behaviour.

Yes, thats why it got changed to unused P1.10

nordic-krch (approved):

PR #2658, opened 2025-03-21
[nrf noup] tests: crypto: Set size for PSA slot

Vge0rge:
noup since this option does not exist upstream.

The Oberon PSA core has an option to manually set
the buffer size of the PSA key slots in bytes.

This option here:
MBEDTLS_PSA_STATIC_KEY_SLOT_BUFFER_SIZE

This option has a default value of 16 since it
directly affects the memory usage.

The crypto tests use a secp256r1 key and thus since the public key for this key is 65 bytes long we need to set the option in the sample.


degjorva (approved):
Looks good to me

frkv (approved):
LGTM

de-nordic (approved):

tomi-font (commented):
Couldn't we instead make that Kconfig option default to 65 in case a secp256r1 key is used? So as to avoid that noup.

Vge0rge:
> Couldn't we instead make that Kconfig option default to 65 in case a secp256r1 key is used? So as to avoid that noup.

I am not fond of this idea. In such case even the applications which use only an AES key will by default get 65 bytes of buffer for each key slot.  And this also means that the many applications will not set this Kconfig since it will just work which is something that I don't want to promote. I thought about this before but decided that the smallest default make the most sense.

tomi-font:
> I am not fond of this idea. In such case even the applications which use only an AES key will by default get 65 bytes of buffer for each key slot. And this also means that the many applications will not set this Kconfig since it will just work which is something that I don't want to promote. I thought about this before but decided that the smallest default make the most sense.

Of course I don't mean to have 65 as the base default. But just to default to 65 if the Kconfig option(s) enabling secp256r1 keys is enabled?

Vge0rge:
Oh no, that I will not do. This needs a very big a complicated Kconfig statement that I don't want to have just to fix this test failure.

PR #2662, opened 2025-03-24
Nrfx none n rf54 l09 align

adamkondraciuk:
Align build system to nRF54L09 CPUFLPR taget


nordic-piks (approved):

nordic-segl (approved):

mstasiaknordic (approved):

nordicjm (changes requested):
why is this not submitted upstream and brought in via fromlist/fromtree? Why do we 2 additional separate noups? This should either be upstreamed or an out-of-tree board extension

nordic-segl:
There is PR in the Zephyr upstream that could get these two commits:
https://github.com/zephyrproject-rtos/zephyr/pull/85310

adamkondraciuk:
> why is this not submitted upstream and brought in via fromlist/fromtree? Why do we need 2 additional separate noups? This should either be upstreamed or an out-of-tree board extension

fixed [nrf fromlist]

nordic-segl (approved):

nordicjm (approved):

PR #2663, opened 2025-03-25
modules: hal_nordic: bump regtool to 9.0.1

jonathannilsen:
Upstream PR #: 88055

Added support for VPR GPIOs and a more concise output print.

### DIFF
--- a/modules/hal_nordic/CMakeLists.txt
+++ b/modules/hal_nordic/CMakeLists.txt
@@ -12,7 +12,7 @@ if(CONFIG_NRF_REGTOOL_GENERATE_UICR)
   list(APPEND nrf_regtool_components GENERATE:UICR)
 endif()
 if(DEFINED nrf_regtool_components)
-  find_package(nrf-regtool 9.0.1
+  find_package(nrf-regtool 9.0.1 REQUIRED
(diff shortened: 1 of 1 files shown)


karstenkoenig (approved):

57300 (approved):

jonathannilsen:
A bug was found before this got integrated into NCS, so I've now updated to 9.0.1 here

rlubos (changes requested):
Please fix the noup commit titles, we don't add special tags for reverts or reapplies, it'd confuse our tooling. The revert commit should use the default title:
`Revert "[nrf noup] moduled: hal_nordic: require nrf-regtool"`
And then the replied commit description should just be an exact copy of the original one:
`[nrf noup] moduled: hal_nordic: require nrf-regtool`

rlubos:
@jonathannilsen The revert still has the redundant prefix, it should be `Revert "[nrf noup] moduled: hal_nordic: require nrf-regtool"`

jonathannilsen:
> @jonathannilsen The revert still has the redundant prefix, it should be `Revert "[nrf noup] moduled: hal_nordic: require nrf-regtool"`

Ah, thanks, forgot about that one, should be fixed now.

PR #2664, opened 2025-03-25
[nrf fromlist] west: runners: add support for cpusec

hakonfam:
Without this it is not possible to program through the secure AP.

Upstream PR #: 87615


hakonfam:
Solved elsewhere

PR #2666, opened 2025-03-25
[nrf noup] ci: Pin GH actions to SHAs

jangalda-nsc:
Jira: NCSDK-32440


karhama:
Should probably wait for upmerge to avoid conflicts, right?

PR #2676, opened 2025-03-27
Pick bmm350 and mss fix

maxd-nordic:
* add bmm350 driver needed for thingy91x
* fix tls_credentials trusted backend for nRF Cloud libs


carlescufi (approved):

gmarull (changes requested):
noups are unnecessary, pull all necessary changes instead.

bjarki-andreasen (changes requested):
I'm ok with the noup, but please revert the noup as the last commit instead of adding a new undoing commit we then have to track at the next upmerge :)

maxd-nordic:
I think I found a subset that applies cleanly now

bjarki-andreasen (approved):

gmarull (approved):

PR #2678, opened 2025-03-28
NCSDK-32650: Revert incompatible memory changes

tomchy:
Due to compatibility reasons, we should either revert those or provide a clear and tested way to migrate between the v2.9.0-nRF54H20-1 and v3.0.0.

### DIFF
--- a/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20-memory_map.dtsi
+++ b/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20-memory_map.dtsi
@@ -7,32 +7,14 @@
 
 / {
 	reserved-memory {
-		cpurad_ram0x_region: memory@2f010000 {
-			compatible = "nordic,owned-memory";
-			reg = <0x2f010000 DT_SIZE_K(4)>;
-			status = "disabled";
-			nordic,access = <NRF_OWNER_ID_RADIOCORE NRF_PERM_RWS>;
-			#address-cells = <1>;
-			#size-cells = <1>;
-			ranges = <0x0 0x2f010000 0x1000>;
-
-			cpusec_cpurad_ipc_shm: memory@0 {
-				reg = <0x0 DT_SIZE_K(2)>;
-			};
-
-			cpurad_cpusec_ipc_shm: memory@800 {
-				reg = <0x800 DT_SIZE_K(2)>;
-			};
-		};
-
-		cpuapp_ram0x_region: memory@2f011000 {
+		cpuapp_ram0x_region: memory@2f010000 {
 			compatible = "nordic,owned-memory";
 			reg = <0x2f010000 DT_SIZE_K(260)>;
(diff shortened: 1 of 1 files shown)


hubertmis:
This is a board configuration - it is application-specific. This configuration is used in samples running on the DK. It's not used in the product PCBs. Is there any reason to maintain compatibility between application-specific configurations for development kits?

tomchy:
> This is a board configuration - it is application-specific. This configuration is used in samples running on the DK. It's not used in the product PCBs. Is there any reason to maintain compatibility between application-specific configurations for development kits?

If we make a board-level change, than it is applied to all of the applications, including those who partition only the MRAM part of the memory. After this change any application that wishes to be compatible (or updateable through DFU) must add an overlay, aligning memory regions. If missed, it will end up in a bus fault.

At this point, it might be easier to revert those changes than handle all of the support tickets, complaining about a failed NCS migration attempt.

The issue was raised by @MarekPieta and discussed with @maciejpietras, so if you want to discuss this change further, please schedule a meeting.

hubertmis:
I'm confused. Is there a requirement to be able to update a sample on nRF54H20-**DK** from NCS 2.9 to 3.0 with DFU? On the development kit, you can update with JLink to be compatible with the new memory map.

Not to mention that 2.9 -> 3.0 is a major version change, so I'm not sure if we intend to make it DFU-able between them. There are more breaking changes than memory maps for Development Kits.

maciejpietras:
> I'm confused. Is there a requirement to be able to update a sample on nRF54H20-**DK** from NCS 2.9 to 3.0 with DFU? On the development kit, you can update with JLink to be compatible with the new memory map.
> 
> Not to mention that 2.9 -> 3.0 is a major version change, so I'm not sure if we intend to make it DFU-able between them. There are more breaking changes than memory maps for Development Kits.

We don't want to make customer life's even harder therefore I proposed to revert this change. This change was mainly triggered by Matter which is not available on H20. Release NCS 3.0 will be the last one in this configuration and starting with NCS 3.1 I anticipate more significant changes therefore this will be good opportunity to modify memory maps then.

maciejpietras (approved):

hubertmis (changes requested):
> This change was mainly triggered by Matter which is not available on H20

This bug was found by Matter, but it applies to all applications. Especially those with bigger memory footprint in the Application CPU

hubertmis (approved):
Discussed offline. For nRF54H20 there will be broken NCS compatibility between NCS versions 3.0 and 3.1. So it's better to revert this change now and reintroduce between 3.0 and 3.1

nordicjm (changes requested):
upstream reverts should go upstream

tomchy:
> upstream reverts should go upstream

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/88121

nordicjm (approved):

PR #2727, opened 2025-04-08
[Backport v4.0.99-ncs1-branch] soc: nordic: nrf54h: Change PM_DEVICE_RUNTIME default

nordic-krch:
Manual backport of #2703


nordic-krch:
Note that commits from #2703 are already merged in upstream so I had to add use `fromtree` (commits in #2703 are `fromlist`).

nordic-piks (approved):

nordicjm (approved):

PR #2730, opened 2025-04-08
Revert "[nrf noup] boards: Add non-secure target for nrf54L15dk"

tomi-font:
Reverting this noup now that we have the 54L15 NS board target upstream, and pulling the needed commits from Zephyr..


nordicjm (approved):

frkv (approved):
LGTM

PR #2737, opened 2025-04-09
[Backport v3.0-branch] NCSDK-32650: Revert incompatible memory changes

tomchy:
Due to compatibility reasons, we should either revert those or provide a clear and tested way to migrate between the v2.9.0-nRF54H20-1 and v3.0.0.


nordic-segl (approved):

nordic-segl:
FYI - compliance fails (as if PR in the upstream was merged in mean time. Thus, this PR is no longer [from list] -> [from tree])

PR #2726, opened 2025-04-08
[nrf fromlist] boards: 54h20: Support IPC between SEC and APP

SebastianBoe:
Support IPC between SEC and APP. This is needed for PSA Crypto support.

This patch is expected to be removed within a couple of weeks by the ongoing work in NCSDK-32158.

Upstream PR #: 88285

Ref: NCSDK-NONE

PR #2752, opened 2025-04-11
[nrf fromlist] soc: add ironside boot report

hakonfam:
non-empty PR description


soc/nordic/nrf54h/ironside_boot_report/ironside_boot_report.c
```diff
@@ -0,0 +1,25 @@
...
+
+#define IRONSIDE_BOOT_REPORT_ADDR DT_REG_ADDR(DT_NODELABEL(cpuapp_ironside_boot_report))
+#define IRONSIDE_BOOT_REPORT_MAGIC CONFIG_SOC_NRF54H_IRONSIDE_BOOT_REPORT_MAGIC
+
+int ironside_boot_report_get(const struct ironside_boot_report **report)
+{
+	const struct ironside_boot_report *tmp_report =
+		(const struct ironside_boot_report *)IRONSIDE_BOOT_REPORT_ADDR;
+
+	if (tmp_report->magic != IRONSIDE_BOOT_REPORT_MAGIC) {
```
hakonfam, line 18: needs a NULL check as per docs

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp_iron.dts
```diff
@@ -22,15 +31,40 @@
...
+			reg = <0x2c000 DT_SIZE_K(736)>;
+		};
+
+		cpuppr_code_partition: partition@e4000 {
+			reg = <0xe4000 DT_SIZE_K(64)>;
+		};
+
+		cpuflpr_code_partition: partition@f4000 {
+			reg = <0xf4000 DT_SIZE_K(48)>;
+		};
```
SebastianBoe, line 43: FYI, I intend to introduce a "memory map" DTS file that both APP and RAD use soon.

This is so that the app can send the correct vector table address to IRONside SE.

We need to make sure we don't conflict with each other more than necessary somehow.

soc/nordic/nrf54h/ironside_boot_report/Kconfig
```diff
@@ -0,0 +1,16 @@
...
+	default y if SOC_NRF54H20_CPUAPP
+	depends on SOC_NRF54H20_IRON
+	help
+	  This option enables parsing of the Boot Report populated by Nordic IRONside SE.
+
+config SOC_NRF54H_IRONSIDE_BOOT_REPORT_MAGIC
+	hex
+	default 0xe176928d
+	help
+	  Constant used to check if an Nordic IRONside SE boot report has been written.
```
SebastianBoe, line 15: Why have a constant in the configuration system?

Wouldn't it be easier for a programmer to know that it is constant if it were a define in a header file?

soc/nordic/nrf54h/ironside_boot_report/include/nrf/ironside_boot_report.h
```diff
@@ -0,0 +1,81 @@
...
+	/** Firmware version of IRONside SE recovery firmware. 8bit MAJOR.MINOR.PATCH.TWEAK */
+	uint32_t ironside_recovery_version_int;
+	/** Human readable extraversion of IRONside SE */
+	char ironside_recovery_extraversion[12];
+	/** Copy of SICR.UROT.UPDATE.STATUS.*/
+	uint32_t ironside_update_status;
+	/** See @ref ironside_boot_report_uicr_error */
+	struct ironside_boot_report_uicr_error uicr_error_description;
+	/** Data passed from booting local domain to local domain being booted */
+	uint8_t local_domain_context[32];
```
SebastianBoe, line 63:
```suggestion
	uint8_t local_domain_context[IRONSIDE_SE_BOOT_REPORT_LOCAL_DOMAIN_CONTEXT_SIZE];
```

Would be nice to have a define so the cpuconf service can refer to this define as the max allowed message size.

soc/nordic/nrf54h/ironside_boot_report/include/nrf/ironside_boot_report.h
```diff
@@ -0,0 +1,81 @@
...
+			uint32_t peripheral;
+			/** The index of the PERIPHCONF entry that triggered the error */
+			int32_t index;
+		} periphconf;
+		/** ITS error */
+		struct {
+			uint32_t err;
+		} its;
+	} description;
+};
```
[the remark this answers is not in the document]
hakonfam: We can break API at any point until we lock it down, is there any gain from not exposing this now? Without this in place I think its harder to verify the functionality.

soc/nordic/nrf54h/ironside_boot_report/Kconfig
hakonfam: We have historically been doing it this way. I guess the idea is that these values are configurations in themselves, hence they should be in kconfig.

Also, the details of their value is of no interest of the developer. Hence it is better to not clutter the .h and .c files with this "useless" information.

Lastly it allows a description of the value itself in case that is important. For instance if it has some special properties (for instance hamming distance to other values, max number of 0's in its binary representation etc) that would otherwise clutter the .c and .h files.

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp_iron.dts
hakonfam: I suggest your change goes in first, then I adapt to that. I need this for the random numbers anyway.

soc/nordic/nrf54h/ironside_boot_report/include/nrf/ironside_boot_report.h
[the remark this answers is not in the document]
hakonfam: Ah, I misunderstood. I thought you were commenting on the whole boot report structure. I still think its worth having something here to simplify testing, but I can always remove it if you want to.

tests/boards/nrf/ironside_se_boot_report/testcase.yaml
```diff
@@ -0,0 +1,19 @@
...
+      type: multi_line
+      ordered: true
+      regex:
+        - "err: 0"
+        - "version: 0x([0-9a-fA-F]+)"
+        - "extraversion: ([0-9a-fA-F]+)"
+    platform_allow:
+      - nrf54h20dk/nrf54h20/cpuapp/iron
+    integration_platforms:
+      - nrf54h20dk/nrf54h20/cpuapp/iron
```
SebastianBoe, line 19: NB: I had to get rid of integration_platforms to make the doc build green.

I don't know if testcase.yaml will have the same problem though.

https://github.com/nrfconnect/sdk-nrf/pull/21770/files

soc/nordic/nrf54h/ironside_se/include/nrf/ironside_se_boot_report.h
```diff
@@ -0,0 +1,65 @@
...
+	/** Firmware version of IRONside SE recovery firmware. 8bit MAJOR.MINOR.PATCH.TWEAK */
+	uint32_t ironside_se_recovery_version_int;
+	/** Human readable extraversion of IRONside SE */
+	char ironside_se_recovery_extraversion[12];
+	/** Copy of SICR.UROT.UPDATE.STATUS.*/
+	uint32_t ironside_update_status;
+	/** See @ref ironside_se_boot_report_uicr_error */
+	struct ironside_se_boot_report_uicr_error uicr_error_description;
+	/** Data passed from booting local domain to local domain being booted */
+	uint8_t local_domain_context[IRONSIDE_SE_BOOT_REPORT_LOCAL_DOMAIN_CONTEXT_SIZE];
```
SebastianBoe, line 47: I see that it's optional for this to be populated.

So I assume we need a local_domain_context_len to communicate how many bytes were written?

(0 when cpusec boots cpuapp for instance)

SebastianBoe: Alternatively we could make this mandatory.

soc/nordic/nrf54h/ironside_se/include/nrf/ironside_se_boot_report.h
```diff
@@ -0,0 +1,65 @@
...
+			uint32_t rfu[4];
+		} rfu;
+	} description;
+};
+
+/** @brief IRONside boot report. */
+struct ironside_se_boot_report {
+	/** Magic value used to identify valid boot report */
+	uint32_t magic;
+	/** Firmware version of IRONside SE. 8bit MAJOR.MINOR.PATCH.TWEAK */
```
jonathannilsen, line 34: We don't have a TWEAK version, we have SEQNUM. So I'd suggest something like
```suggestion
	/** Firmware version of IRONside SE. 8bit MAJOR.MINOR.PATCH.SEQNUM */
```

soc/nordic/nrf54h/ironside_se/include/nrf/ironside_se_boot_report.h
```diff
@@ -0,0 +1,65 @@
...
+	uint32_t ironside_se_version_int;
+	/** Human readable extraversion of IRONside SE */
+	char ironside_se_extraversion[12];
+	/** Firmware version of IRONside SE recovery firmware. 8bit MAJOR.MINOR.PATCH.TWEAK */
```
jonathannilsen, line 38: same here:
```suggestion
	/** Firmware version of IRONside SE recovery firmware. 8bit MAJOR.MINOR.PATCH.SEQNUM */
```

soc/nordic/nrf54h/ironside_se/Kconfig
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2025 Nordic Semiconductor
+# SPDX-License-Identifier: Apache-2.0
+
+config SOC_NRF54H20_IRONSIDE_SE_BOOT_REPORT
+	bool "Nordic IRONside SE boot report"
+	default y if SOC_NRF54H20_CPUAPP
+	depends on SOC_NRF54H20_IRON
```
jonathannilsen, line 7: Right now we depend on building for cpuapp as well - I guess there is no other core that sets the IRON config until cpurad gets added, but should we add a depends on regardless for correctness? Better to see it fail at build time than run time if/when cpurad gets added

soc/nordic/nrf54h/ironside_se/CMakeLists.txt
jonathannilsen, line 1: did we decide on ironside_se vs. ironside/se for the directory layout?

SebastianBoe: Should be ironside/se IIRC

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp_iron.dts
```diff
@@ -6,9 +6,25 @@
...
 
+/* No radio support for IRON variant, don't allocate MRAM. */
+/delete-node/&cpurad_rx_partitions;
+
+/* All of ram0x goes to cpuapp, delete these nodes to re-define below. */
+/delete-node/&cpuapp_ram0x_region;
+/delete-node/&cpurad_ram0x_region;
+
+/* IPC mechanism for IRON variant differs, delete the old node. */
+/delete-node/&cpusec_cpuapp_ipc;
```
SebastianBoe, line 20: Doesn't this break PSA?

https://github.com/nrfconnect/sdk-zephyr/pull/2726/files

soc/nordic/nrf54h/ironside_se/Kconfig
```diff
@@ -0,0 +1,27 @@
...
+	hex
+	default 0x4d69546f
+	help
+	  Constant used to check if an Nordic IRONside SE boot report has been written.
+
+config SOC_NRF54H20_IRONSIDE_SE_BOOT_REPORT_LOCAL_DOMAIN_CONTEXT_SIZE
+	int
+	default 16
+	help
+	  Size of the local domain context in bytes
```
SebastianBoe, line 21: I'm concerned about the the security implications of some third-party code entering the SDK that maliciously provides an alternative Kconfig default without us noticing.

I feel a define would be more robust and readable.

SebastianBoe (approved):

soc/nordic/nrf54h/ironside_se/Kconfig
hakonfam: We use this for CPSEC, I will see if there is some other way to get around that.

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp_iron.dts
hakonfam: Ah yes, I need to fix this now.

soc/nordic/nrf54h/ironside_se/include/nrf/ironside_se_boot_report.h
hakonfam: IMO there is a contract between the image being booted and the image that boots in this situation. For instance, for sec -> app we know that there is no local domain context.

For app -> rad there would be a struct that would communicate within in whether or not there is data of interest within the local domain context.

tests/boards/nrf/ironside_se_boot_report/testcase.yaml
hakonfam: Thanks for the heads up, will remove this if CI starts acting up.

jonathannilsen (approved):

PR #2762, opened 2025-04-14
[nrf noup] dts: nrf21540-fem default tx-en-settle-time-us=26

ankuns:
The default value of the `tx-en-settle-time-us` property is set to 26 to avoid spurious emission issue.

Cherry-pick of commit that went to NSC 3.0 release with https://github.com/nrfconnect/sdk-zephyr/pull/2753
Requires MPSL: https://github.com/nrfconnect/sdk-nrfxlib/pull/1725

### DIFF
--- a/dts/bindings/net/wireless/nordic,nrf21540-fem.yaml
+++ b/dts/bindings/net/wireless/nordic,nrf21540-fem.yaml
@@ -95,7 +95,7 @@ properties:
         This must be present to support SPI control of the FEM.
   tx-en-settle-time-us:
     type: int
-    default: 11
+    default: 26
     description: |
         Settling time in microseconds from state PG to TX.
 


piotrkoziar (approved):

e-rk (approved):

dts/bindings/net/wireless/nordic,nrf21540-fem.yaml
```diff
@@ -95,7 +95,7 @@ properties:
         This must be present to support SPI control of the FEM.
   tx-en-settle-time-us:
     type: int
-    default: 11
+    default: 26
```
nordicjm, line 98: why can't this be submitted upstream and brought in as fromlist/fromtree?

nordicjm (changes requested):

dts/bindings/net/wireless/nordic,nrf21540-fem.yaml
ankuns: @nordicjm  As you see description of this property below
```
Settling time in microseconds from state PG to TX.
Default value is based on Table 6 of the nRF21540 Product
Specification (v1.0), and can be overridden for tuned
configurations.
```
This is exactly this case, tuned configuration for use in NCS with MPSL and protocol drivers we deliver in NCS. The value 11 is perfect from Product Specification point of view. But due the way protocol drivers and use the FEM API the value is increased by 15us to 26 to avoid spurious emission issue.

ankuns: @nordicjm is such justification enough?

carlescufi: Why not make this the default upstream though, is there a downside if we did that?

ankuns: @carlescufi  @nordicjm I can make the change upstream, but I deliberately chose to do it just in downstream sdk-zephyr, because the requirement to enable TX_EN pin for the nrf21540 earlier comes from the way the SDC, 802.15.4 Radio Driver and other protocols/samples work. The TX_EN pin must be enabled is such a moment that the FEM'a PA is ready when the RF power starts to rise (no such parameter in public and even internal product specifications for the nRF SoCs and no such parameter in nordic,nrf-radio compatible that I could add and pass the sum to the MPSL configuration API). All protocols use target time when the FEM should be ready as the RADIO's READY event which is too late (the RF power already is emitted in this moment). 

On the other hand in upstream Zephyr there is no MPSL, SDC, FEM support and the samples that use FEM.
So the justification applies only to downstream. In upstream, the change does not have justification at all. How would I justify it in upstream, should I write "the change is needed in some downstream fork?", that will not pass.

This change is very small, does the job, **fixes an issue important to customers at very low cost**.

ankuns: @carlescufi @nordicjm Is the above justification enough?

ankuns: > so why not fix it in boards? This is a default value, it can be overridden, if you need a different value, set it where it is needed
"don't repeat yourself" rule in action... No.
It is needed just here where I propose. There are for sure boards made by our customers in their own repos which I cannot access and fix for them. The fix I propose fix the issue for all of them (unless they have overridden the default but that's not the case).

ankuns:
Replaced by upstream modification approach and with [nrf fromlist] cherry-picked commit https://github.com/nrfconnect/sdk-zephyr/pull/2833

PR #2776, opened 2025-04-16
[nrf fromlist] soc: nordic: nrf54h: Fix s2ram

nordic-krch:
Cache was not enabled when s2ram did not completed which lead to system malfunction. Always power up cache when returning from s2ram function.

Upstream PR #: 88709


gchwier (approved):
Fixes problem with USB3CV tests with nrf_desktop app (GetDescriptor was not returned proper value)

nika-nordic (approved):

adamkondraciuk (approved):

kl-cruz (approved):

PR #2781, opened 2025-04-17
Enable comparator on nRF54L20

nordic-segl:
Extend comparator drivers with analogue input to GPIO pin mapping.
In comparator test, use other analogue inputs.

manifest-pr-skip


tests/boards/nrf/comp/src/test.c
```diff
@@ -15,20 +15,30 @@ static const struct gpio_dt_spec test_pin_2 = GPIO_DT_SPEC_GET(DT_PATH(zephyr_us
 
 
 struct comp_nrf_comp_se_config comp_se_config = {
+#if defined(CONFIG_BOARD_NRF54LM20APDK)
```
nordic-piks, line 18: There is no such board at upstream, maybe feature flag?

nordic-segl:
Implemented in a different way.

PR #2788, opened 2025-04-17
[nrf fromlist] soc: nordic: common: poweroff: Set VPR to remain in its reset state

jaz1-nordic:
CM33 must write MEMCONF.POWER[VPR].RET = 0 before entering System OFF. This bit drives the vprSavedCtx input to VPR. Forcing it low will disable the Hibernate wake feature in VPR, so that VPR will remain in its reset state when waking from OFF.

Upstream PR #:88760

### DIFF
--- a/soc/nordic/common/poweroff.c
+++ b/soc/nordic/common/poweroff.c
@@ -16,10 +16,16 @@
 #endif
 #if defined(CONFIG_SOC_SERIES_NRF54LX)
 #include <helpers/nrfx_reset_reason.h>
+#include <hal/nrf_memconf.h>
 #endif
 
 #if defined(CONFIG_HAS_NORDIC_RAM_CTRL)
 #include <helpers/nrfx_ram_ctrl.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


soc/nordic/common/poweroff.c
```diff
@@ -59,6 +65,9 @@ void z_sys_poweroff(void)
 #endif
 
 #if defined(CONFIG_SOC_SERIES_NRF54LX)
+	/* Set VPR to remain in its reset state when waking from OFF */
```
lstnl, line 68: to retain probably

lstnl, line 68: to retain probably not remain

jaz1-nordic: I leave it as is because that's how it's already merged upstream.

jaz1-nordic: I leave it as is because that's how it's already merged upstream.

lstnl (approved):

magp-nordic (approved):

PR #2792, opened 2025-04-18
Openthread separate L2 configs

ArekBalysNordic:
Please see commit messages.

**Do not review while it is in draft.**


ArekBalysNordic:
All of those commits will be added in the current SDK-Zephyr upmerge.

PR #2804, opened 2025-04-23
v3.0 rebase branch

bjarki-andreasen:
v3.0 rebase branch


rlubos (commented):
There are some leftover `nrf-squash!`/`fixup!` commits that should be cleaned up.
Also I see quite a few reverts in the rebased history - shouldn't they cancel out with the original commits during the rebase?

rlubos (approved):

rlubos (approved):

PR #2805, opened 2025-04-23
soc: nordic: nrf54h: Disable GPD for MCUBoot

nordic-mik7:
Disable GPD for MCUBoot build, as it cannot be reinitialized later in application (SDFW does not
support reinitialization).

Also, remove the GPD disabling from the mcumgr sample for nRF54H20 iron board app - it was the reinitialization that caused problems.

Upstream PR #: 88958

Signed-off-by: Michal Kozikowski <michal.kozikowski@nordicsemi.no>
(cherry picked from commit bece1c22bbf04d417839635e482ecc9e45bd5c15)

### DIFF
--- a/soc/nordic/nrf54h/gpd/Kconfig
+++ b/soc/nordic/nrf54h/gpd/Kconfig
@@ -6,6 +6,6 @@ config SOC_NRF54H20_GPD
 	imply NRFS
 	imply NRFS_GDPWR_SERVICE_ENABLED
 	select ONOFF
-	default y if SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD
+	default y if !MCUBOOT && (SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD)
 	help
 	  This option enables the Global Power Domain service.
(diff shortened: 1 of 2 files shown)


ahasztag (approved):

soc/nordic/nrf54h/gpd/Kconfig
```diff
@@ -6,6 +6,6 @@ config SOC_NRF54H20_GPD
 	imply NRFS
 	imply NRFS_GDPWR_SERVICE_ENABLED
 	select ONOFF
-	default y if SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD
+	default y if !MCUBOOT && (SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD)
```
tomchy, line 9: Isn't that enabled only in the MCUboot image?
If so, than the override above should rather stay...

nordic-mik7: Yes, it is enabled only for MCUboot image. It is not true that  "no IPC is enabled in SSDFW" but rather that  SDFW is not robust to reinitialization of GPD which happened because previously MCUBoot also used it (we needed to disable GPD in mcuboot OR app). Now, MCUboot will have GPD disabled by default so app can have it  enabled.

tomchy: Does it mean that GPD can be enabled on Iron boards?

PR #2808, opened 2025-04-23
ci: Adding yamllint aciton

DatGizmo:
* The intention is to catch yaml files that are not parsable
* [yamllint action with all the rules definition](https://github.com/nrfconnect/action-yamllint/blob/main/action.yml)

Signed-off-by: Sebastian Wezel <sebastian.wezel@nordicsemi.no>


.github/workflows/lint-yaml.yml
```diff
@@ -0,0 +1,11 @@
+name: Lint yaml files
+
+on:
+  pull_request:
+
+jobs:
+  lint_yaml:
+    runs-on: ubuntu-latest
```
thst-nordic, line 8:
```suggestion
    runs-on: ubuntu-24.04
```

thst-nordic (changes requested):

DatGizmo:
Not needed as ymllint is already done by the compliance check

PR #2810, opened 2025-04-23
LLD fixes, warning and handling

tejlmand:
Cherry-pick of: 88970

This PR fixes warnings and handling happening with lld.

First commit fixes the warning:
```
ld.lld: warning: <internal>:(.eh_frame) is being placed in '.eh_frame'
```
Zephyr doesn't need the `.eh_frame` for C code and even discards the section when converting the elf to bin / hex.
gcc/ld per default discards the section, so there is no reason we should keep it for lld.

Second commit fixes the warning:
```
ld.lld: warning: ignoring memory region assignment for non-allocatable section '.last_section'
```

This is not just a warning but also results in the `LONG(<val>)` command inside the linker script to be ignored by lld.
Result is that the last section id is missing in the final output.

To ensure correct and identical behavior between ld and lld, the last section id is now linked into the final image through C code.


bjarki-andreasen (approved):

bjarki-andreasen:
This PR requires a rebase since the upmerge :)

PR #2828, opened 2025-05-01
Cpuconf3

SebastianBoe:
see commit messages for details


jonathannilsen (approved):
In https://github.com/zephyrproject-rtos/zephyr/pull/88908 there is a commit that enables the cpusec_bellboard for cpuapp, should that be added here as well?

SebastianBoe:
Evidently it is not needed.

I suppose it could trickle down naturally.

I could also cherry-pick it after this PR, but I'd rather not making any unnecessary changes that could delay this PR.

karstenkoenig (approved):

PR #2817, opened 2025-04-24
Upmerge 2025.04.23

jfischer-no:
upstream-head-sha: 9a6f116a6aa9b70b517a420247cd8d33bbbbaaa3

Below are dropped noup/fromlist commits that cannot be applied cleanly and I cannot resolve them because I have no clue about this area or the files upstream are removed/changed. So whoever is responsible for these patches, please take a look at them and if necessary provide me with an updated version of the patches.

@mstasiaknordic 
  ~~`# [nrf noup] Revert "[nrf fromlist] boards: nordic: add FLPR core to nRF54L20pdk"`~~
  ~~c63d142fd807d7bc62c52f37abd94017ee3ae1da~~
  ~~`# [nrf noup] Revert "[nrf fromlist] modules: hal_nordic: nrfx: add nRF54L20 FLPR core"`~~
  ~~8a381aec5c7546759b55401ec45e2d0738d44e7c~~
  ~~`# [nrf noup] Revert "[nrf fromlist] modules: hal_nordic: Add nRF54L09 FLPR"`~~
  ~~b717237137eed32a7788a9dc3912fc899f08950f~~

@frkv 
  ~~`[nrf noup] modules: mbedtls: Disable configurations in Kconfig.tls-generic`~~
  ~~fe0996cbae631a2cb3ce6fa38e5140712a9c9c35~~

@tomchy 
  `[nrf fromlist] Revert "boards: nordic: Fix cpuapp_ram0x_region reg value` https://github.com/zephyrproject-rtos/zephyr/pull/88121/commits
  1bac58ab99a2cc1f2f69500f9db3bbb35b19c761

@krish2718 
  ~~`[nrf fromlist] nrf_wifi: Add nRF71 support` https://github.com/zephyrproject-rtos/zephyr/pull/88816~~
  ~~d8dae50843e60c81b606856270fa4da07d5c4fea~~
  ~~3c86e5ba5b8d25f6b6ffe8e1e1a1cd1343f1dd04~~

PR #2835, opened 2025-05-05
[nrf fromlist] manifest: Update hal_nordic with SPI workaround

nordic-krch:
Update hal_nordic with errata workaround in nrfx_spim.

Upstream PR #: 89489

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -193,7 +193,7 @@ manifest:
       groups:
         - hal
     - name: hal_nordic
-      revision: 119ff5b5ec7a413ca2f64ae1928c79eee1e7b7b2
+      revision: a5a2277d4399939e9aa67a71ca9eb411c85a1a7d
       path: modules/hal/nordic
       groups:
         - hal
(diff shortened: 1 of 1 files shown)


nika-nordic (approved):

mstasiaknordic (approved):

west.yml
```diff
@@ -193,7 +193,7 @@ manifest:
       groups:
         - hal
     - name: hal_nordic
-      revision: 119ff5b5ec7a413ca2f64ae1928c79eee1e7b7b2
+      revision: 37c38a471c042d890180cfb79cf788a32a006b80
```
jfischer-no, line 196: Upstream PR #: 89489 is outdated and will never be merged. Also why do we need two commit to changed the same version? Can you cherry-pick https://github.com/nrfconnect/sdk-zephyr/pull/2817/commits/d463ea1d10fb1d5c6282a59dcc8256a179fdbae5 from the upmerge PR?

jfischer-no (changes requested):

west.yml
nordic-krch: need 2 commits because last commit will not apply without conflicts. Cherry-picked newer commit.

PR #2852, opened 2025-05-13
[nrf noup] Added nrf54h20dk/nrf54h20/cpuapp/iron/mcuboot/b0

ahasztag:
This is one of the solutions for adding an appropriate memory map in case mcuboot and NSIB are to be used.

Note: the nrf54h20dk/nrf54h20/cpuapp/iron/mcuboot/b0 board was added in zephyr. This might not be the final approach. Other ways of specifying the memory map might be used.

### DIFF
--- a/samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml
+++ b/samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml
@@ -92,6 +92,7 @@ tests:
       - rd_rw612_bga
       - nrf52840dk/nrf52840
       - nrf54h20dk/nrf54h20/cpuapp/iron
+      - nrf54h20dk/nrf54h20/cpuapp/iron/mcuboot/b0
       - pinnacle_100_dvk
       - mg100
     integration_platforms:


nordicjm:
Though note that it would probably be better to have this in sdk-nrf as a board extension rather than in sdk-zephyr as a noup which is likely to have merge conflicts in future upmerges - see https://docs.zephyrproject.org/latest/hardware/porting/board_porting.html#board-extensions for details

ahasztag:
> Though note that it would probably be better to have this in sdk-nrf as a board extension rather than in sdk-zephyr as a noup which is likely to have merge conflicts in future upmerges - see https://docs.zephyrproject.org/latest/hardware/porting/board_porting.html#board-extensions for details

You are absolutely right, I did not know of a possibility of such board extension.
For now I am keeping it here until the decision is made how will we handle the static DTS, but I've added a DNM label; I'll probably move this to sdk-nrf

boards/nordic/nrf54h20dk/Kconfig.defconfig
```diff
@@ -27,3 +27,10 @@ config FLASH_LOAD_OFFSET
 	default 0x2c000 if !USE_DT_CODE_PARTITION
 
 endif # BOARD_NRF54H20DK_NRF54H20_CPUAPP_IRON
+
+if BOARD_NRF54H20DK_NRF54H20_CPUAPP_IRON_MCUBOOT_B0
```
tomchy, line 31: I do not have strong opinion, but which of those two you'd consider better and why?
``BOARD_NRF54H20DK_NRF54H20_CPUAPP_IRON_MCUBOOT_B0`` vs ``BOARD_NRF54H20DK_NRF54H20_CPUAPP_IRON_B0_MCUBOOT``

boards/nordic/nrf54h20dk/Kconfig.nrf54h20dk
```diff
@@ -3,12 +3,14 @@
...
 	select SOC_NRF54H20_CPURAD if (BOARD_NRF54H20DK_NRF54H20_CPURAD || \
 				       BOARD_NRF54H20DK_NRF54H20_CPURAD_IRON)
 	select SOC_NRF54H20_CPUPPR if (BOARD_NRF54H20DK_NRF54H20_CPUPPR || \
 				       BOARD_NRF54H20DK_NRF54H20_CPUPPR_XIP)
 	select SOC_NRF54H20_CPUFLPR if (BOARD_NRF54H20DK_NRF54H20_CPUFLPR || \
 				        BOARD_NRF54H20DK_NRF54H20_CPUFLPR_XIP)
 	select SOC_NRF54H20_IRON if (BOARD_NRF54H20DK_NRF54H20_CPUAPP_IRON || \
-				       BOARD_NRF54H20DK_NRF54H20_CPURAD_IRON)
+				       BOARD_NRF54H20DK_NRF54H20_CPURAD_IRON || \
+					   BOARD_NRF54H20DK_NRF54H20_CPUAPP_IRON_MCUBOOT_B0)
```
tomchy, line 16: I think we will need the CPURAD board soon too, but we may handle it in a separate PR.

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp_iron_mcuboot_b0_0_9_0.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2025 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+identifier: nrf54h20dk/nrf54h20/cpuapp/iron/b0
```
tomchy, line 4: Should not that include the `mcuboot` part too?

boards/nordic/nrf54h20dk/Kconfig.defconfig
ahasztag: I prefer `MCUBOOT_B0`, because I assume mcuboot is the more popular choice, thus it seems better to treat B0 as an addition to MCUBOOT. Also, it seems the main usage of B0 is allowing MCUBOOT upgrade

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp_iron_mcuboot_b0.dts
```diff
@@ -0,0 +1,51 @@
...
+	label = "provision";
+};
+
+boot_partition: &cpuapp_boot_partition {
+	label = "mcuboot";
+};
+
+boot_partition_slot1: &cpuapp_boot_slot1_partition {
+	label = "mcuboot-image-1";
+};
```
nordicjm, line 43: as per sdk-nrf PR, needs proper names

nordicjm (changes requested):

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp_iron_mcuboot_b0.dts
ahasztag: No longer applicable, moved board definition to sdk-nrf

samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml
```diff
@@ -92,6 +92,7 @@ tests:
       - rd_rw612_bga
       - nrf52840dk/nrf52840
       - nrf54h20dk/nrf54h20/cpuapp/iron
+      - nrf54h20dk/nrf54h20/cpuapp/iron/mcuboot/b0
```
nordicjm, line 95: move this to sdk-nrf as we do for other tests/samples - see https://github.com/nrfconnect/sdk-nrf/tree/main/samples/zephyr and https://github.com/nrfconnect/sdk-nrf/tree/main/tests/zephyr for examples

nordicjm (changes requested):

ahasztag:
Not needed anymore, everything has been moved to NCS

PR #2863, opened 2025-05-14
fromtree: Add hw-flow-control for UART

MarkusLassila:
Add Asynchronous UART implementation, which does not drop data
when automatic hardware-flow-control is set in the device tree.

With automatic hardware flow control, the CTS pin will be
automatically deactivated when there are no more asynchronous
UART RX buffers available. After buffer space becomes available,
and UART RX is restarted, the CTS pin will be activated.


MarkusLassila:
@bjarki-andreasen: SonarCloud Code Analysis fails on code duplication. Can this be merged despite that or does the upstream need to be reworked?

bjarki-andreasen:
> @bjarki-andreasen: SonarCloud Code Analysis fails on code duplication. Can this be merged despite that or does the upstream need to be reworked?

absolutely no :) code duplication is expected here, I will force merge it after approvals if sonarcloud indeed blocks it

bjarki-andreasen (approved):

PR #2865, opened 2025-05-14
NCSDK-33316: Add radio core support

tomchy:
Add necessary changes to provide a simple, updateable radio image.
    
Ref: NCSDK-33316


share/sysbuild/cmake/modules/sysbuild_extensions.cmake
```diff
@@ -204,19 +204,21 @@ endfunction()
...
-# APP_TYPE <MAIN|BOOTLOADER>: Application type.
+# APP_TYPE <MAIN|BOOTLOADER|UPDATEABLE>: Application type.
 #                             MAIN indicates this application is the main application
 #                             and where user defined settings should be passed on as-is
 #                             except for multi image build flags.
 #                             For example, -DCONF_FILES=<files> will be passed on to the
 #                             MAIN_APP unmodified.
 #                             BOOTLOADER indicates this app is a bootloader
+#                             UPDATEABLE indicates this app is updateable and the bootloader
+#                             settings should be passed on.
```
nordicjm, line 215: no that is not really what this config item is for, this is for an image type. Anyhow you can pass these settings on in sdk-nrf without touching sdk-zephyr, see https://github.com/nrfconnect/sdk-nrf/blob/main/sysbuild/CMakeLists.txt#L344

nordicjm (changes requested):

share/sysbuild/cmake/modules/sysbuild_extensions.cmake
tomchy: File moved: https://github.com/nrfconnect/sdk-nrf/pull/22401

ahasztag (approved):

robertstypa (approved):

PR #2866, opened 2025-05-14
Fromtree zephyr cellular modem fix out of coverage

MarkusLassila:
Previously, it was expected that the modem would immediately
disconnect and close the PPP link, when the network was lost
(+CEREG: 4 is received). Failure to do so, would result in
modem attempting to chat with dlci2 channel, where as cellular
modem driver would drop to dlci1. This would show as dial
script timing out, which would freeze cellular modem driver.

Adding dormant state enables graceful disconnection of PPP-link
when the network is lost. This allows the modems which do
not automatically close with network link loss to close
based on PPP link closure.


MarkusLassila:
Branch is a mess. Creating a new PR from this.

PR #2868, opened 2025-05-14
Bluetooth: Host: ACL/ISO buffers and bt_conn_tx objects mismatch

PavelVPV:
Upstream PR #: 89258

Task: NCSDK-31338


PavelVPV:
Will be downstreamed with upmerge

PR #2873, opened 2025-05-15
[nrf fromtree] boards: nordic: Enable nfct on the supporting boards

dchat-nordic:
Enable NFCT device in dts of the boards that have
NFC connector.

This is part of NCSDK-32456, since it didn't make it before upmerge.

(cherry picked from commit e37b68bfa4b62b7fc19c29eeec6521fd56e6cd30)

### DIFF
--- a/boards/nordic/nrf21540dk/nrf21540dk_nrf52840.dts
+++ b/boards/nordic/nrf21540dk/nrf21540dk_nrf52840.dts
@@ -158,6 +158,10 @@
 	gpio-as-nreset;
 };
 
+&nfct {
+	status = "okay";
+};
+
 &gpiote {
 	status = "okay";
 };
--- a/boards/nordic/nrf52833dk/nrf52833dk_nrf52833.dts
+++ b/boards/nordic/nrf52833dk/nrf52833dk_nrf52833.dts
@@ -135,6 +135,10 @@
 	gpio-as-nreset;
 };
 
+&nfct {
+	status = "okay";
+};
+
 &gpiote {
 	status = "okay";
 };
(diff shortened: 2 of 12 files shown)


kapi-no (approved):

grochu (approved):

pdunaj (changes requested):
I doubt this is a proper way of working with the boards. Enabling nodes enables code which have impact on binary size.
@dchat-nordic , can you confirm that on boards where the node was enabled there was no impact on the size of the resulting binaries for apps/samples that do not use NFC?

pdunaj:
Please check `boards/nordic/nrf54l15dk/nrf54l_05_10_15_cpuapp_common.dtsi` impact on nrf54l05 - you can look at some hid sample. If it does not negatively impact the binary size then the change is ok.

grochu:
> I doubt this is a proper way of working with the boards. Enabling nodes enables code which have impact on binary size. @dchat-nordic , can you confirm that on boards where the node was enabled there was no impact on the size of the resulting binaries for apps/samples that do not use NFC?

Currently, no SW component depends on the nfct node status (through dts scripts). The only exception is a Kconfig option which enables NFC in some BlueBagel application or sample (I don't remember which). Eventually, subsys/nfc/platform.c and related code should depend on this node, but this is not aligned yet. Even when the alignment is there, the binary size will ultimately depend on Kconfig option chosen e.g. if CONFIG_NFC_PLATFORM.

dchat-nordic:
@pdunaj Currently no code or binaries are attached with HAS_NRF_NFCT or similar config resulting from those changes. As @grochu said the NFC libraries and code is selected by Kconfigs CONFIG_NFC_T2T_NRFXLIB CONFIG_NFC_T4T_NRFXLIB and CONFIG_NFC_PLATFORM which are n by default AFAIK.

dchat-nordic:
@pdunaj I checked on nrf54l05 for hids_keyboard and there is no change in binary size.

PR #2875, opened 2025-05-15
[nrf fromlist] ci: Support standalone anchors for reusability

krish2718:
If we want to reuse a set of files for multiple tags, then allow defining a standalone anchor and skip processing it.

e.g., below will work now, as the first entry wifi_common will be ignored

wifi_common: *wifi_common
  - a
  - b

ci_samples_wifi:
  files:
    - c
    - *wifi_common

Upstream PR #: 90028

manifest-pr-skip


krish2718:
This doesn't help.

PR #2907, opened 2025-05-23
[nrf noup] drivers: spi: spi_nrfx_spi: remove multithreading dependency

nordic-mik7:
Replace semaphores with proper atomic flags when used without multithreading enabled.

Related: NCSDK-33699


nordic-mik7:
Work in progress, set as DNM - need to provide some tests config.

nordic-mik7:
@anangl @ahasztag  Please review and approve if you find time; upstream PR already reviewed and pending merge

anangl:
Please change the second commit to fromtree.

drivers/spi/spi_context.h
```diff
@@ -172,29 +185,57 @@ static inline int spi_context_wait_for_completion(struct spi_context *ctx)
...
+				key = irq_lock();
+			}
+
+			ctx->ready = 0;
+			irq_unlock(key);
+		} else {
+			const uint32_t tms = k_uptime_get_32();
+
+			while (!atomic_get(&ctx->ready) && (k_uptime_get_32() - tms < timeout_ms)) {
+				k_busy_wait(1);
```
ahasztag, line 228: Why `k_busy_wait(1)` is needed here? I'd rather say this should be an empty while loop, this does only add unnecessary code

ahasztag (approved):
Approving in advance, just one non-blocking suggestion

drivers/spi/spi_context.h
nordic-mik7: It is unnecessary indeed... Already merged to upstream though

anangl (approved):

PR #2921, opened 2025-05-28
[nrf fromlist] drivers: nrf_ironside dvfs service

lstnl:
Added handling of new IRONside DVFS service.
NRFS DVFS is now not enabled by default.

Upstream PR #: 90755


include/zephyr/drivers/firmware/nrf_ironside/ironside_dvfs.h
```diff
@@ -0,0 +1,59 @@
...
+ * @name IRONside DVFS service error codes.
+ * @{
+ */
+
+/** The requested DVFS oppoint is not allowed. */
+#define IRONSIDE_DVFS_ERROR_WRONG_OPPOINT (EINVAL)
+/** Waiting for mutex lock timed out, or hardware is busy. */
+#define IRONSIDE_DVFS_ERROR_BUSY (EBUSY)
+/** There is configuration error in the DVFS service. */
+#define IRONSIDE_DVFS_ERROR_OPPOINT_DATA (ECANCELED)
```
jonathannilsen: Please use dedicated error codes (starting from 1), we'd rather not depend on a specific errno.h definition.

include/zephyr/drivers/firmware/nrf_ironside/ironside_dvfs.h
jonathannilsen, line 1: nit: we already "namespace" this with `nrf_ironside` in the include path, do we need `ironside_` in the filename as well or could it just be `dvfs.h`? The other service header filenames are not prefixed with ironside_, so that would also match those.

include/zephyr/drivers/firmware/nrf_ironside/ironside_dvfs.h
```diff
@@ -0,0 +1,59 @@
...
+#include <stdint.h>
+#include <stddef.h>
+#include <errno.h>
+
+enum ironside_dvfs_oppoint {
+	IRONSIDE_DVFS_OPP_HIGH = 0,
+	IRONSIDE_DVFS_OPP_MEDLOW = 1,
+	IRONSIDE_DVFS_OPP_LOW = 2,
+	IRONSIDE_DVFS_OPP_COUNT
+};
```
jonathannilsen: nit: `IRONSIDE_OPP_COUNT` is not a valid `enum ironside_dvfs_oppoint`, so it may be better to make it a macro. Maybe it could even be defined internally in the C file as `ARRAY_SIZE(dvfs_hsfll_data)`

drivers/firmware/nrf_ironside/ironside_dvfs.c
```diff
@@ -0,0 +1,203 @@
...
+
+	if (!ironside_dvfs_opp_allowed(dvfs_oppoint)) {
+		return -EINVAL;
+	}
+
+	if (dvfs_oppoint == current_dvfs_oppoint) {
+		return 0;
+	}
+
+	if (!k_mutex_lock(&ironside_dvfs_mutex, K_MSEC(1000))) {
```
jonathannilsen: Should this timeout of 1000ms be configurable? The default here seems quite high, I'd imagine that it would be nice to be able to tweak it.

drivers/firmware/nrf_ironside/ironside_dvfs.c
jonathannilsen, line 1: nit: like the header name, maybe this could just be dvfs.c

drivers/firmware/nrf_ironside/ironside_dvfs.c
```diff
@@ -0,0 +1,203 @@
...
+		requested_dvfs_oppoint = dvfs_oppoint;
```
jonathannilsen: Is there a need for `requested_dvfs_oppoint`? as far as I can see it is only accessed in this function, so it seems like `dvfs_oppoint` could just be used directly.

drivers/firmware/nrf_ironside/ironside_dvfs.c
jonathannilsen: Maybe it would be better to have the check for `dvfs_oppoint == current_dvfs_oppoint` after the mutex lock (and make `current_dvfs_oppoint` no longer volatile)?

Then, if two threads simultaneously try to set e.g. oppoint = HIGH, the first would get the mutex and complete the transition, and the next would block until that is done, see that the oppoint is already HIGH, and exit. With the current code, it seems like the second thread would block but would then still go through with the IPC call unnecessarily once it gets the mutex.

drivers/firmware/nrf_ironside/ironside_dvfs.c
```diff
@@ -0,0 +1,203 @@
...
+ * @param enum oppoint target operation point
+ * @return 0 value indicates no error.
+ * @return -EINVAL invalid oppoint or domain.
+ */
+static int32_t ironside_dvfs_configure_hsfll(enum ironside_dvfs_oppoint oppoint)
+{
+	if (oppoint >= IRONSIDE_DVFS_OPP_COUNT) {
+		return -EINVAL;
+	}
+
```
jonathannilsen: I'd say this check is not really necessary as `ironside_dvfs_change_oppoint` already checks that oppoint is in range.

drivers/firmware/nrf_ironside/ironside_dvfs.c
```diff
@@ -0,0 +1,203 @@
...
+	hsfll_trim.vsup = NRF_FICR->TRIM.APPLICATION.HSFLL.TRIM.VSUP;
+	hsfll_trim.coarse = NRF_FICR->TRIM.APPLICATION.HSFLL.TRIM.COARSE[freq_trim_idx];
+	hsfll_trim.fine = NRF_FICR->TRIM.APPLICATION.HSFLL.TRIM.FINE[freq_trim_idx];
+#if NRF_HSFLL_HAS_TCOEF_TRIM
+	hsfll_trim.tcoef = NRF_FICR->TRIM.APPLICATION.HSFLL.TRIM.TCOEF;
+#endif
+#else
+	/* Currently only application core is supported */
+	return -EINVAL;
+#endif
```
jonathannilsen: Couldn't this just be a `#error`? Better to fail in compile time than have this return an unexpected error at runtime.

drivers/firmware/nrf_ironside/ironside_dvfs.c
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * @brief Configure hsfll depending on selected oppoint
+ *
+ * @param enum oppoint target operation point
+ * @return 0 value indicates no error.
+ * @return -EINVAL invalid oppoint or domain.
+ */
+static int32_t ironside_dvfs_configure_hsfll(enum ironside_dvfs_oppoint oppoint)
+{
```
jonathannilsen: If the two suggestions above are implemented then this could just be a `void` function, and so could the other two functions that call this.

include/zephyr/drivers/firmware/nrf_ironside/ironside_dvfs.h
lstnl: there is already dvfs.c file in /nrf/applications/nrf_desktop/src/modules/dvfs.c that is why I have added ironside. I don't know if this is real problem. I can change that to dvfs if you like.

drivers/firmware/nrf_ironside/ironside_dvfs.c
lstnl: right this was "old" concept. I will remove it

include/zephyr/drivers/firmware/nrf_ironside/ironside_dvfs.h
jonathannilsen: I don't think it should be a problem unless we add `include/zephyr/drivers/firmware/nrf_ironside` to the include path right? The other header files in that directory like `call.h` and `update.h` have pretty generic names that could also easily cause conflicts if we did that. So I think having it as just `dvfs.h` would be fine.

drivers/firmware/nrf_ironside/ironside_dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/**
+ * @brief Check if ABB analog part is locked.
+ *
+ * @param abb Pointer to ABB peripheral.
+ *
+ * @return true if ABB is locked, false otherwise.
+ */
+static inline bool sdfw_dvfs_is_abb_locked(NRF_ABB_Type *abb)
```
jonathannilsen: nit: probably don't need to mention 'sdfw' here
```suggestion
static inline bool dvfs_is_abb_locked(NRF_ABB_Type *abb)
```

drivers/firmware/nrf_ironside/ironside_dvfs.c
jonathannilsen: I still think it could be a minor but nice simplification to make `ironside_dvfs_configure_hsfll`, `ironside_dvfs_prepare_to_scale` and `ironside_dvfs_change_oppoint_complete` into `void` functions and remove the return value now that they return `0` on every path

karl-nordic (approved):

PR #2931, opened 2025-06-02
boards: nrf: nRF54L20: Enable DCDC converter

adamkondraciuk:
Set the vregmain node to use the DCDC converter by default.


nika-nordic (approved):

nordic-piks (changes requested):
This board is no longer used, I suggest to not perform any changes here.
No one is testing this (and using).

PR #2933, opened 2025-06-03
[nrf noup] cmake: nrf54l: iar: fix cmake linker generator for kmu

RobinKastberg:
This adds the required parts for nrf54l kmu in
cmake linker generator.


carlescufi:
@RobinKastberg if this is not applicable upstream, can you please prefix the commit message with `[nrf noup]`? More info here:

https://ncsdoc.z6.web.core.windows.net/latest/nrf/dev_model_and_contributions/code_base.html#oss-repositories-downstream-project-history

You can also `git log` to get an idea of how it works.

Vge0rge:
I tested this (with gcc only) and it seems to work ok. 

One small note, please add this line inside the commit message after the title:
nrf-squash! [nrf noup] soc: nrf54l: Add custom section for KMU

You can see how this commit does it here:
https://github.com/nrfconnect/sdk-zephyr/commit/8feaa71acf41265c163c09a871833bd1bc3bc439


Also a second note is that this logic will be removed in the figure and a pure dts solution will be used. Just sharing this here FYI. 

Add the commit text that I asked for and I will approve this.

RobinKastberg:
How about now?

carlescufi (approved):

PR #2938, opened 2025-06-05
ironside tdd service

karstenkoenig:
Added support for the IronSide TDD service which allows configuring and
powering the trace and debug domain of the nrf54h20.
Also provide option to start the trace and debug domain in the soc start
sequence.

### DIFF
--- a/boards/nordic/nrf54h20dk/support/nrf54h20_cpuapp.JLinkScript
+++ b/boards/nordic/nrf54h20dk/support/nrf54h20_cpuapp.JLinkScript
@@ -1,6 +1,175 @@
 __constant U32 _CPUCONF_ADDR = 0x52011000;
 __constant U32 _CPUCONF_CPUWAIT_OFFSET = 0x50C;
 
+// ATBFUNNEL
+__constant U32 _ATBFUNNEL211_ADDR = 0xBF04D000;
+__constant U32 _ATBFUNNEL212_ADDR = 0xBF04E000;
+__constant U32 _ATBFUNNEL_CTRLREG_OFFSET = 0x0;
+__constant U32 _HOLDTIME_4 = 0x300;
+__constant U32 _ENS0 = 0x1;
+__constant U32 _ENS1 = 0x2;
+__constant U32 _ENS2 = 0x4;
+
+// ATBREPLICATOR
+__constant U32 _ATBREPLICATOR212_ADDR = 0xBF04A000;
+__constant U32 _ATBREPLICATOR213_ADDR = 0xBF04B000;
+__constant U32 _ATBREPLICATOR_IDFILTER0_OFFSET = 0x0;
+__constant U32 _ATBREPLICATOR_IDFILTER1_OFFSET = 0x4;
+__constant U32 _ID_NONE = 0xFFFFFFFF;
+__constant U32 _ID1x = 0xFFFFFFFD;
+
+// TSGEN
+__constant U32 _TSGEN_ADDR = 0xBF041000;
+__constant U32 _TSGEN_CNTCR_OFFSET = 0x0;
+__constant U32 _TSGEN_CNTFID0_OFFSET = 0x20;
+// Clock rate = TDD Freq. / 8
+__constant U32 _TS_CLOCKRATE = 40000000;
+
+// CTI
+__constant U32 _CTI210_ADDR = 0xBF046000;
+__constant U32 _CTICONTROL_OFFSET = 0x0;
+__constant U32 _CTIOUTEN_OFFSET = 0xA0;
+__constant U32 _CTIGATE_OFFSET = 0x140;
+__constant U32 _TPIU_FLUSH_TRIG = 0x2;
+
+// TPIU
+__constant U32 _TPIU_ADDR = 0xBF043000;
+__constant U32 _CURRENTPORTSIZE_OFFSET = 0x4;
+__constant U32 _FFCR_OFFSET = 0x304;
+__constant U32 _FSCR_OFFSET = 0x308;
+__constant U32 _ENFCONT = 0x02;
+__constant U32 _FONFLIN = 0x10;
+__constant U32 _ENFTC = 0x1;
+__constant U32 _TPIU_SYNC_FRAME_COUNT = 0x8;
+__constant U32 _CURRENTPORTSIZE_4 = 0x8;
+
+// TDDCONF
+__constant U32 _TDDCONF_ADDR = 0xBF001000;
+__constant U32 _TRACEPORTSPEED_OFFSET = 0x408;
+__constant U32 _SPEED80MHZ = 0x0;
+
+// CoreSight general
+__constant U32 _CORESIGHT_CLAIMSET_OFFSET = 0xFA0;
+__constant U32 _CORESIGHT_CLAIMCLR_OFFSET = 0xFA4;
+__constant U32 _CORESIGHT_LAR_OFFSET = 0xFB0;
+__constant U32 _CORESIGHT_UNLOCK_KEY = 0xC5ACCE55;
+
+// GPIO P7
+__constant U32 _P7_ADDR = 0x5F938E00;
+__constant U32 _PIN_CNF3_OFFSET = 0x8C;
+__constant U32 _PIN_CNF4_OFFSET = 0x90;
+__constant U32 _PIN_CNF5_OFFSET = 0x94;
+__constant U32 _PIN_CNF6_OFFSET = 0x98;
+__constant U32 _PIN_CNF7_OFFSET = 0x9C;
+__constant U32 _PIN_CNF_TPIU_CLOCK_VALUE = 0x80000503;
+__constant U32 _PIN_CNF_TPIU_DATA_VALUE = 0x00000503;
+
+// Settings
+__constant U32 _DEBUGGER_CLAIM_MASK = 0x2;
+
+// Used to check if we have already set up tracing
+int _needCoresightSetup = 1;
+
+// Unlock a CoreSight peripheral
+void _CSUnlock(U32 addr)
+{
+	JLINK_MEM_WriteU32(addr + _CORESIGHT_LAR_OFFSET, _CORESIGHT_UNLOCK_KEY);
+}
+
+// Lock a CoreSight peripheral
+void _CSLock(U32 addr)
+{
+	JLINK_MEM_WriteU32(addr + _CORESIGHT_LAR_OFFSET, 0);
+}
+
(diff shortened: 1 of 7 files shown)


include/zephyr/drivers/firmware/nrf_ironside/tdd.h
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @brief Control the Trace and Debug Domain (TDD).
+ *
+ * @param config The configuration to be applied.
+ *
+ * @retval 0 on success.
+ * @retval -IRONSIDE_TDD_ERROR_EINVAL on invalid argument.
+ */
+int ironside_tdd(const enum ironside_tdd_config config);
```
jonathannilsen: maybe `ironside_tdd_configure`? Feels a bit like a verb is missing. Though I recognize that this fits with `ironside_cpuconf`.

boards/nordic/nrf54h20dk/support/nrf54h20_cpuapp.JLinkScript
```diff
@@ -1,6 +1,174 @@
...
+
+// Set claim bits in the CoreSight peripheral to indicate to the firmware that it
+// has been configured by the host debugger
+void _CSClaim(U32 addr)
+{
+	JLINK_MEM_WriteU32(addr + _CORESIGHT_CLAIMSET_OFFSET, _DEBUGGER_CLAIM_MASK);
+}
+
+// Set up CoreSight and other necessary configuration so to enable ETM -> TPIU tracing.
+int _SetupSecdomTrace(void)
```
jonathannilsen: nit: just "_SetupTrace" maybe? Or "_SetupETMTrace"?

boards/nordic/nrf54h20dk/support/nrf54h20_cpuapp.JLinkScript
```diff
@@ -1,6 +1,174 @@
...
+
+// TDDCONF
+__constant U32 _TDDCONF_ADDR = 0xBF001000;
+__constant U32 _TRACEPORTSPEED_OFFSET = 0x408;
+__constant U32 _SPEED80MHZ = 0x0;
+
+// SPU
+__constant U32 _SPU131_ADDR = 0x5F920000;
+__constant U32 _FEATURE_GPIO_P7_OFFSET = 0xb80;
+__constant U32 _ATTR_SECDOM_SECURE = 0x10010;
```
jonathannilsen: Unused
```suggestion
```

boards/nordic/nrf54h20dk/support/nrf54h20_cpuapp.JLinkScript
```diff
@@ -1,6 +1,174 @@
...
+
+// CoreSight general
+__constant U32 _CORESIGHT_CLAIMSET_OFFSET = 0xFA0;
+__constant U32 _CORESIGHT_CLAIMCLR_OFFSET = 0xFA4;
```
jonathannilsen, line 53: This usage of the CLAIM register is not standardized, but instead a bespoke solution that we made. Should we leave CLAIM configuration out here? It could always be added later if necessary.

include/zephyr/drivers/firmware/nrf_ironside/tdd.h
```diff
@@ -0,0 +1,42 @@
...
+
+enum ironside_tdd_config {
+	/** Don't touch the configuration */
+	IRONSIDE_TDD_CONFIG_NO_CHANGE = 0, /* This allows future extensions to the TDD service */
+	/** Turn off the TDD */
+	IRONSIDE_TDD_CONFIG_OFF = 1, 
+	/** Turn on the TDD with default configuration */
+	IRONSIDE_TDD_CONFIG_ON_DEFAULT = 2,
+	/** Turn on the TDD with default configuration, but skip TPIU pin configuration */
+	IRONSIDE_TDD_CONFIG_ON_DEFAULT_NO_PINS = 3,
```
jonathannilsen: As pointed out elsewhere, I think it might be best to drop the pins/no pins configurations here

karstenkoenig: dropped

include/zephyr/drivers/firmware/nrf_ironside/tdd.h
karstenkoenig: yea tried aligning with ironside_cpuconf but yea it's an even worse naming scheme for ironside_tdd, changed it to ironside_tdd_configure()

boards/nordic/nrf54h20dk/support/nrf54h20_cpuapp.JLinkScript
karstenkoenig: Given this is in a nordic specific file I feel this is ok, we aren't providing a general jlink script to configure coresight peripherals here

include/zephyr/drivers/firmware/nrf_ironside/tdd.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef ZEPHYR_INCLUDE_ZEPHYR_DRIVERS_FIRMWARE_NRF_IRONSIDE_TDD_H_
+#define ZEPHYR_INCLUDE_ZEPHYR_DRIVERS_FIRMWARE_NRF_IRONSIDE_TDD_H_
+
+#include <zephyr/drivers/firmware/nrf_ironside/call.h>
+
+#include <nrfx.h>
+
+#define IRONSIDE_SE_TDD_SERVICE_ERROR_INVALID_CONFIG (1)
+
+#define IRONSIDE_SE_CALL_ID_TDD_V0 4
```
57300, line 15:
```suggestion
#define IRONSIDE_SE_CALL_ID_TDD_V0 3
```
DVFS service ID should be bumped to 4 if we need to support the TDD service first.

soc/nordic/nrf54h/Kconfig
```diff
@@ -72,11 +71,23 @@ config SOC_NRF54H20_CPURAD_ENABLE
 	default y if NRF_802154_SER_HOST
 	depends on SOC_NRF54H20_CPUAPP
 	select NRF_IRONSIDE_CPUCONF_SERVICE
+	select SOC_LATE_INIT_HOOK
 	help
 	  This will at application boot time enable clock to the
 	  Radiocore, and also power will be requested to the Radiocore
 	  subsystem. The Radiocore will then start executing instructions.
 
+config SOC_NRF54H20_ENABLE_TDD
```
57300: nit:
```suggestion
config SOC_NRF54H20_TDD_ENABLE
```

soc/nordic/nrf54h/soc.c
```diff
@@ -161,10 +158,19 @@ void soc_early_init_hook(void)
...
-#if defined(CONFIG_SOC_NRF54H20_CPURAD_ENABLE)
+#if defined(CONFIG_SOC_LATE_INIT_HOOK)
+
 void soc_late_init_hook(void)
 {
-	int err;
+#if defined(CONFIG_SOC_NRF54H20_ENABLE_TDD)
+	int err_tdd;
+
+	err_tdd = ironside_se_tdd_configure(IRONSIDE_SE_TDD_CONFIG_ON_DEFAULT);
```
57300, line 180: Should this also be invoked by PM state changes, if I'm not misunderstanding?

include/zephyr/drivers/firmware/nrf_ironside/tdd.h
karstenkoenig: DVFS got merged already here with 3
https://github.com/nrfconnect/sdk-zephyr/pull/2921/files#diff-1826e7b5ea3ea650c3d60ec4b8ecff5d10b1bb94ea07f0aec3ee2030648ee028R57

57300: But it's not enabled in IronSide yet, so we're free to change it.

soc/nordic/nrf54h/soc.c
karstenkoenig: In a final solution it can be, right now this will just force on tdd and leave it at that, hence preventing having to reconfigure all the coresight peripherals after a sleep etc, because JLink couldn't handle this.
This can be done when we have the coresight peripherals controlled from the application core but that'll need more time while this allows ETM over TPIU as a static state and is useful like this as is, just not power efficient

include/zephyr/drivers/firmware/nrf_ironside/tdd.h
karstenkoenig: Is there a benefit to it? I'd have to renumber it in this PR and also change the pending PR on the ironside call server side.

soc/nordic/nrf54h/Kconfig
karstenkoenig: Addressed this

include/zephyr/drivers/firmware/nrf_ironside/tdd.h
karstenkoenig: I feel the negatives outside the positives, as far as I can tell the only thing that'd happen is that we consider an id deprecated and would later suddenly use it again, but is that a problem? It only distinguishes itself in the failure code in the ironside call right?

57300: No problem then, we can leave it as is.

soc/nordic/nrf54h/Kconfig
57300: This was undone, but it's fine

57300 (approved):

soc/nordic/nrf54h/Kconfig
karstenkoenig: yea noticed as well and pushed a fix, was an outdated local copy

karstenkoenig:
@57300 I dropped the JLink change as we can pull this into a designated commit and PR to enable ETM

hakonfam (approved):

karstenkoenig:
@nordicjm upstream CI is green after cherry-picking a fix from a different PR, so added this here as well. Can you please check again?
sdk-nrf CI is red but that's because it still uses the outdated `nrf54h20dk/nrf54h20/cpuapp/iron`, this won't be merged before this is green anyways but it'd be nice to clear your change for request here before then.

You will notice a [noup] commit, sadly this can't be done upstream yet as it can go on relies on two different upstream PRs and hence doesn't have a place upstream, we can only do this after those get merged after the feature PRs are allowed again. I opened NCSDK-34289 against myself to follow up on this commit

nordicjm (approved):

PR #2942, opened 2025-06-09
[nrf fromlist] soc: nordic: common: Fix HAS_HW_NRF_NFCT condition

nordic-krch:
Include nordic,nrf-nfct-v2 compatible in the option.

Upstream PR #: 91274


kapi-no:
Link to the upstream PR for convenience:

https://github.com/zephyrproject-rtos/zephyr/pull/91274

PR #2952, opened 2025-06-13
[nrf fromlist] drivers: serial: nrfx_uarte: Reorder suspend

karstenkoenig:
To get a glitchfree UARTE suspend sequence we need to
* load the GPIO configuration for the suspend state
* disable the UARTE so the GPIO configuration is applied
* (optional) enable the GPIO retention as state should stick around

The previous sequence (optionally) locked the analog state of the GPIO so the configuration wouldn't actually get applied. Next it disabled the UART, falling back to the existing GPIO state. Next it loaded the new GPIO state, which could be different from the old GPIO state, leading to a potential transition from TX high to low to high.

Upstream PR #: 91550

### DIFF
--- a/drivers/serial/uart_nrfx_uarte.c
+++ b/drivers/serial/uart_nrfx_uarte.c
@@ -2338,13 +2338,13 @@ static void uarte_pm_suspend(const struct device *dev)
 		wait_for_tx_stopped(dev);
 	}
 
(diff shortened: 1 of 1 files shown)


kl-cruz (approved):

karstenkoenig:
Handled by gpd rework

PR #2963, opened 2025-06-17
Bluetooth: Host: Bonding with the same Central using multiple identities

PavelVPV:
This PR adds a feature to allow bonding with the same Central using multiple identities.

The PR consists of 2 commits.

The first commit adds a new Kconfig option by enabling which Host will keep
bonding with the same Central instead of rejecting pairing.

Brief implementation details:

This implementation adds a new flag to bt_keys struct:
BT_KEYS_ID_CONFLICT. The flag is set, when:
- bonding with the same peer and conflict identified
- when loading conflicting keys from persistent storage.

When bonding and conflict is identified, the new keys aren't added to
the Resolving List immediately. Instead, the old keys stay in the
Resolving List.

When start advertising, Host finds conflicting keys that are already
added to the Resolving List and substitues them.

If, however, there is another advertiser already started for the added
keys, the new request is reject and advertising start function returns
-EPERM.

The second adds the following tests:
- a test that checks undirected advertisements
- a test that checks directed advertisements
- a test that checks undirected advertisement with the Filter Accept List
- a test that checks unpairing
- a test that checks start of multiple advertisements in parallel
- a test that checks recovering BT_KEYS_ID_CONFLICT after reboot

### DIFF
--- a/subsys/bluetooth/host/Kconfig
+++ b/subsys/bluetooth/host/Kconfig
@@ -630,6 +630,24 @@ config BT_ID_UNPAIR_MATCHING_BONDS
 	  link-layer. The Host does not have control over this acknowledgment,
 	  and the order of distribution is fixed by the specification.
 
+config BT_ID_AUTO_SWAP_MATCHING_BONDS
+	bool "Automatically swap conflicting entries in the Resolving List"
+	depends on !BT_ID_UNPAIR_MATCHING_BONDS
+	depends on BT_PRIVACY && BT_PERIPHERAL && !BT_CENTRAL
+	help
+	  If this option is enabled, the Host will not add a new bond with
+	  the same peer address (or IRK) to the Resolving List if there is
+	  already a bond with the same peer address (or IRK) on another local
+	  identity.
+
+	  In case of Peripheral, the Host will swap the existing entry in the
+	  Resolving List with the new one, so that the new bond will be used for
+	  address resolution for the new local identity if the device starts
+	  advertising with the new local identity.
+
+	  Important: this option is supported exclusively in the Peripheral
+	  role. Excluding the Central role.
+
 config BT_ID_ALLOW_UNAUTH_OVERWRITE
 	bool "Allow unauthenticated pairing with same peer with other local identity"
 	depends on !BT_SMP_ALLOW_UNAUTH_OVERWRITE
--- a/subsys/bluetooth/host/adv.c
+++ b/subsys/bluetooth/host/adv.c
@@ -272,6 +272,25 @@ struct bt_le_ext_adv *bt_hci_adv_lookup_handle(uint8_t handle)
 #endif /* CONFIG_BT_BROADCASTER */
 #endif /* defined(CONFIG_BT_EXT_ADV) */
 
+struct bt_le_ext_adv *bt_adv_lookup_by_id(uint8_t id)
+{
+#if defined(CONFIG_BT_EXT_ADV)
+	for (size_t i = 0; i < ARRAY_SIZE(adv_pool); i++) {
+		if (atomic_test_bit(adv_pool[i].flags, BT_ADV_CREATED) &&
+		    adv_pool[i].id == id) {
+			return &adv_pool[i];
+		}
+	}
+#else
+	if (atomic_test_bit(bt_dev.adv.flags, BT_ADV_CREATED) && bt_dev.adv.id == id) {
+		return &bt_dev.adv;
+	}
+#endif
+
+	return NULL;
+}
+
+
 void bt_le_ext_adv_foreach(void (*func)(struct bt_le_ext_adv *adv, void *data),
 			   void *data)
 {
@@ -1021,6 +1040,14 @@ int bt_le_adv_start_legacy(struct bt_le_ext_adv *adv,
 	adv->id = param->id;
 	bt_dev.adv_conn_id = adv->id;
 
+	if (IS_ENABLED(CONFIG_BT_ID_AUTO_SWAP_MATCHING_BONDS)) {
+		err = bt_id_resolving_list_check_and_update(adv->id, param->peer);
+		if (err) {
+			LOG_ERR("Failed to check and update resolving list: %d", err);
+			return err;
+		}
+	}
+
 	err = bt_id_set_adv_own_addr(adv, param->options, dir_adv,
 				     &set_param.own_addr_type);
 	if (err) {
@@ -1336,6 +1363,15 @@ int bt_le_adv_start_ext(struct bt_le_ext_adv *adv,
 	}
 
 	adv->id = param->id;
+
+	if (IS_ENABLED(CONFIG_BT_ID_AUTO_SWAP_MATCHING_BONDS)) {
+		err = bt_id_resolving_list_check_and_update(adv->id, param->peer);
+		if (err) {
+			LOG_ERR("Failed to check and update resolving list: %d", err);
+			return err;
+		}
+	}
+
 	err = le_ext_adv_param_set(adv, param, sd != NULL);
 	if (err) {
 		return err;
@@ -1689,6 +1725,22 @@ int bt_le_ext_adv_start(struct bt_le_ext_adv *adv,
 
 	if (atomic_test_bit(adv->flags, BT_ADV_ENABLED)) {
 		return -EALREADY;
+	}
+
+	if (IS_ENABLED(CONFIG_BT_ID_AUTO_SWAP_MATCHING_BONDS)) {
+		const bt_addr_le_t *peer;
+
+		if (bt_addr_le_eq(&adv->target_addr, BT_ADDR_LE_ANY)) {
+			peer = NULL;
+		} else {
+			peer = &adv->target_addr;
+		}
+
+		err = bt_id_resolving_list_check_and_update(adv->id, peer);
+		if (err) {
+			LOG_ERR("Failed to check and update resolving list: %d", err);
+			return err;
+		}
 	}
 
 	if (IS_ENABLED(CONFIG_BT_PERIPHERAL) &&
--- a/subsys/bluetooth/host/id.c
+++ b/subsys/bluetooth/host/id.c
@@ -944,9 +944,33 @@ void bt_id_pending_keys_update(void)
 	}
 }
 
+static bool keys_conflict_check(const struct bt_keys *candidate, const struct bt_keys *resident)
+{
+	bool addr_conflict;
+	bool irk_conflict;
+
+	addr_conflict = bt_addr_le_eq(&candidate->addr, &resident->addr);
+
+	/* All-zero IRK is "no IRK", and does not conflict with other Zero-IRKs. */
+	irk_conflict = (!bt_irk_eq(&candidate->irk, &(struct bt_irk){}) &&
+			bt_irk_eq(&candidate->irk, &resident->irk));
+
+	if (addr_conflict || irk_conflict) {
+		LOG_DBG("Resident : addr %s and IRK %s", bt_addr_le_str(&resident->addr),
+			bt_hex(resident->irk.val, sizeof(resident->irk.val)));
+		LOG_DBG("Candidate: addr %s and IRK %s", bt_addr_le_str(&candidate->addr),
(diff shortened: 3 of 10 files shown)


subsys/bluetooth/host/Kconfig
```diff
@@ -631,6 +631,24 @@ config BT_ID_UNPAIR_MATCHING_BONDS
 	  link-layer. The Host does not have control over this acknowledgment,
 	  and the order of distribution is fixed by the specification.
 
+config BT_ID_AUTO_SWAP_MATCHING_BONDS
+	bool "Automatically swap conflicting entries in the Resolving List"
+	depends on !BT_ID_UNPAIR_MATCHING_BONDS
+	depends on BT_PRIVACY && BT_PERIPHERAL && !BT_CENTRAL
+	select EXPERIMENTAL
```
PavelVPV: Not sure if I need to keep this option as experimental.

alxelax:
Just curious how are implemented bsim tests going to be checked if CI bsim in downstream isn't run?

PavelVPV:
> Just curious how are implemented bsim tests going to be checked if CI bsim in downstream isn't run?

Manually

PavelVPV:
Added persistent storage test to verify that re-connection works fine after recovering bonding data from flash.

subsys/bluetooth/host/id.c
```diff
@@ -944,9 +944,33 @@ void bt_id_pending_keys_update(void)
...
+		return true;
+	}
+
+	return false;
+}
+
 struct bt_id_conflict {
 	struct bt_keys *candidate;
 	struct bt_keys *found;
+	bool all;
```
KyraLengfeld: Only one nitpick: `all` seems to me a bit ambiguous. Perhaps better to call it `check_all_irk` or similar? Feel free to disregard. Otherwise to me this PR looks good, although with the caveat that I am still lacking quite some knowledge of the host, nor did I manually run the bsim tests you added to check.

subsys/bluetooth/host/adv.c
```diff
@@ -272,6 +272,25 @@ struct bt_le_ext_adv *bt_hci_adv_lookup_handle(uint8_t handle)
...
+	for (size_t i = 0; i < ARRAY_SIZE(adv_pool); i++) {
+		if (atomic_test_bit(adv_pool[i].flags, BT_ADV_CREATED) &&
+		    adv_pool[i].id == id) {
+			return &adv_pool[i];
+		}
+	}
+#else
+	if (atomic_test_bit(bt_dev.adv.flags, BT_ADV_CREATED)) {
+		return &bt_dev.adv;
+	}
```
alwa-nordic: Looks like this is missing `&& bt_dev.adv.id == id` to be symmetric with the other branch of the if-def. If I'm wrong, can you add an assert or code comment to explain why `id` is not used when `!CONFIG_BT_EXT_ADV`?

subsys/bluetooth/host/Kconfig
```diff
@@ -630,6 +630,24 @@ config BT_ID_UNPAIR_MATCHING_BONDS
 	  link-layer. The Host does not have control over this acknowledgment,
 	  and the order of distribution is fixed by the specification.
 
+config BT_ID_AUTO_SWAP_MATCHING_BONDS
+	bool "Automatically swap conflicting entries in the Resolving List"
+	depends on !BT_ID_UNPAIR_MATCHING_BONDS
```
omkar3141, line 635: I think "!BT_CENTRAL" is unnecessary. You may have central which is dealing with other non-private connections with other peripherals, and that would work without any issues.

PavelVPV: I don't even want to think what will happen if `BT_CENTRAL` is enabled. The current use case is for Peripheral role only. Therefore it is easier to prevent compiling with `BT_CENTRAL`. If a customer comes with request to support Central role as well, we will look at this again.

PavelVPV:
Addressed comments. Also found bug when removing last conflicting key during unpairing.

PavelVPV:
Test commit is dropped. Tests are moved to the internal test framework.

subsys/bluetooth/host/Kconfig
Balaklaka: I think experimental should be removed as this goes into production by customer.

subsys/bluetooth/host/Kconfig
```diff
@@ -630,6 +630,24 @@ config BT_ID_UNPAIR_MATCHING_BONDS
...
+	  the same peer address (or IRK) to the Resolving List if there is
+	  already a bond with the same peer address (or IRK) on another local
+	  identity.
+
+	  In case of Peripheral, the Host will swap the existing entry in the
+	  Resolving List with the new one, so that the new bond will be used for
+	  address resolution for the new local identity if the device starts
+	  advertising with the new local identity.
+
+	  Important: this option is only supported by Peripheral role.
```
Balaklaka:
```suggestion
	  Important: this option is supported exclusively in the Peripheral role. Excluding Central role.
```

subsys/bluetooth/host/id.c
```diff
@@ -1254,6 +1344,151 @@ void bt_id_del(struct bt_keys *keys)
...
+ *
+ * If for any of the keys belonging to the given identity a conflict is found and the advertiser for
+ * that key is enabled, the function returns an error.
+ *
+ * @param id The identity ID to check and update.
+ * @param peer The peer address to check against the resolving list.
+ *
+ * @return 0 on success, or a negative error code on failure.
+ * @return -EPERM if a conflict is found and the advertiser for the conflicting key is enabled.
+ */
```
Balaklaka: This should be added to the h-file.

subsys/bluetooth/host/id.c
```diff
@@ -1254,6 +1344,151 @@ void bt_id_del(struct bt_keys *keys)
...
+int bt_id_resolving_list_check_and_update(uint8_t id, const bt_addr_le_t *peer)
+{
+	int err;
+
+	if (!IS_ENABLED(CONFIG_BT_ID_AUTO_SWAP_MATCHING_BONDS)) {
+		return 0;
```
Balaklaka: Should this return error code canceled. Not that it matters, as every call is within option being true. But error code will be good if someone call it by mistake.

subsys/bluetooth/host/id.c
```diff
@@ -1254,6 +1344,151 @@ void bt_id_del(struct bt_keys *keys)
...
+		if (!keys) {
+			LOG_DBG("No keys found for id %u addr %s", id, bt_addr_le_str(peer));
+			return -ENOENT;
+		}
+
+		err = conflict_check_and_replace(id, keys);
+	}
+
+	if (err) {
+		LOG_WRN("Failed to update resolving list for id %u addr %s (err %d)",
```
Balaklaka: LOG_ERR?

subsys/bluetooth/host/keys.c
```diff
@@ -310,16 +310,55 @@ void bt_keys_add_type(struct bt_keys *keys, enum bt_keys_type type)
 	keys->keys |= type;
 }
 
+static void add_id_cb(struct k_work *work)
+{
+	bt_id_pending_keys_update();
+}
+
+static K_WORK_DEFINE(add_id_work, add_id_cb);
```
Balaklaka, line 318:
```suggestion
static K_WORK_DEFINE(add_id_work, add_id_cb);

```

subsys/bluetooth/host/id.c
```diff
@@ -1254,6 +1344,151 @@ void bt_id_del(struct bt_keys *keys)
...
+{
+	/* For the given key check if it has conflicts with other keys in the Resolving List
+	 * (such keys have BT_KEYS_ID_ADDED state and BT_KEYS_ID_CONFLICT flag set). If it does, we
+	 * need to remove the conflicting key from the Resolving List and add the new key.
+	 *
+	 * If the key is not in the Resolving List, we can add the new key right away.
+	 *
+	 * If advertiser for the conflicting key is enabled, we cannot remove the key from the
+	 * Resolving List, so we return an error.
+	 */
```
Balaklaka, line 1339: Should this be added as a function doxygen comment, so it is shown in documentation. Even if it is a static function.

subsys/bluetooth/host/id.c
```diff
@@ -1254,6 +1344,151 @@ void bt_id_del(struct bt_keys *keys)
...
+
+	conflict = bt_id_find_conflict(keys, false);
+	if (conflict == NULL) {
+		/* bt_id_find_conflict returns only keys added to the Resolving List (state is
+		 * BT_KEYS_ID_ADDED). If the key has conflict, but no keys were added (for example,
+		 * if the last added key was removed after bt_unpair()), then this function will
+		 * return NULL.
+		 */
+		LOG_WRN("Conflicting key not in the Resolving List.");
+		goto add;
```
Balaklaka: If we flip to (conflict != NULL), we can remove the goto add; (should try to limit use of goto)

subsys/bluetooth/host/smp.c
```diff
@@ -4048,16 +4048,24 @@ static uint8_t smp_id_add_replace(struct bt_smp *smp, struct bt_keys *new_bond)
...
-	conflict = bt_id_find_conflict(new_bond);
+	conflict = bt_id_find_conflict(new_bond, IS_ENABLED(CONFIG_BT_ID_AUTO_SWAP_MATCHING_BONDS));
 	if (conflict) {
 		LOG_DBG("New bond conflicts with a bond on id %d.", conflict->id);
 	}
 
-	if (conflict && !IS_ENABLED(CONFIG_BT_ID_UNPAIR_MATCHING_BONDS)) {
+	if (conflict && !IS_ENABLED(CONFIG_BT_ID_AUTO_SWAP_MATCHING_BONDS) &&
+	    !IS_ENABLED(CONFIG_BT_ID_UNPAIR_MATCHING_BONDS)) {
 		LOG_WRN("Refusing new pairing. The old bond must be unpaired first.");
```
Balaklaka, line 4058: LOG_ERR?

subsys/bluetooth/host/id.c
```diff
@@ -972,37 +994,105 @@ void find_rl_conflict(struct bt_keys *resident, void *user_data)
...
+ * This function iterates over all keys (added and not added to the Resolving List). If there are
+ * multiple conflicts, this function will return true. Otherwise, it will return false.
+ *
+ * If @c firt_conflict is not NULL, it will be set to the first found conflict.
+ *
+ * @param candidate The candidate key to check for conflicts.
+ * @param first_conflict Pointer to store the first found conflict, if any. Can be NULL.
+ *
+ * @return True if there are multiple conflicts, otherwise it returns false.
+ */
```
Balaklaka: Move description to h-file?

subsys/bluetooth/host/id.c
```diff
@@ -972,37 +994,105 @@ void find_rl_conflict(struct bt_keys *resident, void *user_data)
...
 }
 
-struct bt_keys *bt_id_find_conflict(struct bt_keys *candidate)
+/** * @brief Find a conflict in the resolving list for a candidate IRK.
+ *
+ * @param candidate The candidate keys to check for conflicts.
+ * @param all If true, check all IRKs, otherwise check only added keys.
+ *
+ * @return The conflicting key if there is one, or NULL if no conflict was found.
+ */
```
Balaklaka: Move comment to h-file?

subsys/bluetooth/host/id.c
PavelVPV: I'll keep it here, but will also add the description of the new error code to `bt_le_adv_start` and `bt_le_ext_adv_start` functions.

subsys/bluetooth/host/smp.c
PavelVPV: This is upstream warning. Maybe let it stay as is?

subsys/bluetooth/host/id.c
PavelVPV: Lets keep it as a warning as the application will get `-EPERM` which it will be able to handle.

KyraLengfeld (approved):

PR #2979, opened 2025-06-23
[nrf fromlist] linker: Use zephyr,code-partition value

tomchy:
Enforce usage of values from devicetree if USE_DT_CODE_PARTITION Kconfig symbol is enabled.

Currently even if this symbol is enabled, the FLASH_LOAD_OFFSET value is used in the linker script.
The FLASH_LOAD_OFFSET symbol, even being promptless is such case, may be changed, leading to a very confusing situation, where devicetree
 values have no effect on the build.

A good example of the build that wants to use exactly the same set of
 configs, but a different link address is the A/B update scheme
(known ad Direct XIP in MCUboot), where the same application is linked to two different addresses.
If the source of the address remains in the Kconifg, the current approach forces a usage of two different configurations.

Upstream PR #: 91591


tomchy:
Resolved in: https://github.com/nrfconnect/sdk-zephyr/pull/3880/changes#diff-ca83f8e183c30d9a7224509a357eeff39dd73891b3ea30331477c374dffd430d for the new partition binding schema.

PR #2987, opened 2025-06-25
Device deinit driver support

bjarki-andreasen:
Device deinit driver support


bjarki-andreasen:
The change request is regarding a commit adding stuff to the zephyr release notes, this commit is not cherry picked, so merge with superpowers should be ok :)

bjarki-andreasen:
ping @anangl @kl-cruz @maciejpietras @nika-nordic :)

anangl (approved):

PR #2993, opened 2025-06-26
[nrf fromlist] soc: nordic: nrf54h: s2ram: Add FPU retention

adamkondraciuk:
Add FPU power management for suspend to RAM procedures. Add FPU save/restore procedures when `FPU_SHARING` feature is disabled.


nordic-piks (approved):

zycz (approved):
Tested machine learning application singlecore - it works fine.
There is high power consumption but it is probably due to other known issue.

PR #3000, opened 2025-06-27
[nrf fromtree] net: http_client: Fix handling of POLLHUP

juhaylinen:
POLLHUP event may be returned even if there is still data to read and POLLIN is set. To ensure all data is consumed, check for POLLHUP after handling POLLIN.

(cherry picked from commit ed79675199c321132ebb484895bba85400cad34b)


juhaylinen:
Changed to fromtree PR

PR #3001, opened 2025-06-27
[nrf fromlist] boards: nordic: nrf54h20dk: Fix FLASH_LOAD_SIZE

nordic-mik7:
Fix FLASH_LOAD_SIZE setting for builds with !USE_DT_CODE_PARTITION (for e.g. with no MCUBoot).
Having it wrongly set to 0x0 caused problems when running multiple images (radio core, application core) on the same board.

Upstream PR #: 92191
Related: NCSIDB-1628

### DIFF
--- a/boards/nordic/nrf54h20dk/Kconfig.defconfig
+++ b/boards/nordic/nrf54h20dk/Kconfig.defconfig
@@ -1,6 +1,9 @@
 # Copyright (c) 2024 Nordic Semiconductor ASA
 # SPDX-License-Identifier: Apache-2.0
 
+# Workaround for not being able to have commas in macro arguments
+DT_CHOSEN_Z_CODE_PARTITION := zephyr,code-partition
+
 if BOARD_NRF54H20DK_NRF54H20_CPUAPP
 
 config BT_HCI_IPC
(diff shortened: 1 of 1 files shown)


boards/nordic/nrf54h20dk/Kconfig.defconfig
```diff
@@ -26,6 +29,9 @@ config ROM_START_OFFSET
 config FLASH_LOAD_OFFSET
 	default 0x30000 if !USE_DT_CODE_PARTITION
 
+config FLASH_LOAD_SIZE
+	default $(dt_chosen_reg_size_hex,$(DT_CHOSEN_Z_CODE_PARTITION)) if !USE_DT_CODE_PARTITION
```
ahasztag, line 33: This doesn't seem the right approach, rather a fixed value should be used.

I do not thing that anything from   `zephyr,code-partition` should be used if `!USE_DT_CODE_PARTITION`

Let's say you have flash start at 2c000, and `zephyr-code-partition` and 3c000 - than the `zephyr-code-partition` would be smaller than the actual available size

ahasztag (changes requested):

boards/nordic/nrf54h20dk/Kconfig.defconfig
nordic-mik7: doesn't FLASH_LOAD_SIZE indicate the code parition in fact ? why should we include the rest of available flash in it ?

nordic-mik7:
Closing this PR as a result of conversation with @ahasztag. Please see upstream PR comment : https://github.com/zephyrproject-rtos/zephyr/pull/92191#issuecomment-3018461327

PR #3016, opened 2025-07-03
[nrf fromlist] boards: nordic: nrf54h20dk: ipc to cpusys with unbound

lstnl:
Unbound feature in IPC with cpusys will be detected. This is needed to enable IPC link renewal.

Upstream PR #: 92587


lstnl:
This should be discussed how to introduce it.

PR #3017, opened 2025-07-03
[nrf fromlist] soc: Enable radio core if BT is selected

tomchy:
Automatically enable booting radio core if BT HCI interface is enabled.


samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml
```diff
@@ -230,3 +230,27 @@ tests:
       - nrf5340dk/nrf5340/cpuapp
     integration_platforms:
       - nrf5340dk/nrf5340/cpuapp
+  sample.mcumgr.smp_svr.bt.nrf54h20dk:
```
ahasztag: Shouldn't we add this in `sdk-nrf/samples/zephyr/subsys/mgmt/mcumgr/smp_svr` ?

tomchy: @nordicjm What would you prefer?

nordicjm: sdk-nrf, the nrf52840 one here needs to be moved out there also

tomchy: Moved here: https://github.com/nrfconnect/sdk-nrf/pull/23174

tomchy:
Possibly already handled by: https://github.com/nrfconnect/sdk-zephyr/pull/3027

PR #3026, opened 2025-07-07
[fromtree] bluetooth: fix bug when destroying tx buffers on disconnected

alwa-nordic:
Cherry-pick [commit from upstream #90975](https://github.com/zephyrproject-rtos/zephyr/pull/90975/commits/f4f416dcae6725c7efaed411e8d8abe416cdbdfc)

## Review guide:

The only purpose of this PR is to cherry-pick [commit from upstream #90975](https://github.com/zephyrproject-rtos/zephyr/pull/90975/commits/f4f416dcae6725c7efaed411e8d8abe416cdbdfc).

### So why are there 5 commits in this PR?

- f9e37e59ffac Revert "[nrf noup] bluetooth: conn: Skip buffer ref count check in send_buf"
- 5394bcc8e2fb Revert "[nrf noup] bluetooth: conn: Allow for an extra ref in bt_l2cap_send_pdu"
- 8622c2df7c44 [nrf fromtree] bluetooth: fix bug when destroying tx queue buffers on disconnect
- 2aabbe60c70a [nrf noup] bluetooth: conn: Allow for an extra ref in bt_l2cap_send_pdu
- 3f5429916746 [nrf noup] bluetooth: conn: Skip buffer ref count check in send_buf

The commit in the middle is the cherry-pick we want. The NCS CI required that the diff of all 'fromtree' is the same as the upstream commit. This means it's not possible to alter it, and also precludes a manual conflict resolution when applying that commit.

The upstream commit did not cherry-pick without conflict. It conflicts with some noup commits we have downstream. To resolve this will move the noup changes to after the cherry-pick, by reverting them and reapplying them, doing a manual conflict resolution in the noup changes.

### So what do I need to review?

 - All reverted noup commits are reapplied and the merge is sensible.
 - That the fromtree cherry-pick is not a bad merge. The CI guarantees that the merge was automatic. But an automatic merge is still is a merge that may be broken.

##### [nrf noup] bluetooth: conn: Allow for an extra ref in bt_l2cap_send_pdu

```bash
git range-diff --creation-factor=100 685c53acebd503a6931b8c2811d8a1e74e0e4828^! 2aabbe60c70ab4f3f1a65bc3e9e7764322db2df0^!
```

The diff-diff is a bit unwieldly because of the comments in the code. I have cheated and removed all diff-diff to do with code comments, and get the below:

```diff
1:  685c53acebd5 ! 1:  2aabbe60c70a [nrf noup] bluetooth: conn: Allow for an extra ref in bt_l2cap_send_pdu
@@ Commit message

     Signed-off-by: Marek Pieta <Marek.Pieta@nordicsemi.no>
     (cherry picked from commit d6539f49a567b8539781356ecd03c749702b28ba)
+    (cherry picked from commit 685c53acebd503a6931b8c2811d8a1e74e0e4828)

  ## subsys/bluetooth/host/conn.c ##
 @@ subsys/bluetooth/host/conn.c: static int send_buf(struct bt_conn *conn, struct net_buf *buf,

  	uint16_t frag_len = MIN(conn_mtu(conn), len);

--	__ASSERT_NO_MSG(buf->ref == 1);
-+	if (buf->ref > 1 + (cb ? 1 : 0)) {
+-	__ASSERT_NO_MSG((buf->ref == 1) || (buf->ref == 2));
++	if (buf->ref > 2 + (cb ? 1 : 0)) {
 +		__ASSERT_NO_MSG(false);
 +	}
```

What you see there is that the previous version of this noup changes an assertion to an if-statement, adding `+ (cb ? 1 : 0)` to what's already there. The new version of the noup does the same, but it converts the new ` || (buf->ref == 2)` to just bumping up the number to `2`. This should be equivalent.

##### [nrf noup] bluetooth: conn: Skip buffer ref count check in send_buf

```bash
git range-diff --creation-factor=100 827eb444b1947eb198d4dad28f40ab38e5d3f1eb^! 3f5429916746cc270c43fe6a4c45b78d43fcf5f8^!
```

Again, with the comment-stuff removed:

```diff
1:  827eb444b194 ! 1:  3f5429916746 [nrf noup] bluetooth: conn: Skip buffer ref count check in send_buf
@@ Commit message

     Signed-off-by: Marek Pieta <Marek.Pieta@nordicsemi.no>
     (cherry picked from commit e589307758ca108bd1d3f12501418e0d1d892f5c)
+    (cherry picked from commit 827eb444b1947eb198d4dad28f40ab38e5d3f1eb)

  ## subsys/bluetooth/host/conn.c ##
 @@ subsys/bluetooth/host/conn.c: static int send_buf(struct bt_conn *conn, struct net_buf *buf,

  	uint16_t frag_len = MIN(conn_mtu(conn), len);

--	if (buf->ref > 1 + (cb ? 1 : 0)) {
+-	if (buf->ref > 2 + (cb ? 1 : 0)) {
 -		__ASSERT_NO_MSG(false);
 -	}

-+	__ASSERT_NO_MSG(IS_ENABLED(CONFIG_BT_ATT_SENT_CB_AFTER_TX) || (buf->ref == 1));
++	__ASSERT_NO_MSG(IS_ENABLED(CONFIG_BT_ATT_SENT_CB_AFTER_TX) || (buf->ref == 1) ||
++			(buf->ref == 2));
```

This noup is removing the if-condition and replacing it with an assertion again, so we convert the `> 2` back to a `|| (buf->ref == 2)` in the assertion.

#### The validity of the cherry-pick itself

You can look at the [upstream PR](https://github.com/zephyrproject-rtos/zephyr/pull/90975) to understand the change and note any comments that could indicate potential problems.

It's worth noting that the upstream PR has a test that has not been cherry-picked in this PR. I did not put much though into this other than wanting to keep the cherry-pick small. Less is more. But maybe cherry-picking this test will increase our confidence for this fix?


ludvigsj (approved):
Looks okay. I didn't do a detailed review of the fromtree commit itself, seeing as this has already been through a thorough review upstream. Handling of the noups looks like it's doing the correct thing as far as I can tell.

subsys/bluetooth/host/l2cap.c
```diff
@@ -729,13 +729,17 @@ int bt_l2cap_send_pdu(struct bt_l2cap_le_chan *le_chan, struct net_buf *pdu,
 		return -ENOTCONN;
 	}
 
-	if (pdu->ref != 1) {
+	/* Allow for an additional buffer reference if callback is provided. This can be used to
```
KyraLengfeld: I may have misunderstood your description, but could you elaborate a bit why make this change in this PR and what the change is fixing? While the change itself (increasing the reference count possible if there is a callback) does seem sensible, I am not finding why this is added here in particular, and also whether we have some test in place ensuring this doesn't cause resource issues in some other end?

subsys/bluetooth/host/conn.c
```diff
@@ -702,19 +702,27 @@ static int send_buf(struct bt_conn *conn, struct net_buf *buf,
 
 	uint16_t frag_len = MIN(conn_mtu(conn), len);
 
-	/* Check that buf->ref is 1 or 2. It would be 1 if this was
-	 * the only reference (e.g. buf was removed from the conn
-	 * tx_queue). It would be 2 if the tx_data_pull kept it on
-	 * the tx_queue for segmentation.
+	/* If ATT sent callback is delayed until data transmission
```
KyraLengfeld, line 705: Nitpick: "letter count" (forgot proper word) in one line had been increased. So better to go with the new one. Probably you can just have two long lines and run clang-format.

KyraLengfeld:
@alwa-nordic Have you looked more closely at the test you mention in the description? Because if it is relatively simple to cherry-pick, I'd say we should (we may have nooup changes in some other corner affecting the fix, or will have in the future, so better to have the test also downstream than not imo). Ofc, if the test needs to be amended or drags in lots of other changes, we can decide to rely on it being upstream only, guess we also want to cherry-pick this fix quick.

subsys/bluetooth/host/l2cap.c
KyraLengfeld: I see, I may have commented on the wrong commit. Let me check this. (had the commits open separately)

alwa-nordic:
I cherry-picked the test.

subsys/bluetooth/host/l2cap.c
MarekPieta: Hi
The noups are related to NCS-specific (not accepted in upstream Zephyr) workaround for HID report flow over Bluetooth LE. This PR does not show the complete use-case for the noups. For more details, refer to:

The related Jira bug ticket: https://nordicsemi.atlassian.net/browse/NCSDK-27422
PRs introducing fixes to the NCS: https://github.com/nrfconnect/sdk-nrf/pull/16027 and https://github.com/nrfconnect/sdk-zephyr/pull/1819

subsys/bluetooth/host/conn.c
```diff
@@ -702,19 +702,27 @@ static int send_buf(struct bt_conn *conn, struct net_buf *buf,
...
+	 *
+	 * Otherwise, check that buf->ref is 1 or 2. It would be 1
+	 * if this was the only reference (e.g. buf was removed from
+	 * the conn tx_queue). It would be 2 if the tx_data_pull
+	 * kept it on the tx_queue for segmentation.
 	 */
-	if (buf->ref > 2 + (cb ? 1 : 0)) {
-		__ASSERT_NO_MSG(false);
-	}
+	__ASSERT_NO_MSG(IS_ENABLED(CONFIG_BT_ATT_SENT_CB_AFTER_TX) || (buf->ref == 1) ||
```
MarekPieta, line 724: Can we squash the 2 noups? This one reverts part of the first noup

MarekPieta: Discussed offline with @alwa-nordic . Will not squash in scope of this PR.

MarekPieta (approved):
Change looks good. Tested locally, no regression found

PR #3027, opened 2025-07-07
Enable SOC_NRF54H20_CPURAD_ENABLE and SOC_NRF53_CPUNET_ENABLE by defa…

Mirlenko:
…ult for BLE samples

At the moment these configs should be selected explicitly for each sample. Apply them by default as these configs enable the clock for radio (network) core during application boot time and requests power for radio (network) core. Once powered, radio (network) core begins executing instructions.

### DIFF
--- a/soc/nordic/nrf53/Kconfig
+++ b/soc/nordic/nrf53/Kconfig
@@ -208,7 +208,7 @@ config SOC_NRF53_CPUNET_MGMT
 
 config SOC_NRF53_CPUNET_ENABLE
 	bool "NRF53 Network MCU is enabled at boot time"
-	default y if NRF_802154_SER_HOST
+	default y if NRF_802154_SER_HOST || BT
 	select SOC_NRF53_CPUNET_MGMT
 	help
 	  This option enables releasing the Network 'force off' signal, which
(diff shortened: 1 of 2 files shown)


soc/nordic/nrf53/Kconfig
```diff
@@ -208,7 +208,7 @@ config SOC_NRF53_CPUNET_MGMT
 
 config SOC_NRF53_CPUNET_ENABLE
 	bool "NRF53 Network MCU is enabled at boot time"
-	default y if NRF_802154_SER_HOST
+	default y if NRF_802154_SER_HOST || BT
```
nordicjm, line 211: rather than no-up here, put these in nrf/soc/nordic/nrf5* folders (can just make a Kconfig.defconfig and put e.g.
```
config SOC_NRF53_CPUNET_ENABLE
	default y if BT
```
)

rugeGerritsen, line 211: I would assume this change is not needed? We previously didn't need to set `SOC_NRF53_CPUNET_ENABLE` for Bluetooth applications

Mirlenko: I had a talk with @SebastianBoe , and he suggested to keep `SOC_NRF53_CPUNET_ENABLE` and `SOC_NRF54H20_CPURAD_ENABLE` consistent.

SebastianBoe: If we are observing that 54H needs to set this for BT applications, but 53 for some reason does not, then
I would also support having 54H copy whatever 53 is doing at the moment.

As long as there is no unnecessary divergence between 53 and 54 I am happy :)

PR #3033, opened 2025-07-08
[nrf fromlist] drivers: serial: nrfx_uarte: Add mode with TIMER byte counting

nordic-krch:
Upstream PR 92767

### DIFF
--- a/drivers/serial/uart_nrfx_uarte.c
+++ b/drivers/serial/uart_nrfx_uarte.c
@@ -875,6 +875,78 @@ static void tx_start(const struct device *dev, const uint8_t *buf, size_t len)
 static void rx_timeout(struct k_timer *timer);
 static void tx_timeout(struct k_timer *timer);
 
+static void user_callback(const struct device *dev, struct uart_event *evt)
+{
+	struct uarte_nrfx_data *data = dev->data;
+
+	if (data->async->user_callback) {
+		data->async->user_callback(dev, evt, data->async->user_data);
+	}
+}
+
+static void rx_buf_release(const struct device *dev, uint8_t *buf)
+{
+	struct uart_event evt = {
+		.type = UART_RX_BUF_RELEASED,
+		.data.rx_buf.buf = buf,
+	};
+
+	user_callback(dev, &evt);
+}
+
+static void notify_rx_disable(const struct device *dev)
+{
+	const struct uarte_nrfx_config *cfg = dev->config;
+	struct uart_event evt = {
+		.type = UART_RX_DISABLED,
+	};
+
+	if (LOW_POWER_ENABLED(cfg)) {
+		uint32_t key = irq_lock();
+
+		uarte_disable_locked(dev, UARTE_FLAG_LOW_POWER_RX);
+		irq_unlock(key);
+	}
+
+	user_callback(dev, (struct uart_event *)&evt);
+
+	/* runtime PM is put after the callback. In case uart is re-enabled from that
+	 * callback we avoid suspending/resuming the device.
+	 */
+	if (IS_ENABLED(CONFIG_PM_DEVICE_RUNTIME)) {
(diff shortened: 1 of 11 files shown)


drivers/serial/uart_nrfx_uarte.c
```diff
@@ -728,133 +809,886 @@ static void uarte_periph_enable(const struct device *dev)
...
+}
+#endif /* !defined(CONFIG_UART_NRFX_UARTE_ENHANCED_RX) */
+
+#ifdef CONFIG_UARTE_NRFX_UARTE_COUNT_BYTES_WITH_TIMER
+
+static uint32_t get_byte_cnt(NRF_TIMER_Type *timer)
+{
+	nrf_timer_task_trigger(timer, nrf_timer_capture_task_get(UARTE_TIMER_CAPTURE_CH));
+
+	nrf_barrier_w();
```
nika-nordic, line 1114: I think it might not be needed - accesses (task trigger and register read) toward same bus slave (peripheral) should never be re-ordered

drivers/serial/uart_nrfx_uarte.c
```diff
@@ -728,133 +809,886 @@ static void uarte_periph_enable(const struct device *dev)
...
+		cbwt_data->usr_wr_off = 0;
+
+		if (async_rx->buf_len == 0) {
+			return false;
+		}
+
+		/* Set past value to ensure that event will not expire after clearing but
+		 * before setting the new value.
+		 */
+		nrf_timer_cc_set(cfg->timer_regs, UARTE_TIMER_USR_CNT_CH, cbwt_data->cc_usr - 10);
```
nika-nordic: no risk of underflow on `cbwt_data->cc_usr - 10`?

Maybe would be good to replace `10` with define, so it is more readable why `10` specifically?

drivers/serial/uart_nrfx_uarte.c
```diff
@@ -1086,35 +1896,45 @@ static int uarte_nrfx_rx_enable(const struct device *dev, uint8_t *buf,
 	}
 
 #ifdef CONFIG_HAS_NORDIC_DMM
-	uint8_t *dma_buf;
-	int ret = 0;
+	if (!IS_CBWT(dev)) {
+		uint8_t *dma_buf;
```
nika-nordic: nit: declaring `dma_buf` as `void *` could maybe help avoiding casting to `void**` below

west.yml
```diff
@@ -200,7 +200,7 @@ manifest:
       groups:
         - hal
     - name: hal_nordic
-      revision: 71308dc6d8c021887ce5d98a36cafe9517375a91
+      revision: pull/306/head
```
nika-nordic: is it really needed? I believe MDK accesses are already utilized instead of HAL functions. Maybe `_tx/rx_macnt_set` functions would need to be converted to MDK accesses as well to get rid of the dependency.

Alternatively we could wait for nrfx release

drivers/serial/uart_nrfx_uarte.c
nordic-krch: it is actually needed for VPR. https://github.com/zephyrproject-rtos/zephyr/pull/76955/commits/c9bbc5a076570de771072a6b9aac5e61fc5f25fd

drivers/serial/uart_nrfx_uarte.c
nordic-krch: I changed it to -1, i did not add any magic number as -1 is self explanatory as past value.

drivers/serial/uart_nrfx_uarte.c
nordic-krch: done

west.yml
nordic-krch: replaced usage with MDK register access. It can later on be updated to use hal from nrfx 3.13.

PR #3042, opened 2025-07-10
Nrf pinctrl power domain refactor

bjarki-andreasen:
The power domain and pinctrl refactor touches quite a few files, and was merged upstream before some already fromlisted commits, so there are a bit of reverts and reapplies with no actual code changes to apply the fromtree/fromlist commits cleanly.


anangl:
I think that f5228b90b39edff5bfdc74c176b206f748e961b0 can be dropped permanently. "nordic,nrf-exmif-spi" is no longer used in NCS (since https://github.com/nrfconnect/sdk-nrf/pull/21704 was merged).

PR #3045, opened 2025-07-14
3.1.0 wi-fi upmerge 

rado17:
Wi-Fi only Upmerge for 3.1.0

manifest-pr-skip


krish2718:
Why do we still need this `noup`, it was done last time as were in the release rush. If the certs from upstream don't work out-of-box then we can submit a variant that works out-of-box for nRF platforms.

rado17:
> Why do we still need this `noup`, it was done last time as were in the release rush. If the certs from upstream don't work out-of-box then we can submit a variant that works out-of-box for nRF platforms.

Nordic specific directory in upstream certs then? RSA-3k certs work fine, but we may have to enable additional configs for RSA-2k certs from upstream. 
Also, for MBEDTLS_HEAP_SIZE, the increase is required only with downstream. So, I think, `noup` is justified for that commit.

krish2718:
> Nordic specific directory in upstream certs then? RSA-3k certs work fine, but we may have to enable additional configs for RSA-2k certs from upstream.

I would put a algo specific directory e.g., `rsa2k-no-ebc3`, this way it can be reused in case other platforms have similar constraints.

> Also, for MBEDTLS_HEAP_SIZE, the increase is required only with downstream. So, I think, `noup` is justified for that commit.

We don't want to rely on `noup` unless its absolutely must, why can't we add ` <increased_heap> if SOC_FAMILY_NORDIC_NRF` to upstream?

rado17:
> > Nordic specific directory in upstream certs then? RSA-3k certs work fine, but we may have to enable additional configs for RSA-2k certs from upstream.
> 
> I would put a algo specific directory e.g., `rsa2k-no-ebc3`, this way it can be reused in case other platforms have similar constraints.
> 
> > Also, for MBEDTLS_HEAP_SIZE, the increase is required only with downstream. So, I think, `noup` is justified for that commit.
> 
> We don't want to rely on `noup` unless its absolutely must, why can't we add ` <increased_heap> if SOC_FAMILY_NORDIC_NRF` to upstream?


Let me check this.

krish2718 (approved):

kapbh (approved):

sachinthegreen (approved):

PR #3046, opened 2025-07-14
[nrf noup] boards: nordic: nrf54lm20 adjust ROM_START_OFFSET

michalek-no:
adds kconfig that was probably skipped during migration.


boards/nordic/nrf54lm20dk/Kconfig.defconfig
```diff
@@ -4,6 +4,7 @@
 if BOARD_NRF54LM20DK_NRF54LM20A_CPUAPP
 
 config ROM_START_OFFSET
+	default 0 if PARTITION_MANAGER_ENABLED
```
nordicjm, line 7: this should already be the case https://github.com/nrfconnect/sdk-nrf/blob/main/Kconfig.nrf#L35

michalek-no: This has no effect in our case.

kapi-no: For some reasons, the override is not applied in various projects (e.g. nRF Desktop).

PR #3060, opened 2025-07-18
drivers: timer: grtc: Add uninit function

adamkondraciuk:


drivers/timer/nrf_grtc_timer.c
```diff
@@ -453,6 +453,18 @@ uint32_t sys_clock_elapsed(void)
 	return last_elapsed / CYC_PER_TICK;
 }
 
+void z_nrf_grtc_uninit(void)
```
nordicjm: This should be using https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/timer/system_timer.h#L113 and be submitted upstream

PR #3066, opened 2025-07-22
Wi-Fi upmerge 2 for 3.1.0

kapbh:
manifest-pr-skip


doc/connectivity/networking/api/coap_server.rst
```diff
@@ -243,7 +243,7 @@ following example simply prints when an event occurs.
     #define COAP_EVENTS_SET (NET_EVENT_COAP_OBSERVER_ADDED | NET_EVENT_COAP_OBSERVER_REMOVED | \
                              NET_EVENT_COAP_SERVICE_STARTED | NET_EVENT_COAP_SERVICE_STOPPED)
 
-    void coap_event_handler(uint32_t mgmt_event, struct net_if *iface,
+    void coap_event_handler(uint64_t mgmt_event, struct net_if *iface,
```
krish2718: I don't think we can pull this, @rlubos  tried this earlier in the full upmerge but abandoned due to some limitations.

kapbh: I need to take because of conflicts with @ajayparida  commit.

rlubos: Depends how much you need this, for sure pulling the commit alone is not enough. I have a separate PR (https://github.com/nrfconnect/sdk-nrf/pull/22993) which should be ready I believe, although since we were not doing another upmerge before the release, I've decided to postpone it not to bring a stable API change to the release w/o a need.

But if you need this change, I can refresh my PR an push it forward.

krish2718: Either we pull this change in or do a `noup` for the conflicting commit?

rlubos: Temporary noup is also fine, just leave a note in the commit message it should be reverted on the next upmerge.

D-Triveni (approved):

ajayparida (approved):

PR #3068, opened 2025-07-22
[nrf fromlist] boards: nordic: nrf54l20pdk: Remove obsolete board

nordic-segl:
Board nrf54l20pdk was renamed to nrf54lm20dk.
Remove obsolete board definition.

Upstream PR #: 93483


nordic-segl:
I'm stuck due to `Commit tags` fail.
I think we can wait with this. Board 54L20 will be removed from NCS during next upmerge.

mstasiaknordic:
If this PR is to be merged and not resolved via upmerge then this commit https://github.com/zephyrproject-rtos/zephyr/commit/5c7c3980eb0600a058d4355e3d101078f88d26dc would have to be added

nordic-segl:
> If this PR is to be merged and not resolved via upmerge then this commit [zephyrproject-rtos/zephyr@5c7c398](https://github.com/zephyrproject-rtos/zephyr/commit/5c7c3980eb0600a058d4355e3d101078f88d26dc) would have to be added

To fix `Commit tags` check in this PR I would have to do mini upmerge :cry:

PR #3070, opened 2025-07-23
[nrf fromtree] tests: kernel: timer: timer_behavior: Fix max_stddev c…

nordic-segl:
…alculation

max_stddev calculation was previously patched to be more at least 1 system clock cycle to cover for platforms that use higher frequency system clock (32kHz). On that platform (nRF52) system clock frequency was the same as tick frequency but on nRF54x that is no longer true. Initially, the intention was to use 1 system tick and not cycle. Fixing it now.

(cherry picked from commit 26fe43dd363838e6cfe9c928e72c2af481b84e47)

### DIFF
--- a/tests/kernel/timer/timer_behavior/src/jitter_drift.c
+++ b/tests/kernel/timer/timer_behavior/src/jitter_drift.c
@@ -237,8 +237,8 @@ static void do_test_using(void (*sample_collection_fn)(void), const char *mechan
 		- expected_time_drift_us;
 	double time_diff_us_abs = time_diff_us >= 0.0 ? time_diff_us : -time_diff_us;
 
-	/* If max stddev is lower than a single clock cycle then round it up. */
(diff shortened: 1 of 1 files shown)


nordic-segl:
Pull in changes from https://github.com/zephyrproject-rtos/zephyr/pull/93325/files

nordic-segl:
Fixes one test that is currently in the quarantine
```
09:35:32  INFO    - 1/1 nrf54h20dk@0.9.0/nrf54h20/cpurad tests/kernel/timer/timer_behavior/kernel.timer.timer PASSED (device: 1051164192, 34.676s <zephyr>)
```
```
- scenarios:
    - kernel.timer.timer
  platforms:
    - nrf54h20dk@0.9.0/nrf54h20/cpurad
  comment: "https://github.com/zephyrproject-rtos/zephyr/issues/88491"
```

PR #3074, opened 2025-07-23
[nrf fromlist] cmake: mcuboot: SHA512/pure image signing

nordic-mik7:
Update signing script to use proper arguments for imgtool when SHA512 or pure signature is needed.

Upstream PR #: 93813

### DIFF
--- a/cmake/mcuboot.cmake
+++ b/cmake/mcuboot.cmake
@@ -152,6 +152,13 @@ function(zephyr_mcuboot_tasks)
     set(imgtool_args --align ${write_block_size} ${imgtool_args})
   endif()
 
+  # Set proper hash calculation algorithm for signing
+  if(CONFIG_MCUBOOT_BOOTLOADER_SIGNATURE_TYPE_PURE)
+    set(imgtool_args --pure ${imgtool_args})
+  elseif(CONFIG_MCUBOOT_BOOTLOADER_USES_SHA512)
+    set(imgtool_args --sha 512 ${imgtool_args})
+  endif()
+
   # Extensionless prefix of any output file.
   set(output ${ZEPHYR_BINARY_DIR}/${KERNEL_NAME})
 
--- a/modules/Kconfig.mcuboot
+++ b/modules/Kconfig.mcuboot
@@ -288,11 +288,6 @@ config MCUBOOT_BOOTLOADER_NO_DOWNGRADE
 	  MCUBOOT_DOWNGRADE_PREVENTION option enabled.
 endif
 
-config MCUBOOT_BOOTLOADER_USES_SHA512
-	bool "MCUboot uses SHA512 for image hash"
-	help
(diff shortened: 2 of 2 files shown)


cmake/mcuboot.cmake
```diff
@@ -152,6 +152,13 @@ function(zephyr_mcuboot_tasks)
     set(imgtool_args --align ${write_block_size} ${imgtool_args})
   endif()
 
+  # Set proper hash calculation algorithm for signing
+  if(CONFIG_MCUBOOT_BOOTLOADER_USES_SHA512 AND NOT CONFIG_MCUBOOT_BOOTLOADER_SIGNATURE_TYPE_PURE)
```
tomchy: If found better, you may change the order of `if`s.

tomchy (approved):

cmake/mcuboot.cmake
nordic-mik7: done

nordic-mik7:
> move this (and `config MCUBOOT_BOOTLOADER_USES_SHA512`) upstream

Wasn't there a reason for `config MCUBOOT_BOOTLOADER_USES_SHA512` to be [noup] ? I aligned to existing solution, why now it should be changed ?

nordicjm:
> > move this (and `config MCUBOOT_BOOTLOADER_USES_SHA512`) upstream
> 
> Wasn't there a reason for `config MCUBOOT_BOOTLOADER_USES_SHA512` to be [noup] ? I aligned to existing solution, why now it should be changed ?

That should not have been added to ncs at all, it's not ncs specific and just causes issues in upmerges with conflicts

nordic-mik7:
> > > move this (and `config MCUBOOT_BOOTLOADER_USES_SHA512`) upstream
> > 
> > 
> > Wasn't there a reason for `config MCUBOOT_BOOTLOADER_USES_SHA512` to be [noup] ? I aligned to existing solution, why now it should be changed ?
> 
> That should not have been added to ncs at all, it's not ncs specific and just causes issues in upmerges with conflicts

done

scripts/ci/check_compliance.py
```diff
@@ -1117,6 +1117,8 @@ def check_no_undef_outside_kconfig(self, kconf):
         "LOG_BACKEND_MOCK_OUTPUT_SYST", #Referenced in testcase.yaml of log_syst test
         "LSM6DSO_INT_PIN",
         "MCUBOOT_ACTION_HOOKS",     # Used in (sysbuild-based) test
+        "MCUBOOT_BOOTLOADER_SIGNATURE_TYPE_PURE", # Used in mcuboot.cmake
+        "MCUBOOT_BOOTLOADER_USES_SHA512", # Used in mcuboot.cmake
```
nordicjm: remove as per upstream pr

nordic-mik7: done

nordicjm (approved):

PR #3089, opened 2025-07-25
[nrf fromlist] drivers: serial: nrfx_uarte: Add support for device deinit

mstasiaknordic:
Add device deinit function. Support is optional as it is not widely used and it enables pinctrl sleep state so it impacts memory footprint.

Additionally, added option to create pinctrl sleep state. So far it was only created when power management was used but in case of deinit it is possible that it is used without PM and we still want to put pins into the default state.

Upstream PR #: 93273


nordic-segl (approved):

mstasiaknordic:
Closing, changes arrived with upmerge

PR #3090, opened 2025-07-25
[nrf fromlist] dts: vendor: nordic: nrf54l10: Fix wrong size of NVM

nordicjm:
.

### DIFF
--- a/boards/nordic/nrf54l15dk/nrf54l15dk_nrf54l10_cpuapp.yaml
+++ b/boards/nordic/nrf54l15dk/nrf54l15dk_nrf54l10_cpuapp.yaml
@@ -10,7 +10,7 @@ toolchain:
   - zephyr
 sysbuild: true
 ram: 192
-flash: 230
+flash: 1012
 supported:
   - adc
   - counter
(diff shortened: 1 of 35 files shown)


boards/nordic/nrf54l15dk/nrf54l15dk_nrf54l10_cpuapp.yaml
```diff
@@ -10,7 +10,7 @@ toolchain:
   - zephyr
 sysbuild: true
 ram: 192
-flash: 230
+flash: 1012
```
masz-nordic, line 13: This should be whole NVM allocated to that core or size of image partition? Is that specified anywhere?

anangl (approved):

boards/nordic/nrf54l15dk/nrf54l15dk_nrf54l10_cpuapp.yaml
nordicjm: whole for cpuapp since it doesn't include MCUboot (unless you manually enable it)

kkasperczyk-no (approved):

PR #3091, opened 2025-07-28
[nrf fromlist] drivers: flash_mspi_nor: Apply a few improvements

anangl:
Cherry-pick commits from https://github.com/zephyrproject-rtos/zephyr/pull/93093.

### DIFF
--- a/drivers/flash/flash_mspi_nor.c
+++ b/drivers/flash/flash_mspi_nor.c
@@ -27,7 +27,7 @@ void flash_mspi_command_set(const struct device *dev, const struct flash_mspi_no
 	dev_data->xfer.xfer_mode  = MSPI_PIO;
 	dev_data->xfer.packets    = &dev_data->packet;
 	dev_data->xfer.num_packet = 1;
-	dev_data->xfer.timeout    = 10;
+	dev_data->xfer.timeout    = dev_config->transfer_timeout;
 
 	dev_data->xfer.cmd_length = cmd->cmd_length;
 	dev_data->xfer.addr_length = cmd->addr_length;
@@ -887,6 +887,7 @@ BUILD_ASSERT((FLASH_SIZE_INST(inst) % CONFIG_FLASH_MSPI_NOR_LAYOUT_PAGE_SIZE) ==
 				/ 1000,						\
 		.reset_recovery_us = DT_INST_PROP_OR(inst, t_reset_recovery, 0)	\
 				   / 1000,))					\
+		.transfer_timeout = DT_INST_PROP(inst, transfer_timeout),	\
 		FLASH_PAGE_LAYOUT_DEFINE(inst)					\
 		.jedec_id = DT_INST_PROP(inst, jedec_id),			\
 		.jedec_cmds = FLASH_CMDS(inst),					\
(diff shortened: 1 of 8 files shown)


drivers/flash/flash_mspi_nor_sfdp.h
masz-nordic, line 1: Could the part not specific to mspi be moved to `jesd216.h`? Some defines used here are already in that header.

drivers/flash/flash_mspi_nor.c
```diff
@@ -563,8 +627,29 @@ static int quad_enable_set(const struct device *dev, bool enable)
...
+		if (rc < 0) {
+			return rc;
+		}
+	}
 
-static int default_io_mode(const struct device *dev)
+	set_up_xfer(dev, MSPI_TX);
+	rc = perform_xfer(dev, 0xE7, false);
+	if (rc < 0) {
+		LOG_ERR("Command 0xE7 failed: %d", rc);
```
masz-nordic: `__function__`? 0xE7 seems cryptic.

drivers/flash/flash_mspi_nor.c
```diff
@@ -771,6 +853,51 @@ static int drv_init(const struct device *dev)
...
+			dev->name);
+		return -EINVAL;
+	}
+
+	if (dev_data->cmd_info.pp_cmd == 0) {
+		LOG_ERR("Page Program command not defined for %s, "
+			"use \"write-command\" property to specify it.",
+			dev->name);
+		return -EINVAL;
+	}
```
masz-nordic, line 1090: Could those be build time asserts?

drivers/flash/flash_mspi_nor_quirks.h
```diff
@@ -63,12 +63,6 @@ static inline int mxicy_mx25r_post_switch_mode(const struct device *dev)
 		return 0;
 	}
 
-	/* Wait for previous write to finish */
-	rc = wait_until_ready(dev, K_USEC(1));
-	if (rc < 0) {
-		return rc;
-	}
-
```
masz-nordic: Wrong commit? This change is not mentioned in the commit description.

drivers/flash/flash_mspi_nor.c
anangl: Oops, it should be 0xB7, actually...
But the spec says "issue write enable instruction 06h, then issue instruction B7h", so I'm not sure if it's worth defining a symbol for it.

drivers/flash/flash_mspi_nor.c
anangl: They could be, in the current shape of the driver. But I wanted to already prepare it for possibility of overriding these commands at runtime.

drivers/flash/flash_mspi_nor_quirks.h
anangl: This waiting here is not needed because `quad_enable_set()` waits after writing the status register. But I can mention this in the commit message so there's no doubt.

carlescufi (approved):

drivers/flash/flash_mspi_nor.c
anangl: fixed the command code

drivers/flash/flash_mspi_nor_quirks.h
anangl: Added explanation of the change in the commit message.

drivers/flash/flash_mspi_nor_sfdp.h
anangl: I removed `BFP_DW1_ADDR_*` symbols and used instead corresponding ones from `jesd216.h`, but I don't think there's a point in adding some definitions in this file now. It's better to do it when implementing runtime SFDP processing, as probably some new decoding functions will be also added there.

anangl:
Rebased.

PR #3095, opened 2025-07-28
drivers: adc: nrfx_saadc: Add support for SAADC internal sampling timer

jaz1-nordic:
The SAMPLERATE register can be used as a local timer instead
of triggering individual SAMPLE tasks. When SAMPLERATE.MODE is set
to Timers, it is sufficient to trigger SAMPLE task only once in order
to start the SAADC and triggering the STOP task will stop sampling.
The SAMPLERATE.CC field controls the sample rate.

The SAMPLERATE timer should not be combined with SCAN mode and
only one channel should be enabled when using the internal timer.

Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/91368

### DIFF
--- a/drivers/adc/adc_nrfx_saadc.c
+++ b/drivers/adc/adc_nrfx_saadc.c
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#define ADC_CONTEXT_USES_KERNEL_TIMER
 #include "adc_context.h"
 #include <nrfx_saadc.h>
 #include <zephyr/dt-bindings/adc/nrf-saadc-v3.h>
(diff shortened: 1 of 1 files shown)


drivers/adc/adc_context.h
```diff
@@ -115,6 +115,7 @@ static inline void adc_context_request_next_sampling(struct adc_context *ctx)
 }
 
 #ifdef ADC_CONTEXT_USES_KERNEL_TIMER
+#ifndef ADC_DRIVER_USES_INTERNAL_TIMER
```
masz-nordic: This is okay as a temporary measure, but some refactoring would be needed to suport this mixed case correctly.
See line 33:
```
 * If a given driver uses some dedicated hardware timer to trigger consecutive
 * samplings, it should implement also the following two functions. Otherwise,
 * it should define the ADC_CONTEXT_USES_KERNEL_TIMER macro to enable parts of
 * this module that utilize a standard kernel timer.
```
Currenly SAADC driver is doing both.

drivers/adc/adc_nrfx_saadc.c
```diff
@@ -490,29 +525,78 @@ static int check_buffer_size(const struct adc_sequence *sequence, uint8_t active
...
+/* The internal timer runs at 16 MHz, so to convert the interval in microseconds
+ * to the internal timer CC value, we can use the formula:
+ * interval_cc = interval_us * 16 MHz
+ * where 16 MHz is the frequency of the internal timer.
+ *
+ * The maximum value for interval_cc is 2047, which corresponds to
+ * approximately 7816 Hz ~ 128us.
+ * The minimum value for interval_cc is depends on the SoC.
+ */
+static inline uint16_t interval_to_cc(uint32_t interval_us)
```
masz-nordic:
```suggestion
static inline uint16_t interval_to_cc(uint16_t interval_us)
```
Since this function must  called with `interval_us > 0` and `interval_us <= ADC_INTERNAL_TIMER_INTERVAL_MAX_US` it would be good to assert it.
Then error checking is not necessary.

masz-nordic (approved):

drivers/adc/adc_context.h
jaz1-nordic: I removed the defines and moved the timer support directly to the driver.

jaz1-nordic:
@bjarki-andreasen please, take a look also here.

bjarki-andreasen (approved):

nika-nordic (approved):

PR #3096, opened 2025-07-28
drivers: adc: nrfx: enable negative values for single-ended ADC readings

jaz1-nordic:
Add custom nordic-specific ADC API to enable to read negative values from a single-ended ADC channels.

Based on #3095, only the last commit is in scope of this PR.
Upstream PR: https://github.com/zephyrproject-rtos/zephyr/pull/91704


jaz1-nordic:
Handled here: https://github.com/nrfconnect/sdk-zephyr/pull/3147

PR #3101, opened 2025-07-29
[nrf fromtree] drivers: mspi_dw: Prevent RX FIFO overflows in Single IO mode

anangl:
In Single IO mode, where the clock stretching feature is not available, the RX FIFO could get overflowed if the TX FIFO was filled with more dummy bytes than the RX FIFO could receive data items at a given moment. Such problem could be quite easy hit on nRF54H20, for example for longer RX transfers (like 8 kB) at 8 MHz.
This commit limits accordingly the number of dummy bytes that can be enqueued in the TX FIFO.

Upstream PR #: 93777


carlescufi (approved):

bjarki-andreasen (approved):

anangl:
Rebased.

PR #3147, opened 2025-08-05
[nrf fromlist] drivers: adc: nrfx: enable negative values for single-ended ADC readings

jaz1-nordic:
The ADC driver API already supports ADC readings which can return signed values, these are differential readings. In Nordic's datasheet, we have a mode called "single ended", but its just a name. "Single ended" is a differential reading, with the negative channel tied to GND. This is not compatible with zephyrs definition of a single ended reading.

To support Nordic's "single ended" mode, the user must configure a differential reading, with the negative input tied to ground, which the saadc driver can then use to configure the reading as Nordic SAADC "single ended", and return negative values as expected.

Upstream PR #: 94069


drivers/adc/adc_nrfx_saadc.c
```diff
@@ -187,21 +187,24 @@ static int input_assign(nrf_saadc_input_t *pin_p,
...
 	} else {
 		*pin_n = NRF_SAADC_INPUT_DISABLED;
 	}
 #else
 	*pin_p = channel_cfg->input_positive;
-	*pin_n = channel_cfg->differential ? channel_cfg->input_negative
+	*pin_n = channel_cfg->differential ? (channel_cfg->input_negative == NRF_SAADC_GND
+						      ? NRF_SAADC_INPUT_DISABLED
+						      : channel_cfg->input_negative)
 					   : NRF_SAADC_INPUT_DISABLED;
```
masz-nordic:
```suggestion
	*pin_n = channel_cfg->differential && (channel_cfg->input_negative != NRF_SAADC_GND) ? 
						      : channel_cfg->input_negative
					   : NRF_SAADC_INPUT_DISABLED;
```

jaz1-nordic:
@nika-nordic @bjarki-andreasen please take a look also here.

PR #3156, opened 2025-08-06
[nrf noup] Bluetooth: BAP: Add target latency and phy to codec_cfg

koffes:
The codec configuration operation for unicast (ASCS) contain both a target PHY and a target latency value that was previously just hardcoded by the client and unavailable by the server.

This commit adds them to the bt_audio_codec_cfg struct so that applications can both set and get these values.

The default values used by BT_AUDIO_CODEC_CFG use the same values as the client would previously set.

(cherry picked from commit 8788f8ba260d60e7f0b003324da93d3e29757ae2)


koffes:
I have deleted the change to the documentation. **Conflicts have been resolved locally, history is not preserved**

carlescufi (approved):

Thalley:
https://github.com/zephyrproject-rtos/zephyr/pull/94053 related to this change as well for the BT Tester

koffes:
Downstream run has completed.

PR #3170, opened 2025-08-07
[nrf noup] samples: smp_svr: fix ext_flash configs

michalek-no:
Disable FPROTECT for MCUBoot external flash builds due to NVM size.


nvlsianpu (approved):

samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml
```diff
@@ -181,6 +181,7 @@ tests:
       - mcuboot_EXTRA_DTC_OVERLAY_FILE="boards/nrf54l15dk_nrf54l15_cpuapp_ext_flash.overlay"
       - SB_CONFIG_PM_OVERRIDE_EXTERNAL_DRIVER_CHECK=y
       - SB_CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=y
+      - mcuboot_CONFIG_FPROTECT=n
```
nordicjm, line 184: we should not be making these changes, make it in the mcuboot as a noup fixing the existing one or in ncs in sysbuild config

shanthanordic:
as agrred with @nordicjm  in internal daily, this milestone removed.

PR #3179, opened 2025-08-08
[nrf fromlist] drivers: timer: nrf_rtc: Fix custom wrapping

nordic-krch:
Use channel 0 for RTC wrapping. Skip that channel when processing channels.
Fix algorithm that prevents setting CC=0 when custom wrapping is used.

Upstream PR #: 94259


mstasiaknordic:
Removing backport label as the PR being fixed here was not backported

PR #3180, opened 2025-08-08
soc: nrf54l: Change logic for KMU reserved area

Vge0rge:
nrf-squash! [nrf noup] soc: nrf54l: Add custom section for KMU

Change how the KMU reserved area is included in the linker file. Now there is a separate Kconfig which shows it the there is a dts entry for the KMU
reserved area.

### DIFF
--- a/soc/nordic/nrf54l/CMakeLists.txt
+++ b/soc/nordic/nrf54l/CMakeLists.txt
@@ -7,10 +7,12 @@ zephyr_library_sources(
   )
 zephyr_include_directories(.)
 
+dt_nodelabel(kmu_push_area_node NODELABEL nrf_kmu_reserved_push_area)
+
 # We need a buffer in memory in a static location which can be used by
 # the KMU peripheral. The KMU has a static destination address, we chose
 # this address to be 0x20000000, which is the first address in the SRAM.
-if(NOT CONFIG_BUILD_WITH_TFM AND CONFIG_PSA_NEED_CRACEN_KMU_DRIVER AND CONFIG_XIP)
+if(NOT CONFIG_BUILD_WITH_TFM AND CONFIG_PSA_NEED_CRACEN_KMU_DRIVER AND NOT kmu_push_area_node)
 # Exclamation mark is printable character with the lowest number in ASCII table.
 # We are sure that this file will be included first.
 zephyr_linker_sources(RAM_SECTIONS SORT_KEY ! kmu_push_area_section.ld)


soc/nordic/nrf54l/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ zephyr_include_directories(.)
 # We need a buffer in memory in a static location which can be used by
 # the KMU peripheral. The KMU has a static destination address, we chose
 # this address to be 0x20000000, which is the first address in the SRAM.
-if(NOT CONFIG_BUILD_WITH_TFM AND CONFIG_PSA_NEED_CRACEN_KMU_DRIVER AND CONFIG_XIP)
+if(NOT CONFIG_BUILD_WITH_TFM AND CONFIG_PSA_NEED_CRACEN_KMU_DRIVER AND NOT CONFIG_CRACEN_KMU_RESERVED_MEMORY_IN_DTS)
```
nordicjm: this is referencing a Kconfig that does not exist in this repo, it exists in the sdk-nrf repo, would be better if it was defined in this repo but not a blocker issue since Kconfig checks are not performed here

Vge0rge: It is true but at the moment that is not possible. This already referenced an out of tree Kconfig (namely CONFIG_PSA_NEED_CRACEN_KMU_DRIVER). But to be honest that is not a big issue anyway, I will create a task to remove this linker integration all together and handle everything in dts.

tomi-font (commented):
Not possible to check the DT node directly rather than making a Kconfig option for it in sdk-nrf?

Vge0rge:
> Not possible to check the DT node directly rather than making a Kconfig option for it in sdk-nrf?

I am not aware of any way to do that no. I searched a bit on the net and asked the AI gods but they cannot figure out this riddle either.

nordicjm:
> > Not possible to check the DT node directly rather than making a Kconfig option for it in sdk-nrf?
> 
> I am not aware of any way to do that no. I searched a bit on the net and asked the AI gods but they cannot figure out this riddle either.

what are you trying to do?

Vge0rge:
> > > Not possible to check the DT node directly rather than making a Kconfig option for it in sdk-nrf?
> > 
> > 
> > I am not aware of any way to do that no. I searched a bit on the net and asked the AI gods but they cannot figure out this riddle either.
> 
> what are you trying to do?

In essence I am trying to access DTS information in cmake.

nordicjm:
> > > > Not possible to check the DT node directly rather than making a Kconfig option for it in sdk-nrf?
> > > 
> > > 
> > > I am not aware of any way to do that no. I searched a bit on the net and asked the AI gods but they cannot figure out this riddle either.
> > 
> > 
> > what are you trying to do?
> 
> In essence I am trying to access DTS information in cmake.

yes that works fine, was more meaning show a dts snippet of what information you are trying to get

Vge0rge:
> > > > > Not possible to check the DT node directly rather than making a Kconfig option for it in sdk-nrf?
> > > > 
> > > > 
> > > > I am not aware of any way to do that no. I searched a bit on the net and asked the AI gods but they cannot figure out this riddle either.
> > > 
> > > 
> > > what are you trying to do?
> > 
> > 
> > In essence I am trying to access DTS information in cmake.
> 
> yes that works fine, was more meaning show a dts snippet of what information you are trying to get

```
	soc {
		reserved-memory {
			#address-cells = <1>;
			#size-cells = <1>;
			ranges;

			 nrf_kmu_reserved_push_area: memory@20000000 {
				reg = <0x20000000 0x80>;
				compatible = "zephyr,memory-region", "mmio-sram";
				zephyr,memory-region = "nrf_kmu_reserved_push_area";
				status = "okay";
			 };

		};
	};
```

Here, what the Kconfig does is that it checks if there is a node with the name nrf_kmu_reserved_push_area enabled. Are you aware of any methodology to do that in cmake directly?

Vge0rge:
> Not possible to check the DT node directly rather than making a Kconfig option for it in sdk-nrf?

Apparently Jamie > AI gods so he gave me a suggestion and updated the PR to avoid creating a new Kconfig.

PR #3186, opened 2025-08-11
[nrf fromlist] scripts: fix get_module_setting_root in compliance check

tejlmand:
root settings in zephyr/module.yml are in lower case and without quotes, as seen in following example:
```build:
  settings:
    board_root: .
    dts_root: .
```

remove the `upper()` call and change `_ROOT` --> `_root`. Also remove the not needed `"` stripping.

Upstream PR #: 94343


tejlmand:
need a different fix, see https://github.com/zephyrproject-rtos/zephyr/pull/94343#issuecomment-3174953409

PR #3202, opened 2025-08-15
Misc. ironside work september 3.

SebastianBoe:
Misc. ironside work september 3.

### DIFF
--- a/soc/nordic/common/uicr/sysbuild.cmake
+++ b/soc/nordic/common/uicr/sysbuild.cmake
@@ -8,4 +8,8 @@ ExternalZephyrProject_Add(
 
 # Ensure UICR is configured and built after the default image so EDT/ELFs exist.
 sysbuild_add_dependencies(CONFIGURE uicr ${DEFAULT_IMAGE})
+
 add_dependencies(uicr ${DEFAULT_IMAGE})
+if(DEFINED image)
+  add_dependencies(uicr ${image})
+endif()


soc/nordic/common/uicr/gen_uicr.py
```diff
@@ -169,13 +149,25 @@ def main() -> None:
         "--out-uicr-hex",
         required=True,
         type=argparse.FileType("w", encoding="utf-8"),
-        help="Path to write the generated UICR HEX file to",
+        help="Path to write the generated merged UICR+PERIPHCONF HEX file to (typically zephyr.hex)",
```
jonathannilsen: These artifacts are useful to have separate depending on your application, so the choice to keep them separate was deliberate. Can we keep them as two different outputs and use something like mergehex afterwards instead?

soc/nordic/common/uicr/gen_uicr.py
```diff
@@ -186,49 +178,27 @@ def main() -> None:
...
-
-        try:
-            uicr_node = edt.label2node[UICR_NODELABEL]
-        except LookupError as e:
-            raise ScriptError(
-                "Failed to find UICR node in the devicetree. "
-                f"Expected a DT node with label '{UICR_NODELABEL}'."
-            ) from e
+            uicr.PERIPHCONF.ADDRESS = args.periphconf_address
+            uicr.PERIPHCONF.MAXCOUNT = math.floor(args.periphconf_size / 8)
```
jonathannilsen: nit: noticed this earlier, but this could just be `args.periphconf_size // 8`

jonathannilsen (changes requested):

soc/nordic/common/uicr/gen_uicr.py
SebastianBoe: How are they useful? :)

SebastianBoe: It is possible to have them separate. But it would complicate the build system, and also to a certain degree the user experience.

So I would prefer to not keep them separate unless necessary.

jonathannilsen: The PERIPHCONF area in MRAM can be modified even when UICR is locked, as long as you haven't put it in the PROTECTEDMEM region. So there is a use case for pulling out just the periphconf.hex part and including it in a DFU package without the UICR part.

karstenkoenig: customers might bundle the periphconf in their ota image but not the rest of the uicr

SebastianBoe: I see, I did not think of that, will fix.

SebastianBoe: I suppose even in non-DFU use-cases, you may have locked your UICR, and which to west flash a new periphconf, but not the UICR.

I could make this configurable via KConfig.

soc/nordic/common/uicr/gen_uicr/Kconfig
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: Apache-2.0
+
+source "Kconfig.zephyr"
+
+menu "UICR generator options"
+
+config UICR_SAMPLE_GENERATE_PERIPHCONF
```
karstenkoenig: What does the sample in the name stand for?

karstenkoenig (approved):
Ah super nice, but please split into independent hex files again

soc/nordic/common/uicr/gen_uicr/Kconfig
SebastianBoe: It is trying to communicate to the reader that this Kconfig option is specific to this gen_uicr "sample".

As opposed to most other Kconfigs in ~/ncs/zephyr/soc, which are read by all zephyr images.

Maybe

GEN_UICR_SAMPLE_GENERATE_PERIPHCONF

Would be clearer?

Or do you have another idea?

karstenkoenig: sample in the sense of example? That's what I assume and find confusing, this shouldn't be example code right?
Or in sample the sense of this one unit? image or target? 

Wouldn't `GEN_UICR_GENERATE_PERIPHCONF` be enough? Or would that be confusing because it only applies to the uicr target and not actually be set in the app or radio build?

SebastianBoe: fixed

soc/nordic/common/uicr/gen_uicr.py
SebastianBoe: hex files restored

soc/nordic/common/uicr/Kconfig.sysbuild
```diff
@@ -0,0 +1,7 @@
+config NRF_HALTIUM_GENERATE_UICR
+	bool "Generate UICR file"
+	depends on SOC_SERIES_NRF54HX
+	default y
+	help
+	  Generate UICR HEX file.
+
```
nordicjm: trailing newline

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+
+project(uicr)
+
+# Function to compute partition absolute address and size from devicetree
+function(compute_partition_address_and_size partition_nodelabel output_address_var output_size_var)
+  # Get partition path from devicetree
+  dt_nodelabel(partition_path NODELABEL ${partition_nodelabel})
+  if(NOT DEFINED partition_path)
+    message(FATAL_ERROR "${partition_nodelabel} not found in devicetree")
+  endif()
```
nordicjm:
```suggestion
  dt_nodelabel(partition_path NODELABEL ${partition_nodelabel} REQUIRED)
```

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+  math(EXPR partition_address "${CONFIG_FLASH_BASE_ADDRESS} + ${partition_offset}" OUTPUT_FORMAT HEXADECIMAL)
+
+  # Set output variables in parent scope
+  set(${output_address_var} ${partition_address} PARENT_SCOPE)
+  set(${output_size_var} ${partition_size} PARENT_SCOPE)
+endfunction()
+
+# Use CMAKE_VERBOSE_MAKEFILE to silence an unused-variable warning.
+if(CMAKE_VERBOSE_MAKEFILE)
+endif()
```
nordicjm, line 49: ?

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+
+set(periphconf_args)
+set(periphconf_elfs)
+set(merged_hex_file ${APPLICATION_BINARY_DIR}/zephyr/zephyr.hex)
```
nordicjm: use the Kconfig for this for the bin name

soc/nordic/common/uicr/gen_uicr/Kconfig
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: Apache-2.0
+
+source "Kconfig.zephyr"
```
nordicjm: nit. non-blocking - might want to put this at bottom so that the application Kconfigs appear at the top

soc/nordic/common/uicr/gen_uicr/prj.conf
nordicjm, line 1: don't think prj.conf is needed and it will fallback to empty.conf? Could be wrong on that though

soc/nordic/common/uicr/sysbuild.cmake
```diff
@@ -0,0 +1,11 @@
...
+
+# Add UICR generator as a utility image
+ExternalZephyrProject_Add(
+  APPLICATION uicr
+  SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/gen_uicr
+)
+
+# Ensure UICR is configured and built after the default image so EDT/ELFs exist.
+sysbuild_add_dependencies(CONFIGURE uicr ${DEFAULT_IMAGE})
+add_dependencies(                   uicr ${DEFAULT_IMAGE})
```
nordicjm:
```suggestion
add_dependencies(uicr ${DEFAULT_IMAGE})
```

nordicjm (commented):
Looks good!

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
SebastianBoe: Nice!

Was not aware.

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
SebastianBoe: I got 5 lines of warning in the logs when this variable went unreferenced.

I assume SYSBUILD passes it to each image, and Zephyr usually references it, but since I have this minimal Zephyr it goes unreferenced.

soc/nordic/common/uicr/gen_uicr/prj.conf
SebastianBoe: I got an error without it earlier, but I'll retest.

soc/nordic/common/uicr/gen_uicr.py
```diff
@@ -186,48 +189,43 @@ def main() -> None:
...
-        uicr_hex = IntelHex()
-        uicr_hex.frombytes(bytes(uicr), offset=uicr_node.regs[0].addr)
-
+            uicr.PERIPHCONF.ADDRESS = args.periphconf_address
+            uicr.PERIPHCONF.MAXCOUNT = math.floor(args.periphconf_size / 8)
+
+            # Update UICR hex with the modified UICR data
+            uicr_data = bytes(uicr)
+            uicr_hex = IntelHex()  # Reset and rebuild with updated UICR
+            uicr_hex.frombytes(uicr_data, offset=args.uicr_address)
```
jonathannilsen: I think this might flow a bit cleaner if we populated each hex object just once, instead of populating them with initial data first and then overwriting it. For example:

```python

        periphconf_hex = IntelHex()

        if args.in_periphconf_elfs:  # Check if periphconf data is provided
            periphconf_combined = extract_and_combine_periphconfs(args.in_periphconf_elfs)

            padding_len = args.periphconf_size - len(periphconf_combined)
            periphconf_final = periphconf_combined + bytes([0xFF for _ in range(padding_len)])

            # Add periphconf data to separate and merged hex files
            periphconf_hex.frombytes(periphconf_final, offset=args.periphconf_address)

            uicr.PERIPHCONF.ENABLE = ENABLED_VALUE
            uicr.PERIPHCONF.ADDRESS = args.periphconf_address
            uicr.PERIPHCONF.MAXCOUNT = args.periphconf_size // 8

        uicr_hex = IntelHex()
        uicr_hex.frombytes(bytes(uicr), offset=args.uicr_address)

        merged_hex = IntelHex()
        merged_hex.fromdict(uicr_hex.todict())
        if periphconf_hex:
            merged_hex.fromdict(periphconf_hex.todict())

        # Write the three hex files
        merged_hex.write_hex_file(args.out_merged_hex)
        uicr_hex.write_hex_file(args.out_uicr_hex)
        if args.out_periphconf_hex:
            periphconf_hex.write_hex_file(args.out_periphconf_hex)

```
(I think IntelHex.todict() works there, but haven't tried running it)

soc/nordic/common/uicr/gen_uicr.py
```diff
@@ -186,48 +189,43 @@ def main() -> None:
...
 
-        try:
-            periphconf_partition = edt.label2node[PERIPHCONF_NODELABEL]
-        except LookupError as e:
-            raise ScriptError(
-                "Failed to find a PERIPHCONF partition in the devicetree. "
-                f"Expected a DT node with label '{PERIPHCONF_NODELABEL}'."
-            ) from e
+        if args.in_periphconf_elfs:  # Check if periphconf data is provided
+            periphconf_combined = extract_and_combine_periphconfs(args.in_periphconf_elfs)
```
jonathannilsen: I think we should retain the ability to generate an empty (0xFFFF_FFFFs) periphconf blob if no elf is provided here - an empty periphconf is still valid

soc/nordic/common/uicr/gen_uicr.py
```diff
@@ -165,17 +145,40 @@ def main() -> None:
...
+        default=None,
+        type=lambda s: int(s, 0),
+        help="Absolute flash address of the PERIPHCONF partition (decimal or 0x-prefixed hex)",
+    )
+    parser.add_argument(
+        "--periphconf-size",
+        default=None,
+        type=lambda s: int(s, 0),
+        help="Size in bytes of the PERIPHCONF partition (decimal or 0x-prefixed hex)",
+    )
```
jonathannilsen, line 251: nit: These are required whenever --out-periphconf-hex is used. Should we add a check below here and raise an error if they are not given alongside that? Not that this is a script that is expected to be called by the user directly, but you might end up with some strange python exceptions if for example the CMake code fails to pass these arguments and they end up as None

soc/nordic/common/uicr/gen_uicr.py
SebastianBoe: How about we error-out when it's not divisible by 8?

soc/nordic/common/uicr/gen_uicr/prj.conf
SebastianBoe: I get this error without it:

```
-- Board: nrf54h20dk, Revision: 0.9.0, qualifiers: nrf54h20/cpurad
CMake Error at /home/sebo/ncs/zephyr/cmake/modules/extensions.cmake:2919 (message):
  No prj.conf file(s) was found in the
  /home/sebo/ncs/zephyr/soc/nordic/common/uicr/gen_uicr folder(s), please
  read the Zephyr documentation on application development.
  /home/sebo/ncs/zephyr/cmake/modules/configuration_files.cmake:40 (zephyr_file)
  /home/sebo/ncs/zephyr/cmake/modules/zephyr_default.cmake:131 (include)
  /home/sebo/ncs/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:75 (include)
  /home/sebo/ncs/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
  CMakeLists.txt:15 (find_package)


-- Configuring incomplete, errors occurred!
CMake Error at cmake/modules/sysbuild_extensions.cmake:530 (message):
  CMake configure failed for Zephyr project: uicr

  Location: /home/sebo/ncs/zephyr/soc/nordic/common/uicr/gen_uicr
  cmake/modules/sysbuild_images.cmake:43 (ExternalZephyrProject_Cmake)
  cmake/modules/sysbuild_default.cmake:21 (include)
  /home/sebo/ncs/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:75 (include)
  /home/sebo/ncs/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
  /home/sebo/ncs/zephyr/share/sysbuild-package/cmake/SysbuildConfig.cmake:8 (include)
  template/CMakeLists.txt:10 (find_package)


-- Configuring incomplete, errors occurred!

```

soc/nordic/common/uicr/gen_uicr.py
SebastianBoe: fixed

soc/nordic/common/uicr/gen_uicr.py
SebastianBoe: fixed

soc/nordic/common/uicr/gen_uicr.py
SebastianBoe: fixed

SebastianBoe: much better!

soc/nordic/common/uicr/sysbuild.cmake
SebastianBoe: fixed

soc/nordic/common/uicr/gen_uicr/Kconfig
SebastianBoe: fixed

soc/nordic/common/uicr/gen_uicr/prj.conf
nordicjm: keep the empty file, maybe something to look at in future @tejlmand

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
SebastianBoe: fixed

I couldn't find neither a variable nor a Kconfig that contained the hex suffix unfortunately.

soc/nordic/common/uicr/Kconfig.sysbuild
SebastianBoe: fixed

SebastianBoe:
@karstenkoenig , @jonathannilsen , @nordicjm :

All feedback addressed, will reupload to upstream zephyr soon.

soc/nordic/common/uicr/gen_uicr/zephyr/CMakeLists.txt
jonathannilsen, line 1: Should there be a license header in this file?

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+  foreach(_dir ${_siblings})
+    get_filename_component(_name ${_dir} NAME)
+    if(_name STREQUAL "uicr")
+      # This image is an exception to the rule. It has a zephyr.dts, but
+      # no zephyr.elf
+      continue()
+    endif()
+
+    if(EXISTS ${_dir}/zephyr/zephyr.dts)
+      list(APPEND periphconf_elfs ${_dir}/zephyr/zephyr.elf)
```
jonathannilsen: I guess this should also take into account whether KERNEL_BIN_NAME has been changed in the image it's getting from? Not sure how that is best handled

jonathannilsen (approved):

soc/nordic/common/uicr/gen_uicr/CMakeLists.txt
SebastianBoe: I don't expect this to ever be anything but zephyr

soc/nordic/common/uicr/gen_uicr/zephyr/CMakeLists.txt
SebastianBoe: fixed

SebastianBoe:
These patches are now up for review upstream in https://github.com/zephyrproject-rtos/zephyr/pull/94796

nordicjm (approved):

jonathannilsen (approved):

jonathannilsen:
Superseded by #3276

PR #3207, opened 2025-08-18
[nrf noup] tests: jwt: Add crypto configuration

Vge0rge:
Add the required crypto configuration for the nrf_security to work with this sample. This is a noup because of two reasons: 1) There is a downstream symbol
2) The nrf_security defaults deviate from the upstream ones
   regarding heap


Vge0rge:
> I don't like this very much, can't we just add these options upstream?

Well, that needs to contribute the changes to the mbedlts and zephyr project. I don't think that this is on scope here.

tomi-font:
> Well, that needs to contribute the changes to the mbedlts and zephyr project. I don't think that this is on scope here.

No, I don't quite mean to change how the configuration system works. Just to (more or less) literally make the same additions in Zephyr so that we don't need a noup. Even if they're not strictly required in Zephyr it can be okay to have extra config options in tests. I've already successfully done that.

Vge0rge:
> > Well, that needs to contribute the changes to the mbedlts and zephyr project. I don't think that this is on scope here.
> 
> No, I don't quite mean to change how the configuration system works. Just to (more or less) literally make the same additions in Zephyr so that we don't need a noup. Even if they're not strictly required in Zephyr it can be okay to have extra config options in tests. I've already successfully done that.

But what I am saying here is that one Kconfig is only downstream, it does not exist in Zephyr or Mbedlts. So this cannot be upstreamed because it will break the break the Zephyr CI since the Kconfig doesn't exist, right?

tomi-font:
> But what I am saying here is that one Kconfig is only downstream, it does not exist in Zephyr or Mbedlts. So this cannot be upstreamed because it will break the break the Zephyr CI since the Kconfig doesn't exist, right?

ah :hot_face: annoying

Vge0rge:
No agreement was reached on how to resolve this so this will get abandoned for now.

PR #3211, opened 2025-08-19
drivers: adc: nrfx: synchronize upstream changes

jaz1-nordic:
Rollback of obsolete commits and replacing them with correct ones already merged in upstream.


nordic-segl (approved):

drivers/adc/adc_nrfx_saadc.c
```diff
@@ -575,8 +575,8 @@ static void correct_single_ended(const struct adc_sequence *sequence, nrf_saadc_
 				} else if (*sample < 0) {
 					*sample = 0;
 				}
+				sample++;
```
anangl: This fix is incorrect. Please see the related comment in the upstream PR.

PR #3233, opened 2025-08-28
[nrf noup] soc: nrf54h: work around missing power domain handling

jonathannilsen:
The upcoming release of IronSide SE no longer disables RETAIN
in all GPIO instances on boot, so the application must be able
to handle the hardware default state of RETAIN being enabled.

The GPIO retention is properly handled by changes that are
currently only upstream and will be pulled in by the next upmerge.
This patch exists a workaround to be able to integrate
IronSide SE before the proper solution is pulled in.

### DIFF
--- a/soc/nordic/nrf54h/Kconfig
+++ b/soc/nordic/nrf54h/Kconfig
@@ -95,6 +95,11 @@ config SOC_NRF54H20_CPURAD_ENABLE_CHECK_VTOR
 	help
 	  Verify that VTOR is not 0xFFFFFFFF before booting the Radiocore.
 
+config SOC_NRF54H20_DISABLE_ALL_GPIO_RETENTION_WORKAROUND
+	bool "Disable all GPIO pin retention on boot"
+	default y
+	depends on SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD
+
 config SOC_NRF54H20_CPURAD
 	select SOC_NRF54H20_CPURAD_COMMON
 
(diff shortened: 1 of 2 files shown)


57300 (approved):
Could you link the upstream changes for reference?

I'm wondering if it would complicate things too much if we just cherry-picked them here.

jonathannilsen:
> Could you link the upstream changes for reference?
> 
> I'm wondering if it would complicate things too much if we just cherry-picked them here.

If I'm not mistaken we are missing this, which is only on the 3.1 branch at the moment: #3175 
I believe https://github.com/zephyrproject-rtos/zephyr/pull/90754 is the critical part there with the gpio pad driver for handling retention for the slow pins, no idea if any other changes are strictly needed for this.

I don't have much context other than that, @bjarki-andreasen?

karstenkoenig (approved):

bjarki-andreasen:
> > Could you link the upstream changes for reference?
> > I'm wondering if it would complicate things too much if we just cherry-picked them here.
> 
> If I'm not mistaken we are missing this, which is only on the 3.1 branch at the moment: #3175 I believe [zephyrproject-rtos/zephyr#90754](https://github.com/zephyrproject-rtos/zephyr/pull/90754) is the critical part there with the gpio pad driver for handling retention for the slow pins, no idea if any other changes are strictly needed for this.
> 
> I don't have much context other than that, @bjarki-andreasen?

The proper solution upstream is to enable power domains for the H20 everywhere, which just means not disabling them in upstream MCUBoot. Downstream for now, since we don't have [zephyrproject-rtos/zephyr#90754](https://github.com/zephyrproject-rtos/zephyr/pull/90754), this noup is needed as the power domain drivers which would disable retention simply are not there yet :)

PR #3243, opened 2025-09-02
[nrf noup] samples: smp_svr: add nrf54lm20dk

michalek-no:
adds nrf54lm20dk to sample.yaml


nordicjm (changes requested):
add to https://github.com/nrfconnect/sdk-nrf/tree/main/samples/zephyr/subsys/mgmt/mcumgr/smp_svr

PR #3252, opened 2025-09-05
nRF54h20 pm s2ram resume redirect (bootloader -> App)

nvlsianpu:
Added infrastructure needed for redirect execution to the Application reset vector in case SoC gets woke-up from S2RAM and MCUboot is the first code entity. Thanks to that bootloader can implement its own routine for resume support (CONFIG_SOC_NRF54H20_PM_S2RAM_OVERRIDE) which overrides `zephyr/soc/nordic/nrf54h/pm_s2ram.c` implementation in case of MCUboot.

Introduce additional source of true using additional check over independent source (variable with magic value).

### DIFF
--- a/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
+++ b/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
@@ -343,12 +343,19 @@ zephyr_udc0: &usbhs {
 
 /* Trim this RAM block for making room on all run-time common S2RAM cpu context. */
 &cpuapp_ram0 {
-	reg = <0x22000000 (DT_SIZE_K(32)-32)>;
-	ranges = <0x0 0x22000000 (0x8000-0x20)>;
+	reg = <0x22000000 (DT_SIZE_K(32) - 36)>;
+	ranges = <0x0 0x22000000 (0x8000 - 0x24)>;
 };
 
 / {
 	soc {
+		/* run-time common mcuboot S2RAM support section */
+		mcuboot_s2ram: cpuapp_s2ram@22007fdc  {
+			compatible = "zephyr,memory-region", "mmio-sram";
+			reg = <0x22007fdc 4>;
+			zephyr,memory-region = "mcuboot_s2ram_context";
+		};
+
 		/* run-time common S2RAM cpu context RAM */
 		pm_s2ram: cpuapp_s2ram@22007fe0  {
 			compatible = "zephyr,memory-region", "mmio-sram";
--- a/soc/nordic/nrf54h/Kconfig
+++ b/soc/nordic/nrf54h/Kconfig
@@ -100,6 +100,11 @@ config SOC_NRF54H20_DISABLE_ALL_GPIO_RETENTION_WORKAROUND
 	default y
 	depends on SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD
 
+config SOC_NRF54H20_PM_S2RAM_OVERRIDE
+	bool "Override `pm_s2ram` implementation"
+	help
+	  Override Nordic s2ram implementation.
+
 config SOC_NRF54H20_CPURAD
 	select SOC_NRF54H20_CPURAD_COMMON
 
--- a/soc/nordic/nrf54h/CMakeLists.txt
+++ b/soc/nordic/nrf54h/CMakeLists.txt
@@ -8,7 +8,9 @@ if(CONFIG_ARM)
   endif()
 endif()
 
-zephyr_library_sources_ifdef(CONFIG_PM_S2RAM pm_s2ram.c)
+if(NOT CONFIG_SOC_NRF54H20_PM_S2RAM_OVERRIDE)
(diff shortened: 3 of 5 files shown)


pizi-nordic (approved):

tomchy (approved):

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
```diff
@@ -343,12 +343,19 @@ zephyr_udc0: &usbhs {
 
 /* Trim this RAM block for making room on all run-time common S2RAM cpu context. */
 &cpuapp_ram0 {
-	reg = <0x22000000 (DT_SIZE_K(32)-32)>;
-	ranges = <0x0 0x22000000 (0x8000-0x20)>;
+	reg = <0x22000000 (DT_SIZE_K(32)-36)>;
```
tomchy: Note that in upstream, there has been a reformatting commit that added spaces around `-`..

nordic-bami (approved):

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
nvlsianpu: Will be covered elswhere,

nordicjm:
why can this upstream fix not be used https://github.com/zephyrproject-rtos/zephyr/pull/95914 and why does this need noups?

nvlsianpu:
@nordicjm The upstream PR you mentioned is bugfix for another S2RAm related  Issue.
This PR implements new scheme of resume form S2RAM sleep.

soc/nordic/nrf54h/Kconfig
```diff
@@ -100,6 +100,13 @@ config SOC_NRF54H20_DISABLE_ALL_GPIO_RETENTION_WORKAROUND
 	default y
 	depends on SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD
 
+config SOC_NRF54H20_PM_S2RAM_OVERRIDE
+	bool "Override `pm_s2ram` implementation"
+	default n
+	help
+	  Override Nordic s2ram implementation.
+
```
nordicjm, line 107:
```suggestion
```

soc/nordic/nrf54h/Kconfig
```diff
@@ -100,6 +100,13 @@ config SOC_NRF54H20_DISABLE_ALL_GPIO_RETENTION_WORKAROUND
 	default y
 	depends on SOC_NRF54H20_CPUAPP || SOC_NRF54H20_CPURAD
 
+config SOC_NRF54H20_PM_S2RAM_OVERRIDE
+	bool "Override `pm_s2ram` implementation"
+	default n
```
nordicjm:
```suggestion
```

nordicjm (commented):
This can be moved into sdk-nrf rather than being a noup, the original file can be marked as a header only file which means it will not be compiled - see https://github.com/nrfconnect/sdk-nrf/blob/main/tests/subsys/net/lib/nrf_cloud/cloud/CMakeLists.txt#L40 for an example of how to do that, you can then substitute in a modified file to be compiled in place of it

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
nvlsianpu: done

nordicjm (approved):

nordicjm:
in future when opening PRs to sdk-mcuboot/sdk-zephyr and you are creating your own manifest PR then put `manifest-pr-skip` in the body so it doesn't automatically create a manifest that uses CI resources

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
```diff
@@ -343,12 +343,19 @@ zephyr_udc0: &usbhs {
...
-	reg = <0x22000000 (DT_SIZE_K(32)-32)>;
-	ranges = <0x0 0x22000000 (0x8000-0x20)>;
+	reg = <0x22000000 (DT_SIZE_K(32) - 36)>;
+	ranges = <0x0 0x22000000 (0x8000 - 0x24)>;
 };
 
 / {
 	soc {
+		/* run-time common mcuboot S2RAM support section */
+		mcuboot_s2ram: cpuapp_s2ram@22007fdc  {
```
nordicjm, line 353: having a look at the sdk-mcuboot PR this is basically completely duplicating retention's boot mode feature, any reason that wasn't used given it already has checksum/header validity and is already supported by mcuboot/doesn't need duplicate maintenance?

nvlsianpu: Reason are that it is not another boot. From application execution it is perceived as resume from very deep sleep.  Operation must be quick and low level. It must use limited resources. This section is for anty-glitch magic value recommended by Architect group. It is used for double check decision.

PR #3257, opened 2025-09-09
scripts: Remove SUIT-specific scripts

tomchy:
Remove SUIT west tool extentions.


tomchy:
Already remove here: https://github.com/zephyrproject-rtos/zephyr/pull/94527

PR #3265, opened 2025-09-10
Nrfx 8369 align sdk zephyr to bsp

magp-nordic:


mib1-nordic (approved):

modules/hal_nordic/nrfx/CMakeLists.txt
```diff
@@ -11,16 +11,33 @@ if(NOT DEFINED NRFX_DIR)
...
+set(BSP_DIR ${NRFX_DIR}/bsp)
 set(HELPERS_DIR ${NRFX_DIR}/helpers)
+set(SOC_DIR ${BSP_DIR}/${CONFIG_SOC_NORDIC_BSP_NAME})
+set(MDK_DIR ${SOC_DIR}/mdk)
+
+zephyr_compile_definitions(NRFX_BSP_NRF_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/mdk/nrf.h")
+zephyr_compile_definitions(NRFX_BSP_ERRATAS_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/mdk/nrf_erratas.h")
+zephyr_compile_definitions(NRFX_BSP_SOC_IRQS_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/soc/nrfx_irqs.h")
+zephyr_compile_definitions(NRFX_BSP_NRFX_EXT_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/nrfx_ext.h")
+zephyr_compile_definitions(NRFX_BSP_NRFX_COREDEP_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/soc/nrfx_coredep_defs.h")
```
tejlmand: no, we should not start passing a bunch of compiler definitions to all source files.

This means even `main.c` is now compiled with:
```
$ /opt/zephyr-sdk-0.16.9/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc ...
 -DNRFX_BSP_ERRATAS_PATH=\"stable/mdk/nrf_erratas.h\"
 -DNRFX_BSP_NRFX_COREDEP_PATH=\"stable/soc/nrfx_coredep_defs.h\"
 -DNRFX_BSP_NRFX_EXT_PATH=\"stable/nrfx_ext.h\"
 -DNRFX_BSP_NRF_PATH=\"stable/mdk/nrf.h\"
 -DNRFX_BSP_SOC_IRQS_PATH=\"stable/soc/nrfx_irqs.h\"
  ... -c .../samples/hello_world/src/main.c
```

not to mention the fact that if you change the value of a compile definition, then all sources will rebuild on an incremental build because they now all have the compile definition applied.

masz-nordic: This should not be an issue for targets in-tree, those defines have defaults that evaluate to `stable` path.
@magp-nordic can you add condition that does not set them if `CONFIG_SOC_NORDIC_BSP_NAME` is `stable`?

For out of tree targets, those defines will be needed and that is a design decision. Since development on those targets will be limited, it should be acceptable for that audience to live with a few additional defines.

Incremental rebuild should not be an issue since those Kconfigs are set once and never changed.

magp-nordic: Added

PR #3272, opened 2025-09-12
[nrf noup] dts: choose a psa-rng for entropy for 54lm20a

ppryga-nordic:
Set PSA as the entropy source for nRF54lm20a target. PSA is the only NCS-supported interface to CRACEN. There is no other entropy source in 54lm20a than CRACEN.

The commit also disables `rng` compatible with
`nrf-cracen-ctrdrbg`, the nrfx based interface to CRACEN that is used in upstream Zephyr. Only one CRACEN interface may be enabled.

### DIFF
--- a/dts/arm/nordic/nrf54l_05_10_15_cpuapp.dtsi
+++ b/dts/arm/nordic/nrf54l_05_10_15_cpuapp.dtsi
@@ -26,7 +26,7 @@ nvic: &cpuapp_nvic {};
 	};
 
 	rng: rng {
-		status = "okay";
+		status = "disabled";
 		compatible = "nordic,nrf-cracen-ctrdrbg";
 	};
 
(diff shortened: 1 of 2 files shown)


hermabe:
Do we need the same change for nrf54LV10?

ppryga-nordic:
> Do we need the same change for nrf54LV10?

As of now there is not nRF54LV10 board in sdk-zephyr. It is available here: https://github.com/nrfconnect/sdk-nrf/blob/fe3d5953976bb1272df2178e15280c64025c3912/dts/arm/nordic/nrf54lv10a_enga_cpuapp.dtsi#L21. 

I hope when the nRF54LV10 is moved to sdk-zephyr a team responsible for that makes all changes correctly including entropy source configuration.

PR #3275, opened 2025-09-12
[nrf fromlist] arch: Allow to specify memory for S2RAM resume

tomchy:
If the area, dedicated for the interrupt stack is not available, allow to specify a memory region that will be used as the stack for the S2RAM resume logic.

Upstream PR #: 95914

manifest-pr-skip

ref.: NCSDK-35410

### DIFF
--- a/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
+++ b/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
@@ -343,17 +343,24 @@ zephyr_udc0: &usbhs {
 
 /* Trim this RAM block for making room on all run-time common S2RAM cpu context. */
 &cpuapp_ram0 {
-	reg = <0x22000000 (DT_SIZE_K(32) - 48)>;
-	ranges = <0x0 0x22000000 (0x8000 - 0x30)>;
+	reg = <0x22000000 (DT_SIZE_K(32) - 52)>;
+	ranges = <0x0 0x22000000 (0x8000 - 0x34)>;
 };
 
 / {
 	soc {
 		/* temporary stack for S2RAM resume logic */
-		pm_s2ram_stack: cpuapp_s2ram_stack@22007fd0  {
+		pm_s2ram_stack: cpuapp_s2ram_stack@22007fcc  {
 			compatible = "zephyr,memory-region", "mmio-sram";
-			reg = <0x22007fd0 16>;
+			reg = <0x22007fcc 16>;
 			zephyr,memory-region = "pm_s2ram_stack";
+		};
+
+		/* run-time common mcuboot S2RAM support section */
+		mcuboot_s2ram: cpuapp_s2ram@22007fdc  {
+			compatible = "zephyr,memory-region", "mmio-sram";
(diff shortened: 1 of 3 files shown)


kl-cruz (approved):

michalek-no (approved):

kapi-no (approved):

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
```diff
@@ -343,8 +343,8 @@ zephyr_udc0: &usbhs {
 
 /* Trim this RAM block for making room on all run-time common S2RAM cpu context. */
 &cpuapp_ram0 {
-	reg = <0x22000000 (DT_SIZE_K(32)-32)>;
```
nordicjm: temp. block until input from @carlescufi - this places random burdens on VV commit 79bffff39afb384061a6cb0a543bb2b8bf9b762d:
```
[nrf noup] boards: Align board DTS with upstream.

This alignment cannot be done through nrf fromlist commit, since it was
introduced in the tree-wide commit:
2d22884f52a01236d404fb2e6910d3683f680af6
Once tried to pull all changes to merge it cleanly, the chain ended up
in the doc/releases/migration-guide-4.3.rst file, which would bring
nearly all of the changes from the Zephyr 4.3 release, making a small PR
with just 4 white characters a Zephyr upmerge PR.
Feel free to drop this commit during the next Zephyr upmerge.
```

nordicjm (changes requested):

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20_cpuapp.dts
carlescufi: @tomchy can't you wait until the current upmerge is complete?

tomchy: No, we need it for the release that will be tomorrow.

tomchy: Keep in mind that it is a separate commit that adds 4 spaces and should be more than trivial to handle during the upmerge (either way - noup commits create conflicts).

nvlsianpu: @nordicjm - I don't understand why you block this PR. It is bugfix. Its urgent, needed for preview tag tomorrow. There is ref. to internal ticket even you can check.

nordicjm: This places the blame of going through noup's on whomever is doing the upmerge, we have over 100 noup commits in sdk-zephyr, it is not nor should it be their responsibility to go through these 100+ commits, read their descriptions, apply possible action and then try to get them to apply. If things are needed then the required commits should be cherry-picked with them, if that takes 1 or 30 commits isn't VV's problem, it's on the person submitting to deal with and bring those in. An exception is being made for this due to urgency, future usage will be blocked and users will be asked to fromlist/fromtree cherry pick the required commits for a commit to be cleanly applied

PR #3287, opened 2025-09-19
[nrf fromlist] cmake: Use temp. for edt pickle CMake output

nordicjm:
This potentially fixes an issues whereby when sysbuild projects import dts configuration from images and images are reconfigured, that sysbuild will needlessly rerun even if the output has not changed by using a temporary output file and then only updating the actual file if the contents have changed

Upstream PR #: 96259


Vge0rge:
I tested this with a multi image build with mcuboot + tfm + app which was always rebuilding everything the second time that you invoked a build even without any changes. This fixes it and it does not rebuild anymore.

57300 (approved):

PR #3288, opened 2025-09-19
Align nRF errata handling to nrfx 4.0

nomo-nordic:


drivers/pwm/pwm_nrfx.c
```diff
@@ -22,10 +22,10 @@
 
 LOG_MODULE_REGISTER(pwm_nrfx, CONFIG_PWM_LOG_LEVEL);
 
-/* NRFX_PWM_NRF52_ANOMALY_109_WORKAROUND_ENABLED can be undefined or defined
+/* NRF52_ERRATA_109_ENABLE_WORKAROUND can be undefined or defined
```
masz-nordic: The undefined scenario is now not possible, right?

nomo-nordic: As a last resort, MDK will define it, so yes.

nomo-nordic: Do we want to use the new `NRF_ERRATA_STATIC_CHECK` macros here? (They work)

nomo-nordic: Also, do we want to align other files to STATIC/DYNAMIC macro checks?

masz-nordic:
Replaced with https://github.com/nrfconnect/sdk-zephyr/pull/3301

PR #3291, opened 2025-09-19
drivers: clock_control: Separated nrf hfclk shim from nrf clock shim.

mif1-nordic:
Separated clock_control_nrf_hfclk shim from clock_control_nrf shim.

### DIFF
--- a/dts/arm/nordic/nrf52833.dtsi
+++ b/dts/arm/nordic/nrf52833.dtsi
@@ -69,6 +69,13 @@
 			status = "okay";
 		};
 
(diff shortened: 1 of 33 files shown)


dts/arm/nordic/nrf52833.dtsi
```diff
@@ -69,6 +69,13 @@
 			status = "okay";
 		};
 
+	hfclk: hfclk@40000000 {
```
magp-nordic, line 72: Nit: alignment. Also in other files.

PR #3301, opened 2025-09-24
soc: nordic: remove nRF52 Series anomaly 58 workaround

nomo-nordic:


soc/nordic/nrf52/CMakeLists.txt
```diff
@@ -3,8 +3,8 @@
 zephyr_library_sources(soc.c)
 zephyr_include_directories(.)
 
-if(CONFIG_SOC_NRF52832_ALLOW_SPIM_DESPITE_PAN_58 AND CONFIG_SPI_NRFX_SPIM)
-  message(WARNING "Both SOC_NRF52832_ALLOW_SPIM_DESPITE_PAN_58 and an NRF SPIM driver are enabled, therefore PAN 58 will apply if RXD.MAXCNT == 1 and TXD.MAXCNT <= 1")
+if(CONFIG_NRF52_ANOMALY_58_WORKAROUND AND CONFIG_SPI_NRFX_SPIM)
+  message(WARNING "Both CONFIG_NRF52_ANOMALY_58_WORKAROUND and an NRF SPIM driver are enabled, therefore PAN 58 will apply if RXD.MAXCNT == 1 and TXD.MAXCNT <= 1")
```
anangl: This doesn't really make sense. The warning that anomaly 58 can occur should be issued when the workaround is NOT enabled.

samples/drivers/led/led_strip/boards/nrf52dk_nrf52832.conf
```diff
@@ -1,3 +1,3 @@
 # This driver only uses spi_write() with the SPIM instance it allocates,
 # so PAN 58 doesn't matter, because the RX length is always 0.
-CONFIG_SOC_NRF52832_ALLOW_SPIM_DESPITE_PAN_58=y
+CONFIG_NRF52_ANOMALY_58_WORKAROUND=y
```
anangl:
```suggestion
CONFIG_NRF52_ANOMALY_58_WORKAROUND=n
```
or maybe better remove this file altogether.

soc/nordic/nrf52/CMakeLists.txt
masz-nordic: Previous logic was:
- If `ALLOW_DESPITE_PAN` is not enabled, the driver will not build.
- If it is enabled, it will apply workaround and build but print this warning.

My understanding is that changes in this file preserve behaviour from second bullet point (if `WORKAROUND` is enabled).
What is not preserved is the first bullet point, but I am not sure if that is necessary.

samples/drivers/led/led_strip/boards/nrf52dk_nrf52832.conf
masz-nordic: Removed

soc/nordic/nrf52/CMakeLists.txt
anangl: > Previous logic was:
> 
> * If `ALLOW_DESPITE_PAN` is not enabled, the driver will not build.
> * If it is enabled, it will apply workaround and build but print this warning.

The workaround was applied only for those instances that had the `anomaly-58-workaround` property set in dts. Hence the warning that for the others the anomaly would occur.
Now the workaround, if enabled, is used for all instances, so the warning in its current form is misleading.

masz-nordic: How about removing it completely as it is now handled in a standard way and users should be aware of anomalies?

anangl: Sounds good. The workaround is enabled by default, so users deciding to disable it should know what they're doing.

masz-nordic: Removed

mstasiaknordic:
Adding commits missing from main to fix failing tests

PR #3319, opened 2025-09-29
[nrf fromlist] drivers: audio: dmic_nrfx_pdm: introduce prescaler cal…

mstasiaknordic:
…culation

Added usage of new API responsible for calculating PDM prescaler values.

Upstream PR #: 96703

### DIFF
--- a/drivers/audio/dmic_nrfx_pdm.c
+++ b/drivers/audio/dmic_nrfx_pdm.c
@@ -21,7 +21,6 @@ LOG_MODULE_REGISTER(dmic_nrfx_pdm, CONFIG_AUDIO_DMIC_LOG_LEVEL);
 
 #if CONFIG_SOC_SERIES_NRF54HX
 #define DMIC_NRFX_CLOCK_FREQ MHZ(16)
-#define DMIC_NRFX_CLOCK_FACTOR 8192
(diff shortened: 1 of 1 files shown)


drivers/audio/dmic_nrfx_pdm.c
```diff
@@ -499,11 +283,23 @@ static int dmic_nrfx_pdm_configure(const struct device *dev,
...
+		.base_clock_freq = (NRF_PDM_HAS_SELECTABLE_CLOCK && drv_cfg->clk_src == ACLK)
+					? DMIC_NRFX_AUDIO_CLOCK_FREQ
+					: DMIC_NRFX_CLOCK_FREQ,
+		.sampling_rate = config->streams[0].pcm_rate,
+		.output_freq_min = config->io.min_pdm_clk_freq,
+		.output_freq_max = config->io.max_pdm_clk_freq
+	};
+	nrfx_pdm_prescalers_t prescalers;
+
+	if (nrfx_pdm_prescalers_calc(&output_config, &prescalers) != NRFX_SUCCESS) {
```
anangl: Is there an advantage of using another local variable for prescalers? Why not just:

```suggestion
	if (nrfx_pdm_prescalers_calc(&output_config, &nrfx_cfg.prescalers) != NRFX_SUCCESS) {
```

mstasiaknordic: You are right, there is none

PR #3323, opened 2025-09-29
[nrf noup] bluetooth: att: Fix `CONFIG_BT_ATT_SENT_CB_AFTER_TX`

MarekPieta:
The `delay ATT sent callback until data transmission is done by controller` feature introduces an extra `net_buf` reference which is removed on sent callback. Make sure to remove the reference on BLE disconnection to free the buffer as the registered sent callback will not be called.

Jira: NCSDK-35650

### DIFF
--- a/subsys/bluetooth/host/l2cap.c
+++ b/subsys/bluetooth/host/l2cap.c
@@ -273,6 +273,21 @@ static void l2cap_chan_del(struct bt_l2cap_chan *chan)
 	 * `l2cap_chan_destroy()` as it is not called for fixed channels.
 	 */
 	while ((buf = k_fifo_get(&le_chan->tx_queue, K_NO_WAIT))) {
+		bt_conn_tx_cb_t cb = closure_cb(buf->user_data);
+
+		if (cb) {
+			void *user_data = closure_data(buf->user_data);
+
+			/* When bt_l2cap_send_pdu() succeeds, the stack takes ownership
+			 * and must invoke the callback eventually. Since these PDUs will
+			 * never be transmitted, invoke the callback now with an error.
+			 * Note: We cannot use conn_tx_destroy() here because no bt_conn_tx
+			 * struct has been allocated yet - the closure is still in the
+			 * buf->user_data.
+			 */
+			cb(chan->conn, user_data, -ESHUTDOWN);
+		}
+
 		net_buf_unref(buf);
 	}
 
--- a/subsys/bluetooth/host/l2cap.c
+++ b/subsys/bluetooth/host/l2cap.c
@@ -761,13 +761,19 @@ int bt_l2cap_send_pdu(struct bt_l2cap_le_chan *le_chan, struct net_buf *pdu,
 		return -ENOTCONN;
 	}
 
-	if (pdu->ref != 1) {
+	/* If ATT sent callback is delayed until data transmission is done by BLE controller
+	 * (CONFIG_BT_ATT_SENT_CB_AFTER_TX), the `chan_send` function from `att.c` introduces an
+	 * additional reference. The reference is used to extend lifetime of the net buffer until
+	 * the data transmission is confirmed by ACK of the remote (the reference is removed when
+	 * the TX callback passed to `bt_l2cap_send_pdu` is called).
+	 */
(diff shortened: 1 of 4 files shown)


MarekPieta:
Pure rebase

kapi-no (approved):

subsys/bluetooth/host/l2cap.c
```diff
@@ -273,6 +273,24 @@ static void l2cap_chan_del(struct bt_l2cap_chan *chan)
...
+		 * The TX callbacks of the buffers removed here are not called, therefore the
+		 * additional reference needs to be removed here.
+		 */
+		if (IS_ENABLED(CONFIG_BT_ATT_SENT_CB_AFTER_TX)) {
+			__ASSERT_NO_MSG(buf->ref <= 2);
+
+			if (buf->ref == 2) {
+				net_buf_unref(buf);
+			}
+		}
```
alwa-nordic: Detecting ATT PDUs by checking buf->ref == 2 is fragile because future changes might introduce other scenarios with extra references for different reasons. I'm not even sure that it's not the case today. Instead of using the ref count as a proxy, I suggest comparing the cb:

```c
if (IS_ENABLED(CONFIG_BT_ATT_SENT_CB_AFTER_TX)) {
    bt_conn_tx_cb_t cb = closure_cb(buf->user_data);
    
    if (cb == chan_sent_cb) {
        void *user_data = closure_data(buf->user_data);
        /* Release the extra reference attached to user_data */
        cb(chan->conn, user_data, -ESHUTDOWN);
    }
}
```

But I realized it's a bug that we don't invoke the callbacks here. We should be invoking all the callbacks here just like in `process_unack_tx()`. I made a PR upstream to fix this just now:

https://github.com/zephyrproject-rtos/zephyr/pull/97056

I suggest we cherry pick that general fix instead. Can you test that it works for you when it passes the upstream CI?

alwa-nordic (changes requested):

subsys/bluetooth/host/l2cap.c
MarekPieta: The `bt_l2cap_send_pdu` and `bt_l2cap_dyn_chan_send` both seem to demand that `ref` counter is `1` for buffers placed on the TX queue. Also it seems that the `net_buf_unref` here is expected to remove the last reference (please correct me if I overlooked something).

Using the `closure_cb` as in the code snippet above would require to pass the `chan_sent_cb` function (through header or using `extern` keyword). The function would no longer be static for the `att.c` file. I tried to avoid that as it might not be a clean solution.

Invoking the sent callback here might help. I will test your commit locally and let you know if it works.

alwa-nordic: > The `bt_l2cap_send_pdu` and `bt_l2cap_dyn_chan_send` both seem to demand that `ref` counter is `1` for buffers placed on the TX queue. Also it seems that the `net_buf_unref` here is expected to remove the last reference (please correct me if I overlooked something).

That's just to protect against user error. We did not foresee that a stack-internal user would keep an extra reference and wait for the callback. We just saw no use for the extra reference and it is indicative of a leaked reference. The stack does not rely on this fact AFAIK. We can remove all these checks. Maybe we should.

> Using the `closure_cb` as in the code snippet above would require to pass the `chan_sent_cb` function (through header or using `extern` keyword). The function would no longer be static for the `att.c` file. I tried to avoid that as it might not be a clean solution.

I prefer that. We are referring to ATT anyway (to explain the code), so including the header is just being explicit about the dependency.

> Invoking the sent callback here might help. I will test your commit locally and let you know if it works.

Thanks! It passed CI checks. Invoking the sent callback is a generalization of my code snippet here, so I'm sure it will work. Let's just look out for unintentional breakage elsewhere.

MarekPieta: I confirmed your fix from the upstream PR. As soon as the upstream PR is merged, I will update this PR.

MarekPieta:
Cherry-picked the https://github.com/zephyrproject-rtos/zephyr/pull/97056 to trigger NCS CI tests

alwa-nordic (approved):

MarekPieta:
Pure rebase

kapi-no (approved):

PR #3329, opened 2025-09-30
[nrf fromlist] drivers: watchdog: wdt_nrfx: use standard instantiation

mstasiaknordic:
Used API for standard instantiation and replaced nrfx_err_t error values with errno.

Removed Kconfig symbols for each WDT instance which are no longer needed.


drivers/watchdog/wdt_nrfx.c
```diff
@@ -138,11 +134,11 @@ static int wdt_nrf_install_timeout(const struct device *dev,
...
 	}
 
-	err_code = nrfx_wdt_channel_alloc(&config->wdt,
+	err_code = nrfx_wdt_channel_alloc(&data->wdt,
 					  &channel_id);
 
-	if (err_code == NRFX_ERROR_NO_MEM) {
-		return -ENOMEM;
+	if (err_code == -ENOMEM) {
+		return err_code;
```
magp-nordic, line 141: Doing it like in `wdt_nrf_disable` would make more sense, and will work if more error codes are added to `nrfx_wdt_channel_alloc`, but can be left as is.

PR #3330, opened 2025-09-30
[nrf noup] soc: nordic: Support TF-M for poweroff

Vge0rge:
Enable going to system off when the TF-M API is available. This
    calls the relevant TF-M platform API to enter system off mode using
    TF-M. This is a noup because the TF-M service for system off is
    currently located in sdk-nrf only.

    This is meant to be a short lived commit, work is already ongoing
    to upstream an official TF-M system off solution that can be be
    added to upstream Zephyr.


rlubos:
Integrated in https://github.com/nrfconnect/sdk-zephyr/pull/3553

PR #3331, opened 2025-09-30
 Bluetooth: nRF Connect SDK v2.6.4 NCSIDB-1718 cherry-picks

cvinayak:
nRF Connect SDK v2.6.4 NCSIDB-1718 cherry-picks.


alwa-nordic:
Instructions for Copilot reviewer: These are cherry-picks. Do NOT suggest edits. Process mandates that the cherry-picks auto merge with no manual resolutions or edits. DO evaluate whether the automatic merge is semantically correct, and if it's not, which parent commits from the source need to be added to avoid an error.

alwa-nordic:
Doesn't look like copilot reads PR comments.

PR #3332, opened 2025-09-30
[nrf fromlist] drivers: spi: spi_nrfx_spim: use standard instantiation

mstasiaknordic:
Used API for standard instantiation and replaced nrfx_err_t error values with errno.

Removed Kconfig symbols for each SPIM instance which are no longer needed.


mstasiaknordic:
> @mstasiaknordic why is this a draft? Is it not ready for review yet? Only PRs in upstream were supposed to be drafts, as they will not be merged yet, but PRs in NCS can be regular ones.

I will get it out of draft once I confirm it works in all cases, probably when CI is more green

mstasiaknordic:
CI seems to be green now, apart from unrelated fails. Ready for review :)

@magp-nordic

modules/hal_nordic/nrfx/nrfx_kconfig.h
```diff
@@ -486,24 +486,9 @@
...
-#ifdef CONFIG_NRFX_SPIM1
-#define NRFX_SPIM1_ENABLED 1
-#endif
-#ifdef CONFIG_NRFX_SPIM2
-#define NRFX_SPIM2_ENABLED 1
-#endif
-#ifdef CONFIG_NRFX_SPIM3
-#define NRFX_SPIM3_ENABLED 1
 #ifdef CONFIG_NRF52_ANOMALY_198_WORKAROUND
 #define NRF52_ERRATA_198_ENABLE_WORKAROUND 1
```
magp-nordic, line 490: Won't that cause enabling the workaround on all instances? Is this something that can be done?
Maybe definition of `NRF52_ANOMALY_198_WORKAROUND` should have `depends on $(dt_nodelabel_enabled_with_compat,spi3,$(DT_COMPAT_NORDIC_NRF_SPIM))` added.

mstasiaknordic: Done

PR #3336, opened 2025-09-30
Nrfx 8370 test external bsp

magp-nordic:
DNM, this is only to test CI.

Based on https://github.com/nrfconnect/sdk-zephyr/pull/3335

PR #3342, opened 2025-10-01
[nrf fromlist] modules: hal_nordic: align to BSP outside of nrfx

magp-nordic:
Align sdk-zephyr to having external BSP outside of nrfx directory.

Upstream PR #:


modules/hal_nordic/nrfx/CMakeLists.txt
```diff
@@ -15,19 +15,26 @@ if(NOT DEFINED CONFIG_SOC_NORDIC_BSP_NAME)
...
-  zephyr_compile_definitions(NRFX_BSP_NRF_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/mdk/nrf.h")
-  zephyr_compile_definitions(NRFX_BSP_ERRATAS_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/mdk/nrf_erratas.h")
-  zephyr_compile_definitions(NRFX_BSP_SOC_IRQS_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/soc/nrfx_irqs.h")
-  zephyr_compile_definitions(NRFX_BSP_NRFX_EXT_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/nrfx_ext.h")
-  zephyr_compile_definitions(NRFX_BSP_NRFX_COREDEP_PATH="${CONFIG_SOC_NORDIC_BSP_NAME}/soc/nrfx_coredep_defs.h")
+  zephyr_compile_definitions(NRFX_BSP_NRF_PATH="mdk/nrf.h")
+  zephyr_compile_definitions(NRFX_BSP_ERRATAS_PATH="mdk/nrf_erratas.h")
+  zephyr_compile_definitions(NRFX_BSP_SOC_IRQS_PATH="soc/nrfx_irqs.h")
+  zephyr_compile_definitions(NRFX_BSP_NRFX_EXT_PATH="nrfx_ext.h")
+  zephyr_compile_definitions(NRFX_BSP_NRFX_COREDEP_PATH="soc/nrfx_coredep_defs.h")
```
masz-nordic, line 37: This implicitly removes need for those compile definitions in BSP dispatcher.
Is that by design?

magp-nordic: @masz-nordic yes, it is. Chosen BSP dir is added to include path anyway above, and paths containing BSP are no longer valid since they can be placed anywhere in NCS, even in directories that are not in include path (for example, when I tested it with https://github.com/nrfconnect/sdk-zephyr/pull/3336 I set the BSP external path to `modules/hal/bspexternal` and building failed on those lines before I removed `CONFIG_SOC_NORDIC_BSP_NAME`).

masz-nordic: Then a PR removing compile definitions usage in nrfx should be created too.

PR #3344, opened 2025-10-01
[nrf fromlist] drivers: timer: nrf_grtc_timer: Optimize to reduce reg…

nordic-krch:
…ister access

Speed up execution of the interrupt handler and sys_clock_set_timeout(). Sys_clock_set_timeout() can be called in two scenarios: from previous timeout expiration handler or freely. If the former case fast path can be used since CC value in the GRTC register just expired and it can be used as a reference for CCADD setting. This is only a single register write so it's much faster. In the latter a longer procedure is applied which also happens in two variants. If value which is set in CC is further in the future (e.g. K_FOREVER was set before) then CC can be safely overwritten with a new value without a risk of triggering unexpected COMPARE event. If value in CC is earlier than the new CC value (if earlier timeout was aborted) then there is a risk of COMPARE event happening while it is being overwritten. That case requires long and safer procedure of setting CC.

Update hal_nordic with changes in the nrfx_grtc driver which are needed for nrf_grtc_timer changes.

Upstream PR #: 87944


nordic-krch:
`DNM` as it requires alignment in SDFW.

nordic-krch:
Removing `DNM` as cpusec is already aligned so next ironside release will be ready and it should cooperate fine with existing ironside which is using GRTC driver without those changes.

nordic-krch:
It is another attempt to get those optimizations in after #3059 got reverted. It was reverted because it broke GRTC on cpusec which is not fixed.

PR #3346, opened 2025-10-01
[nrf noup] boards: nordic: Enable PSA RNG for nrf54h20

Vge0rge:
Noup since Ironside not available upstream and it is required for PSA RNG.

This enables the PSA RNG as the default Zephyr entropy provider for the nrf54h20dk cpuapp and cpurad targets.


SebastianBoe:
ironside **PSA** is **currently** not available upstream, just to clarify for others reading the PR description.

subsys/secure_storage/Kconfig
```diff
@@ -4,6 +4,7 @@
 menuconfig SECURE_STORAGE
 	bool "Secure storage subsystem"
 	depends on !BUILD_WITH_TFM
+	depends on !NRF_IRONSIDE
```
tomi-font: missing `#` in `Upstream PR #`

modules/mbedtls/Kconfig.psa.logic
```diff
@@ -1,8 +1,35 @@
 # Copyright (c) 2024 BayLibre SAS
 # SPDX-License-Identifier: Apache-2.0
 
-# This file extends Kconfig.psa (which is automatically generated) by adding
-# some logic between PSA_WANT symbols.
+config PSA_CRYPTO
+	bool "PSA Crypto API"
+	help
+	  Enable a PSA Crypto API provider in the build. If TF-M is enabled then
+	  it will be used for this scope, otherwise Mbed TLS will be used.
```
SebastianBoe, line 8: The help text doesn't mention out-of-tree providers.

Vge0rge: Sure, but this is a fromlist commit so I cannot update it here. Please send comments here if you want: 
https://github.com/zephyrproject-rtos/zephyr/pull/96415

seko-nordic:
can we remove CONFIG_FAKE_ENTROPY_NRF_PRNG and all associated to it now?

Vge0rge:
> FAKE_ENTROPY_NRF_PRNG

I am not sure about this, what I can tell is that with this PR the option FAKE_ENTROPY_NRF_PRNG cannot be enabled without modifying the device tree nodes. I am not sure if there are use cases that still require it for some reason.

PR #3362, opened 2025-10-06
[nrf noup] scripts: ci: check_compliance: Check Kconfigs for enable

nordicjm:
Allows checking sdk-nrf for Kconfig prompts that start with "Enable" as these should not be allowed

### DIFF
--- a/scripts/ci/check_compliance.py
+++ b/scripts/ci/check_compliance.py
@@ -1232,6 +1232,7 @@ Missing SoC names or CONFIG_SOC vs soc.yml out of sync:
                           ":!/doc/nrf/releases_and_maturity",
                           ":!/doc/nrf/libraries/bin/lwm2m_carrier/CHANGELOG.rst",
                           ":!/doc/nrf/app_dev/device_guides/nrf70/wifi_advanced_security_modes.rst",
+                          ":!/doc/nrf-bm/release_notes",
                           cwd=GIT_TOP)
 
         # splitlines() supports various line terminators


rugeGerritsen:
Why is this a no-up? Not using "Enable" should be in the style guide allowing us to upstream this commit

rugeGerritsen:
@nordicjm , now that you are anyways helping out adding style checks to CI, I would appreciate if you could approve https://github.com/nrfconnect/sdk-zephyr/pull/3360

nordicjm:
> Why is this a no-up? Not using "Enable" should be in the style guide allowing us to upstream this commit

It's problematic because zephyr modules do not follow this because they are external projects not zephyr-specific projects, so upon enabling it in zephyr it fails because of 3 modules and it's unlikely that they will be fixed in their upstream repos, so this has to be enabled here like this so it actually runs in sdk-nrf due to how sdk-nrf works (it is added as a zephyr module)

rugeGerritsen:
> > Why is this a no-up? Not using "Enable" should be in the style guide allowing us to upstream this commit
> 
> It's problematic because zephyr modules do not follow this because they are external projects not zephyr-specific projects, so upon enabling it in zephyr it fails because of 3 modules and it's unlikely that they will be fixed in their upstream repos, so this has to be enabled here like this so it actually runs in sdk-nrf due to how sdk-nrf works (it is added as a zephyr module)

I know I'm not maintaining this part, but I anyways want to help out with some suggestions on how to address this:
1. Extend the commit to allow using an include- or exclude-list for modules where this check is enabled.
2. Create a separate check in the compliance script where Kconfig files are read as text. That will ensure that we only enforce this new style check on modified files.

PR #3364, opened 2025-10-06
[nrf fromlist] samples: boards: nordic: spis_wakeup: Fix sample on nr…

nordic-segl:
…f54h20

Add missing property to spi130 node.
Add `zephyr,pm-device-runtime-auto` since the sample enables PM.

Upstream PR #: 97044


nordic-piks (approved):

nordic-segl:
After recent changes to PM on nrf54h20 this is no longer needed.

PR #3613, opened 2025-12-05
[nrf noup] drivers: flash: nrf_mram: wait until mramc is ready before write

rghaddab:
When writes are issued to the mram controller without verifying that it is ready, it can cause the reads to be stalled until all writes finish. Fix this by adding a wait busy loop before each MRAM_WORD aligned write and disable autopowerdown before starting the write.

### DIFF
--- a/drivers/flash/soc_flash_nrf_mram.c
+++ b/drivers/flash/soc_flash_nrf_mram.c
@@ -9,6 +9,13 @@
 #include <zephyr/drivers/flash.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/sys/barrier.h>
+#include <zephyr/cache.h>
+#if defined(CONFIG_MRAM_LATENCY)
+#include "../soc/nordic/common/mram_latency.h"
+#endif
+#if defined(CONFIG_NRF_IRONSIDE_BOOT_REPORT)
+#include "../soc/nordic/ironside/include/nrf_ironside/boot_report.h"
+#endif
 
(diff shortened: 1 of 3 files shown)


nordicjm (approved):

bjarki-andreasen (approved):

PerMac (changes requested):
this is wrong file (should be removed). We use this one for zephyr tests https://github.com/nrfconnect/sdk-nrf/blob/main/scripts/quarantine_zephyr.yaml

rghaddab:
@PerMac can you remove your change request ?

PerMac (approved):

David-Garcia-Polo (approved):

karstenkoenig (approved):

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -186,5 +269,31 @@ static DEVICE_API(flash, nrf_mram_api) = {
...
+
+	rc = ironside_boot_report_get(&report);
+
+	if (rc) {
+		LOG_ERR("Failed to get Ironside boot report");
+		return rc;
+	}
+
+	nrf_mram_data->ironside_se_ver = FIELD_GET(IRONSIDE_SE_VER_MASK,
+						   report->ironside_se_version_int);
```
jonathannilsen: @rghaddab Checking only this last part is not entirely forward compatible, as this number is planned to wrap around at 127.

Comparing against the full version is recommended, see https://github.com/zephyrproject-rtos/zephyr/pull/97691#discussion_r2592100564

PR #3684, opened 2026-01-14
Rghaddab/mram weak words

rghaddab:


karstenkoenig:
How/where does the busfault handler return? As far as I can tell this only preps the mram flash driver so it has a clear layout that is easy to jump back to, but at the moment the busfault would continue to kill the thread doing the mram access.

Could an alternative be to just have the mram read happen in a designated thread that either returns successfully at the end or gets killed by the busfault handler if there was a busfault?

PR #3749, opened 2026-02-05
drivers: nrf_mram: verify writes using MVDMA

rghaddab:
Fix the case where the MRAM writes get corrupted in some conditions. After each MRAM_WORD_SIZE write a read of the written/erased MRAM word is performed using the MVDMA.
If an error is detected in the MVDMA registers, retry the write for a maximum of 3 times.

### DIFF
--- a/drivers/flash/Kconfig.nrf_mram
+++ b/drivers/flash/Kconfig.nrf_mram
@@ -8,6 +8,7 @@ config SOC_FLASH_NRF_MRAM
 	bool "Nordic Semiconductor flash driver for MRAM"
 	default y
 	depends on DT_HAS_NORDIC_MRAM_ENABLED
+	depends on HAS_NRFX
 	select FLASH_HAS_DRIVER_ENABLED
 	select FLASH_HAS_PAGE_LAYOUT
 	select FLASH_HAS_NO_EXPLICIT_ERASE
@@ -20,3 +21,15 @@ config SOC_FLASH_NRF_MRAM
 	  Note that MRAM words are auto-erased when written to, but writing to a
 	  pre-erased area is faster. Hence, the erase API is not required, but
 	  it can be used to amortize write performance for some use cases.
+
+config NRF_MRAM_MAX_RETRIES
+	int "Maximum number of retries for MRAM write/erase operations"
+	default 20
+	range 0 255
+	depends on SOC_FLASH_NRF_MRAM
+	help
+	  This option sets the maximum number of retry attempts for MRAM
+	  write and erase operations when verification fails. Each retry
+	  attempts to rewrite or re-erase the data to ensure reliability.
+	  Higher values increase reliability but may impact performance
+	  in case of persistent failures.
--- a/drivers/flash/soc_flash_nrf_mram.c
+++ b/drivers/flash/soc_flash_nrf_mram.c
@@ -142,22 +142,9 @@ static int nrf_mram_mvdma_read(uint32_t addr)
 	return 0;
 }
 
-static int nrf_mram_write_and_verify(uint32_t addr, const void *data, size_t len)
+static int nrf_mram_write_and_verify_word(uint32_t addr, const void *data, size_t len)
 {
 	uint8_t retries = CONFIG_NRF_MRAM_MAX_RETRIES;
-#if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
-	uint8_t rbuf[MRAM_WORD_SIZE];
-
-	if (len % MRAM_WORD_SIZE) {
-		/* For partial word writes, we need to read the existing data first to avoid
-		 * overwriting the unwritten bytes in the same word.
-		 */
-		memcpy(rbuf + len, (void *)(addr + len), MRAM_WORD_SIZE - len);
-		memcpy(rbuf, data, len);
-		data = rbuf;
-		len = MRAM_WORD_SIZE;
-	}
-#endif
 
 	while (retries--) {
 		memcpy((void *)addr, data, len);
@@ -172,27 +159,29 @@ static int nrf_mram_write_and_verify(uint32_t addr, const void *data, size_t len
 	return -EIO;
 }
 
-static int nrf_mram_erase_and_verify(uint32_t addr, size_t len)
-{
-	uint8_t retries = CONFIG_NRF_MRAM_MAX_RETRIES;
 #if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
+static int nrf_mram_write_and_verify_partial(uint32_t addr, const void *data, size_t len)
+{
 	uint8_t rbuf[MRAM_WORD_SIZE];
 
-	if (len % MRAM_WORD_SIZE) {
-		/* For partial word writes, we need to read the existing data first to avoid
-		 * overwriting the unwritten bytes in the same word.
-		 */
-		memcpy(rbuf, (void *)(addr + len), MRAM_WORD_SIZE - len);
-	}
+	/* For partial word writes, we need to read the existing data first to avoid
+	 * overwriting the unwritten bytes in the same word.
+	 */
+	memcpy(rbuf, (void *)(addr & ~MRAM_WORD_MASK), MRAM_WORD_SIZE);
+	memcpy(rbuf + (addr & MRAM_WORD_MASK), data, len);
+
+	return nrf_mram_write_and_verify_word(addr & ~MRAM_WORD_MASK, rbuf, MRAM_WORD_SIZE);
+}
 #endif
+
+static int nrf_mram_erase_and_verify_word(uint32_t addr, size_t len)
+{
+	uint8_t retries = CONFIG_NRF_MRAM_MAX_RETRIES;
 
 	while (retries--) {
 		memset((void *)addr, ERASE_VALUE, len);
-#if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
-		if (len % MRAM_WORD_SIZE) {
-			memcpy((void *)addr, rbuf, MRAM_WORD_SIZE - len);
-		}
-#endif
+		/* before using MVDMA make sure that the cache is flushed */
+		sys_cache_data_flush_range((void *)addr, len);
 		if (!nrf_mram_mvdma_read(addr)) {
 			return 0;
 		}
@@ -203,6 +192,21 @@ static int nrf_mram_erase_and_verify(uint32_t addr, size_t len)
 
 	return -EIO;
 }
+
+#if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
+static int nrf_mram_erase_and_verify_partial(uint32_t addr, size_t len)
+{
+	uint8_t rbuf[MRAM_WORD_SIZE];
+
+	/* For partial word erases, we need to read the existing data first to avoid
+	 * overwriting the unwritten bytes in the same word.
+	 */
+	memcpy(rbuf, (void *)(addr & ~MRAM_WORD_MASK), MRAM_WORD_SIZE);
+	memset(rbuf + (addr & MRAM_WORD_MASK), ERASE_VALUE, len);
+
+	return nrf_mram_write_and_verify_word(addr & ~MRAM_WORD_MASK, rbuf, MRAM_WORD_SIZE);
+}
+#endif
 
 #if defined(CONFIG_MRAM_LATENCY)
 static inline uint32_t nrf_mram_ready(uint32_t addr, uint8_t ironside_se_ver)
@@ -262,6 +266,50 @@ static int nrf_mram_read(const struct device *dev, off_t offset, void *data, siz
 	return 0;
 }
 
+#if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
+/**
+ * @brief Structure to hold divided byte lengths for MRAM operations.
+ */
+struct nrf_mram_length_breakdown {
+	size_t first_word_bytes; /**< Number of bytes in the first partial MRAM word */
+	size_t aligned_bytes;    /**< Number of bytes in fully aligned MRAM words */
+	size_t last_word_bytes;  /**< Number of bytes in the last partial MRAM word */
+};
+
+/**
+ * @brief Divide operation length into first word, aligned words, and last word components.
+ *
+ * Calculates how many bytes belong to:
+ * - The first partial MRAM word (if offset is not word-aligned)
+ * - Fully aligned MRAM words in the middle
+ * - The last partial MRAM word (if total length doesn't end on word boundary)
+ *
+ * @param offset Relative offset into memory.
+ * @param len    Number of bytes for the intended operation.
+ *
+ * @return Structure containing the divided byte lengths.
+ */
+static inline struct nrf_mram_length_breakdown nrf_mram_divide_length(off_t offset, size_t len)
+{
+	struct nrf_mram_length_breakdown len_break = {0};
+	size_t first_word_bytes = MRAM_WORD_SIZE - (offset & MRAM_WORD_MASK);
+	size_t remaining = 0;
+
+	if (len > first_word_bytes) {
+		len_break.first_word_bytes = first_word_bytes;
+		remaining = len - first_word_bytes;
+		len_break.aligned_bytes = remaining & ~MRAM_WORD_MASK;
+		len_break.last_word_bytes = remaining & MRAM_WORD_MASK;
+	} else {
+		len_break.first_word_bytes = len;
+		len_break.aligned_bytes = 0;
+		len_break.last_word_bytes = 0;
+	}
+
+	return len_break;
+}
+#endif
+
 static int nrf_mram_write(const struct device *dev, off_t offset, const void *data, size_t len)
 {
 	struct nrf_mram_data_t *nrf_mram_data = dev->data;
@@ -288,29 +336,48 @@ static int nrf_mram_write(const struct device *dev, off_t offset, const void *da
 #endif
 	}
 
+#if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
+	struct nrf_mram_length_breakdown len_break = nrf_mram_divide_length(offset, len);
+
+	/* First, write the partial bytes from the first word */
+	if (len_break.first_word_bytes) {
+		while (!nrf_mram_ready(addr, ironside_se_ver)) {
+			/* Wait until MRAM controller is ready */
+		}
+		ret = nrf_mram_write_and_verify_partial(addr, data, len_break.first_word_bytes);
+		if (ret) {
+			goto latency_release;
+		}
+		/* align to the next word boundary */
+		len = len_break.aligned_bytes;
+		data = (const void *)((uintptr_t)data + len_break.first_word_bytes);
+		addr += len_break.first_word_bytes;
+	}
+#endif
 	for (uint32_t i = 0; i < (len / MRAM_WORD_SIZE); i++) {
 #if defined(CONFIG_MRAM_LATENCY)
 		while (!nrf_mram_ready(addr + (i * MRAM_WORD_SIZE), ironside_se_ver)) {
 			/* Wait until MRAM controller is ready */
 		}
 #endif
-		ret = nrf_mram_write_and_verify(addr + (i * MRAM_WORD_SIZE),
-						(void *)((uintptr_t)data + (i * MRAM_WORD_SIZE)),
-						MRAM_WORD_SIZE);
+		ret = nrf_mram_write_and_verify_word(
+			addr + (i * MRAM_WORD_SIZE),
+			(void *)((uintptr_t)data + (i * MRAM_WORD_SIZE)), MRAM_WORD_SIZE);
 		if (ret) {
 			goto latency_release;
 		}
 	}
 #if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
-	if (len % MRAM_WORD_SIZE) {
+	if (len_break.last_word_bytes) {
 #if defined(CONFIG_MRAM_LATENCY)
-		while (!nrf_mram_ready(addr + (len & ~MRAM_WORD_MASK), ironside_se_ver)) {
+		while (!nrf_mram_ready(addr + len_break.aligned_bytes, ironside_se_ver)) {
 			/* Wait until MRAM controller is ready */
 		}
 #endif
-		ret = nrf_mram_write_and_verify(addr + (len & ~MRAM_WORD_MASK),
-						(void *)((uintptr_t)data + (len & ~MRAM_WORD_MASK)),
-						len & MRAM_WORD_MASK);
+		ret = nrf_mram_write_and_verify_partial(
+			addr + len_break.aligned_bytes,
+			(void *)((uintptr_t)data + len_break.aligned_bytes),
+			len_break.last_word_bytes);
 		if (ret) {
 			goto latency_release;
 		}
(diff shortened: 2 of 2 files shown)


drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +48,151 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
 BUILD_ASSERT((ERASE_BLOCK_SIZE % WRITE_BLOCK_SIZE) == 0,
 	     "erase-block-size expected to be a multiple of write-block-size");
 
+/** Maximum number of retries for MRAM write/erase operations. */
```
pizi-nordic: This is a bit nitpicking, but IMHO MVDMA definitions and declarations should be in file not having "flash" in a name...

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +48,151 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+
+	/* Issue a dummy write, since we didn't have anything to write here.
+	 * Doing this lets us finalize our changes before we exit the driver API.
+	 */
+	sys_write8(sys_read8(addr_end), addr_end);
+}
+
+static int nrf_mram_mvdma_read(uint32_t addr)
+{
+	uint8_t rbuf[MRAM_WORD_SIZE] __aligned(CONFIG_DCACHE_LINE_SIZE);
```
pizi-nordic: There should be a configuration, which allows us to read data but not copy it somewhere (like calculating CRC32). This way we can reduce stack usage.

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +48,151 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+	nrf_mvdma_task_trigger(NRF_MVDMA, NRF_MVDMA_TASK_START0);
+	while (!nrf_mvdma_event_check(NRF_MVDMA, NRF_MVDMA_EVENT_COMPLETED0)) {
+		/* Wait until transfer is complete */
+		if (nrf_mvdma_event_check(NRF_MVDMA, NRF_MVDMA_EVENT_SOURCEBUSERROR)) {
+			nrf_mvdma_event_clear(NRF_MVDMA, NRF_MVDMA_EVENT_SOURCEBUSERROR);
+			LOG_ERR("MVDMA source bus error");
+			return -EIO;
+		}
+
+		if (nrf_mvdma_event_check(NRF_MVDMA, NRF_MVDMA_EVENT_SINKBUSERROR)) {
```
pizi-nordic: This should never happen.

drivers/flash/soc_flash_nrf_mram.c
rghaddab: @pizi-nordic yes these definitions will be in a separate file once we confirm that this fix works

drivers/flash/soc_flash_nrf_mram.c
rghaddab: Yes, but I chose the simplest configuration cause I know that it works. I am not sure if the other configurations of the MVDMA have been tested.
Let's confirm that the fix works then we can change the configuration mode for the MVDMA

drivers/flash/soc_flash_nrf_mram.c
rghaddab: Yes normally that should never happen and I will remove it later.
I was curious if during the tests we can see these errors messages

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +52,185 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+	sys_write8(sys_read8(addr_end), addr_end);
+}
+
+static int nrf_mram_mvdma_read(uint32_t addr)
+{
+	uint8_t rbuf[MRAM_WORD_SIZE] __aligned(CONFIG_DCACHE_LINE_SIZE);
+	struct mvdma_basic_desc desc __aligned(CONFIG_DCACHE_LINE_SIZE) =
+		NRF_MVDMA_BASIC_MEMCPY_INIT(rbuf, addr, MRAM_WORD_SIZE);
+
+	sys_cache_data_flush_range(&desc, sizeof(desc));
```
travis3630, line 127: Why do you need to flush the cache? And why need to care about the cache?

rghaddab: @travis3630 that's a limitation of the MVDMA, the descriptor must be aligned with DCACHE and flushed otherwise the transfer won't happen

travis3630: And why are you not using the nrf_mram_read and use mvdma_read to verify the write? Just curious to know?

rghaddab: @travis3630 that's a long discussion to explain it.
But I can try to briefly explain it : when using nrf_mram_read and a bus fault occurs all the system halts and no recovery is possible (not possible to resume the current thread).
However when using MVDMA the bus error is signaled through its registers and we can handle it and recover the write

travis3630: and using nrf_mram_read to read on a weak MRAM word can trigger a bus fault error?

rghaddab: Yes

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +56,149 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
 BUILD_ASSERT((ERASE_BLOCK_SIZE % WRITE_BLOCK_SIZE) == 0,
 	     "erase-block-size expected to be a multiple of write-block-size");
 
+/** Maximum number of retries for MRAM write/erase operations. */
+#define MAX_MRAM_RETRIES 3
```
karstenkoenig: Unsure on policy, but shouldn't this be a kconfig option?

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: They are still here, do you plan to keep them in here now? I think breaking them out into a custom file might be cleaner, but maybe it is easier getting it into upstream if contained in here and not put into some vendor specific driver?

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +56,149 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+}
+#endif
+
+static int nrf_mram_mvdma_read(uint32_t addr)
+{
+	uint8_t rbuf[MRAM_WORD_SIZE] __aligned(CONFIG_DCACHE_LINE_SIZE);
+	struct mvdma_basic_desc desc __aligned(CONFIG_DCACHE_LINE_SIZE) =
+		NRF_MVDMA_BASIC_MEMCPY_INIT(rbuf, addr, MRAM_WORD_SIZE);
+
+	sys_cache_data_flush_range(&desc, sizeof(desc));
```
karstenkoenig, line 127: This flush here is correct, but you also need cache handling for _dst as a subsequent transaction to the same location would then just return the cached data. 
Maybe the cache handling in itself is reason enough to move this to a designated file given the cache handling is far from trivial

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +56,149 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+
+	return 0;
+}
+
+static int nrf_mram_write_and_verify(uint32_t addr, const void *data, size_t len)
+{
+	uint8_t retries = MAX_MRAM_RETRIES;
+
+	while (retries--) {
+		memcpy((void *)addr, data, len);
```
karstenkoenig, line 150: You need to read the mram data word first if len != mram_word_size*n, then copy in the data you want updated and then write that result to mram. That way you won't get corruption on the follow up write operation because there will be no internal read-modify-write operation. You always need to write the full 16 byte mram word. That'll also make the commit call obsolete at the same time

karstenkoenig (changes requested):

drivers/flash/soc_flash_nrf_mram.c
rghaddab: There is no such policy for Zephyr.
It could be defined here if there is no need for the user to change it.
Why do you thing such Kconfig should exist ?

drivers/flash/soc_flash_nrf_mram.c
rghaddab: @karstenkoenig so according to this there is a difference in the MRAMC between handling for example :
- write of 10 bytes => read 6 bytes => write of 6 bytes
And
- read of 6 bytes => write of 10 bytes => write of 6 bytes
Is that write ?

drivers/flash/soc_flash_nrf_mram.c
rghaddab: I think it is simpler to keep this in the mram driver as there is no interest in creating another file in the flash driver folder which is used only by the mram driver.
I don't think that separating this MVDMA workaround (using HAL) in a separate file will be accepted upstream.
It should remain in the same file until the MVDMA driver gets merged upstream

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: case a) len = 16 
write 16 bytes -> just straight memcpy to mram

case b) len < 16
read 16B aligned mram word into 16 byte tmp_buffer, replace everything in the tmp buffer that the write operation wants to write with the content from the write operation, write tmp_buffer via memcpy to mram. If it fails, write tmp_buffer again
Effectively do a read-modify-write in software for the 16byte mram word instead of letting the hardware do it. If the hardware does it and the write fails you can't recover from it

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: Ah I was just unsure because it looks like a user tunable thing, but I don't foresee them changing this either, just leave as is then

drivers/flash/soc_flash_nrf_mram.c
rghaddab: @karstenkoenig I just discovered another problem.
When write-block-size is not a multiple of mram-word-size then when calling nrf_mram_write() the offset can start from a position in the middle of an mram word.
This will break all the logic of writing first chunks of mram_words, as the starting address is sometimes not aligned to an mram_word boundary.
I will add this check in the validate_and_map_addr() otherwise it will complicate all the write operations

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: ok, but anyways you'll need dcache handling for dst, otherwise you might have dst already cached in the cpu dcache.
That means you'll need to flush all affected lines for dst before calling the mvdma memcpy and a flush_and_invalidate after

rghaddab: The problem of cache is only for the descriptor (not dst). The descriptor contains all the information for the MVDMA to start the transfer. Once the descriptor is flushed (copied to the controller), the transfer starts.
Why do you think we should handle cached data ?

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: Hmm yes sorry I wasn't clear there. You need to make sure that on the write/erase paths all it always does mram-size-aligned accesses to the actual mram, so 16B alignment on address and size. Then you can also drop commit() as well as you know you've written to the last u32_t in the physical mramword

rghaddab: @karstenkoenig yes that will be fixed.
But in addition to that I will add a check to make sure that the starting address for nrf_mram_write is also aligned on MRAM_WORD_SIZE.
Let's suppose that write-block-size=4 and mram-word-size=16. Currently you can call nrf_mram_write with an offset 0x08 and a length of 64 bytes.
The current behavior is wrong as we need normally to write 8 bytes (from the 16 mram-word-size) then 3 chunks of 16 bytes (48 bytes) then the last 8 bytes.
Either we rework all the code to make sure that this usecase works or we need to assert that the starting address is aligned on the mram-word-size boundary

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -80,42 +225,16 @@ static uintptr_t validate_and_map_addr(off_t offset, size_t len, bool must_align
...
-	 */
-	addr_end |= MRAM_WORD_MASK;
-
-	/* Issue a dummy write, since we didn't have anything to write here.
-	 * Doing this lets us finalize our changes before we exit the driver API.
-	 */
-	sys_write8(sys_read8(addr_end), addr_end);
-}
-
 static int nrf_mram_read(const struct device *dev, off_t offset, void *data, size_t len)
```
karstenkoenig, line 252: You also need to wrap regular reads into the mvdma access, using the flash driver shouldn't be able to cause a busfault when accessing the data

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: If you read dst before then you won't see the update when mvdma has written it. You can try it out by writing something else to dst just before and then have mvdma make the copy, it'll be your old data. And you need to do the flush before as well, otherwise when you do a write+invalidate after you'll write back the dcache content, overwriting the mvdma written content

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +56,145 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+{
+	uint8_t retries = CONFIG_NRF_MRAM_MAX_RETRIES;
+#if (WRITE_BLOCK_SIZE & MRAM_WORD_MASK)
+	uint8_t rbuf[MRAM_WORD_SIZE];
+
+	if (len % MRAM_WORD_SIZE) {
+		/* For partial word writes, we need to read the existing data first to avoid
+		 * overwriting the unwritten bytes in the same word.
+		 */
+		memcpy(rbuf + len, (void *)(addr + len), MRAM_WORD_SIZE - len);
```
karstenkoenig: Can you rework the PR to support non-16byte aligned writes and lengths before we take another round through this? All the read-modify-write code will need to be rewritten for it and I believe will become cleaner if taken into account from the start.
Like at the moment it isn't clear if `nrf_mram_write_and_verify` takes length <=16 or not, as you do some % comparison here and there is no assert or documentation for this etc

My suggestion is to turn this function into a `nrf_mram_write_and_verify_mam_word()` that expects the data to fit in into the mram word, so addr and addr+length are in the same mram word. Then you can just check if the length is 16 and if yes just write that straight up and if not read the full existing mram word to rbuf and then update in place in the rbuf and write that to mram.

karstenkoenig (changes requested):
This should be either full way allow non-mram word size aligned addresses and lengths or don't allow it.

drivers/flash/soc_flash_nrf_mram.c
rghaddab: The current nrf_mram_write_and_verify takes as well length < 16 bytes (only if write-block-size is not aligned to mram-word-size).
For partial writes we don't need to read the entire mram word. Only the unkown part is needed to be read

drivers/flash/soc_flash_nrf_mram.c
rghaddab: I don't get your point here.
Why should we wrap regular mram_read with MVDMA.
This is not part of the mram weak words during a 'write'

karstenkoenig: The write could have failed eventually, and then the flash driver still shouldn't busfault accessing this address. Returning -EIO sounds a lot cleaner as a result of flash_read() than a busfault

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: At the moment you are requiring address to 16B aligned. Only that's why it works at the moment, because the 'to be read part' is at the back of the 16B. If you suddenly have the data to be written from byte 3 to byte 7 in the mram word you are doing memcpy with byte sized transactions, instead of just loading the whole thing in 4 single uint32_t transaction. The mram word is always buffered anyways in the mramc, the transaction will be fast enough and the code more readable than trying to stitch together the final word from 3 different reads.

That's also why I recommended to rework this to take unaligned addresses, then it becomes apparent its easier to just read the 16bytes and overwrite the parts that the write wants to actually write

drivers/flash/soc_flash_nrf_mram.c
rghaddab: If the number of write retries exceeds the maximum then a Bus fault will be triggered eventually in the write function which means that the memory is damaged.
Currently this is a workaround for the write function, I do not agree that the read function should be using MVDMA.
In a future PR we can make read/write functions use only the MVDMA for their transfers as it is much faster than the memcpy which is doing them byte by byte in the Bus

drivers/flash/soc_flash_nrf_mram.c
rghaddab: I don't mind making this change.
But just notice that memcpy is different than sys_read32. To load 16 bytes that's not 4 transactions but 16 transactions (byte by byte)

karstenkoenig: doesn't that get optimized into just 32bit transactions if size and alignment of the data allows for it? If not all the more reason to just use sys_read32 4 times for reading and then sys_write32 4 times for writing the whole thing back

rghaddab: Yes normally using sys_read32 and sys_write32 will make the transactions faster.
But this is out of scope for the PR.
Improving the performance of the read/write should be dealt with in a separate PR

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: You still need to handle the cached data. If dst is in your dcache it will not show the new data. Also you need to flush the dcache line for dst to make sure all pending writes to the cacheline are done, otherwise you won't be able to invalidate and flush after the mvdma transaction is done

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +56,155 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+ * specified address. This operation serves as a verification mechanism after
+ * write/erase operations by checking if a bus fault occurs or if the operation
+ * completes successfully.
+ *
+ * @param addr The MRAM address to read from.
+ * @return 0 on success, -EIO if a bus error is detected.
+ */
+static int nrf_mram_mvdma_read(uint32_t addr)
+{
+	uint8_t rbuf[MRAM_WORD_SIZE] __aligned(CONFIG_DCACHE_LINE_SIZE);
```
karstenkoenig: rbuf doesn't need to be aligned to cache_line_size, but 4 byte aligned to satisfy cache dataunit size and the mvdma transaction needs.

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +56,155 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+	struct mvdma_basic_desc desc __aligned(CONFIG_DCACHE_LINE_SIZE) =
```
karstenkoenig: again, 4 byte alignment is enough. A cacheline is 64B, that means your existing alignment worst case has your stack growing by 124B + sizeof(struct mvdma_basic_desc)

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: Ah I misunderstood, you aren't actually reading the data back here, you are just using mvdma to check if you get an event. No need to cache handle the dst then, but as soon as you want to process the rbuf you'll need to add it

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -133,10 +261,55 @@ static int nrf_mram_read(const struct device *dev, off_t offset, void *data, siz
...
+	size_t first_word_bytes = MRAM_WORD_SIZE - (offset & MRAM_WORD_MASK);
+	size_t remaining = 0;
+
+	if (len > first_word_bytes) {
+		len_break.first_word_bytes = first_word_bytes;
+		remaining = len - first_word_bytes;
+		len_break.aligned_bytes = remaining & ~MRAM_WORD_MASK;
+		len_break.last_word_bytes = remaining & MRAM_WORD_MASK;
+	} else {
+		len_break.first_word_bytes = 0;
```
karstenkoenig: you return all 0 here? is this really working? for an offset of 0 and a length of 4 this returns just 0 for all lengths.

karstenkoenig (changes requested):

drivers/flash/soc_flash_nrf_mram.c
nordic-krch: isn't cache line is 32 bytes? you need to do cache flush for `desc` so aligning it to 32 allows to keep it in a single line.

drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -46,10 +56,155 @@ BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero
...
+}
+
+static int nrf_mram_write_and_verify_word(uint32_t addr, const void *data, size_t len)
+{
+	uint8_t retries = CONFIG_NRF_MRAM_MAX_RETRIES;
+
+	while (retries--) {
+		memcpy((void *)addr, data, len);
+		/* before using MVDMA make sure that the cache is flushed */
+		sys_cache_data_flush_range((void *)addr, len);
```
nordic-krch: Is this needed? `addr` is in MRAM space, is it D-cachable?

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: Sorry yes you are right it is 32B, I mixed it up with the L2CACHE
You can still have it just aligned to 4B as the dataunit is 4B and the cache tracks the dirty and invalid status based on the dataunit granularity. 
align doesn't guarantee full ownership of the cacheline anyways, so given this is the stack it is likely there will be more data still in the same cacheline.

drivers/flash/soc_flash_nrf_mram.c
rghaddab: yes I will remove the alignment here

drivers/flash/soc_flash_nrf_mram.c
rghaddab: This is taken from the MVDMA driver that was introduced a while ago in zephyr but got rejected.
https://github.com/zephyrproject-rtos/zephyr/pull/95912/changes#diff-8bc8063d15c4bd7916dbadf5e4e427c7319ef1da28f0fff43aec12d52854f592R140
@nordic-krch maybe knows the reason for this alignment

drivers/flash/soc_flash_nrf_mram.c
rghaddab: this is fixed now

drivers/flash/soc_flash_nrf_mram.c
nordic-krch: yes, functionally, it's ok to have 4 byte alignment. I was thinking about software flushing operation that will need to do the cache flushing loop twice if structure is not 32B aligned. So it's a trade-off between stack usage and speed. 
If speed is important then it could be static and function would only need to modify address field before each DMA transfer. Optimizations like that can be done in the future so i'm ok with have 4B alignment.

drivers/flash/Kconfig.nrf_mram
```diff
@@ -20,3 +21,15 @@ config SOC_FLASH_NRF_MRAM
 	  Note that MRAM words are auto-erased when written to, but writing to a
 	  pre-erased area is faster. Hence, the erase API is not required, but
 	  it can be used to amortize write performance for some use cases.
+
+config NRF_MRAM_MAX_RETRIES
+	int "Maximum number of retries for MRAM write/erase operations"
+	default 5
+	range 1 255
```
nordic-krch: what about support for retry=0. Is the risk of write failure real in all conditions? Is it a valid scenario that user would like to turn it off to speed up the writing (e.g. DFU process?)?

drivers/flash/soc_flash_nrf_mram.c
nordic-krch: we already had some community issues when trying to add mvdma driver so it might be easier to get that in when keeping it in one file. I hope to come back at some point to the generic mvdma driver and it would then replace it.

rghaddab:
@karstenkoenig @nordic-krch 
Can you have a look again at this PR ?
We agreed to merge this in 3.2 branch

nordic-krch:
@rghaddab I saw in upstream the PR with blocking bus fault interrupt. Do we still want to use MVDMA for verification?

rghaddab:
> @rghaddab I saw in upstream the PR with blocking bus fault interrupt. Do we still want to use MVDMA for verification?

well the upstream version still has problems related to Risc-V linker script and I still don't know how to resolve it
We agreed that for 3.2 branch we will keep this version (verified) until the upstream one gets merged.
Then I will revert this

nordic-krch (approved):

karstenkoenig (approved):

PR #3750, opened 2026-02-06
[nrf fromlist] dts: vendor: nordic: fix nrf54lm20a/*[/ns] sram size

Damian-Nordic:
reg attribute is supposed to hold start address and size instead of start address and end address. This caused a crash in malloc_prepare().

Upstream PR #: 103417

manifest-pr-skip


nordic-piks (approved):

de-nordic (approved):

Damian-Nordic:
It's very difficult to pull this change cleanly into as fromlist due to recent tree-wide changes in Zephyr.

PR #3802, opened 2026-02-24
Revert "[nrf noup] bluetooth: Temporary Kconfig fix for BT RPC config…

Damian-Nordic:
…uration"

This reverts commit a87ca4c92469a6ef8118c29785e8f5ac36111b75.

It appears the noup commit is no longer needed to build BT RPC client. Enabling `BT_DRIVERS` doesn't actually enable anything if `BT_HCI` is not set (which is the case for RPC client).


LuDuda (approved):

drivers/bluetooth/Kconfig
```diff
@@ -12,7 +12,7 @@
 menuconfig BT_DRIVERS
 	bool "Bluetooth drivers"
 	default y
```
nordicjm, line 14: Add `nrf-squash!` with the title of the original commit to the commit message so that both this and the original commit vanish when the release rebase is done

nordicjm (changes requested):

drivers/bluetooth/Kconfig
Damian-Nordic: good to know! Done.

nordicjm (approved):

Damian-Nordic:
It seems on nRF53 and nRF54H the noup is still needed and without that twister fails. Added an alternative change which I'll try to upstream before merging this PR but first adding it to this PR to check internal CI results.

PR #3865, opened 2026-03-13
[nrf fromtree] scripts: west: nrf_common: Provision slot invalidation…

Vge0rge:
… json

The nRF54L series devices require to reserve 2 KMU slots which are filled with random data used for invalidating the Protected RAM. Until now the provisioning of these slots happends in firmware which costs in both volatile and non volatile memory.

Instead of doing this nrfutil recently added the funcionality to provision these slots using a .json file which means that everything can be handled by the build system now.

This is preparation work, that is needed before the full integration arrives. The change checks if a json file with the name: "prot_ram_inv_slots.json" exists and if it does it invokes nrfutil with the provisioning command.

(cherry picked from commit ee6fda031108a5ab599ded90c494ab785ea8f02f)


Vge0rge:
This is already merged so closing.

PR #3876, opened 2026-03-17
linker: arm: account for ROM_END_OFFSET in partition manager path

nrbrook:
When USE_PARTITION_MANAGER is active, the ARM linker script defines ROM_SIZE as PM_SIZE (the full app partition) without subtracting CONFIG_ROM_END_OFFSET. This means the MCUboot swap trailer overhead computed by MCUboot's sysbuild module and propagated via CONFIG_ROM_END_OFFSET is not enforced by the linker.

As a result, an image can link successfully but be too large for MCUboot to swap at runtime, causing silent DFU failures where the device reports success but remains on the old firmware.

Subtract CONFIG_ROM_END_OFFSET from PM_SIZE in the USE_PARTITION_MANAGER branch, matching the behavior of the non-PM path. This ensures the linker produces an overflow error if the image exceeds the usable slot space.


nordicjm (commented):
the rom end offset generated is completely bogus when PM is enabled so this is not valid since it reads from dummy dts partitions

nrbrook:
> the rom end offset generated is completely bogus when PM is enabled so this is not valid since it reads from dummy dts partitions

@nordicjm Thanks for the review. The concern about DTS-derived values is fair as a follow-up but doesn't apply to the common path:

`mcuboot_image_footer_size` only reads DTS slot sizes when `CONFIG_BOOT_MAX_IMG_SECTORS_AUTO=y`. Otherwise it only uses the DTS parent-flash `erase-block-size` / `write-block-size`, which are hardware properties PM cannot override. For any config that keeps both slots on the same internal flash (the vast majority of nRF52 / nRF53 / nRF54 builds) the offset is exactly correct.

I attached a reproducer [dfu_size_overflow.zip](https://github.com/user-attachments/files/26890420/dfu_size_overflow.zip), a minimal Zephyr app with `SWAP_USING_MOVE` + PM + a Kconfig-sized bloat array. On stock sdk-zephyr it links, signs, and packages a DFU zip that MCUboot cannot swap at runtime (image 60,624 B slot - 8,192 = 57,344 B; `imgtool`'s default 128-sector trailer estimate of 1,584 B does not catch it either). With this PR the same build fails the link with `region 'FLASH' overflowed by 3942 bytes`.

The two configurations where the DTS value genuinely drifts from PM are (a) `BOOT_MAX_IMG_SECTORS_AUTO=y` with DTS slot sizes that straddle different erase-block counts than PM, and (b) `PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=y` where DTS slot1 stays on internal flash while PM puts it on QSPI. Both need a PM-aware recomputation of `CONFIG_ROM_END_OFFSET` in NCS before the linker runs. Happy to do that as a follow-up PR — but even in those cases this PR doesn't make things worse than today (where the check is simply absent), and for the majority of users it removes a well-disguised footgun. 

Also worth noting: `imgtool`'s own `check_trailer()` is *never* going to save us. NCS doesn't pass `--max-sectors`, so imgtool uses the 128-sector default and computes a trailer (~1.5 KiB) that has no relationship to the real SWAP_USING_MOVE overhead (~8 KiB on nRF52). The repro shows imgtool green-lighting a ~60 KiB payload into a 64 KiB slot that MCUboot can't touch.

Another thing to note: my [related zephyr PR](https://github.com/zephyrproject-rtos/zephyr/pull/105592) catches this issue by checking the real binary post-build.

PR #3880, opened 2026-03-18
Bring in mapped partition binding

nordicjm:
Including reverts and pre-requisites, some commits have been converted to no-ups with permission of Shantha to avoid bringing in many commits for other vendors that have no use in NCS

manifest-pr-skip


nordic-segl (approved):

subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt.c
```diff
@@ -43,72 +43,77 @@
 
 #if !defined(CONFIG_MCUBOOT_BOOTLOADER_MODE_RAM_LOAD)
 
+#if defined(CONFIG_FLASH_USES_MAPPED_PARTITION)
+#define PARTITION_IS_RUNNING_APP_PARTITION(label)				\
+	DT_SAME_NODE(DT_CHOSEN(zephyr_code_partition), DT_NODELABEL(label))
```
tomchy, line 48: The idea of having range checks was to support merged slots (i.e. inside FW loader on multi-core devices). Since we have a copy of this file in sdk-nrf, I do not see it as a necessity to add this support in the upstream Zephyr though.

subsys/dfu/img_util/flash_img.c
```diff
@@ -23,18 +23,23 @@ LOG_MODULE_REGISTER(flash_img, CONFIG_IMG_MANAGER_LOG_LEVEL);
 #include <bootutil/bootutil_public.h>
 #endif
 
-#define FIXED_PARTITION_IS_RUNNING_APP_PARTITION(label)                                            \
-	DT_SAME_NODE(FIXED_PARTITION_NODE_MTD(DT_CHOSEN(zephyr_code_partition)),                   \
-		FIXED_PARTITION_MTD(label)) && (FIXED_PARTITION_ADDRESS(label) <=                  \
+#if defined(CONFIG_FLASH_USES_MAPPED_PARTITION)
+#define PARTITION_IS_RUNNING_APP_PARTITION(label) \
+	DT_SAME_NODE(DT_CHOSEN(zephyr_code_partition), DT_NODELABEL(label))
```
tomchy, line 27: In this instance, I'm not sure if it is OK to remove the range check.

samples/subsys/fs/fs_sample/boards/nrf54l15dk_nrf54l15_cpuapp.overlay
```diff
@@ -14,19 +14,18 @@
 
 &cpuapp_rram {
 	partitions {
```
tomchy, line 16: I see that in most place, when the `fixed-partitions` compatible was removed, the `ranges` property was added. Please double-check that indeed in this instance the `ranges` propagates from the board-level definitions.

dts/vendor/nordic/nrf54l_05_10_15.dtsi
```diff
@@ -745,14 +745,18 @@
 		rram_controller: rram-controller@5004b000 {
 			compatible = "nordic,rram-controller";
 			reg = <0x5004b000 0x1000>;
+			ranges;
 			#address-cells = <1>;
 			#size-cells = <1>;
 			interrupts = <75 NRF_DEFAULT_IRQ_PRIORITY>;
 
-			cpuapp_rram: rram@0 {
+			cpuapp_rram: flash@0 {
```
tomchy: Why `flash` instead of `rram`?

dts/vendor/nordic/nrf54l15_cpuapp_partition.dtsi
```diff
@@ -6,28 +6,51 @@
...
 		#address-cells = <1>;
 		#size-cells = <1>;
 
 		boot_partition: partition@0 {
+			compatible = "zephyr,mapped-partition";
 			label = "mcuboot";
 			reg = <0x0 DT_SIZE_K(64)>;
 		};
 
+#ifdef WITH_FLPR_PARTITIONS
```
tomchy, line 18: It is a little bit strange that `WITH_FLPR_PARTITIONS` does not define the FLPR partition.

cmake/linker_script/arm/linker.cmake
```diff
@@ -33,23 +33,38 @@ endif()
...
-     OUTPUT_FORMAT HEXADECIMAL
-)
-
-if(CONFIG_FLASH_LOAD_SIZE GREATER 0)
+if(CONFIG_FLASH_USES_MAPPED_PARTITION)
+  dt_chosen(chosen_partition_path PROPERTY "zephyr,code-partition")
+  dt_reg_addr(chosen_partition_address PATH "${chosen_partition_path}")
+  dt_reg_size(chosen_partition_size PATH "${chosen_partition_path}")
+
+  math(EXPR FLASH_ADDR
```
tomchy: Why the `CONFIG_FLASH_LOAD_OFFSET` is skipped in this configuration?

include/zephyr/arch/arm/cortex_m/scripts/linker.ld
```diff
@@ -67,23 +67,28 @@ _image_1_primary_slot_id = PM_S1_ID;
...
-#endif
-
 #if defined(CONFIG_ROM_END_OFFSET)
 #define ROM_END_OFFSET CONFIG_ROM_END_OFFSET
 #else
 #define ROM_END_OFFSET 0
 #endif
 
+#if defined(CONFIG_FLASH_USES_MAPPED_PARTITION)
+#define ROM_ADDR (DT_REG_ADDR(DT_CHOSEN(zephyr_code_partition)))
```
tomchy, line 77: I'm not entirely sure if it is expected to point the `zephyr,code-partition` into RAM region if `CONFIG_XIP` is set to `n`.

subsys/storage/flash_map/flash_map_default.c
```diff
@@ -50,30 +71,48 @@ const struct flash_area *flash_map = default_flash_map;
...
+		.fa_off = PARTITION_NODE_OFFSET(part),					\
+		.fa_dev = DEVICE_DT_GET(DT_MTD_FROM_PARTITION(part)),			\
+		.fa_size = DT_REG_SIZE(part),						\
+	};
+
+DT_FOREACH_STATUS_OKAY(zephyr_mapped_partition, DEFINE_PARTITION)
+
+#undef DEFINE_PARTITION
+#undef DEFINE_PARTITION_1
+#undef DEFINE_PARTITION_0
```
tomchy, line 89: What's the point of having definitions in lines 73, 74 and 77?

include/zephyr/storage/flash_map.h
```diff
@@ -541,6 +598,19 @@ DT_FOREACH_STATUS_OKAY(fixed_partitions, FOR_EACH_PARTITION_TABLE)
...
+	((const struct flash_area *)&DT_CAT(global_zephyr_mapped_partition_ORD_, ord))
+
+#define DECLARE_MAPPED_PARTITION(node) DECLARE_MAPPED_PARTITION_0(DT_DEP_ORD(node))
+#define DECLARE_MAPPED_PARTITION_0(ord) \
+	extern const struct flash_area DT_CAT(global_zephyr_mapped_partition_ORD_, ord);
+
+DT_FOREACH_STATUS_OKAY(zephyr_mapped_partition, DECLARE_MAPPED_PARTITION)
+
+#undef DECLARE_MAPPED_PARTITION
+#undef DECLARE_MAPPED_PARTITION_0
```
tomchy, line 608: Why define things to `#undef` them right in the next line?

include/zephyr/storage/flash_map.h
```diff
@@ -353,97 +356,131 @@ uint8_t flash_area_erased_val(const struct flash_area *fa);
...
-		(DT_PROP_BY_IDX(DT_PARENT(DT_NODELABEL(label)), reg, 0) + \
-		 DT_PROP_BY_IDX(DT_NODELABEL(label), reg, 0)), \
-		(DT_PROP_BY_IDX(DT_NODELABEL(label), reg, 0)))
+#define PARTITION_OFFSET(label)									\
+	COND_CODE_1(DT_NODE_HAS_COMPAT(DT_NODELABEL(label), zephyr_mapped_partition),		\
+		    (DT_MAPPED_PARTITION_OFFSET(DT_NODELABEL(label))),				\
+		    (COND_CODE_1(DT_FIXED_SUBPARTITION_EXISTS(DT_NODELABEL(label)),		\
+				(DT_PROP_BY_IDX(DT_PARENT(DT_NODELABEL(label)), reg, 0) +	\
+				 DT_PROP_BY_IDX(DT_NODELABEL(label), reg, 0)), \
+				(DT_PROP_BY_IDX(DT_NODELABEL(label), reg, 0)))))
```
tomchy: Why `PARTITION_OFFSET` checks for `FIXED_SUBPARTITION` compatibility while the `PARTITION_NODE_OFFSET` does not?

tomchy (commented):
There are a few lines that I find suspicious. Please take a look at them as there might be some hidden bugs in this PR.

cmake/linker_script/arm/linker.cmake
tomchy: Already discussed: the only Kconfig that needs to be handled is the `CONFIG_ROM_OFFSET` and this part of the logic can be safely removed.

scripts/dts/gen_defines.py
```diff
@@ -998,6 +1068,25 @@ def write_global_macros(edt: edtlib.EDT):
 
                         out_dt_define(macro, val)
                         out_dt_define(macro + "_EXISTS", 1)
+            elif compat == "zephyr,mapped-partition":
+                parent = node.parent
+
+                while parent and "soc-nv-flash" not in parent.compats:
```
tomchy, line 1074:
```suggestion
                while parent and "soc-nv-flash" not in parent.compats and "nordic,mram" not in parent.compats:
```
If I correctly remember, @57300 knows the reason behind it.

include/zephyr/devicetree/mapped-partition.h
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * @brief Get the node identifier of the NVM memory for a partition.
+ * @param node_id node identifier for a zephyr,mapped-partition node.
+ * @return the node identifier of the internal memory that contains the zephyr,mapped-partition
+ *         node, or @ref DT_INVALID_NODE if it doesn't exist.
+ */
+#define DT_MEM_FROM_MAPPED_PARTITION(node_id)						\
+	COND_CODE_1(DT_NODE_HAS_COMPAT(DT_CAT(node_id, _NVM_DEVICE), soc_nv_flash),	\
+				       (DT_CAT(node_id, _NVM_DEVICE)),			\
+				       (DT_INVALID_NODE))
```
tomchy, line 97:
```suggestion
				       (COND_CODE_1(DT_NODE_HAS_COMPAT(DT_CAT(node_id, _NVM_DEVICE), nordic_mram), (DT_CAT(node_id, _NVM_DEVICE)), (DT_INVALID_NODE))))
```

include/zephyr/devicetree/mapped-partition.h
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * @brief Get the node identifier of the NVM controller for a partition.
+ * @param node_id node identifier for a zephyr,mapped-partition node.
+ * @return the node identifier of the memory technology device that contains the
+ *         zephyr,mapped-partition node.
+ */
+#define DT_MTD_FROM_MAPPED_PARTITION(node_id) DT_PARENT(DT_MEM_FROM_MAPPED_PARTITION(node_id))
```
tomchy, line 105: If `DT_MEM_FROM_MAPPED_PARTITION` returns a reference to the `*_NVM_DEVICE`, why there is still a need to call `DT_PARENT(..)` to get the MTD?

tomchy (changes requested):

subsys/dfu/img_util/flash_img.c
nordicjm: it is fine to do, this checks the node, ranges do not matter

dts/vendor/nordic/nrf54l15_cpuapp_partition.dtsi
nordicjm: it means with _allowance_ left for flpr partitions to be defined, they are defined in the snippet. Not that the way it's been done is all that great, but that's what has been inherited by the people that first created it

dts/vendor/nordic/nrf54l_05_10_15.dtsi
nordicjm: ah that is a silly typo

subsys/storage/flash_map/flash_map_default.c
nordicjm: they are called by the dt macros, so are needed for that, they are not needed after the macro's have been called

include/zephyr/storage/flash_map.h
nordicjm: as above, they are called by the macros

include/zephyr/storage/flash_map.h
nordicjm: seems like a bug, will remove the bad one and make the label one call the node one

scripts/dts/gen_defines.py
nordicjm: the mram device is being reworked so this is no longer an issue

include/zephyr/devicetree/mapped-partition.h
nordicjm: as above, mram is being reworked, changes will be brought in

include/zephyr/devicetree/mapped-partition.h
nordicjm: mem is the memory node, the parent is memory controller node

dts/vendor/nordic/nrf54l_05_10_15.dtsi
nordicjm: fixed

include/zephyr/arch/arm/cortex_m/scripts/linker.ld
nordicjm: fixed

include/zephyr/storage/flash_map.h
nordicjm: updated

include/zephyr/devicetree/mapped-partition.h
tomchy: I might've received incorrect results due to the odd MRAM node definition.

PR #3900, opened 2026-03-24
[nrf fromlist] cmake: modules: move DTC_OVERLAY_FILE boilerplate watch

lucasvdp:
Move the zephyr_boilerplate_watch() call for DTC_OVERLAY_FILE from configuration_files.cmake to dts.cmake so that it fires right before the variable is consumed rather than right after it was configured automatically.

This avoids false warnings when DTC_OVERLAY_FILE is modified after the automatic configuration.

Upstream PR #:106193


tejlmand (changes requested):
blocked upstream.

PR #3906, opened 2026-03-25
[nrf fromtree] drivers: serial: nrfx_uarte: Add support for variable frame size

nordic-krch:
Add various frame sizes in UART driver. Extend `uart_async_api` test to check those frame sizes.


nika-nordic (approved):

shanthanordic (approved):
approved for scope

PR #3933, opened 2026-03-31
Cleanup bsp references

magp-nordic:
Cleanup references to BSPs. It is no longer required to define a path to BSP via Kconfigs; BSP should now have its own CMake that will be processed by Zephyr.
Also, added a new Kconfig (`NRF_DONT_USE_STABLE_BSP`) that determines whether the stable BSP should be used or not, and ifdefs to hal_nordic/nrfx/CMakeLists.txt to avoid conflicts between BSP's and hal_nordic/nrfx cmakes. `NRF_DONT_USE_STABLE_BSP` should be selected for all SoCs that are defined in external BSPs, and not selected for SoCs defined in Zephyr.


scripts/ci/undef_kconfig_allowlist.txt
```diff
@@ -163,3 +163,77 @@ ZEPHYR_TRY_MASS_ERASE  # MCUBoot setting described in sysbuild documentation
 ZTEST_FAIL_TEST_  # regex in tests/ztest/fail/CMakeLists.txt
 ZVFS_OPEN_ADD_SIZE_  # Used as an option matching prefix
 # zephyr-keep-sorted-stop
+# NCS-specific allow list
```
nordicjm: this file should probably now just live in ncs without needing to be a noup, would need to look

nordicjm: drop this noup commit

nika-nordic (approved):

adamkondraciuk (approved):

mstasiaknordic (approved):

nordic-segl (approved):

jaz1-nordic (approved):

thst-nordic (approved):

PR #3940, opened 2026-04-01
bluetooth: host: Support nRF Connect SDK Bluetooth Host

alwa-nordic:
This PR adds include redirects and build exclusions for CONFIG_BT_HOST_NORDIC. When enabled, the Zephyr host source and drivers are replaced by the nRF Connect SDK Bluetooth Host from sdk-nrf.

Companion sdk-nrf PR: https://github.com/nrfconnect/sdk-nrf/pull/27863


include/zephyr/bluetooth/addr.h
```diff
@@ -2,6 +2,14 @@
  *  @brief Bluetooth device address definitions and utilities.
  */
 
+#if defined(CONFIG_BT_HOST_NORDIC)
```
aescolar, line 5: Major: These noups would create a really big permanent downmerge burden.

include/zephyr/bluetooth/addr.h
```diff
@@ -2,6 +2,14 @@
...
+
+#define ZEPHYR_INCLUDE_BLUETOOTH_ADDR_H_
+
+#include <bluetooth/host/addr.h>
```
aescolar, line 9: Major: This included header is an exact copy of the zephyr upstream one...
We cannot add a merge burden/noup to do effectively nothing.

This is the case for quite a few of the other headers which are being ifdef'd out in these noups.

include/zephyr/bluetooth/addr.h
```diff
@@ -2,6 +2,14 @@
  *  @brief Bluetooth device address definitions and utilities.
  */
 
+#if defined(CONFIG_BT_HOST_NORDIC)
+
+#define ZEPHYR_INCLUDE_BLUETOOTH_ADDR_H_
```
aescolar, line 7: defining a header guard instead of doing an #else is not something one would expect, it is less evident, and for maintenance (merging) it would seem worse (that header guard may change upstream)

aescolar (changes requested):

include/zephyr/bluetooth/addr.h
KyraLengfeld: As far as I can see there is a clear scheme upstream for header guards though? So I would expect this issue to pop up (if) either through a large scale change or due to files being moved or renamed, so not something that would easily "slip through the crags". This is just a personal opinion, but the nature of this change is that we have the exact same header in another place we use instead, which a header guard is for: to ensure the header isn't included multiple times.

include/zephyr/bluetooth/addr.h
KyraLengfeld: I see the issue (aka "both files were modified", right?), but I am not sure if there is a solution that does not have such a burden. Perhaps we can come up with a script, as these changes are quite uniform.

include/zephyr/bluetooth/addr.h
aescolar: (note this is a minor thing, so I'd not want to lose focus from what matters)
Something like this instead would be a bit briefer and hopefully more evident that it is either one or the other.
```diff
  /*
  * Copyright (c) 2019 Nordic Semiconductor ASA
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-#ifndef ZEPHYR_INCLUDE_BLUETOOTH_ADDR_H_
+#if defined(CONFIG_BT_HOST_NORDIC)
+
+#include <bluetooth/host/addr.h>
+
+#elif !defined(ZEPHYR_INCLUDE_BLUETOOTH_ADDR_H_)
 #define ZEPHYR_INCLUDE_BLUETOOTH_ADDR_H_
 
 #include <stdint.h>
```

include/zephyr/bluetooth/addr.h
aescolar: I would just not do this for any file for which we do not yet need to do it.

PavelVPV: I can't forsee issues with these particular noups as they touch part of file that will unlikely be changed, thus I belive git will manage to resolve conflicts automatically. Please give examples how the auto-mergeing can brake here.

> I would just not do this for any file for which we do not yet need to do it.

Please elaborate why you think we should do as you suggest.

include/zephyr/bluetooth/addr.h
PavelVPV: I believe keeping the public headers in one place (the Nordic host tree in sdk-nrf) while preserving the existing Zephyr include paths is better. This avoids splitting the same API across repos, which in turns increases maintenance burden, reviews and user experience.

include/zephyr/bluetooth/addr.h
PavelVPV: I understand you suggestion, but the activate guard pattern is used on purpose here: when `CONFIG_BT_HOST_NORDIC` is set, it defines `ZEPHYR_INCLUDE_...` and pulls in the Nordic header so the original Zephyr body is skipped without wrapping it in `#else`. This is done to keep the change in a small block at the top of each file and this avoids changing the rest of the headers. I think this should make upmerges less noisy than folding everything under a single `#ifndef - #if - #else`. I doubt the guards will be changed so that this pattern gets broken.

aescolar: With what I proposed above if the guards don't change nothing breaks either. But if they do change, there is a merge conflict during the rebase so it is immediately evident, instead of later.
The change is just those 3 lines in the diff above. There is no extra #if wrapping.

PavelVPV: > instead of later

BLE Host CI will catch this immediately as it runs on any ble host file change.

> The change is just those 3 lines in the diff above. There is no extra #if wrapping.

It is not a question about whether the change is big or not, I understand your suggestion as I wrote above. The question is why we should do this.

aescolar: As mentioned before, this suggestion is a minor thing that would make the code more readable/immediately evident. It is not the reason for the rejection. If you guys have attachment to how you have done it, you can just ignore this suggestion.
The point of the suggestion is to make it clear the other header is included ignoring the remaining, instead of requiring people to realize that first if above the copyright notice, is defining the header guard used below, which is not typical.

include/zephyr/bluetooth/assigned_numbers.h
```diff
@@ -2,6 +2,14 @@
  *  @brief Bluetooth Assigned Numbers, codes and identifiers.
  */
 
+#if defined(CONFIG_BT_HOST_NORDIC)
```
tejlmand, line 5: Which header files that are included depends on the include path order.

If you ensure `<sdk-nrf>/include/bluetooth/host/` are first in path before `<zephyr>/include/bluetooth` and you keep the same include folder structure then you wouldn't need to have those patches.

Folder structure is important because Zephyr code includes bluetooth headers like:
`#include <zephyr/bluetooth/assigned_numbers.h>`

So if you keep the copied headers under a: `zephyr/bluetooth/` structure, and don't add host you can avoid those patches. 

To make things in general a bit safer, you can make a single line patch here:
https://github.com/nrfconnect/sdk-zephyr/blob/47d2ffc20cf00511e022bb075f461f0538a7719c/subsys/bluetooth/CMakeLists.txt#L18

where you only add this folder as include dir if `!CONFIG_BT_HOST_NORDIC`.

Note, the CMake part will not ensure the handling when includes are done like: `#include <zephyr/bluetooth/assigned_numbers.h>`. For that, the correct include dir order is important.

subsys/bluetooth/CMakeLists.txt
```diff
@@ -4,17 +4,23 @@ add_library(subsys__bluetooth INTERFACE)
 
 target_include_directories(subsys__bluetooth INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
 
-add_subdirectory(common)
-add_subdirectory_ifdef(CONFIG_BT_HCI host)
+if(NOT CONFIG_BT_HOST_NORDIC)
+  add_subdirectory(common)
+  add_subdirectory_ifdef(CONFIG_BT_HCI host)
```
tejlmand, line 9: this indicates that `BT_HOST_NORDIC` and `BT_HCI` are not mutual exclusive.

Shouldn't `BT_HOST_NORDIC` not be a part of the choice ?
Then that would make them mutual exclusive, and the `if` would only need to cover the common, which could then be done as:
```
add_subdirectory_ifndef(CONFIG_BT_HOST_NORDIC common)
```
and thus a smaller and less risk of merge conflicting.

subsys/bluetooth/CMakeLists.txt
```diff
@@ -4,17 +4,23 @@ add_library(subsys__bluetooth INTERFACE)
...
 if(CONFIG_BT_LL_SW_SPLIT)
   add_subdirectory(controller)
 endif()
 
 zephyr_include_directories(${ZEPHYR_BASE}/subsys/bluetooth)
 
-add_subdirectory(lib)
+if(NOT CONFIG_BT_HOST_NORDIC)
+  add_subdirectory(lib)
+endif()
```
tejlmand, line 26: single line changes has less risk of merge conflicting, and when it does happen, is usually easier to resolve:
```suggestion
add_subdirectory_ifndef(CONFIG_BT_HOST_NORDIC lib)
```

include/zephyr/bluetooth/assigned_numbers.h
PavelVPV: Zephyr [include](https://github.com/nrfconnect/sdk-zephyr/blob/47d2ffc20cf00511e022bb075f461f0538a7719c/CMakeLists.txt#L125-L130) goes before [sdk-nrf one](https://github.com/nrfconnect/sdk-nrf/blob/e267d1a1dfed944907d172fcf8d2e4a045a50c7f/CMakeLists.txt#L29) and the header files are included just through the include folder. We can't dictated in CMake to include host files in sdk-nrf first, then zephyr include and then rest sdk-nrf include folders. Exactly because of this issue, this patch is added in every file. And we can't change order of sdk-nrf and sdk-zephyr includes (we can, but this is not a solution, I would say).

We can't exclude `zephyr_include_directories(${ZEPHYR_BASE}/subsys/bluetooth)` if `!CONFIG_BT_HOST_NORDIC` because mesh and audio rely on this include.

subsys/bluetooth/CMakeLists.txt
PavelVPV: BT_HCI is plays bigger role than just enabling/disabling specific Host as it kind of includes both, Host and HCI interface with all the dependencies if I'm not mistaken. But ideally yes, it should be mutually exclusive, like with controllers, but this requires changing Kconfigs on Zephyr side first.

subsys/bluetooth/host/hci_core.c
```diff
@@ -4087,6 +4087,7 @@ static void bt_dev_show_info(void)
 {
 	int i;
 
+	LOG_INF("Zephyr Bluetooth Host");
```
aescolar, line 4090: Have you tried getting this change upstream? (noups should be a last resort)

KyraLengfeld: No, we haven't, but this is used during testing of the NCS host whether we compile the right host (hence a difference in logs to upstream is useful). I am thinking this is really a minor thing though, so removing this log line or adding a general one upstream is fine for us to do.

shanthanordic (changes requested):
Want to align on scope

PR #3942, opened 2026-04-02
[nrf fromlist] drivers: mbox: nrf: retrigger VEVIF task TX on nRF54LM20

jaz1-nordic:
Back-to-back VEVIF task notifications on nRF54L could be collapsed into a single observed event, which caused the second mailbox packet to be missed by the remote side. Extend the existing delayed retrigger workaround used on nRF54H20 so task-based mailbox sends are delivered reliably on nRF54LM20 as well.

Upstream PR #: [106782](https://github.com/zephyrproject-rtos/zephyr/pull/106782)

manifest-pr-skip

### DIFF
--- a/drivers/mbox/mbox_nrf_vevif_task_tx.c
+++ b/drivers/mbox/mbox_nrf_vevif_task_tx.c
@@ -41,11 +41,11 @@ static int vevif_task_tx_send(const struct device *dev, uint32_t id, const struc
 
 	nrfy_vpr_task_trigger(config->vpr, nrfy_vpr_trigger_task_get(id));
 
-#ifdef CONFIG_SOC_NRF54H20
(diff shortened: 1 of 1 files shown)


drivers/mbox/mbox_nrf_vevif_task_tx.c
```diff
@@ -41,11 +41,11 @@ static int vevif_task_tx_send(const struct device *dev, uint32_t id, const struc
 
 	nrfy_vpr_task_trigger(config->vpr, nrfy_vpr_trigger_task_get(id));
 
-#ifdef CONFIG_SOC_NRF54H20
+#if defined(CONFIG_SOC_NRF54H20) || defined(CONFIG_SOC_NRF54LM20A) || defined(CONFIG_SOC_NRF54LM20B)
```
nordic-piks, line 44: Can it be changed to feature flag?
Are other chips affected? like L15 and LV10?

PR #3957, opened 2026-04-09
Revert "[nrf fromlist] boards: nordic: nrf54l15dk: Fix missing flash …

nordicjm:
…runner groups"

This reverts commit fe2e7d0129fa5d4ab4dbd13e0066e8cf02558ef9.


boards/nordic/nrf54l15dk/board.yml
```diff
@@ -23,9 +23,6 @@ runners:
         run: first
         groups:
           - boards:
```
tomchy, line 25: This change seems to be unrelated to the removal of `ranges;` property.

scripts/kconfig/kconfigfunctions.py
```diff
@@ -384,10 +374,13 @@ def dt_chosen_reg(kconf, name, chosen, index=0, unit=None):
         return hex(_dt_chosen_reg_addr(kconf, chosen, index, unit))
 
 
-def _dt_chosen_partition_addr(kconf, chosen, unit=None):
+def _dt_chosen_partition_addr(kconf, chosen, index=0, unit=None):
```
tomchy, line 377: Isn't this the issue, from which all of this started?
I mean - this function does not seem to work for the `fixed-subpartitions` nodes.

https://github.com/zephyrproject-rtos/zephyr/pull/102730

tomchy (changes requested):

scripts/kconfig/kconfigfunctions.py
nordicjm: Yes, and so be it, that is no different to previous versions of NCS, it will be fixed in future released with mapped bindings

boards/nordic/nrf54l15dk/board.yml
nordicjm: this is a squash of multiple commits, the commit details are wrong, hence why it is marked as draft, will need to check that these are all correct later

scripts/kconfig/kconfigfunctions.py
tomchy: If you do not want to fix it here, you need to remove the usage of `fixed-subpartitions` as the code partition in the sdk-nrf in all samples.

nordicjm: Removed

tomchy (approved):

de-nordic (approved):

PR #3998, opened 2026-04-17
[nrf noup] drivers: nrf_mram: Fix write-block-size != 16 handling

karstenkoenig:
We support writing to unaligned mram locations and sizes, so the asserts were invalid.
That also means addr can't be constant as it gets updated when reworking the unaligned arguments into aligned accesses via read-modify-write steps.
The unlock label can now also be reached from a mram-latency unrelated path, so removed the ifdef for the label.

### DIFF
--- a/drivers/flash/soc_flash_nrf_mram.c
+++ b/drivers/flash/soc_flash_nrf_mram.c
@@ -38,10 +38,6 @@ LOG_MODULE_REGISTER(flash_nrf_mram, CONFIG_FLASH_LOG_LEVEL);
 #define IRONSIDE_SE_SUPPORT_READY_VER IRONSIDE_SE_V23_1_2_21
 
 BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero");
-BUILD_ASSERT((WRITE_BLOCK_SIZE % MRAM_WORD_SIZE) == 0,
-	     "write-block-size expected to be a multiple of MRAM_WORD_SIZE");
-BUILD_ASSERT((ERASE_BLOCK_SIZE % WRITE_BLOCK_SIZE) == 0,
-	     "erase-block-size expected to be a multiple of write-block-size");
 
 struct nrf_mram_data_t {
 	uint8_t ironside_se_ver;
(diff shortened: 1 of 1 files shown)


jonathannilsen:
Would it be better to revert the existing no-up and reapply it with this patch?

57300:
> Would it be better to revert the existing no-up and reapply it with this patch?

Or, you can add this line to the new commit:
```
nrf-squash! [nrf noup] drivers: nrf_mram: handle non aligned write-block-size
```

karstenkoenig:
> > Would it be better to revert the existing no-up and reapply it with this patch?
> 
> Or, you can add this line to the new commit:
> 
> ```
> nrf-squash! [nrf noup] drivers: nrf_mram: handle non aligned write-block-size
> ```

done after messing up initially :smile:

carlescufi (approved):

tomchy (approved):

de-nordic (approved):
OK
But I do not agree with the simplicity of the statement "We support writing to unaligned mram locations and sizes, so the asserts were invalid.", because it is half-true, the wbs is there for a reason and it is baseline for endurance. The check is for ensuring baseline endurance as specified for full word writes.

This is the same issue as eeproms are sold with, where endurance is stated often as some 1M writes per word and writes can be done with one byte alignment; both parts of a sentence are true but once you start writing by on byte, you are rewriting entire word and the endurance is cut by the number of bytes in word. Example would be 25aa512 with 1M write endurance and 1 byte alignment, but each write rewrites 128 page, and endurance is calculated for page writes.

karstenkoenig:
> OK But I do not agree with the simplicity of the statement "We support writing to unaligned mram locations and sizes, so the asserts were invalid.", because it is half-true, the wbs is there for a reason and it is baseline for endurance. The check is for ensuring baseline endurance as specified for full word writes.
> 
> This is the same issue as eeproms are sold with, where endurance is stated often as some 1M writes per word and writes can be done with one byte alignment; both parts of a sentence are true but once you start writing by on byte, you are rewriting entire word and the endurance is cut by the number of bytes in word. Example would be 25aa512 with 1M write endurance and 1 byte alignment, but each write rewrites 128 page, and endurance is calculated for page writes.

I didn't mean to downplay the effects and problems with read-modify-write, you are right they are fairly severe. It is just that we decided to carry support for read-modify-write here and hence block it via a build_assert is wrong

PR #4028, opened 2026-04-28
[nrf fromtree] tests: net: socket: tls: Improve tests interoperability

rlubos:
Update TLS tests configuration and runtime socket configuration to improve tests reliability regardless of platform used:

* Disable unneeded ciphersuite to reduce the number of mbed TLS features enabled an thus overall image size
* Make sure tests always use the same ciphersuite, regardless of extra mbed TLS features enabled (that was the case with TFM which enabled some extra mbed TLS features by default)

Ref: NCSDK-39085


tomi-font (approved):

shanthanordic (approved):
approved for scope

PR #4039, opened 2026-05-04
modules: mbedtls: update to v4.1.0

tomi-font:
- Revert noups that were made to stick to Mbed TLS 3.6 in the Zephyr upmerge.
- Convert fromlists to fromtrees.
- Cherry-pick additional needed changes from upstream.

manifest-pr-skip


carlescufi:
@rlubos @Vge0rge could you please review?

rlubos (approved):

rlubos:
Approved, but have to make one comment (non-blocking one as this could just be my preference). 

There are many `noup` commits that actually disguise `fromtree`/`fromlist` commits that do not apply cleanly. IMO such approach just makes it harder for whoever makes the next upmerge as those won't be identified automatically and would require manual (often non-obvious as it's usually just a conflict) inspection, each. I'd personally pretty much prefer to tag those commits properly and just accept the fact we have to ignore "Commit tags" check.

tomi-font:
> Approved, but have to make one comment (non-blocking one as this could just be my preference).
> 
> There are many `noup` commits that actually disguise `fromtree`/`fromlist` commits that do not apply cleanly. IMO such approach just makes it harder for whoever makes the next upmerge as those won't be identified automatically and would require manual (often non-obvious as it's usually just a conflict) inspection, each. I'd personally pretty much prefer to tag those commits properly and just accept the fact we have to ignore "Commit tags" check.

Sure then, if we can bypass the CI check I'm happy to convert them. The reason I turned them into noups is to make the check happy.

anangl (approved):

Vge0rge:
Looks good,  legacy mbedtls related noups which got dropped make sense.
The reapplied noups are ok, one minor change in the commit: 
"[nrf noup] mbedtls: Remove unsupported algorithms in PSA crypto"
make sense since it removes more unsupported algorithms in the benchmark. 

Fromlist/fromtrees should be checked by CI that they are good so it should be alright.

Vge0rge (approved):

PR #4048, opened 2026-05-07
[nrf fromlist] debug: mipi_stp_decoder: Refactor to avoid unaligned 6…

nordic-krch:
…4 bit access

When optimized for speed it was possible that compiler for ARM Cortex-M used double word read for code that was reading two words from an array. However, it was possible that this array was not word-aligned and double word access was resulting in a bus fault. Refactored the function to ensure that only word access is used (word access on Cortex-M4 and later supports unaligned access).

Upstream PR #: 108644


nika-nordic (approved):

shanthanordic (approved):
approved for scope

PR #4057, opened 2026-05-11
[nrf fromtree] bluetooth: mesh: fix BLOB IO flash write alignment for…

alxelax:
… non-erase devices

The wr_chunk function had a fast path for devices without explicit erase (e.g. nRF54L RRAM) that wrote chunk data directly without aligning the offset or size to the flash write block size. On RRAM with write-block-size=16, this caused -EINVAL from the flash driver for every chunk write since chunk_size (161) and most chunk offsets are not 16-byte aligned.

The failed writes caused the BLOB server to never mark chunks as received, resulting in block_status_rsp always reporting all chunks missing. This led to infinite retransmission of block 0 and eventual DFU timeout.

Fix by using a read-modify-write approach for non-erase devices: read existing data at the aligned boundaries, overlay the chunk data, and write back the full aligned buffer. This satisfies the driver's alignment requirements while preserving neighboring bytes.

(cherry picked from commit a9ad258c844928d59d48aeef94661a274357ced0)

### DIFF
--- a/subsys/bluetooth/mesh/blob_io_flash.c
+++ b/subsys/bluetooth/mesh/blob_io_flash.c
@@ -143,17 +143,6 @@ static int wr_chunk(const struct bt_mesh_blob_io *io,
 	const struct flash_parameters *fparam = flash_get_parameters(fdev);
 
 	/*
-	 * If device has no erase requirement then write directly.
-	 * This is required since trick with padding using the erase value will
-	 * not work in this case.
-	 */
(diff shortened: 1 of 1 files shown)


omkar3141 (approved):

ludvigsj (approved):

shanthanordic (approved):

shanthanordic:
approved for scope,

PR #4060, opened 2026-05-11
[nrf fromtree] drivers: nrf_qspi_nor: Fix EXTXIP + Flash write race c…

nvlsianpu:
…ondition

When parts of code are running from EXTXIP and at the same time QSPI external flash writes/erases are happenning there is a high risk of a race condition where a thread running EXTXIP code starts executing while an external flash operation is in progress; This can lead to execution stalling or bus faults/usage faults.

The issue was observed when parts of MCUMGR were relocated to EXTXIP and DFU was performed.

To fix the issue added the CONFIG_NORDIC_QSPI_NOR_XIP_FLASH_SCHED_LOCK option which should be used when executing parts of code from EXTXIP. It ensures the blocking variant of nrfx_qspi_ is used and that no preemption happens while an external flash write is in progress (by making qspi_wait_while_writing blocking)

Signed-off-by: Artur Hadasz <artur.hadasz@nordicsemi.no>

Upstream PR #: 108747

ref.: NCSIDB-1895, NCSDK-39208


anangl (approved):

shanthanordic (approved):
approved for scope

anangl:
rebased

tejlmand:
please rebase.

PR #4088, opened 2026-05-19
[nrf fromtree] drivers: regulator: nrf_vregusb: fix USB not working a…

MarekPieta:
…fter warm reboot

After sys_reboot or debugger reset on nRF54LM20, USB controller fails to initialize because the VREGUSB peripheral retains state across warm resets. Two issues are addressed:

1. Stale VBUS_DETECTED/VBUS_REMOVED events from the previous boot remain set in the event registers. When interrupts are enabled without clearing them first, the ISR fires immediately with potentially wrong event type (VBUS_REMOVED can overwrite VBUS_DETECTED in the handler).

2. When USB cable stays connected across a reboot, no new VBUS_DETECTED edge occurs because the peripheral was never stopped. The regulator callback never fires, so the wrapper never posts NRF_USBHS_PHY_READY, and usbhs_enable_core() blocks indefinitely or times out.

Fix by clearing stale events before enabling interrupts, and by reading the undocumented STATUS register (base + 0x400, bit 2) after TASK_START to detect VBUS-already-present and synthesize the REGULATOR_VOLTAGE_DETECTED callback. This matches the approach used in the original inline vendor quirks added in commit 66b938f93a40 ("drivers: udc_dwc2: Add nRF54LM20A vendor quirks") before the refactoring to a separate regulator driver.

(cherry picked from commit f72c0cd1ee542d9f2ae5c9f96886a47f91ed4ac2)

### DIFF
--- a/drivers/regulator/regulator_nrf_vregusb.c
+++ b/drivers/regulator/regulator_nrf_vregusb.c
@@ -8,6 +8,7 @@
 #include <stdint.h>
 #include <zephyr/drivers/regulator.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/sys/sys_io.h>
 #include <hal/nrf_vregusb.h>
 
 LOG_MODULE_REGISTER(vregusb, CONFIG_REGULATOR_LOG_LEVEL);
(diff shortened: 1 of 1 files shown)


tmon-nordic (approved):

jema-nordic (approved):

shanthanordic (approved):
approved for scope

PR #4089, opened 2026-05-19
[nrf fromlist] drivers: serial: nrfx_uarte: Fix instance initializati…

nordic-krch:
…on priority

If UARTE instance is using GPPI during the initialization and GPPI is using Ironside then that instance need to be initialized once Ironside and GPPI is initialized. Delay initialization to POST_KERNEL and derive init priority from Ironside call init priority (+2 as +1 is used by GPPI).

Upstream PR #: 109400

### DIFF
--- a/drivers/serial/uart_nrfx_uarte.c
+++ b/drivers/serial/uart_nrfx_uarte.c
@@ -3337,6 +3337,31 @@ static int uarte_instance_deinit(const struct device *dev)
 		)							       \
 	)
 
+/* If UARTE instance is using GPPI during the initialization and GPPI is using
+ * Ironside then that instance needs to be initialized once Ironside and GPPI is
+ * ready. GPPI is initialized with Ironside priority +1 so UARTE instance
+ * is initialized just after that.
+ *
+ * Macro determines if delayed initialization needs to be applied.
+ */
+#define UARTE_INIT_AFTER_GPPI(idx)					       \
+	COND_CODE_1(UTIL_AND(CONFIG_UARTE_NRFX_UARTE_COUNT_BYTES_WITH_TIMER,   \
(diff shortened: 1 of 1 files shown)


nordic-segl (approved):

drivers/serial/uart_nrfx_uarte.c
```diff
@@ -3337,6 +3337,31 @@ static int uarte_instance_deinit(const struct device *dev)
...
+ */
+#define UARTE_INIT_AFTER_GPPI(idx)					       \
+	COND_CODE_1(UTIL_AND(CONFIG_UARTE_NRFX_UARTE_COUNT_BYTES_WITH_TIMER,   \
+			     UARTE_HAS_PROP(idx, timer)),		       \
+		    (UTIL_AND(IS_ENABLED(CONFIG_NRFX_GPPI_SD2PPI_GLOBAL),      \
+			      IS_ENABLED(CONFIG_IRONSIDE_SE_CALL))), (0))
+
+/* Init phase is delayed to POST_KERNEL if it relies on Ironside+GPPI being ready. */
+#define UARTE_INIT_PHASE(idx)						       \
+	COND_CODE_1(UARTE_INIT_AFTER_GPPI(idx), (POST_KERNEL), (PRE_KERNEL_1))
```
nika-nordic, line 3355: why not always init in `post_kernel` ? because of other dependencies like logger?

nika-nordic (approved):

drivers/serial/uart_nrfx_uarte.c
nordic-krch: we usually want to have uart as early as possible, for printk, logging etc..

shanthanordic (commented):
approved for scope

PR #4100, opened 2026-05-20
Fixes required for pts qualification

m-alperen-sener:
These two commits are needed for PTS qualification.

Those commits are already in main after latest upmerged but also needed here for tests cases:
L2CAP/COS/CED/BI-29-C
GAP/CONN/DCON/BV-05-C


shanthanordic (approved):
approved for scope

cvinayak (approved):

PR #4102, opened 2026-05-20
[nrf noup] images: bootloader: Kconfig: RSA phaseout

michalek-no:
remove RSA from availible signature types


nordicjm (changes requested):
noup for this is not acceptable

PR #4108, opened 2026-05-22
net: l2: wifi: Add Kconfig option to control automatic DHCPv4 start

krish2718:
Add a new Kconfig option WIFI_STA_AUTO_DHCPV4 to allow applications to control whether DHCPv4 client should be automatically started by the Wi-Fi driver or supplicant after STA connection is established. This addresses the use case where applications need to switch between static IP addressing and DHCP, or prefer to manually control when the DHCP client starts.
When CONFIG_WIFI_STA_AUTO_DHCPV4=n, the application layer becomes responsible for either manually starting the DHCPv4 client or configuring a static IPv4 address after Wi-Fi connection is established. The option defaults to 'y' to maintain backward compatibility with existing behavior.

manifest-pr-skip

### DIFF
--- a/subsys/net/l2/wifi/Kconfig
+++ b/subsys/net/l2/wifi/Kconfig
@@ -165,6 +165,18 @@ endif # WIFI_SHELL_RUNTIME_CERTIFICATES
 
 endif # WIFI_NM_WPA_SUPPLICANT_CRYPTO_ENTERPRISE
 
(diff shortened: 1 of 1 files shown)


PavelVPV (approved):
I confirm that this fixes ipv4 discovery issue on Wi-Fi.

rado17 (approved):

shanthanordic (approved):
approve for scope

PR #4130, opened 2026-05-28
[nrf noup] drivers: nrf_mram: handle non aligned write-block-size

rghaddab:
When write-block-size is not aligned to the mram-word-size source addresses could be non aligned to the start of an mram word. To write correctly the data, a read-modify-write operation needs to be performed for data that is partially written in an mram word.

### DIFF
--- a/drivers/flash/soc_flash_nrf_mram.c
+++ b/drivers/flash/soc_flash_nrf_mram.c
@@ -40,10 +40,6 @@ LOG_MODULE_REGISTER(flash_nrf_mram, CONFIG_FLASH_LOG_LEVEL);
 #define IRONSIDE_SE_SUPPORT_READY_VER IRONSIDE_SE_V23_1_2_21
 
 BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero");
-BUILD_ASSERT((WRITE_BLOCK_SIZE % MRAM_WORD_SIZE) == 0,
-	     "write-block-size expected to be a multiple of MRAM_WORD_SIZE");
-BUILD_ASSERT((ERASE_BLOCK_SIZE % WRITE_BLOCK_SIZE) == 0,
-	     "erase-block-size expected to be a multiple of write-block-size");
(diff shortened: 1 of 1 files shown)


drivers/flash/soc_flash_nrf_mram.c
```diff
@@ -40,10 +40,6 @@ LOG_MODULE_REGISTER(flash_nrf_mram, CONFIG_FLASH_LOG_LEVEL);
 #define IRONSIDE_SE_SUPPORT_READY_VER IRONSIDE_SE_V23_1_2_21
 
 BUILD_ASSERT(MRAM_START > 0, "nordic,mram: start address expected to be non-zero");
-BUILD_ASSERT((WRITE_BLOCK_SIZE % MRAM_WORD_SIZE) == 0,
```
karstenkoenig, line 43: Wasn't there a warning instead of error at some point? I think it is fair to throw the warning to at least try to indicate the problem of a read/modify/write cycle getting introduced

karstenkoenig (commented):
This is just a copy of the previous one right? Or does this need a thorough review?

drivers/flash/soc_flash_nrf_mram.c
rghaddab: @karstenkoenig I haven't modified the previously merged PR in 3.3-branch.
I just squashed your commit with mine.
Let me double check if there is a warning in 3.3-branch

rghaddab: @karstenkoenig we didn't add a warning in the previously merged PR.
Do you think it is worth adding it ?

karstenkoenig (approved):

drivers/flash/soc_flash_nrf_mram.c
karstenkoenig: I'd prefer one yes, at least it can highlight a potential risk to the customer

carlescufi (approved):

PR #4148, opened 2026-06-01
nrf71: Increase RAM to accommodate Wi-Fi when TF-M is enabled

krish2718:
*  Increase RAM to accommodate Wi-Fi when TF-M is enabled
* Build error with TF-M

manifest-pr-skip

### DIFF
--- a/dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
+++ b/dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
@@ -57,7 +57,7 @@
 		 * needs to happen both in the flash_layout.h and in this file at the same time.
 		 */
 
-		slot0_partition: partition@0 {
+		slot0_s_partition: partition@0 {
 			compatible = "zephyr,mapped-partition";
 			label = "image-0";
 			reg = <0x0000000 DT_SIZE_K(512)>;
--- a/dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
+++ b/dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
@@ -17,7 +17,7 @@
  * needs to happen both in the flash_layout.h and in this file at the same time.
  */
 &cpuapp_sram {
-	sram0_s: image_s@0 {
+	sram0_s: sram@0 {
 		#address-cells = <1>;
 		#size-cells = <1>;
 		/* Secure image memory */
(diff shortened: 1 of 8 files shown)


dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
```diff
@@ -8,29 +8,28 @@
...
 	};
+};
 
-	sram0_ns: image_ns@40000 {
-		#address-cells = <1>;
-		#size-cells = <1>;
-		/* Non-Secure image memory */
-		reg = <0x40000 DT_SIZE_K(256)>;
-		ranges = <0x0 0x40000 DT_SIZE_K(256)>;
+/ {
```
rob-robinson-14: I am not sure RAM01 should be used here by default (i.e. top 256kB). Total RAM for application should be 512kB

dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
```diff
@@ -8,29 +8,28 @@
...
  */
 &cpuapp_sram {
-	sram0_s: image_s@0 {
+	sram0_s: sram@0 {
 		#address-cells = <1>;
 		#size-cells = <1>;
 		/* Secure image memory */
-		reg = <0x0 DT_SIZE_K(256)>;
-		ranges = <0x0 0x0 DT_SIZE_K(256)>;
+		reg = <0x0 DT_SIZE_K(64)>;
```
travis3630: Can you keep it at 128Kb for TF-m to accomodate future changes

dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
krish2718: We can do that, and leave 384K for the NS parts, and leave it to per-sample in case they want to reclaim RAM01 memory.

simonduq: It would be good to have a chat soon on default RAM allocation on Wezen.

Today with TFM the net samples with (D)TLS stacks are absolutely at the limit. Samples needs tweaking to fit into the available partition, and things are fragile (small changes result in a sample running out of RAM).

simonduq (approved):

rob-robinson-14 (approved):

krish2718:
Changes merged as part of https://github.com/nrfconnect/sdk-zephyr/pull/4196 , closing

PR #4170, opened 2026-06-04
[nrf noup] tests: drivers: spi: add nrfx SPIM no-multithreading build test

thst-nordic:
## Summary

Adds a build-only Twister test (`drivers.spi.nrfx_spim.no_multithreading`) that links the `nordic,nrf-spim` driver with `CONFIG_MULTITHREADING=n`, matching MCUboot-style images (e.g. Matter sysbuild MCUboot on nRF54 with external SPI NOR).

Catches linker failures such as `undefined reference to z_impl_k_sem_init` / `z_impl_k_sem_give` when SPIM wake synchronization used kernel semaphores without multithreading enabled.

Depends on the merged driver fix in #4165 (`spi_nrfx_spim` `wake_atom` path).

## Test plan

- [ ] `./scripts/twister -T tests/drivers/spi/spi_nrfx_spim_no_multithreading -p nrf54lm20dk/nrf54lm20a/cpuapp --build-only`
- [ ] CI Twister on Nordic platforms in `platform_allow`

### DIFF
--- a/tests/drivers/spi/spi_nrfx_spim_no_multithreading/src/main.c
+++ b/tests/drivers/spi/spi_nrfx_spim_no_multithreading/src/main.c
@@ -1,0 +1,25 @@
+/*
+ * Copyright (c) 2025 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/kernel.h>
+
+#if DT_NODE_HAS_STATUS(DT_NODELABEL(spi00), okay)
+
+static const struct device *const spi_bus = DEVICE_DT_GET(DT_NODELABEL(spi00));
+
+int main(void)
+{
+	if (!device_is_ready(spi_bus)) {
+		return -ENODEV;
+	}
+
+	return 0;
+}
+
+#else
+#error "This test requires an enabled spi00 (nordic,nrf-spim) devicetree node"
+#endif
--- a/tests/drivers/spi/spi_nrfx_spim_no_multithreading/CMakeLists.txt
+++ b/tests/drivers/spi/spi_nrfx_spim_no_multithreading/CMakeLists.txt
@@ -1,0 +1,8 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+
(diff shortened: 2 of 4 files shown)


tests/drivers/spi/spi_nrfx_spim_no_multithreading/src/main.c
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2025 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/kernel.h>
+
+#if DT_NODE_HAS_STATUS(DT_NODELABEL(spi00), okay)
```
bjarki-andreasen, line 10: You can do
```
#if DT_HAS_COMPAT_STATUS_OKAY(nordic_nrf_spim)

static const struct device *const spi_bus =
        DEVICE_DT_GET(DT_COMPAT_GET_ANY_STATUS_OKAY(nordic_nrf_spim));

...

#endif
```

To get an instance if present, that way the test is not relying on the `spi00` nodelabel specifically, it could be any enabled spi instance like `spi3` on an nrf5340.

tests/drivers/spi/spi_nrfx_spim_no_multithreading/src/main.c
```diff
@@ -0,0 +1,25 @@
...
+{
+	if (!device_is_ready(spi_bus)) {
+		return -ENODEV;
+	}
+
+	return 0;
+}
+
+#else
+#error "This test requires an enabled spi00 (nordic,nrf-spim) devicetree node"
```
bjarki-andreasen, line 24: This could be updated to "This test requires an enabled node with compatible nordic,nrf-spim"

bjarki-andreasen (approved):

nordic-piks (changes requested):
I do not think noup is needed here.
This no multithreading feature was introduced at upstream, thus the test should be there.
Any of this can be extended https://github.com/nrfconnect/sdk-zephyr/tree/main/tests/drivers/spi (ie its upstream version).
If we want to have this in NCS, then better place is https://github.com/nrfconnect/sdk-nrf/tree/main/tests/zephyr/drivers/spi , ie use existing code of some spi tests from zephyr but compile it with no multithreading option.

PR #4173, opened 2026-06-05
[nrf fromlist] drivers: spi: nrfx_spim: chunk transfers exceeding Eas…

krish2718:
…yDMA MAXCNT

The reworked SPIM driver passes the whole continuous chunk returned by spi_context_max_continuous_chunk() straight to the common transfer layer, which rejects anything larger than the controller MAXCNT with -EINVAL. On SoCs with a small EasyDMA MAXCNT (e.g. nRF91 SPIM with 13 bits, max 8191 bytes) a transfer of 8192 bytes or more fails outright.

The old monolithic driver used to cap the chunk to the MAXCNT limit and let the event handler walk the rest of the buffer. That behaviour was lost in the rework, so any caller pushing a buffer larger than MAXCNT now breaks. One example is the nRF70 firmware patch download, which copies the patch in 8192 byte chunks and ends up corrupting RPU memory, leading to a boot signature mismatch.

Restore the cap in transfer_start() so an oversized chunk is split across multiple EasyDMA transfers, with the remainder handled by the next call from the event handler once the current chunk completes.

Upstream PR #: 110621

Assisted-by: Cursor:Opus4.8

manifest-pr-skip

### DIFF
--- a/drivers/spi/spi_nrfx_spim.c
+++ b/drivers/spi/spi_nrfx_spim.c
@@ -43,6 +43,7 @@ static void transfer_end(const struct device *dev, int ret)
 static void transfer_start(const struct device *dev)
 {
 	struct driver_data *dev_data = dev->data;
+	const struct driver_config *dev_config = dev->config;
 	size_t chunk_len;
 	const uint8_t *tx_buf;
 	size_t tx_buf_len;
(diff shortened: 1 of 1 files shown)


sachinthegreen (approved):

bjarki-andreasen (approved):

krish2718:
The SPIM driver (and others) are different in v3.4-branch and the chunking mechanism is there in v3.4-branch, so, backporting of this fix is not applicable.

PR #4188, opened 2026-06-10
[nrf fromlist] drivers: spi: nrf_spim_common: init sck pin using nrf …

bjarki-andreasen:
…gpio

The SPI SCK pin needs to be initialized to match the current SPI transfer CPOL setting. The SPIM peripheral is enabled after CS is toggled, thus all pins need to be configured to drive correct initial state using gpio. Since the SCK pin is defined in pinctrl rather than as a gpio pin, nrfy is used to get the pin number and set the pin rather than gpio peripheral driver.

Without this addition, CPOL=1 may cause an incorrect clock to be registered at the start of the transfer, shifting the data 1 bit.

Signed-off-by: Bjarki Arge Andreasen <bjarki.andreasen@nordicsemi.no>

Upstream PR #: 110942


MarekPieta (approved):
PWM3360 issue in nRF Desktop is fixed. Thanks!

nika-nordic (approved):

PR #4199, opened 2026-06-15
hostap: Legacy crypto rejig

krish2718:
Fix WPA-PSK association needed for quicktrack. While at it, rejig the legacy crypto, move WEP/TKIP (RC4) or HMAC-MD5 (WPA) to Legacy crypto, keep defaults to ensure the PR is backwards compatible, but also mark these as both deprecated and not secure symbol, in the future WEP symbol will be removed and we will disable legacy crypto by default (with enough notice) to make OOB experience secure.

manifest-pr-skip


kapbh (approved):

sachinthegreen (approved):

krish2718:
FYI, a bit of history cleanup as few commits that were backported had additional changes compared to main, so, this syncs them both for easier cherry-picks.

PR #4215, opened 2026-06-18
[v3.4-branch] Fix Wi-Fi WEP regression

rlubos:
Re-cherry-pick the commit to align with the upstream and fix the build failures


krish2718 (approved):
Thanks.

PR #4218, opened 2026-06-18
[nrf fromlist] kconfig: Allow to get partition MTD path

tomchy:
The partition MTD path is required to correctly check if two partitions overlap with each other.

Upstream PR #: 111599

### DIFF
--- a/scripts/kconfig/kconfigfunctions.py
+++ b/scripts/kconfig/kconfigfunctions.py
@@ -1118,6 +1118,32 @@ def dt_highest_controller_irq_number(kconfig, _, path, irq_cell_name):
 
     return str(max(irqns))
 
+def dt_partition_mtd(kconf, _, path):
+    """
+    This function takes a 'path' and returns the mtd node for that path.
+    """
+    if doc_mode or edt is None:
+        return ""
+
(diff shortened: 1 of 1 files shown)


de-nordic (approved):

scripts/kconfig/kconfigfunctions.py
```diff
@@ -1118,6 +1118,32 @@ def dt_highest_controller_irq_number(kconfig, _, path, irq_cell_name):
...
+    except edtlib.EDTError:
+        return ""
+
+    if node is None:
+        return ""
+
+    # Look for the parent node compatible with "soc-nv-flash"
+    parent = node.parent
+
+    while parent and "soc-nv-flash" not in parent.compats:
```
nvlsianpu, line 1139: is it only node binding we need to support?

nvlsianpu (approved):

scripts/kconfig/kconfigfunctions.py
tomchy: Based on `zephyr/scripts/dts/gen_defines.py` to match the logic behind `mapped-partition` compatible.

PR #4224, opened 2026-06-19
tree: Revert partition manager patches

nordicjm:
manifest-pr-skip


include/zephyr/arch/arm/cortex_m/scripts/linker.ld
```diff
@@ -97,23 +64,6 @@ _image_1_primary_slot_id = PM_S1_ID;
 #define RAM_ADDR DT_CHOSEN_SRAM_ADDR
 #define RAM_SIZE DT_CHOSEN_SRAM_SIZE
 
-#endif /* USE_PARTITION_MANAGER */
-
-#if DT_NODE_HAS_STATUS(DT_CHOSEN(zephyr_ccm), okay)
```
tomchy, line 102: I do not know how ccm/itcm/dtcm relate to the partition manager, possible too much was removed here?

nordicjm: Nope, those do not exist in the zephyr file, it's likely that it was messed up being reapplied during an upmerge with a conflict

tomchy (approved):

nordic-segl (approved):

PR #4232, opened 2026-06-24
nordic: nrf7120: initial support for tfm mcuboot ncs downstream

travis3630:
Cherry-picking: upstream commits

Changes:

1. Disable WiFi to be booted in MCUboot code.
2. Remove BUILD_WITH_TFM and TFM_FLASH_MERGED_BINARY kconfig in defconfig as it is not needed, since it will be enabled by the build system.
3. Utilise full flash allocation for NS application, as currently it did not make full use of the NS application flash.
4. Remove Wi-Fi patch reserved area as alternative solutions will be propose in the future.
5. Reduce sram_s allocation for tfm secure to 96KiB for furtehr NS app RAM free up.

Related manifest:
https://github.com/nrfconnect/sdk-trusted-firmware-m/pull/259
https://github.com/nrfconnect/sdk-nrf/pull/29836

### DIFF
--- a/boards/nordic/nrf7120dk/Kconfig.defconfig
+++ b/boards/nordic/nrf7120dk/Kconfig.defconfig
@@ -19,15 +19,4 @@ config BOARD_NRF7120DK
 config HAS_BT_CTLR
 	default BT
 
-# By default, if we build for a Non-Secure version of the board,
-# enable building with TF-M as the Secure Execution Environment.
-config BUILD_WITH_TFM
-	default y
-
-# By default, if we build with TF-M, instruct build system to
-# flash the combined TF-M (Secure) & Zephyr (Non Secure) image
-config TFM_FLASH_MERGED_BINARY
-	default y
-	depends on BUILD_WITH_TFM
-
 endif # BOARD_NRF7120DK_NRF7120_CPUAPP_NS
--- a/dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
+++ b/dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
@@ -45,28 +45,28 @@
 			ranges = <0x0 0x0 DT_SIZE_K(416)>;
 		};
 
-		sram01_ns: sram0_ns@60000 {
+		sram01_ns: sram0_ns@68000 {
 			#address-cells = <1>;
 			#size-cells = <1>;
 			/* Non-Secure RAM */
-			reg = <0x60000 DT_SIZE_K(256)>; /* 0x2008_0000 */
-			ranges = <0x0 0x60000 DT_SIZE_K(256)>;
+			reg = <0x68000 DT_SIZE_K(256)>; /* 0x2008_0000 */
+			ranges = <0x0 0x68000 DT_SIZE_K(256)>;
 		};
 
-		sram02_ns: sram0_ns@a0000 {
+		sram02_ns: sram0_ns@a8000 {
 			#address-cells = <1>;
 			#size-cells = <1>;
 			/* Non-Secure RAM */
-			reg = <0xa0000 DT_SIZE_K(128)>; /* 0x200C_0000 */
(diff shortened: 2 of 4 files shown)


boards/nordic/nrf7120dk/Kconfig.defconfig
```diff
@@ -10,7 +10,7 @@ config HAS_BT_CTLR
 	default BT
 
 config ROM_START_OFFSET
-	default 0x800 if BOOTLOADER_MCUBOOT
+	default 0x800 if BOOTLOADER_MCUBOOT && !BUILD_WITH_TFM
```
nordicjm: no to no-ups

nordicjm (changes requested):

boards/nordic/nrf7120dk/Kconfig.defconfig
travis3630: Will bring this to upstream later, for this, just to show you if you are fine for me to add the snippet this way?

nordicjm: not really following why a new snippet is needed, the partitions should be defined normally (using partitions inside of partitions) then you just use the ones depending upon the mode you are in

travis3630: When TFM is built with MCUboot, other platforms can do it because they have partition manager but we dont have? I tried build something NS with MCUboot, the boot-partition is needed so we need the first snippet: WITH_MCUBOOT_PARTITIONS.

When I build something NS without the MCUBOOT enabled, I will have to offset the tfm image back to address 0x0000_0000, which I am not able to do that WITH_MCUBOOT_PARTITIONS. TF-M does enable DT_USE_CODE_PARTITION where it needs to enforce the use of DTS, and there is not other ways to do it without the snippet.

dhsu-nordic (approved):

dts/vendor/nordic/nrf7120_cpuapp_ns_partition.dtsi
```diff
@@ -24,23 +26,23 @@
...
+		ranges = <0x0 0x18000 DT_SIZE_K(920)>;
 
 		sram00_ns: sram0_ns@0 {
 			#address-cells = <1>;
 			#size-cells = <1>;
 			/* Non-Secure RAM */
-			reg = <0x0 DT_SIZE_K(384)>; /* 0x2002_0000 */
-			ranges = <0x0 0x0 DT_SIZE_K(384)>;
+			reg = <0x0 DT_SIZE_K(416)>; /* 0x2001_8000 */
+			ranges = <0x0 0x0 DT_SIZE_K(416)>;
```
simonduq, line 45: Now that sram00_ns is larger by 0x8000 don't we need to offset sram01_ns, sram02_ns, sram03_ns accordingly?
I mean start stam01_ns at 0x68000 instead of 0x60000 etc.

travis3630: What we doing is increasing ram00_ns only? Size and offset for each sramX is always fixed, so the total size of ram00 is fixed, when we reduce sram00_s, then we increase sram00_ns, sram01 should remain fixed in their location.

simonduq: This is the part I'm confused about: `sram01_ns: sram0_ns@60000`
Do we really want to keep 0x60000 (== 384K, the old ram00 size)?

travis3630: ok that is 68000 in sram01_ns, will get it corrected.

simonduq: And then sram02/03 also need shifting by 0x8000 as far as I understand

nordicjm (approved):

simonduq (approved):
Thank you for addressing my comment!

PR #4252, opened 2026-07-01
[nrf fromtree] drivers: nrf_wifi: Support to set extended sleep interval

kapbh:
Extended sleep interval is the period during which STA will stop all types of TX/RX. Through this command user can configure the quiet period.

Upstream PR #: 112368

manifest-pr-skip

### DIFF
--- a/drivers/wifi/nrf_wifi/src/wifi_util.c
+++ b/drivers/wifi/nrf_wifi/src/wifi_util.c
@@ -1112,7 +1112,7 @@ static int nrf_wifi_util_req_extended_sleep(const struct shell *sh,
 
 	val = strtoul(argv[1], &ptr, 10);
 
-	if ((val < 0) || (val > UINT_MAX)) {
+	if (val > UINT_MAX) {
 		shell_fprintf(sh,
 			      SHELL_ERROR,
 			      "Invalid value(%lu).\n",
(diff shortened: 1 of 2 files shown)


sachinthegreen (approved):

drivers/wifi/nrf_wifi/src/wifi_util.c
```diff
@@ -1097,6 +1102,41 @@ static int nrf_wifi_util_dump_rpu_stats_mem(const struct shell *sh,
...
+					    size_t argc,
+					    const char *argv[])
+{
+	enum nrf_wifi_status status = NRF_WIFI_STATUS_FAIL;
+	char *ptr = NULL;
+	unsigned long val = 0;
+
+	val = strtoul(argv[1], &ptr, 10);
+
+	if ((val < 0) || (val > UINT_MAX)) {
```
rado17, line 1115: Concern raised by SonarCloud is legit. Needs to be fixed.

rado17 (changes requested):

drivers/wifi/nrf_wifi/src/wifi_util.c
kapbh: done

rado17 (approved):

PR #4260, opened 2026-07-02
[nrf fromlist] boards: nordic: nrf7120dk: Add Wi-Fi XO frequency offset

krish2718:
The Wi-Fi firmware needs a crystal oscillator (XO) frequency offset trim for its calibration. This is a board-level value that depends on the crystal, so express it in devicetree rather than sourcing it from a fixed register.

Add a nordic,wifi-xo-freq-offset property to the nordic,nrf7120-wifi binding (default 0) and set it on the nRF7120 DK.

Upstream PR #: 112610

Assisted-by: Claude:claude-opus-4.8

manifest-pr-skip

### DIFF
--- a/boards/nordic/nrf7120dk/nrf7120_cpuapp_common.dtsi
+++ b/boards/nordic/nrf7120dk/nrf7120_cpuapp_common.dtsi
@@ -151,6 +151,7 @@
 
 &wifi {
 	status = "okay";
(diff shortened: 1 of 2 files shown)


sachinthegreen (approved):

krish2718:
Linked manifest PR is now closed, this will be pulled as part of upmerge.

PR #4263, opened 2026-07-03
[nrf fromlist] tests: drivers: spi: spi_controller_peripheral: Incorr…

nordic-pikr:
…ect SCK

Added a set of test cases to verify the SPI controller's behavior when the SCK line is in an incorrect state (the opposite of the CPOL setting) just before transmission. These tests will be skipped if the GPIO SCK is not defined in the device tree.

Upstream PR #: 112705

### DIFF
--- a/tests/drivers/spi/spi_controller_peripheral/src/main.c
+++ b/tests/drivers/spi/spi_controller_peripheral/src/main.c
@@ -68,6 +68,11 @@ struct test_data {
 
 static struct test_data tdata;
 
+struct k_work_q spim_spis_work_q;
+
+static K_KERNEL_STACK_DEFINE(spim_spis_work_q_stack,
+			CONFIG_SPI_CONTROLLER_PERIPHERAL_WORKQUEUE_STACK_SIZE);
+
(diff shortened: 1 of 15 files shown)


tests/drivers/spi/spi_controller_peripheral/src/main.c
```diff
@@ -68,6 +75,11 @@ struct test_data {
 
 static struct test_data tdata;
 
+struct k_work_q spim_spis_work_q;
```
mstasiaknordic, line 78: nit: If the tests are to be used by other vendors, Nordic-specific SPIM or SPIS naming could be replaced by something more general

mstasiaknordic (approved):

tests/drivers/spi/spi_controller_peripheral/src/main.c
nordic-pikr: the commit with this change is already merged (fromtree) and I am not its author :)

nordic-segl (approved):

PR #4273, opened 2026-07-08
History cleanup on ncs-v3.4-branch

rlubos:
History cleanup


rlubos:
Diff is empty compared to the current `ncs-v3.4-branch`, but please double check.

PR #4278, opened 2026-07-09
[nrf noup] drivers: ironside: re-ring bell to fix notify-before-dequeue

sigvartmh:
IronSide SE servers predating NCSDK-39877 notify the client before dequeuing the request. In that window the Zephyr client can complete the call and post a new request in the same buffer (release, alloc, write, ring the bell) before the server dequeues. The server's ironside_call_recv scan then skips the buffer because it still looks queued, so the new request is never enqueued and the client blocks forever in k_event_wait(K_FOREVER).

Wait CONFIG_IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS instead of forever and re-ring the bell on timeout. Once the server has (belatedly) dequeued, the re-ring makes it rescan and pick up the pending request. Re-ringing a healthy or merely slow server is harmless because the server-side enqueue is idempotent. Set the timeout to 0 to restore the wait-forever behaviour on servers known to include the NCSDK-39877 fix.

new commit message:
>Some IronSide SE versions notify the client before dequeuing the
request. A request posted in that window is skipped and would otherwise
never be answered, leaving the client waiting forever.

>Wait CONFIG_IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS instead of forever and
post the request again on timeout. Posting again is harmless because the
enqueue is idempotent. Set the timeout to 0 to wait forever on IronSide
SE versions that include the fix.

Ref: NCSDK-39877

### DIFF
--- a/modules/hal_nordic/ironside/se/Kconfig
+++ b/modules/hal_nordic/ironside/se/Kconfig
@@ -50,6 +50,19 @@ config IRONSIDE_SE_CALL_MINIMAL
 
 endchoice
 
+config IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS
+	int "IronSide call response timeout (ms)"
+	default 1000
+	depends on IRONSIDE_SE_CALL_ZEPHYR
+	help
+	  Time to wait for a response before posting the request again,
+	  in milliseconds. This is a workaround for IronSide SE versions
+	  that notify the client before dequeuing the request: a request
+	  posted in that window is skipped and would otherwise never be
+	  answered. Posting the request again is harmless because the
+	  enqueue is idempotent. Set to 0 to wait forever and disable the
+	  workaround on IronSide SE versions that include the fix.
+
 config IRONSIDE_SE_CALL_INIT_PRIORITY
 	int "IronSide calls' driver initialization priority"
 	default 41
--- a/modules/hal_nordic/ironside/se/call.c
+++ b/modules/hal_nordic/ironside/se/call.c
@@ -112,6 +112,11 @@ struct ironside_se_call_buf *ironside_se_call_alloc(void)
 void ironside_se_call_dispatch(struct ironside_se_call_buf *buf)
 {
 	const uint32_t buf_bit = BIT(buf - bufs);
+#if CONFIG_IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS > 0
+	const k_timeout_t rsp_timeout = K_MSEC(CONFIG_IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS);
+#else
+	const k_timeout_t rsp_timeout = K_FOREVER;
+#endif
 	int err;
 
 	buf->status = IRONSIDE_SE_CALL_STATUS_REQ;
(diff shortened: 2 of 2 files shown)


modules/hal_nordic/ironside/se/call.c
```diff
@@ -124,7 +124,19 @@ void ironside_se_call_dispatch(struct ironside_se_call_buf *buf)
 	err = mbox_send_dt(&mbox_tx, NULL);
 	__ASSERT_NO_MSG(err == 0);
 
+#if CONFIG_IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS > 0
+	/* Workaround for IronSide SE servers predating NCSDK-39877, which
+	 * notify the client before dequeuing the request. Re-ringing a healthy
+	 * (or merely slow) server is harmless, as the enqueue is idempotent.
+	 */
```
57300:
```suggestion
```
Maybe drop this and refer to Kconfig instead, which has a more complete description of the bug.

modules/hal_nordic/ironside/se/call.c
```diff
@@ -124,7 +124,19 @@ void ironside_se_call_dispatch(struct ironside_se_call_buf *buf)
...
+	while (k_event_wait(&rsp_evts, buf_bit, false,
+			    K_MSEC(CONFIG_IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS)) == 0) {
+		err = mbox_send_dt(&mbox_tx, NULL);
+		__ASSERT_NO_MSG(err == 0);
```
57300, line 134: nit: I'd prefer to rewrite the existing code into a do-while, to avoid repetition of `mbox_send_dt` etc.

It's okay to check the result of `k_event_wait` even in the `K_FOREVER` case.

modules/hal_nordic/ironside/se/Kconfig
```diff
@@ -50,6 +50,18 @@ config IRONSIDE_SE_CALL_MINIMAL
 
 endchoice
 
+config IRONSIDE_SE_CALL_RESPONSE_TIMEOUT_MS
+	int "IronSide call response re-ring timeout (ms)"
+	default 1000
+	depends on IRONSIDE_SE_CALL_ZEPHYR
+	help
+	  Time to wait for a response before re-notifying (re-ringing) the
+	  server, in milliseconds. This is a workaround for IronSide SE servers
```
57300:
```suggestion
	  Time to wait for a response before posting the request again,
	  in milliseconds. This is a workaround for IronSide SE versions
```
Prefer not to say "IronSide SE server". IronSide SE is the server.

Amend the commit message accordingly.

modules/hal_nordic/ironside/se/Kconfig
```diff
@@ -50,6 +50,18 @@ config IRONSIDE_SE_CALL_MINIMAL
...
+	  that notify the client before dequeuing the request (fixed in
+	  NCSDK-39877): a request posted in that window is skipped by the server
```
57300: If we intend to upstream this, then we shouldn't mention Jira tickets.

Normally, we'd put a fix version (not yet available) or affected versions of ISE instead.

Amend the commit message accordingly.

modules/hal_nordic/ironside/se/Kconfig
```diff
@@ -50,6 +50,18 @@ config IRONSIDE_SE_CALL_MINIMAL
...
+	  and would otherwise never be answered. Set to 0 to wait forever and
```
57300: Mention the idempotent enqueue somewhere here.

57300 (commented):
Thanks for taking this on.

Let's cut back on the server-side implementation details in the commit message.

Also, put `modules: hal_nordic: ironside:` in the title.

sigvartmh:
@57300 Tried to address your comments and moved logic into `do while`

57300 (approved):

karstenkoenig (approved):

sigvartmh:
Included by https://github.com/nrfconnect/sdk-zephyr/pull/4454

PR #4293, opened 2026-07-17
Enable ram retention for nrf7120

erdemsimsek:
This PR covers;

- Brings commits from upstream zephyr. 
- Extend `system_off` test for nRF120/ns variant via no-up commit
- Configures ram retention just before `system_off` for nRF7120/ns via no-up commit


samples/boards/nordic/system_off/boards/nrf7120dk_nrf7120_cpuapp_ns.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_TFM_SFN=y
+CONFIG_TFM_NRF_SYSTEM_OFF_SERVICE=y
```
rob-robinson-14: Should system_off_service just be a default for ns builds, rather than sample config?

erdemsimsek: System-off service and support level are marked as experimental in the Kconfigs.
Also, `CONFIG_TFM_SFN` isn't enabled automatically for `nRF7120`,. At the moment, the default config for `nRF7120` is `CONFIG_TFM_IPC`

drivers/retained_mem/Kconfig.nrf
```diff
@@ -14,5 +14,6 @@ config RETAINED_MEM_NRF_RAM_CTRL
 	default y
 	depends on DT_HAS_ZEPHYR_RETAINED_RAM_ENABLED && RETAINED_MEM_ZEPHYR_RAM && POWEROFF
 	depends on HAS_NORDIC_RAM_CTRL
```
rob-robinson-14, line 16: `NRF_TFM_RAM_CTRL_SERVICE` depends on `HAS_NORDIC_RAM_CTRL`. So do you need HAS_NORDIC_RAM_CTRL here as well?

samples/boards/nordic/system_off/sample.yaml
```diff
@@ -149,6 +149,34 @@ tests:
         - "Active Ticks:"
         - "Entering system off; wait 2 seconds to restart"
         - "Wakeup from System OFF by GRTC."
+  sample.boards.nrf.system_off.retained_mem.tfm:
```
rob-robinson-14: consider adding grtc_wakeup to name to align with other tests?
Is there value in a test for gpio wakeup as well? Even if there is, could be scope for future work.

drivers/retained_mem/Kconfig.nrf
erdemsimsek: No additional guard is needed here because `RETAINED_MEM_NRF_RAM_CTRL` already has a separate depends on `HAS_NORDIC_RAM_CTRL`.

samples/boards/nordic/system_off/sample.yaml
erdemsimsek: Updated the test name to `sample.boards.nrf.system_off.retained_mem.grtc_wakeup.tfm` to align with the existing GRTC wakeup test naming.

ioannisg:
@erdemsimsek you can update the description of the PR, as the commits are now from-tree.

samples/boards/nordic/system_off/boards/nrf7120dk_nrf7120_cpuapp_ns.overlay
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/ {
+	chosen {
+		zephyr,sram = &app_sram;
+	};
+
+	app_sram: memory@20020000 {
```
rob-robinson-14: Should delete sram0_ns or cpuapp_sram since nodes overlap? Same for retained_ram section with sram01_ns?

erdemsimsek: The overlap is intentional. 
`cpuapp_sram` defines the complete physical RAM available to CPUAPP. `sram0_ns` is a child of that node and defines which part of CPUAPP RAM is non-secure. 
TF-M uses `sram0_ns` to determine the complete non-secure RAM range, so neither node should be deleted.

samples/boards/nordic/system_off/boards/nrf7120dk_nrf7120_cpuapp_ns.conf
nordicjm: noup not acceptable

nordicjm (changes requested):

samples/boards/nordic/system_off/boards/nrf7120dk_nrf7120_cpuapp_ns.conf
rob-robinson-14: noup dropped - will extend in sdk-nrf as a follow up PR

rob-robinson-14 (approved):

nordicjm (approved):

PR #4307, opened 2026-07-24
[nrf fromtree] Soc flash nrf mramc read secure mem downstream

travis3630:
The noup commit is to support the dts method of getting address of slot0_ns_partition. And also soc_secure_flash_range_is_secure() function unavailable in downstream.

similar to commit that done for soc_flash_nrf and soc_flash_nrf_rram: https://github.com/nrfconnect/sdk-zephyr/commit/9fb02fe66fa005b5a12fe69bf72d4616c1c70c2e

manifest-pr-skip
Manifest update PR in sdk-nrf: https://github.com/nrfconnect/sdk-nrf/pull/30752

### DIFF
--- a/drivers/flash/soc_flash_nrf_mramc.c
+++ b/drivers/flash/soc_flash_nrf_mramc.c
@@ -12,6 +12,7 @@
 
 #if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
 #include "tfm_ioctl_core_api.h"
+#include <soc_secure.h>
 #else
 #include <nrfx_mramc.h>
 #endif
@@ -180,6 +181,10 @@ static int nrf_mramc_read(const struct device *dev, off_t offset, void *data, si
 	/* Buffer read number of bytes and store in data pointer.
 	 */
 #if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
+	if (soc_secure_flash_range_is_secure((uintptr_t)addr, len)) {
+		LOG_DBG("read secure mem: %x:%zu", addr, len);
+		return soc_secure_mem_read(data, (void *)addr, len);
+	}
 	memcpy(data, (void *)addr, len);
 #else
 	nrfx_mramc_buffer_read(data, addr, len);


drivers/flash/soc_flash_nrf_mramc.c
```diff
@@ -181,10 +181,12 @@ static int nrf_mramc_read(const struct device *dev, off_t offset, void *data, si
 	/* Buffer read number of bytes and store in data pointer.
 	 */
 #if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE)
-	if (soc_secure_flash_range_is_secure((uintptr_t)addr, len)) {
+#if DT_NODE_EXISTS(DT_NODELABEL(slot0_ns_partition))
```
rob-robinson-14: why is this a noup - slot0_ns_partition is in upstream zephyr?

How do you guarantee that there are no secure addresses after slot0_ns_partition?

travis3630: If you look at the source file soc/Nordic/common/soc_secure.h function : [upstream][zephyr/soc/nordic/common/soc_secure.h at 8ba1698fbf8eb76401aa4a39b8c476453dcb1d30 · zephyrproject-r…](https://github.com/zephyrproject-rtos/zephyr/blob/8ba1698fbf8eb76401aa4a39b8c476453dcb1d30/soc/nordic/common/soc_secure.h#L52) vs [downstream] [sdk-zephyr/soc/nordic/common/soc_secure.h at main · nrfconnect/sdk-zephyr](https://github.com/nrfconnect/sdk-zephyr/blob/main/soc/nordic/common/soc_secure.h) They are very different... and missing of soc_secure_flash_range_is_secure()

Same goes to: drivers/flash/soc_flash_nrf_rram.c doing same thing, and nrf5340 soc_flash_nrf.c
 
Downstream: [sdk-zephyr/drivers/flash/soc_flash_nrf_rram.c at main · nrfconnect/sdk-zephyr](https://github.com/nrfconnect/sdk-zephyr/blob/main/drivers/flash/soc_flash_nrf_rram.c#L402) vs upstream: [zephyr/drivers/flash/soc_flash_nrf_rram.c at main · zephyrproject-rtos/zephyr](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/flash/soc_flash_nrf_rram.c#L391)

So the sdk-zephyr maybe not having the update and behind it, it is not rational to bring it downstream ourself since the changes also affect others. We can commit this as a noup to make things work first, and everyone is doing it. Or wait until next sdk upmerge which is not deterministic whether it will be present.

travis3630:
Depends on: https://github.com/nrfconnect/sdk-nrf/pull/30168

FYI: @rob-robinson-14

rob-robinson-14 (approved):

nordicjm (approved):

PR #4333, opened 2026-07-31
[nrf noup] ci: cancel in-progress manifest PR workflows

gminn:
If a PR is merged while a manifest-PR workflow is running, the running workflow should be canceled in favor of the new workflow triggered by the merge. This can help prevent race conditions between workflow instances both writing to the same sdk-nrf branch.

### DIFF
--- a/.github/workflows/manifest-PR.yml
+++ b/.github/workflows/manifest-PR.yml
@@ -9,6 +9,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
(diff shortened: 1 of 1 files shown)


rlubos (commented):
I think it makes sense but it'd need `[nrf noup] ` tag in the title and 
`nrf-squash! [nrf noup] ci: Enable action-manifest-pr`
line in the commit message.

rlubos (approved):

andu-nordic (approved):

thst-nordic (approved):

PR #4334, opened 2026-08-03
[noup] VPR pin control support

lucasvdp:
- Allow pins to be assigned to the VPR cores in a generic way (not MSPI specific)
- Allow pin configuration for VPR cores using devicetree.


lucasvdp:
There was only MSPI specific pincontroll for the VPR cores in nrfconnect/sdk-zephyr, this is not in zephyrproject-rtos/zephyr. It might be nice to have the generic pincontroll for the VPR cores in upstream Zephyr. The Laucher is there, it just  doesn't allow you to assign pins to the VPR cores.

Changing code that was added with [noup] with code that is also upstream might be tricky so I left it noup for now.

lucasvdp:
@anangl You added the HPF MSPI pin support, it makes sense to only have that as [NOUP]. Do you think this PR makes sense? And should we add only this generic VIO pin to upstream Zephyr?

lucasvdp:
Needed for: https://github.com/nrfconnect/sdk-nrf/pull/30487

lucasvdp:
Replaced by #4461

PR #4337, opened 2026-08-04
POC [nrf noup] cmake: Kconfig: strict mode

michalek-no:
it would be nice to have build system option to fail if Kconfigs given by the user via CLI and/or prj.conf's are being changed in the process. This PR investigates scope and impact.


nordicjm (changes requested):
not acceptable in any way, shape or form

PR #4347, opened 2026-08-06
Fix nrf71 soc.c non secure initialsation issue downstream

travis3630:
The soc.c non-secure build initialisation was using NRF_TFM, which is meant to be for sdk-nrf only. New ifdef function is updated to replace the secure initialisation condition. So that now power related, wifi, wicr setup is done in secure environment for non secure build.

Also updated SystemCoreClock as global variable with retrival from the DT. SystemCoreClock setting will be done in non secure environment aligning to other nordic soc.

Corresponds to:
https://github.com/nrfconnect/sdk-trusted-firmware-m/pull/268

Manifest PR:
https://github.com/nrfconnect/sdk-nrf/pull/30578

manifest-pr-skip


dhsu-nordic (approved):

ioannisg:
@rob-robinson-14 pls add your review

rlubos (approved):

PR #4352, opened 2026-08-07
[nrf fromlist] modules: tf-m: make Crypto partition auto-enabled by PSA_CRYPTO

tomi-font:
In order to have a uniform way to enable and use PSA Crypto both with and without TF-M, make `CONFIG_TFM_PARTITION_CRYPTO` promptless and enable it automatically when `CONFIG_PSA_CRYPTO=y`.

manifest-pr-skip


magnev (approved):

samples/tfm_integration/tfm_psa_test/prj.conf
```diff
@@ -4,10 +4,14 @@
...
 CONFIG_QEMU_ICOUNT_SHIFT=1
 
-# Needed for CRYPTO and INITIAL_ATTESTATION
-CONFIG_MAIN_STACK_SIZE=4096
+# CONFIG_PSA_CSPRNG_GENERATOR gets automatically enabled due to CONFIG_PSA_CRYPTO
+# (needed to enable the Crypto partition), but compilation of random_psa.c fails
+# because TF-M doesn't expose the psa/crypto.h header when CONFIG_TFM_USE_NS_APP.
+# Work around this by enabling the test CSPRNG to allow the discarded NS app to build.
+CONFIG_TEST_RANDOM_GENERATOR=y
+CONFIG_TEST_CSPRNG_GENERATOR=y
```
degjorva, line 17: This seems rather problematic no? Is it just an issue with this specific test or can users face this as well?

degjorva (approved):

samples/tfm_integration/tfm_psa_test/prj.conf
tomi-font: Yeah this test overall can be problematic, we migrated the TF-M regression tests away from `CONFIG_TFM_USE_NS_APP` but this one still uses it. Users can't face it, it's purely a thing that now happens only with this test. see also comments on the upstream PR

PR #4379, opened 2026-08-17
[nrf noup] ci: exclude Bluetooth Mesh from CI-dfu-test

alxelax:
CI-dfu-test triggered on any change under subsys/bluetooth and on the blanket include/zephyr glob, so Bluetooth Mesh changes enabled the whole DFU test scope.

Mesh implements its own firmware update stack, built on the BLOB Transfer and Firmware Update models under subsys/bluetooth/mesh/dfu. It shares no code with the mcumgr/MCUboot DFU path that CI-dfu-test exercises, so mesh changes cannot regress those tests.

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -67,7 +67,6 @@
   - "dts/riscv/nordic/nrf54h*"
   - "include/dfu/**/*"
   - "include/mgmt/mcumgr/**/*"
-  - "include/zephyr/**/*"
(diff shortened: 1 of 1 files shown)


alxelax:
Only closing and reopening PR again retriggers Manifest step of CI correctly. Just re-run is always failed since step depends on "PR open" step.

ludvigsj (approved):

omkar3141 (approved):

PR #4393, opened 2026-08-20
[nrf fromtree] tests: arch: arm: arm_thread_swap_tz: shorten preempti…

degjorva:
…on delay

The 1 ms delay is longer than a hardware-accelerated secure call (e.g. ~0.5 ms for a 4 KB SHA-256), so the preemption lands after the call returns and the test fails. Use a single tick (tens of microseconds) instead.

(cherry picked from commit 6d4c0854a2052bfedde9b64243d27be1bf42d8a6)


frkv (approved):
LGTM

tomi-font (approved):

sigvartmh (approved):

PR #4406, opened 2026-08-24
[nrf fromlist] scripts: cmake: style: Load mixed-case commands from allow-list files

nordicjm:
Replace the hardcoded MIXED_CASE_COMMANDS set with an allow-list file
    shipped next to the script, and add a repeatable --mixed-case-file
    argument that appends extra files to it. This gives downstream projects
    a way to extend the lowercase-rule exceptions with their own mixed-case
    commands.

manifest-pr-skip


tejlmand:
upstream is merged, please update to `fromtree`.

carlescufi (approved):

tejlmand (approved):

PR #4409, opened 2026-08-25
[DNM][DNR] nrf71: bringup super PR

krish2718:
Draft super PR to be used for bringup.


soc/nordic/nrf71/soc.c
```diff
@@ -205,6 +357,12 @@ void soc_early_init_hook(void)
 #endif
 
 #if defined(CONFIG_SOC_NRF71_WIFI_BOOT)
+	/* Bring up the 64 MHz crystal oscillator the Wi-Fi core depends on. */
+	hfxo64m_setup();
+	hfxo64m_start();
```
Hi-Im-David, line 370: FYI these HFXO settings only need to be set because all the DK's are currently in use are in "production" life cycle state. The next batch (and all future SOC's shipped to customers) will be "empty" LCS so these will be set in the ROM instead.

krish2718: and IIUC, once the clock is started, its management is handled by the HW i.e., turning off and on based on incoming requests? So, no SW involvement?

PR #4410, opened 2026-08-25
nRF93M1DK specific updates from upstream

SeppoTakalo:
* Migrate to new nrf-clock DT-layout
* Fix runner confics
* Add missing DTR and RI pins
* Fix the flash-chip variant on the DK v0.4 board


maxd-nordic (approved):

boards/nordic/nrf93m1dk/nrf93m1dk_nrf54l15_cpuapp_common.dtsi
```diff
@@ -103,6 +103,18 @@
...
 };
 
+&xo {
+	status = "okay";
+};
+
+&lfclk {
+	status = "okay";
+
+	/* DK does no have 32kHz crystal, use the internal RC oscillator */
```
SyverHaraldsen, line 113:
```suggestion
	/* DK does not have 32kHz crystal, use the internal RC oscillator */
```

SyverHaraldsen (approved):

boards/nordic/nrf93m1dk/nrf93m1dk_nrf54l15_cpuapp_common.dtsi
SeppoTakalo: Thanks, but these are cherry-picked from upstream, so we are not supposed to modify the content, otherwise we would just deviate from Zephyr, and make future upmerges and cherry-picks harded.

It should be fixed upstream.

PR #4415, opened 2026-08-27
[nrf fromlist] drivers: wifi: nrf_wifi: Fix raw TX with controlled-po…

kapbh:
…rt dormant logic

Hold STA interfaces dormant until the 802.1X controlled port is authorized, so the IP stack cannot transmit during the 4-way handshake (EAPOL still flows out-of-band via the control port). Raw TX injection has no association or controlled port, so track it in tx_injection_active and let it bring the interface operational once the carrier is on, keeping oper-state consistent with TX admission.

Centralize the operational-state decision in the net-iface work handler and route both set_supp_port() and the TX-injection config path through it. Submitting the work item only queues it, so the dormant toggle runs later on the system workqueue, outside vif_lock.

Upstream PR #: 117295


drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -231,12 +231,36 @@ void nrf_wifi_rpu_recovery_cb(void *vif_ctx_handle,
 }
 #endif /* CONFIG_NRF_WIFI_RPU_RECOVERY */
 
+#ifdef CONFIG_NRF70_RAW_DATA_TX
+/* The FMAC VIF context holds the only txinjection_mode the firmware keeps in
+ * sync, so it is what the dormant gate and the teardown path both work on.
+ */
+static struct nrf_wifi_fmac_vif_ctx *nrf_wifi_get_fmac_vif_ctx(
```
krish2718: Good helper. Suggest going one step further: wrap the null-checks + `txinjection_mode` read in a `nrf_wifi_txinjection_active(vif)` predicate that returns false when `CONFIG_NRF70_RAW_DATA_TX` is off. That lets the work handler below drop its `#ifdef` and the local `fmac_vif_ctx`. Also, this getter and the refresh helper are needed in `wifi_mgmt.c` as well, so consider moving them to a shared header as `static inline` rather than file-static here.

drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -248,17 +272,31 @@ static void nrf_wifi_net_iface_work_handler(struct k_work *work)
...
-			net_if_dormant_off(vif_ctx_zep->zep_net_if_ctx);
-		}
-	} else if (vif_ctx_zep->if_carr_state == NRF_WIFI_FMAC_IF_CARR_STATE_OFF) {
+	/* A station must not go operational on association alone: that would let
+	 * the IP stack transmit during the 802.1X handshake, which the TX path
+	 * then has to drop. EAPOL flows out-of-band via
+	 * nrf_wifi_wpa_tx_control_port(), so wait for the controlled port.
+	 * Other interface types only depend on the carrier.
+	 */
+	operational = (vif_ctx_zep->if_carr_state == NRF_WIFI_FMAC_IF_CARR_STATE_ON) &&
```
krish2718: Right place to centralize, but the expression is doing a lot inline. Consider extracting `bool nrf_wifi_iface_operational(vif)` so the handler reads as policy:

```c
if (nrf_wifi_iface_operational(vif_ctx_zep))
	net_if_dormant_off(vif_ctx_zep->zep_net_if_ctx);
else
	net_if_dormant_on(vif_ctx_zep->zep_net_if_ctx);
```

The RAW_DATA_TX term then becomes `carrier_on && nrf_wifi_txinjection_active(vif)` with no `#ifdef` and no local `fmac_vif_ctx` variable in the handler.

drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -987,6 +1025,15 @@ int nrf_wifi_if_start_zep(const struct device *dev, struct net_if *iface)
 
 	k_mutex_init(&vif_ctx_zep->vif_lock);
 	vif_ctx_zep->if_type = add_vif_info.iftype;
+	/* The VIF was just recreated with a zeroed FMAC context, so match it and
+	 * drop whatever the previous session left behind.
+	 */
```
krish2718: This block is duplicated verbatim in `nrf_wifi_if_stop_zep` (and a third `txinjection_mode = false` reset lives in both stop and `wifi_mgmt.c`). Suggest one `nrf_wifi_clear_session_state(vif)` helper that guards `authorized` / `if_carr_state` / `txinjection_mode` once, called from both start and stop.

drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -1078,6 +1125,12 @@ int nrf_wifi_if_start_zep(const struct device *dev, struct net_if *iface)
 	}
 out:
 	k_mutex_unlock(&vif_ctx_zep->vif_lock);
+#ifdef CONFIG_NRF70_DATA_TX
```
krish2718: This `#ifdef`-guarded submit + comment is repeated at 5 sites (start, stop, set_config, wifi_mgmt, wpa_supp_if). Wrap it in `static inline void nrf_wifi_refresh_oper_state(vif)` (no-op when `!CONFIG_NRF70_DATA_TX`) and every call site loses its `#ifdef`, leaving just `nrf_wifi_refresh_oper_state(vif_ctx_zep);`.

drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -1110,13 +1166,27 @@ int nrf_wifi_if_stop_zep(const struct device *dev, struct net_if *iface __unused
 		goto unlock;
 	}
 
+#if defined(CONFIG_NRF70_STA_MODE) || defined(CONFIG_NRF70_RAW_DATA_TX)
```
krish2718: Duplicate of the block in `nrf_wifi_if_start_zep` - fold into the shared `nrf_wifi_clear_session_state()` helper suggested there.

drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -1110,13 +1166,27 @@ int nrf_wifi_if_stop_zep(const struct device *dev, struct net_if *iface __unused
...
+
 	rpu_ctx_zep = vif_ctx_zep->rpu_ctx_zep;
 	if (!rpu_ctx_zep || !rpu_ctx_zep->rpu_ctx) {
 		LOG_ERR("%s: rpu_ctx_zep is NULL",
 			__func__);
 		goto unlock;
 	}
 
+#ifdef CONFIG_NRF70_RAW_DATA_TX
+	fmac_vif_ctx = nrf_wifi_get_fmac_vif_ctx(vif_ctx_zep);
```
krish2718: Same `txinjection_mode = false` reset is duplicated in `wifi_mgmt.c` `nrf_wifi_mode()`. If it moves into `nrf_wifi_clear_session_state()`, both sites collapse to one call.

drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -1313,6 +1389,7 @@ int nrf_wifi_if_set_config_zep(const struct device *dev,
 		    config->txinjection_mode) {
 			LOG_INF("%s: Driver TX injection setting is same as configured setting",
 				__func__);
+			ret = 0;
```
krish2718: Nit: this `ret = 0;` is an unrelated bugfix - the same-setting early-out previously fell through and returned -1. Worth calling out in the commit body (or splitting), since it is not part of the dormant refactor.

drivers/wifi/nrf_wifi/src/net_if.c
```diff
@@ -1336,6 +1413,14 @@ int nrf_wifi_if_set_config_zep(const struct device *dev,
 			LOG_ERR("%s: Mode set operation failed", __func__);
 			goto unlock;
 		}
+
+		sys_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx]->txinjection_mode =
```
krish2718: With the refresh helper this reduces to:

```c
sys_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx]->txinjection_mode = config->txinjection_mode;
nrf_wifi_refresh_oper_state(vif_ctx_zep);
ret = 0;
goto unlock;
```

no `#ifdef` at the call site.

drivers/wifi/nrf_wifi/src/wifi_mgmt.c
```diff
@@ -788,6 +788,9 @@ int nrf_wifi_mode(const struct device *dev,
 	struct peers_info *peer = NULL;
 	int i = 0;
 	int ret = -1;
+#ifdef CONFIG_NRF70_DATA_TX
+	bool refresh_dormant = false;
```
krish2718: If `nrf_wifi_refresh_oper_state()` is a no-op when data TX is off, this bool needs no `#ifdef` guards (here or below): a plain `bool refresh = false;` works and the three `#ifdef CONFIG_NRF70_DATA_TX` blocks in this function collapse to a single trailing `if (refresh) nrf_wifi_refresh_oper_state(vif_ctx_zep);`.

drivers/wifi/nrf_wifi/src/wifi_mgmt.c
```diff
@@ -854,6 +857,14 @@ int nrf_wifi_mode(const struct device *dev,
 			goto out;
 		}
 
+#ifdef CONFIG_NRF70_RAW_DATA_TX
```
krish2718: Same `txinjection_mode = false` reset as `nrf_wifi_if_stop_zep` - candidate for the shared clear-session helper. And `refresh_dormant = true` here loses its own `#ifdef` once the refresh helper is a safe no-op.

drivers/wifi/nrf_wifi/src/wpa_supp_if.c
```diff
@@ -1192,11 +1192,7 @@ int nrf_wifi_wpa_set_supp_port(void *if_priv, int authorized, char *bssid)
 	 * uses the control port), avoiding transmits into the closed port.
 	 */
 	if (update_dormant) {
-		if (authorized) {
-			net_if_dormant_off(vif_ctx_zep->zep_net_if_ctx);
-		} else {
-			net_if_dormant_on(vif_ctx_zep->zep_net_if_ctx);
-		}
+		k_work_submit(&vif_ctx_zep->nrf_wifi_net_iface_work);
```
krish2718: This is the cleanest example of the centralization - the caller just requests a re-evaluation instead of duplicating the on/off branch. With the wrapper it becomes `nrf_wifi_refresh_oper_state(vif_ctx_zep);` and reads self-documenting.

krish2718 (commented):
Review moved to the upstream PR (this is a [nrf fromlist] backport): https://github.com/zephyrproject-rtos/zephyr/pull/117295#pullrequestreview-5112383686 - please follow up there.

sachinthegreen (approved):

PR #4465, opened 2026-09-14
[nrf noup] soc: nordic: uicr: Update PERIPHCONF register file

jtguggedal:
Update the nRF9220 SoC PERIPHCONF register file.


nordicjm (changes requested):
strong justification needed as to why a no-up is needed for this rather than the proper way of doing it with Kconfig that can be extended out of tree e.g. like how sysbuild does it https://nrfconnectdocs.nordicsemi.com/ncs/latest/nrf/app_dev/config_and_build/sysbuild/sysbuild_images.html#adding-to-a-single-project

shanthanordic (approved):
apporved @jtguggedal

PR #4487, opened 2026-09-18
[nrf fromlist] net: ipv6: Accept NA and RA packets without options

rlubos:
RFC 4861 sets the minimum ICMP length of a Neighbor Advertisement to 24 octets and of a Router Advertisement to 16 octets, the fixed header only. The input length checks also required room for one option header, so valid option-free NA and RA packets were dropped. This broke NUD against peers answering a unicast NS without a TLLAO option and ignored RAs that carry only the fixed header.

Upstream PR #: 119499

Assisted-by: Claude:claude-fable-5.1


ArekBalysNordic (approved):
LGTM

jukkar (approved):

